PR #4, opened 2017-08-01
Implements KISS-GP for scalable regression

jrg365:
Implements KISS-GP (http://proceedings.mlr.press/v37/wilson15.pdf) for scalable regression. The internals of this implementation could still use a decent amount of basic optimization (the unit test with 100 training points takes about 6 seconds), but all operations should be linear space and time in the number of training points. 

To do this, covariance matrices returned by a `Kernel` can now be `LazyVariables` that implicitly wrap a structured matrix. Then, functions defined in the top level of the package (e.g., `gpytorch.add_diag`) have different behavior depending on whether their inputs are of type `Variable` or some `LazyVariable` type. In particular, `gpytorch.exact_gp_marginal_log_likelihood` can be evaluated much more quickly when `covar` is a `ToeplitzLazyVariable`.

I also moved most of the internals of `_ExactGPPosterior.forward` to `gpytorch._exact_predict` so that this pattern can continue: `gpytorch._exact_predict` is more efficient when the various covariance matrix inputs are `ToeplitzLazyVariables`.

To accomplish these efficiency gains, gpytorch now requires access to a good FFT algorithm and accompanying operations on Toeplitz matrices. As a result, GPyTorch now requires libfftw3 (http://www.fftw.org/).

Right now, `_VariationalGPPosterior` does not work with `LazyVariables`


gpytorch/lazy/toeplitz_lazy_variable.py
```diff
@@ -0,0 +1,150 @@
...
+                WTW = torch.dsmm(W_left, W_right_T.t())
+        else:
+            WTW = utils.toeplitz(self.c.data, self.r.data)
+
+        if self.added_diag is not None:
+            WTW = WTW + torch.diag(self.added_diag.data)
+
+        return WTW
+
+    def explicit_interpolate_T(self, J, C):
```
gpleiss: We should rename this

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Project specific
+gpytorch/libfft
```
gpleiss: FIx build file

gpytorch/__init__.py
```diff
@@ -1,37 +1,250 @@
...
+
+    N1 is parameterized by a mean vector and the Cholesky decomposition of the covariance matrix.
+    N2 is parameterized by a mean vector and covariance matrix.
+
+    For more information, see the MVNKLDivergence function documentation.
+    """
     return MVNKLDivergence()(mean_1, chol_covar_1, mean_2, covar_2)
+
+
+def interpolated_toeplitz_mul(c, W_left, W_right, v, noise_diag=None):
```
gpleiss: Move to util

gpytorch/__init__.py
```diff
@@ -1,37 +1,250 @@
...
+    WTWt_v = torch.dsmm(W_left, TWt_v).squeeze()
+
+    if noise_term is not None:
+        # Get (W_{l}TW_{r}^{T} + \sigma^{2}I)v
+        WTWt_v = WTWt_v + noise_term
+
+    return WTWt_v
+
+
+def _exact_predict(test_mean, test_test_covar, train_y, train_mean,
```
gpleiss: It would be good to improve this, so that all the branching logic is handled in separate classes.

gpytorch/__init__.py
```diff
@@ -1,37 +1,250 @@
...
+        # TODO: Add a diagonal only mode / use implicit math
+        train_test_covar = train_test_covar.evaluate()
+        test_test_covar = test_test_covar.evaluate()
+
+        test_test_covar_correction = test_train_mul_closure(LinearCG().solve(train_mul_closure, train_test_covar))
+        test_test_covar = Variable(test_test_covar.sub(test_test_covar_correction))
+    else:
+        train_y_var = Variable(train_y)
+        if alpha is None:
+            alpha = invmv(train_train_covar, train_y_var - train_mean)
```
gpleiss: This should use function forms of invmm and invmv once they're implemented

gpytorch/inference/posterior_models/exact_gp_posterior.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # If there's data, use it
+        if n:
+            train_mean = full_mean[:n]
+            train_train_covar = gpytorch.add_diag(full_covar[:n, :n], log_noise.exp())
+            train_test_covar = full_covar[:n, n:]
+            test_train_covar = full_covar[n:, :n]
+
+            if hasattr(self, 'alpha') and self.alpha is not None:
+                GRV, alpha = gpytorch._exact_predict(test_mean, test_test_covar, self.train_y, train_mean,
```
gpleiss: Rename GRV

gpytorch/inference/posterior_models/exact_gp_posterior.py
```diff
@@ -0,0 +1,75 @@
...
+                                                     train_train_covar, train_test_covar, test_train_covar, self.alpha)
+            else:
+                GRV, alpha = gpytorch._exact_predict(test_mean, test_test_covar, self.train_y, train_mean,
```
gpleiss: In conjunction with refactoring `_exact_predict`, we should have the alpha caching being handled by some object that encapsulates the ugliness.

Something like a factory, that produces a `PosteriorMeanFunction` (or `ToeplitzPosteriorMeanFunction`), and the factory is in charge of the cache.

gpytorch/lazy/toeplitz_lazy_variable.py
```diff
@@ -0,0 +1,150 @@
...
+            n_left = len(self.J_left)
+            n_right = len(self.J_right)
+            W_left = utils.index_coef_to_sparse(self.J_left, self.C_left, len(self.c))
+            W_right = utils.index_coef_to_sparse(self.J_right, self.C_right, len(self.c))
+            if n_left <= n_right:
+                W_left_T = self.explicit_interpolate_T(self.J_left, self.C_left).data
+                WTW = torch.dsmm(W_right, W_left_T.t()).t()
+            else:
+                W_right_T = self.explicit_interpolate_T(self.J_right, self.C_right).data
+                WTW = torch.dsmm(W_left, W_right_T.t())
```
gpleiss: Change WTW

gpytorch/math/functions/interpolated_toeplitz_gp_marginal_log_likelihood.py
```diff
@@ -0,0 +1,98 @@
...
+            WTWt_v = WTWt_v + noise_diag * v.squeeze()
+
+            return WTWt_v
+        self.save_for_backward(c, y, noise_diag)
+
+        mat_inv_y = LinearCG().solve(mv_closure, y)
+        # Inverse quad form
+        res = mat_inv_y.dot(y)
+        # Log determinant
+        ld, tr_inv = SLQLogDet(num_random_probes=10).logdet(mv_closure, len(y))
```
gpleiss: Some ideas to refactor SLQ stuff

gpytorch/utils/__init__.py
```diff
@@ -41,4 +46,118 @@ def wrapped_function(*args, **kwargs):
...
+        - T[i,j], where T is the Toeplitz matrix specified by c and r.
+    """
+    index = i - j
+    if index < 0:
+        return r[abs(index)]
+    else:
+        return c[index]
+
+
+def sym_toeplitz(c):
```
gpleiss: It's probably time to break off several util files.

gpleiss:
Let's make all these comments into trello tasks. But let's get this beast in.

PR #5, opened 2017-08-05
Rename classes

gpleiss:
- Change `ObservationModel` to `gpytorch.GPModel`
- Change `Distribution` to `gpytorch.Module`

## Notes:
- I'm think we should favor importing from gpytorch/torch base packages rather than importing classes direcly. E.g.

```python
import gpytorch
from torch import nn

...

gpytorch.Module
nn.Module
nn.Parameter
```

- Also, the notebooks no longer work, but I'll fix that later.

### DIFF
--- a/gpytorch/__init__.py
+++ b/gpytorch/__init__.py
@@ -3,7 +3,7 @@ from torch.autograd import Variable
 from .lazy import ToeplitzLazyVariable
 from .random_variables import GaussianRandomVariable
 from .module import Module
-from .observation_model import ObservationModel
+from .gp_model import GPModel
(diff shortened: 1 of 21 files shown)


jrg365:
Since nothing is super pressing and I do a decent amount of my testing out of the notebooks, do you think it would be better to merge this in now or after the notebooks are fixed?

gpleiss:
I'm going to keep making update pretty continuously today, so I think it's better to merge now. After the next change I'll update the notebooks.

jrg365:
Alright, no problem.

PR #6, opened 2017-08-11
Parameters

gpleiss:
## Removed parameter groups

- Parameters are now registered with `module.register_parameter(param, bounds=(lower_bound, upper_bound))`
  - `param` is a nn.Parameter
  - You cannot set a parameter directly with `__setattr__`
- Bounds are handled at a module level. The module ensures the parameter never exceeds the bounds.
- Parameter groups no longer exist

### Methods

- `module.parameters()` and `module.named_parameters()` functions in a way similar to `nn.Modules`
- `module.parameter_bounds()` and `module.named_parameter_bounds()` get the bounds
- `module.bound_for(name)` gets the bound for a particular parameter

## Train/eval mode

- Models when they come out of inference are in "eval" mode
- Models should be set to "train" mode for optimization

## Optimization

- Optimization now occurs outside the inference loop. A user must implement it manually
- All the tests use Adam by default (not LBFGS)
- Removing our custom version of LBFGS, since it causes nothing but trouble
  - (Also it was okay when optimization was under the hood. Now that its not, it should be handled separately.)

## Other changes

- Renamed `posterior_model.observation_model` to `posterior_model.prior_model`
- Renamed 'gp_model.observation_model` to `gp_model.likelihood`


jrg365:
What do you think an implementation of hyperparameter priors will look like? Should those be handled entirely user side, or is there any merit to registering them with the `GPModel`? I agree the obviously choice is to have them handled user side (e.g., just add the log prior for each parameter to the loss). 

The argument for the latter is that bounded parameters could also be handled by having MAP be the default inference choice with something like a uniform prior on [a,b] for each parameter representing the bounds. This has two advantages over bounds as hard constraints:

1. Having both priors and bound constraints could silently lead to strange behavior with poorly set bounds. For example, imagine I use a N(5,1) prior for a hyperparameter, but set the bounds to [-2,2] or something. Then, the parameters are hard constrained to lie on the tail of the prior. 
2. With something like the SmoothBoxPrior from GPML, users could control how hard or soft they want the bounds to be -- maybe in certain cases they should be hard constraints, or maybe there is some flexibility in others.

Either way, let's merge this in for now and just make a Trello card to either (1) add an example notebook that uses priors in the external optimization loop, or (2) add hyperparameter priors to the GPModel.

PR #7, opened 2017-08-16
Param encapsulation

gpleiss:
Modules now define their own parameters.

For example, the RBFKernel is now initialized as follows:

```python
covar_module = RBFKernel(log_lengthscale_bounds=(-3, 3))
```

Alternatively, the bounds can be set using `covar_module.set_bounds(log_lengthscale=(-3, 3))`.
The parameter values can be set using `covar_module.initialize(log_lengthscale=1.1)`.


jrg365:
I actually have a couple concerns about this pull request reducing flexibility around parameters.

1. I don't think this PR actually fully solves its problem. As it stands, users would have to mix declaring parameters in their `GPModel` like they used to anyways and declaring them in `Modules`. For example, the `RBFKernel` module as written does not have a `log_outputscale` parameter. If you want to do regression on a task where the label can be outside the range [-1,1], a `log_outputscale` is required. So before the PR I'd do something like
```python
self.covar_module = RBFKernel(log_lengthscale_bounds=(-3,3))
self.register_parameter('log_outputscale', nn.Parameter(...), bounds=...)
covar_x = self.log_outputscale * self.covar_module.forward(x)
```
Basically every stationary kernel is going to need an outputscale. If I do that, then I run in to the problem where, if I want a product of kernels, one `Kernel` is going to be responsible for the outputscale of all `Kernels` (see point 2). I also have an additional problem where I have a useless parameter that I have to set trivial bounds for like `(1,1)` if I specifically do not want an output scale. 

2.  Before this PR, tying parameters together was easy. For example:
```python
self.register_parameter('my_first_param', nn.Parameter(torch.zeros(1, 1)), bounds=my_bounds)
self.register_parameter('my_second_param1', nn.Parameter(torch.zeros(1, 1)), bounds=my_bounds)
self.register_parameter('my_second_param2', nn.Parameter(torch.zeros(1, 1)), bounds=my_bounds)
covar1 = covar_module_one.forward(x, my_first_param=self.my_first_param, my_second_param=self.my_second_param1)
covar2 = covar_module_two.forward(x, my_first_param=self.my_first_param, my_second_param=self.my_second_param2)
covar_x = covar_1 + covar_2
```
After the pull request, I guess I'd have to do something like:
```python
covar_module_two.my_first_param = covar_module_one.my_first_param
```
I do not think this is elegant. If I have many covariance modules (for example, I have additive structure on a many dimensional problem), I'm basically saying that one of the `Kernels` is responsible for the (e.g.) `log_lengthscale` of all other covariance modules. This breaks the intended convention of this PR of 'An `RBFKernel` has a `log_lengthscale`,' since there is one `log_lengthscale` across all `RBFKernels`. That is to say, the `GPModel` has a log_lengthscale, which is how it is handled now. 

I do not think parameter tying is a minor use case: any time I take the product of two kernels, the output scales of those kernels should be tied together, as the product of two output scales is the same as having just one.

3. I do not know if this is an actual use case, but before this PR doing something like "my `spectral mixture weights` are actually the output of some other model like a 3 layer neural network" would theoretically be possible. After the PR, `RBFKernel` insists that `log_lengthscale` is a single parameter.

jrg365:
Actually, we can ignore point 2 above which was my biggest concern. I've been looking a bit in to how PyTorch handles parameter tying, and they seem to prefer something like 
```python
module_one.param = module_two.param
```
So in that sense this PR follows PyTorch convention.

gpleiss:
W.R.T. point 1 - we can make a kernel scaling module. Alternatively, we could even use some built-in pytorch modules to accomplish this. This would follow pytorch convention.

W.R.T. point 3 - would the neural net output the kernel parameters, or the new feature space?

jrg365:
Point 1 - Sure, having a separate module that scales kernels makes sense to me.

Point 3- I was just pointing out that having parameters be in modules limits us to having each parameter be a single Variable that is set by optimization. This pattern is broken, e.g., if a user wants to implement support for heteroskedastic noise or input varying lengthscales. These are things some people care about.

PR #11, opened 2017-08-18
KISS-GP Classification

jrg365:
This PR implements KISS-GP classification. The primary changes are (1) implementing ToeplitzMVNKLDivergence, (2) changing the classification code to use gpytorch function calls so that LazyVariables can define their own versions of the various required functions, and (3) pulling out the internals of prediction in to a _variational_predict function that serves a similar purpose to _exact_predict now. Additional stability improvements were made to classification in general.

@wrh14 is working on a rewrite of MVNKLDivergence for standard (non KISS-GP) classification that incorporates some of the stability and speed improvements, but that is orthogonal to the goal of implementing KISS-GP classification, and this PR is long enough.


gpytorch/__init__.py
```diff
@@ -119,7 +119,41 @@ def mvn_kl_divergence(mean_1, chol_covar_1, mean_2, covar_2):
...
-    return MVNKLDivergence()(mean_1, chol_covar_1, mean_2, covar_2)
+    if isinstance(covar_2, LazyVariable):
+        return covar_2.mvn_kl_divergence(mean_1, chol_covar_1, mean_2)
+    else:
+        return MVNKLDivergence()(mean_1, chol_covar_1, mean_2, covar_2)
+
+
+def add_jitter(covar):
+    if isinstance(covar, LazyVariable):
+        covar.jitter_()
```
gpleiss: Naming should be consistent between `add_jitter` and `jitter_`

gpytorch/__init__.py
```diff
@@ -119,7 +119,41 @@ def mvn_kl_divergence(mean_1, chol_covar_1, mean_2, covar_2):
...
+    else:
+        covar.data.add_(1e-3 * torch.eye(len(covar)))
+
+
+def initialize_variational_parameters(var_output):
```
gpleiss: Does this belong in `gpytorch.__init__`? This file is getting super bloated, and this doesn't seem like a top-level function.

gpytorch/__init__.py
```diff
@@ -193,3 +227,43 @@ def test_train_mul_closure(v):
         test_test_covar = test_test_covar.sub(test_test_covar_correction)
 
     return GaussianRandomVariable(test_mean, test_test_covar), alpha
+
+
+def _variational_predict(n, full_covar, variational_mean, chol_variational_covar, alpha=None):
```
gpleiss: I really don't like this here. I'm about to refactor away `_exact_predict` anyways.

gpytorch/inference/inference.py
```diff
@@ -34,24 +35,16 @@ def restore_from_state_dict(self, state_dict):
...
         train_y = Variable(state_dict[train_y_keys[0]], volatile=True)
-        posterior_model = self.run(train_x, train_y, inducing_points=None)
+        posterior_model = self.run(train_x, train_y)
 
         # Copy over parameters to posterior model
         posterior_model.load_state_dict(state_dict)
         return posterior_model
 
-    def run_(self, train_x, train_y, inducing_points=None, max_inference_steps=20, **kwargs):
+    def run_(self, train_x, train_y, max_inference_steps=20, **kwargs):
```
gpleiss: Why is `max_inference_steps` still there? I removed it from an earlier commit

gpytorch/__init__.py
```diff
@@ -193,3 +227,43 @@ def test_train_mul_closure(v):
...
+
+        # test_covar = K_{mn}K_{nn}^{-1}(S - K_{nn})
+        test_covar = torch.mm(test_train_covar, invmm(train_train_covar, test_covar))
+
+        # right_factor = K_{nn}^{-1}K_{nm}
+        right_factor = invmm(train_train_covar, train_test_covar)
+
+        # test_covar = K_{mn}K_{nn}^{-1}(S - K_{nn})K_{nn}^{-1}K_{nm}
+        test_covar = full_covar[n:, n:] + test_covar.mm(right_factor)
+        GRV = GaussianRandomVariable(test_mean, test_covar)
```
gpleiss: Another name besides `GRV`.

gpytorch/math/functions/lazy_toeplitz/toeplitz_mvn_kl_divergence.py
```diff
@@ -0,0 +1,126 @@
...
+            grad_toeplitz_column = grad_toeplitz_column / self.num_samples
+
+            grad_toeplitz_column = grad_toeplitz_column - sym_toeplitz_derivative_quadratic_form(mat_inv_y.squeeze(),
+                                                                                                 mat_inv_y.squeeze())
+            grad_toeplitz_column.mul_(grad_output_value)
+
+        return grad_mu_diff, grad_cholesky_factor, grad_toeplitz_column
+
+
+class ToeplitzMVNKLDivergence(Function):
```
gpleiss: This class might not be necessary with the latest push to master.
I got rid of the MVNKLDivergence function.

test/toeplitz/test_apply_interpolation.py
```diff
@@ -0,0 +1,23 @@
+import torch
+from gpytorch.kernels import RBFKernel, GridInterpolationKernel
+from gpytorch import utils
+from gpytorch.utils.toeplitz import index_coef_to_sparse
+from torch.autograd import Variable
+
+
+def test_apply_interpolation():
```
gpleiss: This should belong in the new `test/utils/toeplitz_test.py` file.
The convention for test files is: replace `gpytorch` with `test`, and replace `.py` with `_test.py`.
I'm not even sure pytest is picking up on this file, since it is picky about naming conventions.

gpytorch/__init__.py
```diff
@@ -119,7 +119,41 @@ def mvn_kl_divergence(mean_1, chol_covar_1, mean_2, covar_2):
 
     For more information, see the MVNKLDivergence function documentation.
     """
-    return MVNKLDivergence()(mean_1, chol_covar_1, mean_2, covar_2)
+    if isinstance(covar_2, LazyVariable):
+        return covar_2.mvn_kl_divergence(mean_1, chol_covar_1, mean_2)
```
gpleiss: I understand why the arguments are in this order, but it's weird that `covar_2` has the `mvn_kl_divergence` function, since normally the first argument is the one that makes the function call.

If there's not a better way, we can keep it this way, but this is a bit of an antipattern.

gpytorch/inference/posterior_models/variational_gp_posterior.py
```diff
@@ -56,32 +59,23 @@ def forward(self, *inputs, **params):
         gaussian_rv_output = self.prior_model.forward(*full_inputs, **params)
         full_mean, full_covar = gaussian_rv_output.representation()
 
-        if not has_posterior:
-            test_mean = full_mean
-            test_covar = full_covar
+        if not self.training:
+            if hasattr(self, 'alpha') and self.alpha is not None:
+                GRV, alpha = gpytorch._variational_predict(self.num_inducing,
```
gpleiss: Please rename `GRV`.

gpytorch/lazy/lazy_variable.py
```diff
@@ -1,9 +1,87 @@
...
+        plus the diagonal noise term (added using add_diag above) is stored as this lazy variable
+
+        Args:
+            - target (vector n) - training label vector to be used in the marginal log likelihood calculation.
+        Returns:
+            - scalar - The GP marginal log likelihood where (K+\sigma^{2}I) is represented by this LazyVariable.
+        """
+        raise NotImplementedError
+
+    def mvn_kl_divergence(self, mean_1, chol_covar_1, mean_2):
```
gpleiss: This might not be necessary with the latest push into master.
See the new `mvn_kl_divergence` in `gpytorch.__init__`

gpytorch/lazy/lazy_variable.py
```diff
@@ -1,9 +1,87 @@
...
+        """
+        Adds a diagonal component to this lazy variable in a potentially lazy fashion.
+        That is, if this lazy variable represents some matrix A, add_diag should return
+        a LazyVariable of the same type that represents A+dI.
+
+        Args:
+            - diag (scalar) - diagonal component to add.
+        Returns:
+            - LazyVariable representing this current LazyVariable plus diag*I.
+        """
```
gpleiss: If you're just adding docs like this (i.e. not specific to this PR), just commit it directly into master. It's kind of distracting for me to go thorough this stuff in a review, and I end up repeating work like this because I don't know you're doing it.

gpytorch/lazy/toeplitz_lazy_variable.py
```diff
@@ -2,17 +2,22 @@
...
 from gpytorch.utils import toeplitz
 from .lazy_variable import LazyVariable
-from gpytorch.math.functions.lazy_toeplitz import InterpolatedToeplitzGPMarginalLogLikelihood
+from gpytorch.math.functions.lazy_toeplitz import InterpolatedToeplitzGPMarginalLogLikelihood, \
+    ToeplitzMVNKLDivergence
 
 
 class ToeplitzLazyVariable(LazyVariable):
-    def __init__(self, c, r, J_left=None, C_left=None, J_right=None, C_right=None, added_diag=None):
+    def __init__(self, c, r, inducing_points=None,
```
gpleiss: I'm not a fan of inducing points living here. We should talk about another way of doing this.

gpytorch/__init__.py
```diff
@@ -193,3 +227,43 @@ def test_train_mul_closure(v):
...
+def _variational_predict(n, full_covar, variational_mean, chol_variational_covar, alpha=None):
+    test_train_covar = full_covar[n:, :n]
+    train_test_covar = full_covar[:n, n:]
+    if isinstance(test_train_covar, LazyVariable):
+        test_covar = chol_variational_covar.t().mm(chol_variational_covar)
+        add_jitter(test_covar)
+        test_covar = train_test_covar.apply_right_interpolation(test_covar)
+        test_covar = test_train_covar.apply_left_interpolation(test_covar)
+
+        test_mean = test_train_covar.apply_left_interpolation(variational_mean.unsqueeze(1)).squeeze()
```
gpleiss: We can wrap this in a Function. I was about to refactor the `_exact` stuff. Let's talk about it later.

gpytorch/lazy/lazy_variable.py
```diff
@@ -1,9 +1,87 @@
...
+
+    def jitter_(self):
+        """
+        Adds jitter (i.e., a small diagonal component) to the matrix this LazyVariable represents.
+        This could potentially be implemented as a no-op, however this could lead to numerical instabilities,
+        so this should only be done at the user's risk.
+        """
+        raise NotImplementedError
+
+    def apply_left_interpolation(self, M):
```
gpleiss: This name is confusing, and perhaps overly-specific.

We should rename `apply_right_interpolation` to `dsmm`, which is the more generic form, and matches the `dsmm` function I already added in my latest refactor.

`apply_left_interpolation` can be done with `dsmm` and a transpose.

gpytorch/math/functions/lazy_toeplitz/interpolated_toeplitz_gp_marginal_log_likelihood.py
```diff
@@ -20,31 +20,31 @@ def __init__(self, W_left, W_right, num_samples=10):
 
         self.num_samples = num_samples
 
-    def forward(self, c, y, noise_diag):
+    def forward(self, toeplitz_column, labels, noise_diag):
```
gpleiss: Better names :)

gpytorch/math/functions/lazy_toeplitz/toeplitz_mvn_kl_divergence.py
jrg365: The new MVNKLDivergence is faster and more stable with the merged forward and custom backward function, as it avoids explicitly computing K^{1}M and merges the trace derivative with the log det derivative.

jrg365: My intention was that the new MVNKLDivergence get merged directly in to master, not tacked on to this branch. We should revert the removal of MVNKLDivergence and cherry pick Ruihan's commit in to master.

gpleiss: I like this new method of computing things. What I'm saying is that we don't need `MVNKLDivergence` Function objects. It can be written as a lower-case function, since it's doing operations on variables. See the new lower-case function in master as an example.

jrg365: Right, but the new MVNKLDivergence does operate on tensors via `_TraceLogDetQuadForm` . Do you just want to make that a nonprivate function that gpytorch.mvn_kl_divergence uses? The reason I left it as MVNKLDivergence is because that is a more useful `Function` over all than the very specific `TraceLogDetQuadForm`.

PR #21, opened 2017-09-03
Inv_matmul, Matmul and ExactGPMLL

wrh14:
@gpleiss @jrg365 
This new branch is for implementations of `Inv_matmul`,  `Matmul` and `ExactGPMLL` in `function_factory.py`.

Also I did some small modifications. 
1. I pass the `inducing_covar` rather than `train_covar` into  `gpytorch.mvn_kl_divergence` in `marginal_log_likelihood` in `variational_gp_posterior`, which should be more reasonable according to the definition.
2. I rewrite `_derivative_quadratic_form_factory` for LazyVariable, which was only for Toeplitz matrix before and is for total LazyVariable now (Add the consideration for interpolated part).

### DIFF
--- a/gpytorch/utils/function_factory.py
+++ b/gpytorch/utils/function_factory.py
@@ -9,10 +9,6 @@ def _default_matmul_closure_factor(mat):
     return mat
 
 
-def _default_grad_fn(grad_output, rhs):
-    return rhs.t().matmul(grad_output),
-
-
 def _default_derivative_quadratic_form_factory(mat):
     return lambda left_vector, right_vector: (left_vector.ger(right_vector),)
 
--- a/gpytorch/utils/function_factory.py
+++ b/gpytorch/utils/function_factory.py
@@ -33,8 +33,6 @@ def inv_matmul_factory(matmul_closure_factory=_default_matmul_closure_factor,
             self.args = args
 
         def forward(self, *args):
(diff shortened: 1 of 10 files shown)


gpytorch/utils/function_factory.py
```diff
@@ -26,48 +26,59 @@ def closure(matmul_closure, tr_inv, mat_inv_labels, labels, num_samples):
...
 
-def inv_matmul_factory(matmul_closure_factory=_default_matmul_closure_factor, grad_fn=_default_grad_fn):
+def inv_matmul_factory(matmul_closure_factory=_default_matmul_closure_factor,
+                       derivative_quadratic_form_factory=_default_derivative_quadratic_form_factory):
     class InvMatmul(Function):
         def __init__(self, *args):
             self.args = args
 
         def forward(self, *args):
+            if derivative_quadratic_form_factory is None:
```
jrg365: This check should probably be in backward, so that we don't throw an error about missing derivatives unless the user actually calls backward.

jrg365:
It seems like a very large amount of GPyTorch's internals is now housed in `function_factory.py`. Do you guys think it might be better to split that out a bit (e.g., `functions/` -> `function_factories/` or something).

gpleiss:
@jrg365 that's probably a good idea! At the same time, I don't see us adding any more function factories for the time being.

PR #24, opened 2017-09-09
Make the trace computation be a batch operation

wrh14:
@gpleiss @jrg365 
In this branch, I modify those computations of trace. In detail, I write them into a batch operation rather than `for` loop. The training speed is improved a lot.

### DIFF
--- a/gpytorch/utils/function_factory.py
+++ b/gpytorch/utils/function_factory.py
@@ -20,7 +20,7 @@ def _default_derivative_quadratic_form_factory(mat):
             right_factor = right_vectors
         left_factor.unsqueeze_(2)
         right_factor.unsqueeze_(1)
-        res =  (left_factor * right_factor).sum(dim=0).squeeze_()
+        res = (left_factor * right_factor).sum(dim=0).squeeze_()
         return res,
     return closure
 
@@ -206,7 +206,6 @@ def trace_logdet_quad_form_factory(matmul_closure_factory=_default_matmul_closur
                                                                                              right_vectors_1))
                     quad_derivative_2 = list(derivative_quadratic_form_factory(*covar2_args)(left_vectors_2,
                                                                                              right_vectors_2))
-
 
                     for i in range(len(covar2_args)):
                         if grad_covar2_args[i] is not None:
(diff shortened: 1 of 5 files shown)


jrg365:
Awesome! This is a great change. I wonder if we could write a batch version of Toeplitz matmul that would let us get rid of the new for loop in the derivative quadratic form function, too.

gpytorch/utils/function_factory.py
```diff
@@ -192,19 +181,12 @@ def backward(self, grad_output):
...
-                    for z in sample_matrix:
-                        quad_derivative = deriv_quad_form_closure(z)
-                        for i in range(len(covar2_args)):
-                            if grad_covar2_args[i] is not None:
-                                grad_covar2_args[i].add_(quad_derivative[i])
+                    sample_matrix = torch.sign(torch.randn(len(mu_diff), gpytorch.functions.num_trace_samples))
+                    rhs_vectors = chol_covar1.t().mm(chol_covar1.mm(sample_matrix))
+                    I_minus_Tinv_M_vectors = sample_matrix - LinearCG().solve(covar2_matmul_closure, rhs_vectors)
+                    Tinv_vectors = LinearCG().solve(covar2_matmul_closure, sample_matrix)
+                    grad_covar2_args = list(derivative_quadratic_form_factory(*covar2_args)(Tinv_vectors.t(),
```
jrg365: Not for this PR, but at some point we should go through and evaluate whether any of the transpose operations we do can be removed by storing matrices better as their transposes already. Doing operations with the transpose of a matrix can be quite the slowdown due to poor memory access patterns.

jrg365: Alternatively if we must transpose a matrix, it might be worth making some of them contiguous before performing a lot of operations with them.

wrh14: I think so. I am going through those transpose operations to see if anything can be simplified. I will also try to find if it possible to write a batch version for `toeplitz_matmul`.

gpleiss:
This is awesome! I agree with @jrg365 - at some point let's try to get rid of the last for loops in Toeplitz. But this is a huge change!

PR #25, opened 2017-09-09
Refactor: remove inference module

gpleiss:
## Changes to `gpytorch.Module`
- Modules now have a `condition(*data)` method, which allows the model to
  enter posterior mode
- Models have a property `posterior`, which says 1) if the model
  has conditioned on data and 2) if the model is not in training mode

## Changes to `gpytorch.GPModel`
- The ExactGPPosterior and VariationalGPPoseterior classes were absorbed
  into GPModel
- GPModel has an `exact_inference` property, which basically checks if
  there is a GaussianLikelihood or not
- GPModel caches the `alpha` vector, in a way that allows it to be saved
- Conditioning on data clears the `alpha` vector and the variational
  parameters

## Other random changes
- The internals of parameter bounds has changed a bit to allow for
  parameter sizes that change on-the-fly.
- `gpytorch.Modules` implement their own version of `load_state_dict`
  which automatically resizes parameters as necessary. This is necessary
  so that the alpha cache and variational parameters can be set
  on-the-fly.

@jrg365 @wrh14 the only major changes are to `gpytorch/module.py` and `gpytorch/gp_model.py`. Everything else is removing old files and updating tests.

I'll push a commit in a bit that updates the example notebooks.


gpleiss:
Okay @jrg365 @wrh14 - examples updated. This PR is ready to go!

PR #27, opened 2017-09-14
Simpler Initialization  for Spectral Mixture Kernel

wrh14:
@gpleiss  @jrg365 The simpler initialization has been implemented. This implementation is according to Andrew Wilson's matlab code `initSMhypers.m` https://people.orie.cornell.edu/andrew/code/initSMhypers.m. However, the stability becomes worse in `test/examples/spectral_mixture_gp_regression_test.py`. Also, the implementation of log_mixture_scales in `initSMhypers.m` is not consistent to truncated Gaussian distribution with mean proportional to the max range of the data, which is described in his page https://people.orie.cornell.edu/andrew/code/.

We may consider whether to use this initialization.

### DIFF
--- a/gpytorch/kernels/spectral_mixture_kernel.py
+++ b/gpytorch/kernels/spectral_mixture_kernel.py
@@ -2,18 +2,39 @@ import math
 import torch
 from torch import nn
 from .kernel import Kernel
+from torch.autograd import Variable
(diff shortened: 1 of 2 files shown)


gpleiss:
We're going to change how initialization works post-alpha.

PR #31, opened 2017-09-19
MulLazyVariable

wrh14:
The functions `SumLazyVariable` has are all implemented in MulLazyVariable except `add_diag()` and `add_jitter()`. If they are not needed, then this branch is ready for merge.

### DIFF
--- a/gpytorch/lazy/mul_lazy_variable.py
+++ b/gpytorch/lazy/mul_lazy_variable.py
@@ -1,0 +1,216 @@
+from .lazy_variable import LazyVariable
+from ..posterior import DefaultPosteriorStrategy
+import torch
+from torch.autograd import Variable
+from ..utils.trace import trace_components
+
+
+class MulLazyVariable(LazyVariable):
+    def __init__(self, *lazy_vars, **kwargs):
+        if not all([isinstance(lazy_var, LazyVariable) for lazy_var in lazy_vars]):
+            raise RuntimeError('All arguments of a MulLazyVariable should be lazy variables')
+        self.lazy_vars = lazy_vars
+        if 'added_diag' in kwargs:
+            self.added_diag = kwargs['added_diag']
+        else:
+            self.added_diag = None
+
+    def _matmul_closure_factory(self, *args):
+        len_mul_args = 0
+        for lazy_var in self.lazy_vars:
+            len_repr = len(lazy_var.representation())
+            len_mul_args = len_mul_args + len_repr
+        if len(args) > len_mul_args:
+            added_diag = args[-1]
+            args = args[:-1]
+        else:
+            added_diag = None
+
+        if len(self.lazy_vars) == 1:
+            def closure(rhs_mat):
+                if_vector = False
+                if rhs_mat.ndimension() == 1:
+                    rhs_mat = rhs_mat.unsqueeze(1)
+                    if_vector = True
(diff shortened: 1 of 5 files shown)


gpytorch/lazy/toeplitz_lazy_variable.py
```diff
@@ -155,18 +156,18 @@ def mul(self, other):
...
             if len(self.c) != len(other.c):
-                raise RuntimeError('Can only multiply two ToeplitzLazyVariables if they correspond to the same grid.')
+                return MulLazyVariable(self, other)
             if self.J_left is not None:
                 if other.J_left is None:
-                    raise RuntimeError('Cannot multiply interpolated ToeplitzLazyVariables by non-interpolated ones.')
+                    return MulLazyVariable(self, other)
                 if not (approx_equal(self.C_left, other.C_left) and approx_equal(self.C_right, other.C_right)):
-                    raise RuntimeError('Cannot multiply two ToeplitzLazyVariables with different',
-                                       'left interpolation matrices.')
```
jrg365: I think with `MulLazyVariable`, the check we want is simply if the sizes are the same, since that's the only requirement for elementwise multiplication to work. It should be on the user's head if they elementwise two `ToeplitzLazyVariables` they didn't intend to. In other words, if `other` is a `ToeplitzLazyVariable` of the same size we can return a `MulLazyVariable`.

jrg365: For now we can leave it throwing an error if they both have added diagonal components -- the logic of FOILing (A+p*I)*(B+q*I) will be much simpler once we implement added diagonal components as `DiagLazyVariable`

gpytorch/lazy/mul_lazy_variable.py
```diff
@@ -0,0 +1,157 @@
...
+from ..posterior import DefaultPosteriorStrategy
+import gpytorch
+import torch
+from torch.autograd import Variable
+
+
+class MulLazyVariable(LazyVariable):
+    def __init__(self, *lazy_vars):
+        if not all([isinstance(lazy_var, LazyVariable) for lazy_var in lazy_vars]):
+            raise RuntimeError('All arguments of a SumLazyVariable should be lazy variables')
```
jrg365: Should be "All arguments of a MulLazyVariable"

jrg365: Let's add a Trello card for support of add and mul of LazyVariables with standard Torch Variables.

gpytorch/lazy/mul_lazy_variable.py
```diff
@@ -0,0 +1,157 @@
...
+    def add_diag(self, diag):
+        raise NotImplementedError
+
+    def add_jitter(self):
+        return NotImplementedError
+
+    def diag(self):
+        res = Variable(torch.ones(self.size()[0]))
+        for lazy_var in self.lazy_vars:
+            res = res * lazy_vars.diag()
```
jrg365: This should probably be self.lazy.vars

wrh14:
The improved `matmul()` for >=3 lazy variables is under working.

wrh14:
@jrg365 @gpleiss 
The trace has been modified!

jrg365:
Great! Let's merge this and I'll take a pass at simplifying the logic of `ToeplitzLazyVariable.mul`

PR #32, opened 2017-09-19
Stochatic trace refactor

gpleiss:
New util: `gpytorch.utils.trace_components`.

Given a matrix `A^T B` that we want to compute the trace for, this function returns `A'` and `B'` such that `tr(A^T B) = sum(A * B) = sum(A' * B')`.

This function can also be used for efficiently computing the diagonal elements of a matrix, as well as the left and right vectors for derivative quad form (which has been refactored).

@wrh14 @jrg365

### DIFF
--- a/gpytorch/utils/trace.py
+++ b/gpytorch/utils/trace.py
@@ -1,0 +1,134 @@
+import torch
+import gpytorch
+import math
+from torch.autograd import Variable
+
+
+def _identity(x):
+    return x
+
+
+def trace_components(left_matmul_closure, right_matmul_closure, size=None, num_samples=None,
+                     estimator_type='mub', tensor_cls=None, use_vars=False):
+    """
+    Create components for stochastic trace estimation
+    Given two matrices `A` and `B` (represented by closures that define matrix-multiplication
+    operations), returns two matrices `A'` and `B'` such that
+
+    ```
+    trace( A^T B ) = sum(A' * B') (elementwise operation)
+    ```
+
+    If A and B are regular matrices, then this function simply returns A and B,
+    since `trace( A^T B ) = sum(A * B)`
+
+    However, if `A` and `B` are more efficiently represented by matrix multiplication closures
+    (as is the case if A is a  Toeplitz matrices, if A is the inverse of a PSD matrix, etc.)
+    then this will be far more efficient, since `A' and `B'` will have significantly lower dimensionality.
+
+    Args:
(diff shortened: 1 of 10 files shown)


gpleiss:
Oh and there's some bonus cuda compatibility.

jrg365:
Everything looks great, except for one issue Ruihan and I discussed earlier: the space complexity for the trace function as written is O(ns), which could be as bad as O(n^2) if we use I (i.e. we compute the exact trace). This is solvable by batching the vectors (for example using 10 rows of I at a time).

What do you think, Geoff?

gpleiss:
I'm not sure I understand. What do you mean by batching? And what's the problem we're trying to solve?

gpytorch/utils/trace.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if tensor_cls is None:
+        tensor_cls = torch.Tensor
+
+    # Return A and B if we're using deterministic mode
+    if not gpytorch.functions.fastest or size < num_samples:
+        eye = tensor_cls(size).fill_(1).diag()
+        if use_vars:
+            eye = Variable(eye)
+        return left_matmul_closure(eye), right_matmul_closure(eye)
```
jrg365: This line right here is the problem (or an example of it). If we need to compute Tr(K^{-1}), left_matmul_closure(eye) will return an n-by-n matrix, which destroys the linear space complexity of KISS-GP. This can be solved by doing it in large pieces instead of all at once, e.g. estimating the trace with left_matmul_closure(eye[:, :20]), left_matmul_closure(eye[:, 20:40]), etc by performing the final multiply-and-sum parts one at a time.

jrg365: Or in general with the stochastic version the space complexity is O(ns). In principle we can just average over 10-20 probe vectors, then average over the next 10-20, and so on, rather than computing with all s at once.

Ruihan and I simply think we should be able to set a batch size so that when n is very large the space requirement doesn't get out of control.

gpleiss: I still don't understand. We'll only enter this loop if the matrix is smaller than 20x20 (or if the user specifies not to do stochastic trace estimation).

gpleiss: And what is the purpose of batching? I think trying to implement a batched version will probably be significantly slower than what's here right now.

wrh14: Our purpose is to make space complexity in our control. The space complexity is O(sn) now and it will be O(n^2) in worst case, where n is the number of training data in exact likelihood in regression task.

PR #35, opened 2017-09-22
Improve cuda compatibility

gpleiss:
- Simple GP classification and Kiss-gp regression (1d) are cuda compatible


jrg365:
This PR is the hero we need.

PR #42, opened 2017-10-02
Lazy variables and random variables can be cuda/cpu'd

gpleiss:
Lazy variables and random variables must now call `super(<Class>, self).__init__(*args, **kwargs)`, but this allows us easily to construct cpu and cuda versions of these objects, using the same interface as for Variables and Tensors.


gpleiss:
Oops debugging don't review yet...

PR #44, opened 2017-10-30
Fast predictive variances

gpleiss:
Predictive variances are now handled by a lazy variable. There are two modes:

- `gpytorch.functions.fast_pred_var = True` (default) - uses Lanczos decomposition to compute extremely fast predictive variances. Good when the number of training points and the number of test points are both large
- `gpytorch.functions.fast_pred_var = False` - explicitly evaluates the test/train covariance matrix. Still much faster than before though!

Additional changes:
- `evaluate()` is now a default lazy variable function. It multiplies the lazy variable with the identity matrix, but does it in a smart way now. It constructs an identity matrix that is the size of the smaller dimension. This way, if you have lots of test points and few train points (or vice versa), evaluation will not be an expensive operation.

Other note: I found that, for the Lanczos decomposition mode, calling `lanczos_t_mat.inverse().matmul(...)` was much more stable than `gpytorch.inv_matmul(lanczos_t_mat, ...)`.


gpytorch/posterior/posterior_strategy.py
```diff
@@ -26,6 +42,34 @@ def exact_posterior_alpha(self, train_mean, train_y):
...
+        else:
+            train_train_matmul = self.var.data.matmul
+            tensor_cls = type(self.var.data)
+
+        n_train = self.var.size()[0]
+        max_iter = min(n_train, gpytorch.functions.max_lanczos_iterations)
+        lq_object = StochasticLQ(cls=tensor_cls, max_iter=max_iter)
+        init_vector = tensor_cls(n_train, 1).normal_()
+        init_vector /= torch.norm(init_vector, 2, 0)
+        q_mat, t_mat = lq_object.lanczos_batch(train_train_matmul, init_vector)
```
jrg365: We might want to add the eigenvalue binary search for T.

gpytorch/functions/__init__.py
```diff
@@ -12,5 +12,7 @@
 ]
 
 fastest = True
+fast_pred_var = True
```
jrg365: In my opinion, let's make this False by default for a little while until the Lanczos stuff has been tested on kernels other than Matern and RBF.

gpytorch/posterior/posterior_strategy.py
gpleiss: As far as I can tell, this is the same LQ code that we use in mul lazy variable. Since it's not going to be the default option (and hopefully we'll have more stable LQ code soon) should we just merge as is?

gpytorch/functions/__init__.py
gpleiss: Fixed this.

PR #45, opened 2017-11-11
Inducing point refactor

gpleiss:
This makes a number of changes to the interface of GPyTorch. If you are doing simple GP regression (no inducing points), then everything remains the same. Otherwise, there's some changes.

## Summary
- `model.condition` should not be called, until you're ready to compute the predictive variance. This will make things a lot nicer, because you don't actually have to add training data to use a model in prior mode.
- There are new modules -- `InducingPointModule` and `GridInducingPointModules` -- which are required for doing any inducing point work
- Variational inference, even if you're not using inducing points, requires an `InducingPointModule`
- `GridInterpolationKernel` should no longer be used explicitly, but is instead used through an `InducingPointModule`.

All in all, it only requires a few lines of different code for a given model (see the example notebooks). Under the hood though, it's going to make it SO MUCH easier to work with inducing points.

### Change to `condition`
For a while, we were requiring that `condition` be called after creating a model. The reason for this was kind of subtle: for GP classification with no inducing points, we were treating the training data as inducing points. Therefore, we needed a way to know what the training data was before doing anything with the model.

By explicitly forcing the user to define inducing points for GP classification, we no longer need to know about the training data until we're ready to make predictions.

### `InducingPointModule`
The inducing point interface was a little strange. For regression, only the kernel needs to know about inducing points. However, for classification, the model also needs to know about inducing points. This became especially complicated when you wanted a model with multiple sets of inducing points - e.g. different inducing points for each dimension.

`InducingPointModule` is now a atomic module for inducing points. It is a wrapper for a latent GP that uses a set of inducing points. If you want to use inducing points, then your `GPModel` wraps an `InducingPointModule`. If you want to use multiple sets of inducing points (e.g. one per dimension), then your `GPModel` wraps multiple `InducingPointModules`.

The super constructor for an InducingPointModule takes in the inducing points, which is a nicer interface IMO.

### `GridInducingPointModule`
This is a subtype for inducing points on a grid. It automatically wraps any kernel functions in a `GridInterpolationKernel`. Now you no longer have to specifically create a GridInterpolationKernel, which is kind of nice.


jrg365:
This PR I don't have any real comments for -- I'm happy to merge it after the WIP commits get rebased in some fashion. I left a bunch of comments on the `InterpolatedLazyVariable` PR, though.

gpleiss:
Don't merge this yet - I think I have an idea to make GridInducingPointModule much better.

PR #46, opened 2017-11-11
InterpolatedLazyVariable, ConstantMulLazyVariable

gpleiss:
A refactor, which is currently only used for ToeplitzLazyVariables.
Merge this after #45


gpytorch/lazy/toeplitz_lazy_variable.py
```diff
@@ -191,113 +51,78 @@ def repeat(self, *sizes):
...
+        batch_sizes = list(column.size()[:-1])
+        row_iter = column.data.new(column.size(-1)).long()
+        torch.arange(0, self.column.size(-1), out=row_iter)
+
+        left_interp_indices = row_iter[left_index].unsqueeze(-1)
+        right_interp_indices = row_iter[right_index].unsqueeze(-1)
+
+        left_interp_len = len(left_interp_indices)
+        right_interp_len = len(right_interp_indices)
+        for i in xrange(ndimension - 2):
```
jrg365: Should be `range` for Python 3 compatability

gpytorch/lazy/interpolated_lazy_variable.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        Repeat elements of the Variable.
+        Right now it only works to create a batched version of a ToeplitzLazyVariable.
+
+        e.g. `var.repeat(3, 1, 1)` creates a batched version of length 3
+        """
+        if not len(sizes) == 3 and sizes[1] == 1 and sizes[2] == 1:
+            raise RuntimeError('Repeat only works to create a batched version at the moment.')
+        if self.J_left is None:
+            raise RuntimeError('Repeat only works when the toeplitz variable is interpolated.')
```
jrg365: This probably isn't a great error message now that this lives in `InterpolatedLazyVariable`

gpytorch/lazy/interpolated_lazy_variable.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def mul(self, other):
+        """
+        Multiplies the matrix by a constant, or elementwise the matrix by another matrix
+        """
+        if not (isinstance(other, Variable) or isinstance(other, LazyVariable)) or \
+               (isinstance(other, Variable) and other.numel() == 1):
+            base_lazy_variable = ConstantMulLazyVariable(self.base_lazy_variable, other)
+
+            return self.__class__(base_lazy_variable, self.left_interp_indices, self.left_interp_values,
```
jrg365: Does it matter whether we do `W(cT)W'` vs `cWTW'`? I think it would be cleaner to just call `super` here and use `LazyVariable`'s `mul`.

gpytorch/lazy/lazy_variable.py
```diff
@@ -215,11 +217,41 @@ def size(self):
         """
         raise NotImplementedError
 
+    def transpose(self, dim1, dim2):
+        """
```
jrg365: The implementation below makes me think this was supposed to be in `InterpolatedLazyVariable`, not `LazyVariable` (e.g., it expects `self.left_interp_indices` to exist)

gpytorch/lazy/mul_lazy_variable.py
```diff
@@ -15,9 +16,14 @@ def __init__(self, *lazy_vars, **kwargs):
             - max_iter (int) - the maximum iteration in lanczos decomposition in when matmul_mode=approximate
             - num_samples (int) - the samples number when matmul_mode=stochastic
         '''
+        lazy_vars = list(lazy_vars)
+        for i, lazy_var in enumerate(lazy_vars):
+            if not isinstance(lazy_var, LazyVariable):
+                if isinstance(lazy_var, Variable):
+                    lazy_vars[i] = NonLazyVariable(lazy_var)
+                else:
+                    raise RuntimeError('All arguments of a SumLazyVariable should be lazy variables or vairables')
```
jrg365: -> All arguments of a MulLazyVariable

gpytorch/lazy/sum_lazy_variable.py
```diff
@@ -73,7 +72,7 @@ def __add__(self, other):
 
     def __getitem__(self, i):
         sliced_lazy_vars = [lazy_var.__getitem__(i) for lazy_var in self.lazy_vars]
-        return SumLazyVariable(*sliced_lazy_vars)
+        return sum(*sliced_lazy_vars)
```
jrg365: I don't know precisely what sum does internally, but is there a chance that using `sum(lazy_var_list)` is riskier than using `SumLazyVariable(lazy_var_list)`? I notice in __init__ you explicitly fix the case where someone is mixing and matching `Variables` and `LazyVariables` -- will sum do something silly where if a Variable appears before a `LazyVariable` in the list it will do `the_variable + the_lazy_variable`, which could fail because the add method for `Variables` does not expect `LazyVariables`?

gpytorch/lazy/toeplitz_lazy_variable.py
```diff
@@ -191,113 +51,78 @@ def repeat(self, *sizes):
...
+            left_interp_indices.unsqueeze_(0)
+            right_interp_indices.unsqueeze_(0)
+
+        left_interp_indices = left_interp_indices.expand(*(batch_sizes + [left_interp_len, 1]))
+        left_interp_values = left_interp_indices.new(left_interp_indices.size()).fill_(1).float()
+        right_interp_indices = right_interp_indices.expand(*(batch_sizes + [right_interp_len, 1]))
+        right_interp_values = right_interp_indices.new(right_interp_indices.size()).fill_(1).float()
+
+        res = InterpolatedLazyVariable(ToeplitzLazyVariable(column), Variable(left_interp_indices),
```
jrg365: I'm confused -- why is `__getitem__` on a `ToeplitzLazyVariable` returning an `InterpolatedLazyVariable`?

jrg365:
Can we also squash WIP commits before merging?

gpytorch/lazy/toeplitz_lazy_variable.py
jrg365: Do the "interpolation" matrices basically just pick out the correct elements?

gpytorch/lazy/interpolated_lazy_variable.py
gpleiss: The reason for this has to do with the new GridInducingPointModules, but I think after I make some changes to GridInducingPointModule it won't matter.

gpytorch/lazy/sum_lazy_variable.py
gpleiss: This is intentional. `__getitem__` may return LazyVariables, or it may return Variables. If the returned items are LazyVariables, then we'll get a sum lazy variable. Otherwise we'll get a LazyVariable.

gpytorch/lazy/toeplitz_lazy_variable.py
gpleiss: Yeah. My plan was to make this the default `__getitem__` behavior. It's such an annoying method to write for every lazy variable.

gpleiss: I think I forgot to drag it out to LazyVariable though.

gpytorch/lazy/mul_lazy_variable.py
gpleiss: Will change.

gpytorch/lazy/lazy_variable.py
gpleiss: Lol I honestly have no idea what I was thinking here!
I think I was hoping to write a generic transpose method that calls the `_transpose_no_batch` method

gpleiss: I'll figure out what was going on here.

gpytorch/lazy/toeplitz_lazy_variable.py
gpleiss: Ah good catch

gpytorch/lazy/interpolated_lazy_variable.py
gpleiss: Will fix.

gpleiss:
@jrg365 all comments have been addressed.
The only one that didn't change is the `sum` in `SumLazyVariable` - see the specific comment on it.

A few additional changes thrown in this PR (sorry!):
- `gpytorch.functions.use_toeplitz` is by default True. If you set it to False, it explicitly computes the Toeplitz matrix. Gets a speed up on the GPU for <500 inducing points (which is almost always the case for variational inference).
- Variational strategies for `SumLazyVariables`
- Some slight modifications to `GridInducingPointModule`

PR #49, opened 2017-12-19
Refactor variational inference

gpleiss:
Lots of other changes in here, but now there is a significantly better way to handle variational inference.

Modules can call `register_variational_strategy`. This adds a variational strategy to the module. When the variational parameters and/or the prior distribution are updated, then the module can call `update_variational_strategy` (this is typically handled by one of the latent function modules, so users won't have to worry about this unless they want to add their own variational parameters).

When optimizing, in `marginal_log_likelihood`, the module calls `variational_strategies`, which returns all strategies associated with the module (and its children). It can then call `kl_divergence` on all of them and sums them up.

**Note** this assumes that all the registered variational strategies are all independent, or rather that the joint variational distribution factorizes. This was the assumption we were making before, but now it's a bit more explicit.

This will make it much easier to add additional variational distributions.

## Variational strategies
- I'm changing up what these objects do. Now, a variational strategy is defined ONLY for a specific distribution. Right now, we have `MVNVariationalStrategy`.
- Variational strategies only define a `kl_divergence` method.

## Some other changes

- Lazy variables now have `chol_matmul` for multiplying by an (approximate) cholesky decomposition. We'll use this to do sampling for MC integration, rather than what we had previously.
  - Right now, it's only defined for `CholLazyVariables`, but that's okay because that's all we need right now for feature parity. For other lazy variables, we can use Lanczos.
- `GaussianRandomVariables` now do sampling.
- `SumInterpolatedLazyVariable` is now replaced by the more general `SumBatchLazyVariable`, which basically wraps any lazy variable. It basically represents a batch of lazy variables that we eventually sum over.


gpleiss:
Alright, everything is green now and ready to go

Issue #58, opened 2018-01-15
ImportError: cannot import name 'SumVariationalStrategy'


Pinletrot:
Here's a minimal test code. 

```Python
import torch

from gpytorch.random_variables import GaussianRandomVariable
from torch.autograd import Variable

a = GaussianRandomVariable(Variable(torch.rand(3)), Variable(torch.rand(3, 3)))
b = GaussianRandomVariable(Variable(torch.rand(3)), Variable(torch.rand(3, 3)))

print(a + b)
```
The output is
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "~/packages/anaconda3/lib/python3.6/site-packages/gpytorch/random_variables/gaussian_random_variable.py", line 56, in __add__
>     from ..variational import SumVariationalStrategy
> ImportError: cannot import name 'SumVariationalStrategy'

---

The reason may be in the method "gpytorch.random_variables.gaussian_random_variable.__add__", where in line 56, it tries to 

```Python
from ..variational import SumVariationalStrategy
```

However, in "gpytorch/gpytorch/variational/__init__.py", there's no SumVariationalStrategy being imported.

```Python
from .variational_strategy import VariationalStrategy
from .mvn_variational_strategy import MVNVariationalStrategy


__all__ = [
    VariationalStrategy,
    MVNVariationalStrategy,
]
```

jrg365:
Good catch, just pushed a quick fix to master.

jrg365:
I'm actually reopening this -- we don't even have `SumVariationalStrategy` any more, so it is the code in `gaussian_random_variable.py` that needs fixing. Sorry about that -- the import fix seemed to work because I had an old .pyc file lying around.

gpleiss:
Fixed by #59

PR #59, opened 2018-01-16
Remove variational strategy checks from gaussian_random_variable (Fix for #58)

jrg365:
Removes some seemingly old checks in `gaussian_random_variable.py` for variational strategies -- I think this is holdover code from before one of the recent refactors. I don't think `RandomVariables` have variational strategies now.

#58 

@gpleiss


gpleiss:
The linter is failing

jrg365:
deb98404 fixes that

PR #63, opened 2018-01-23
Lazy var roots

gpleiss:
This adds `root_decomposition` to lazy variables.

This allows for fast predictive variances, MVN sampling from arbitrary lazy variables, and eventually a MulLazyVariable refactor.


gpleiss:
Additionally, lazy variables now have to define `t_matmul_closure_factory`, which defines how to do a transpose matrix multiplication. This was always necessary for `matmul`'s derivative. We were just never doing it :p

gpleiss:
Now fast predictive variances use `root_decomposition`.
I also added `root_inv_decomposition`, which does what it sounds like!

Issue #54, opened 2018-01-10
Merge all SKIP changes into master


gpleiss:
Includes refactor of MulLazyVariable

PR #70, opened 2018-02-15
MulLazyVariable refactor

gpleiss:
`MulLazyVariable` has been updated, and now works properly.

There is also a `MultiplicativeGridInterpolationKernel`, which does an extremely fast batch version of `MulLazyVariable`.

This incorporates all the SKIP changes.

@jrg365


gpleiss:
This fixes #54

PR #78, opened 2018-03-08
Switch unit testing over to the unittest 

darbour:
As the title would imply, this switches from pytest to unittest. There are also some small formatting changes. The same number of tests are passing now as before. The only real ugliness is all the file renaming, which was necessary to make tests easily discoverable from unittest.


Balandat:
do you mind squashing the commits into one?

gpleiss:
Looks good to me!

PR #79, opened 2018-03-08
Fixes to gpytorch to be compatible with PyTorch master

colesbury:
This retains compatiblity with PyTorch 0.3.

The file gpytorch/utils/function_factory.py deserves some attention: The
`t_matmul_closure_factory` does not always return the correct shape for
the gradient. This is a bug: gradients should have the exact same shape
as the differnetiated variable. PyTorch is missing some checks for this
behavior, but future versions of PyTorch are likely to raise an
exception if shapes don't exactly match.

(This depends on https://github.com/pytorch/pytorch/pull/5643 in PyTorch)

### DIFF
--- a/build.py
+++ b/build.py
@@ -31,5 +31,6 @@ ffi = create_extension(
     library_dirs=library_dirs,
     with_cuda=with_cuda,
     package=True,
+    extra_compile_args=["-std=c99"],
     relative_to=__file__,
 )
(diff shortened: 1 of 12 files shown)


darbour:
@gpleiss @jrg365 the failed test appears to be transient. We should either set the seed on that test or loosen the bound for failure.

gpleiss:
Thanks @colesbury ! I'll take a look at this PR later tonight.
And yeah... we have a few flaky tests in the repo right now. @jrg365 and I are slowly but surely trying to fix them.

gpleiss:
The changes to function_factory_test and interp_test need to be updated from pytest to unittest style, but I can fix that tomorrow morning.

Thanks @colesbury!

Issue #82, opened 2018-03-13
Add support for ARD


Balandat:
The current Kernels all use a single length scale. Adding support for ARD is straightforward by adding a Parameter for every dimension. E.g. for the RBFKernel this could be something like the following:

```
class RBFKernel(Kernel):

    def __init__(
        self,
        log_lengthscale_bounds=(-10000, 10000),
        eps=1e-5,
        ARD=False,
        ndim=None,
    ):
        super(RBFKernel, self).__init__()
        if ARD and ndim is None:
            raise ValueError("Must provide ndim if ARD=True")
        self.eps = eps
        nparams = ndim if ARD else 1
        if ARD:
            log_lengthscale_bounds = (
                torch.tensor(log_lengthscale_bounds[0]).repeat(1, nparams),
                torch.tensor(log_lengthscale_bounds[1]).repeat(1, nparams),
            )
        self.register_parameter(
            'log_lengthscale',
            nn.Parameter(torch.zeros(1, nparams)),
            bounds=log_lengthscale_bounds,
        )
```

However, we don't want to have to re-implement this for every Kernel, so we should come up with a better abstraction for this.

gpleiss:
@jrg365 just talked about this. We're thinking it would be good to have lengthscales live on the base kernel class:

```python
class Kernel(Module):
    def __init__(self, has_lengthscale=True, ard_num_dims=None, log_lengthscale_bounds=(-10000, 10000)):
        self.ard_num_dims = ard_num_dims
        if has_lengthscale:
            lengthscale_num_dims = 1 if ard_num_dims is None else ard_num_dims
            self.register_parameter('log_lengthscale', torch.nn.Parameter(torch.Tensor(1, 1, ard_num_dims)),
                                    bounds=log_lengthscale_bounds)

    @property
    def lengthscale(self):
        if 'log_lengthscale' in self.named_parameters().keys():
            return self.log_lengthscale.exp()
        else:
            return None


# ...


class RBFKernel(Kernel):
    def __init__(self, **kwargs):
        kwargs['has_lengthscale'] = True
        super(RBFKernel, self).__init__(**kwargs)
        # ...
```

Thoughts?

darbour:
I agree with @gpleiss and @jrg365 about the length-scales living in the base Kernel class. Since we're likely to re-use them for a variety of kernels, it's much cleaner to just have everything use the same base representation.

Balandat:
[...]

PR #83, opened 2018-03-14
Add ability to take sum and product of kernels directly, add linear and periodic kernels, create `active_dims`

darbour:
This has two new features:
* You can now add two kernels using `+` and `*` and an AdditiveKernel or ProductKernel is returned. 
* `active_dims` can be sent via the constructor to subset dimensions, mirroring the capabilities of GPy

Small items: adding linear and periodic kernel, and adding `six` imports to make sure there aren't python 2.x / 3.x discrepancies

### DIFF
--- a/gpytorch/kernels/linear_kernel.py
+++ b/gpytorch/kernels/linear_kernel.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 
 import torch
 from torch import nn
-from ..lazy.matmul_lazy_variable import MatmulLazyVariable
+from ..lazy import MatmulLazyVariable, RootLazyVariable
 from .kernel import Kernel
 
 
@@ -32,9 +32,14 @@ class LinearKernel(Kernel):
         )
 
     def forward(self, x1, x2):
-        prod = MatmulLazyVariable(
-            x1 - self.offset,
-            (x2 - self.offset).transpose(2, 1)
-        )
+        if x1.size() == x2.size() and torch.equal(x1, x2):
+            # Use RootLazyVariable when x1 == x2 for efficiency when composing
+            # with other kernels
+            prod = RootLazyVariable(x1 - self.offset)
+        else:
+            prod = MatmulLazyVariable(
+                x1 - self.offset,
+                (x2 - self.offset).transpose(2, 1)
+            )
 
         return prod + self.variance.expand_as(prod)
--- a/gpytorch/kernels/linear_kernel.py
+++ b/gpytorch/kernels/linear_kernel.py
@@ -1,0 +1,40 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import torch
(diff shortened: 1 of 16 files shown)


jrg365:
Other than that I don't have any issues. @gpleiss what do you think about the active_dims stuff? I don't mind it since it'll be default behavior for every kernel. 

That said, can we fold `active_dims` into something like `**kwargs` that kernels pass to the super constructor? This would prevent future blow up of arguments to kernel constructors that only need to make their way to the base constructor, like the proposed changes for #82.

Balandat:
How many kwargs do we expect a Kernel to have? I'm generally a fan of being explicit with the args, rather than using **kwargs, since it's more transparent what your options are (also I'm lazy and don't want to write extensive docstrings describing which kwargs get used...)

gpytorch/kernels/linear_kernel.py
```diff
@@ -0,0 +1,34 @@
...
+        self.register_parameter(
+            'offset',
+            nn.Parameter(torch.zeros(1, 1)),
+            bounds=offset_bounds,
+        )
+        self.register_paramter(
+            'coefficients',
+            nn.Parameter(torch.zeros(1, num_dimensions)),
+            bounds=linear_coefficient_bounds,
+        )
```
gpleiss: This will probably change once we add the ARD option (since "lengthscale" will probably become a parameter on the Kernel superclass), but it's fine for now!

Also, I think these parameters should probably be 3D tensors rather than matrices (i.e. `torch.zeros(1, 1, num_dimensions)`. The base Kernel class is designed to operate on batches of data (batch_size x n_data x n_dims).

gpytorch/kernels/periodic_kernel.py
```diff
@@ -0,0 +1,40 @@
...
+        eps=1e-5,
+        active_dims=None,
+    ):
+        super(PeriodicKernel, self).__init__(active_dims=active_dims)
+        self.eps = eps
+        self.register_parameter(
+            'log_lengthscale',
+            nn.Parameter(torch.zeros(1, 1)),
+            bounds=log_lengthscale_bounds,
+        )
```
gpleiss: Same thing here: the parameters should be `torch.zeros(1, 1, 1)`.

Balandat:
@gpleiss The kernels that don't support ard/lenghtscales of course won't have the kwargs, and I'll choose the default values in such a way that you don't have to provide anything to the superclass constructor that you don't want explicitly supported. Will send out a pull request as soon as this one goes in.

darbour:
@jrg365 @gpleiss updated to incorporate your comments. Let me know if this looks good.

jrg365:
Made a minor change so that `LinearKernel` uses a `RootLazyVariable` (basically a `MatmulLazyVariable` but restricted to assume `rhs = lhs.t()`) when `x1 == x2`. This will be more efficient for inference in a few scenarios.

Other than that, I'll merge this in when the build finishes!

PR #88, opened 2018-03-20
[Codemod] Future imports to improve py2/py3 compatibility

Balandat:
Ran a script to add the following future imports to each file:
```
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals
```

Also fix some windows-style line breaks (e.g. in bernoulli_likelihood.py)


gpleiss:
Thanks @Balandat !

PR #90, opened 2018-03-21
[Codemod] Clean up test files

Balandat:
Rename redundantly named files, add future imports, fix windows-style line breaks, make empty line after class definition consistent.


gpleiss:
The failures and linting issues should be fixed in master now, so I'm going to go ahead and merge this. Thanks @Balandat !

PR #93, opened 2018-03-22
Fix some more lint issues

Balandat:
DOS lines, loop control


Balandat:
self-review is the best.

PR #95, opened 2018-03-23
Fix test failures on pytorch master

Balandat:
Indexing a zero-dimensional tensor does not return a numeric value anymore, but a tensor. This caused test failures with `assertAlmostEqual` of the kind `TypeError: type Variable doesn't define __round__ method`.

This explicitly converts the troublesome arguments to float, so tests pass on pytorch master.


gpleiss:
Unfortunately this PR now causes the tests to fail on PyTorch 0.3 😰 
I'm think I have a solution that works for both versions. I'll make a new PR out of it (#97).

Balandat:
argh

gpleiss:
Closing this, since it was fixed in #97

PR #97, opened 2018-03-23
Function factory tests work on pytorch 0.3 and 0.4

gpleiss:
On PyTorch master (0.4-alpha), Indexing a zero-dimensional tensor does not return a numeric value anymore, but a tensor. This caused test failures with assertAlmostEqual of the kind TypeError: type Variable doesn't define __round__ method.

This PR fixes these assertion errors, so they work on stable PyTorch (0.3) and PyTorch master.

### DIFF
--- a/test/util/test_function_factory.py
+++ b/test/util/test_function_factory.py
@@ -15,7 +15,7 @@ from gpytorch.lazy import NonLazyVariable
 
 
 class PyTorchCompatibleTestCase(unittest.TestCase):
-    # Writing a separate function for compatability with PyTorch 0.3 and PyTorch 0.4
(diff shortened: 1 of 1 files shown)


test/util/test_function_factory.py
```diff
@@ -14,8 +14,28 @@
 from gpytorch.lazy import NonLazyVariable
 
 
-class TestMatmulNonBatch(unittest.TestCase):
-
+class PyTorchCompatibleTestCase(unittest.TestCase):
+    # Writing a separate function for compatability with PyTorch 0.3 and PyTorch 0.4
```
Balandat: spelling: compatibility

gpleiss: Whoops 😁

PR #98, opened 2018-03-26
Implement Preconditioning

jrg365:
This pull request implements a (currently optional) preconditioning feature for linear solves, using a pivoted Cholesky decomposition as a preconditioner. 

Preconditioning can significantly reduce the number of CG iterations required to get accurate solves. While this may not substantially affect training (other than a potential speed improvement), this can significantly improve test accuracy / speed where previously a significant number of CG iterations were required.

## Interface Changes
Preconditioning can now be enabled for linear solves by using the context
```python
with gpytorch.settings.max_preconditioner_size(value)
```
with value > 0 (default is 0, which disables preconditioning). 

## Internal Changes
- `LazyVariable` now defines a `_preconditioner()` method, which by default returns `None`. For certain `LazyVariables`, this can return a closure that takes a tensor and returns `P^{-1}v`, where `P` is a preconditioner for `self`.
- `LazyVariable.inv_matmul()` and `LazyVariable.inv_quad_log_det()` now pass the result of `self._preconditioner()` to their respective `Functions`.
- `add_diag` now creates a new `AddedDiagLazyVariable`, which extends `SumLazyVariable` but allows us to define the `_preconditioner()` method.
- The marginal likelihood calculations now always convert the prior covariance to a `NonLazyVariable`. This allows preconditioning to be applied when not using KISS-GP.
- When using preconditioning, the log determinant calculation in `inv_quad_log_det()` for an `AddedDiagLazyVariable` is adjusted to account for this. As a result, the loss calculations may be slightly less accurate when using preconditioning. However, the loss __gradient__ calculations are likely __more__ accurate, so convergence may be improved.
- `NonLazyVariable` now uses `super`'s `diag()` function if it is in batch mode.

### DIFF
--- a/gpytorch/lazy/__init__.py
+++ b/gpytorch/lazy/__init__.py
@@ -12,6 +12,7 @@ from .root_lazy_variable import RootLazyVariable
 from .sum_lazy_variable import SumLazyVariable
 from .sum_batch_lazy_variable import SumBatchLazyVariable
 from .toeplitz_lazy_variable import ToeplitzLazyVariable
+from .added_diag_lazy_variable import AddedDiagLazyVariable
 
 
 __all__ = [
@@ -29,4 +30,5 @@ __all__ = [
     SumLazyVariable,
     SumBatchLazyVariable,
     ToeplitzLazyVariable,
+    AddedDiagLazyVariable,
 ]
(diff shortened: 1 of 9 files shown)


gpleiss:
Looks good!

PR #105, opened 2018-04-27
Allow to change training targets for exact GP after fitting

Balandat:
This is useful e.g. for "fantasizing", i.e. performing posterior prediction under different realizations of the training target, without re-fitting the model parameters (see example below).

Note that this is written under pytorch 0.4+ and may not work under pytorch 0.3.1 (feel free to make that compatible)

```
def posterior_mean(gp, X):
    return gp(X).mean()

def fantasize(gp, fun, X, Ydist, nsamp):
    Ybase = gp.train_targets.clone()
    for i in range(nsamp):
        gp.set_train_targets(Ydist.sample())
        yield fun(gp, X)
    gp.set_train_targets(Ybase)

Ydist = Normal(
    loc=training_data.Y[metric],
    scale=training_data.Y_sem[metric],
)

gp.train_targets[:2]

0.1653
 0.3505
[torch.FloatTensor of size (2,)]

for m in evaluate_fantasies(gp, posterior_mean, training_data.X[metric], Ydist, 2):
    print(m[:2])

-0.4004
 0.6623
[torch.FloatTensor of size (2,)]

-0.0739
 0.4799
[torch.FloatTensor of size (2,)]

gp.train_targets[:2]

 0.1653
 0.3505
[torch.FloatTensor of size (2,)]
```

### DIFF
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -13,11 +13,14 @@ from ..likelihoods import GaussianLikelihood
 
 
 class ExactGP(Module):
+
     def __init__(self, train_inputs, train_targets, likelihood):
         if torch.is_tensor(train_inputs):
             train_inputs = train_inputs,
         if not all(torch.is_tensor(train_input) for train_input in train_inputs):
-            raise RuntimeError('Train inputs must be a tensor, or a list/tuple of tensors')
+            raise RuntimeError(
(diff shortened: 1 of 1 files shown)


Balandat:
Actually, I'll modify this to be able to set both training X and training Y. This is a quick way of conditioning on new data if we don't expect the hyperparameters to change much.

PR #107, opened 2018-04-27
Use torch.fft in place of libfftw (Requires PyTorch 0.4.0!)

jacobrgardner:
Alright, so here's a bit of a turning point. This pull request completely removes our dependency on libfftw in favor of the new `torch.fft` added in 0.4.0. 

Pros:
- The unit tests are about twice as fast, and the FFTs are as much as 10x faster in some cases in my testing. This scaling probably gets even better on large data regimes where we use even more inducing points.
- Our installation procedure is vastly simplified, since many of the install issues we've had raised in the past come from issues getting libfftw installed.

Cons:
- `master` becomes PyTorch 0.4 only.

I know @Balandat @gpleiss and I are using 0.4 at this point anyways. My vote would be to take the plunge, and split off the current master as a 'release' supporting both.


Balandat:
You read my mind - this was a thing that I was going to request :)
This would also vastly simplify deploying gpytoch in our infra, right now this super hacky...

Balandat:
While talking about changes in 0.4 - the support for distributions is vastly improved - would this allow simplifications / perf gains for the random variables?

jacobrgardner:
Yeah, replacing random_variables is more tricky because we have some custom stuff in there. The most obvious example is that we need it to be possible for the covariance of a Gaussian to be a LazyVariable.

It may be possible for us to extend the PyTorch distributions, but we've run in to trouble with the PyTorch backend in the past when trying to extend things (e.g. our original idea was for LazyVariable to extend Variable, but this turns out to just not be possible).

environment_cuda.yml
```diff
@@ -1,9 +1,7 @@
 name: gpytorch
 channels:
-  - conda-forge
   - pytorch
 dependencies:
   - cffi
-  - fftw
   - pytorch>=0.3.0
   - torchvision
```
gpleiss: This file should be deleted. It's the same as the other file.

Balandat:
nice!

Issue #110, opened 2018-05-02
Get rid of Variable


Balandat:
In pytorch 0.4, variables are just a legacy concept. Instead, tensors now take the `requires_grad` argument. Since master is now 0.4+, we should clean things up and replace all occurrences of `Variable` in the code with the appropriate Tensor equivalent.

Balandat:
yeah no rush, just putting it on the to do list

Balandat:
I'm going to put up a PR to kill this off once and for all, should be relatively straightforward in terms fo the codemod based on #265, now that there are no LazyVariables anymore....

PR #115, opened 2018-05-03
Change definition of RBF Kernel

Balandat:
Previous: `k(x, y) = exp(-|x-y|^2 / l)`
Now: `k(x,y) = exp(-0.5 * |(x-y)/l|^2)`

This doesn't really change any functionality whatsoever, but makes the definition consistent with some of the literature (e.g. Rasmussen & Williams) and the implementations in other packages (e.g. scipy and GPy)

Also, making baby steps towards getting rid of variables #110


jacobrgardner:
Awesome! This is something we'd wanted to change for a while but had always gotten stability problems in the unit tests. I wonder if something about 0.4 solved the issue.

PR #121, opened 2018-05-08
modify expand_as  to expand for 0.4

soumith:
expand_as(prod) errors out now, because it checks that the input is a Tensor, and the check fails because it's of type MatmulLazyVariable. So instead move to expand(size)


gpleiss:
Thanks @soumith !

PR #124, opened 2018-05-09
Use black to auto-format test code

Balandat:
I've been successfully using black (https://github.com/ambv/black) for code auto-formatting.
It's super easy to use and works well. I figured standardization in code formatting can't hurt.

I also added W503 to flake8 ignore, I think having the operator at the beginning of the line is so much better for readability (and this is what black uses).

If this looks ok to you guys we can successively move other parts of the code base to the new style.
In case you're worried about rebasing local commits, it's pretty simple to first auto-format any of the locally changed files, then rebase, that usually doesn't cause too much friction.


gpleiss:
This looks fantastic. Should we setup black to run as a pre-commit hook?

PR #127, opened 2018-05-09
[codemod] Format all code with black

Balandat:
Codemod!

### DIFF
--- a/gpytorch/lazy/mul_lazy_variable.py
+++ b/gpytorch/lazy/mul_lazy_variable.py
@@ -11,21 +11,26 @@ from ..utils import prod
 
 
 class MulLazyVariable(LazyVariable):
+
     def __init__(self, *lazy_vars):
-        '''
+        """
         Args:
             - lazy_vars (A list of LazyVariable) - A list of LazyVariable to multiplicate with.
-        '''
+        """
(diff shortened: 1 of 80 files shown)


gpytorch/lazy/mul_lazy_variable.py
```diff
@@ -102,13 +114,25 @@ def _matmul(self, rhs):
...
             left_res = (rhs.unsqueeze(-2) * left_root.unsqueeze(-1))
-            left_res = left_res.view(n, rank * m) if batch_size is None else left_res.view(batch_size, n, rank * m)
+            left_res = left_res.view(
+                n, rank * m
+            ) if batch_size is None else left_res.view(
+                batch_size, n, rank * m
+            )
             left_res = self.right_lazy_var._matmul(left_res)
-            left_res = left_res.view(n, rank, m) if batch_size is None else left_res.view(batch_size, n, rank, m)
+            left_res = left_res.view(
```
Balandat: There are a few things like this that can be written more nicely like

```
left_res = (
    left_res.view(n, rank, m)
    if batch_size is None
    else left_res.view(batch_size, n, rank, m)
)

```
but we can fix those later.

PR #132, opened 2018-05-11
A few tiny tweaks to linear_cg

Balandat:
The modified check for convergence is a little cleaner and also slightly faster. The other change is just formatting.


gpleiss:
It looks like torch.all wasn't working on python 3.6 for some reason? Switching it to `(...).all()` seems to work.

gpleiss:
Thanks @Balandat !

Balandat:
Hmm that's odd that it wouldn't work on 3.6, the python version should have nothing to do with it. I'm assuming the same pytorch version is used for both the 2.7 and the 3.6 tests?

PR #133, opened 2018-05-11
RBF stability improvements

### LINKED ISSUE #119: RBF Kernel Change Breaks Testing Code
The change to RBFKernel in 84fccd898c45c08279fb5c109e6e234f3a47588a may break something about our prediction code. 

I am not totally sure what the problem is yet, but I isolated this as the problem with `git bisect` and have a reasonable test case where results are significantly worse with the commit in compared to after a revert commit. 

It seems like the stability issues we encountered when making this change in the past don't come up in the unit tests, but do on some real datasets.

I can try to push my test case to a branch as well, although it relies on a UCI dataset.

@Balandat @gpleiss

gpleiss:
A few changes:

1. We divide the lengthscale by 2 before summing the squared dimension-wise distances. This seems to be more numerically stable than dividing by 2 afterwards.

1. Rather than adding 1e-5 to the lengthscale (to prevent division by zero errors), we're instead using a clamp operation.

Fixes #119

### DIFF
--- a/gpytorch/kernels/rbf_kernel.py
+++ b/gpytorch/kernels/rbf_kernel.py
@@ -3,6 +3,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import math
 from .kernel import Kernel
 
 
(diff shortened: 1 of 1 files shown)


jacobrgardner:
I can confirm that this fixes #119

PR #138, opened 2018-05-18
Lints, formatting

Balandat:


Balandat:
Sure. I had to put some of the stuff (priors) I was working on on the back burner myself, but will get to it soon.

Issue #113, opened 2018-05-03
White noise kernel?


stanbiryukov:
Hi there,

Thanks for the excellent library.  Is it possible to either add a white noise kernel or have an option to include input variance to the diagonal of the kernel matrix during fitting?  If I have known variance on my input data, it's not clear to me how I can incorporate that in the model examples.

Balandat:
@darbour is working on implementing that. Pull request should be incoming soon.

Balandat:
Ok, so I was trying to get this to work, but I'm having a hard time getting the correct behavior in all the different contexts (training, test, eval, forward, lazy, non-lazy, batch, non-batch, in usage with AdditiveKernel, etc.). @gpleiss, @jacobrgardner, any thoughts? 

The common use case is to define a new AdditiveKernel, e.g. `RBFKernel() + HeteroscedasticWhiteKernel(Y_train_target_var)`

```
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals

import torch

from gpytorch.kernels import Kernel
from gpytorch.lazy.diag_lazy_variable import DiagLazyVariable


class HeteroscedasticWhiteKernel(Kernel):

    def __init__(self, variances):
        super(HeteroscedasticWhiteKernel, self).__init__()
        self._variances = variances

    def forward(self, x1, x2, **params):
        if (
            x1.shape == x2.shape
            and x1.shape[0] == self._variances.shape[0]
            and torch.equal(x1, x2)
        ):
            return DiagLazyVariable(self._variances)
        else:
            return x1.new_zeros(x1.shape[0], x2.shape[0])

    def __call__(self, x1_, x2_=None, **params):
        if x2_ is None:
            x2_ = x1_
        return self.forward(x1_, x2_, **params)
```

Balandat:
Hmm so while #141 allows to subset the kernel matrix without fully computing it, I'm still not sure how to implement the logic for the heteroscedastic white kernel in a clean & efficient way. 

With all the laziness and efficient transposing for diag computations there is a lot going on. It's also not super straightforward to debug if the error you're making somewhere doesn't surface until way later in the code b/c of the lazy evaluation :(

Anyway, here's is a very dumb indefficient barebones idea for the heteroschedastic white Kernel - this trains and evaluates, but will fail when evaluating `.var()` b/c @jacobrgardner was trying to be smart in #144 😄 (`.covar().evaluate()` works). 

```
[...]

PR #139, opened 2018-05-20
Add priors [WIP]

Balandat:
This is an early attempt at adding priors. Lots of callsites in the code aren't updated yet, so this will fail spectacularly.

The main thing we need to figure out is how to properly do the optimization using standard gpytorch optimizers that don't support bounds. We should probably modify the smoothed uniform prior so it has full support and is differentiable everywhere but decays rapidly outside the given bounds. Does this sound reasonable?


jacobrgardner:
I am not sure what GPy and GPflow do for this, but GPML has this smooth box prior: https://github.com/alshedivat/gpml/blob/master/prior/priorSmoothBox2.m

Which is basically a uniform prior with tails that decay quickly as you move away from the range. Something like this could make a decent replacement for bounds.

Balandat:
So this is in a much better shape now. All the non-example tests should be passing, and all callsites should be fixed. There are two main outstanding issues:

1. What to do with the more complex parameters (e.g. `variational_mean`) - should we choose priors for those? 
2. How to deal with the cudaification of the model. Should priors be `nn.Module` classes and children of the model so we can utilize the `.cuda()` call? This seems ideal, not entirely sure how to best set this up though. 

Welcoming a more in-depth review.

jacobrgardner:
I'll take a look through the changes. To the two specific points you raised:

1. I think variational parameters ought to just not have priors, probably. They get used in such a specific thing (variational inference) that adding random extra terms to the ELBO seems strange to me.

2. I don't really have any strong preferences on how this is done, other than that on the user side it should behave as you'd expect (e.g., calling `.cuda()` on the model should work). At a glance, this might require some custom handling, as it seems like `torch.Distribution` doesn't define a `cuda()` function??? If that's true, I don't know what the cleanest way to do things is.

Making priors modules makes sense. One possibility is for `gpytorch.Module` to overwrite the `_apply` function (which `torch.nn.Module.cuda()` calls) and have it also loop through `self._priors` after calling `super`.

gpytorch/models/abstract_variational_gp.py
```diff
@@ -21,14 +20,14 @@ def __init__(self, inducing_points):
         self.register_buffer("inducing_points", inducing_points)
         self.register_buffer("variational_params_initialized", torch.zeros(1))
         self.register_parameter(
-            "variational_mean",
-            nn.Parameter(torch.zeros(n_inducing)),
-            bounds=(-1e4, 1e4),
+            name="variational_mean",
+            parameter=torch.nn.Parameter(torch.zeros(n_inducing)),
+            # TODO Add prior bounds: (-1e4, 1e4)
```
jacobrgardner: I mentioned this in my comment already, but I'd say let's just leave priors off of variational parameters for now (and remove their bounds). It's kind of strange to me to add extra terms to the variational ELBO.

test/examples/test_kissgp_additive_classification.py
```diff
@@ -30,10 +31,12 @@ def __init__(self):
         super(GPClassificationModel, self).__init__(
             grid_size=16, grid_bounds=[(-1, 1)], n_components=2
         )
-        self.mean_module = ConstantMean(constant_bounds=[-1e-5, 1e-5])
-        self.covar_module = RBFKernel(log_lengthscale_bounds=(-5, 6))
+        self.mean_module = ConstantMean(prior=SmoothedBoxPrior(-1e-5, 1e-5))
+        self.covar_module = RBFKernel(log_lengthscale_prior=SmoothedBoxPrior(-5, 6))
```
jacobrgardner: Should this (and a few other instance) be `log_lengthscale_prior=SmoothedBoxPrior(exp(-5), exp(6), log_transform=True)`? Placing a uniform distribution directly over log space leads to a distribution over the actual parameter that is not uniform.

test/examples/test_kissgp_additive_regression.py
```diff
@@ -45,8 +46,10 @@ class GPRegressionModel(gpytorch.models.ExactGP):
 
     def __init__(self, train_x, train_y, likelihood):
         super(GPRegressionModel, self).__init__(train_x, train_y, likelihood)
-        self.mean_module = ConstantMean(constant_bounds=(-1, 1))
-        self.base_covar_module = RBFKernel(log_lengthscale_bounds=(-3, 3))
+        self.mean_module = ConstantMean(prior=SmoothedBoxPrior(-1, 1))
+        self.base_covar_module = RBFKernel(
+            log_lengthscale_prior=SmoothedBoxPrior(-3, 3)
```
jacobrgardner: (e.g., `log_transform=True` here as well probably)

Balandat:
one thing I noticed is that the log_prob of the pytorch distributions that I call isn't behaving the way I want. E.g. I was hoping that when I specify a `n = Normal(loc=[m1, m2], scale=[s1, s2])`, then evaluating `n.log_prob([x1, x2]) would give me a tensor of log probs, one for the pdf of a N(m1, s1) at x1, and one for the pdf of N(m2, s2) at x2, but this is not what happens. Instead, even a scalar tensor will evaluate to a vector of log probs, one for each of the loc/scale elements. I'll need to rethink how to do this.

Balandat:
I guess instead i should use a MultivariateNormal as the basis. While this is unnecessary (working with full covariance matrices), it would allow to specify a prior jointly over multi-dimensional parameters (e.g. lengthscales). It's probably also quite inefficient in the case of independent parameters, b/c it works with the dense matrix, but the cost for this should be pretty tiny compared to the cost of computing the posterior....

Balandat:
Another, simpler way would be to for now just work with univariate distributions for each dimension and simply add the log probs (ie assuming independence). This would probably be the easiest for now.

Balandat:
argh it's the weird torch.any bug under 3.6 again... though this is the least of the problems with the current state of this code

Balandat:
Opened a pytorch issue for cudaification of distributions here: https://github.com/pytorch/pytorch/issues/7795

Balandat:
For now what I think I will do is register the parameters of the distribution as module parameters. This way the cudaification should work automatically, now that I'm making Priors Modules. The side effect is that we'll get hyperpriors for free, since we can pass in a prior for these priors :) so we'll need to somehow ensure that we don't set a prior on a prior that itself sets a prior that sets a prior.... you get the idea.

Balandat:
Well, importantly, I need to overwrite the `parameters()` function from the pytorch Module class so that we don't optimize over our prior parameters :)

gpleiss:
@Balandat is it possible to use buffers instead for the prior parameters?

Balandat:
yep that would make more sense, wouldn't it?

Balandat:
although I was pretty stoked about the hyperpriors :)

Balandat:
Priors are now also Modules, attached as children to the module to which
the parameter the prior is over belongs. I think this is pretty clean,
and the pytorch representation *i.e. just call `model`) makes things very clear as to which parameters have priors. I also found a reasonably elegant way to overwrite `_apply` in the priors in a minimally verbose fashion. Basically, all parameters and buffers are updated using the pytorch internals, but the distributions that live in the prior objects are re-initialized upon call of` _apply`. This makes
it such that calling `model.cuda()` on the top level model will just work.

There is still more work to be done - lots of test failures in the examples, some collision of `Module.initialize()` with `SpectralMixtureKernel.initialize()`, and the need to explicitly pass in
the size/dimension to the priors we already have in the examples (and probably much more stuff)

Balandat:
So the challenge with using the SmoothedBox prior in the original parameter space in combination with the log transformed parameters is that when you want to allow very small lengthscales you have to do a very abrupt transition to zero near zero for the SmoothedBox prior. Right now the way SmoothedBox prior is implemented is that it actually "spills over" the specified bounds, so it may happen that the distribution's support is not on the positive reals. I will change the implementation so that the support is always contained between the bounds a and b.

jacobrgardner:
I think I see what you are saying -- basically if the lower bound is `10e-5`, we need a really sharp tail, otherwise `10e-6` will have a decent amount of support still? Are there numerical issues with the tails falling off quickly? The negative support might not be an issue, since the exponentiation at least guarantees that the values entering `log_prior` are positive.

All that being said, in practice if a user has a lengthscale of `10e-5` or whatever, that probably means they should rescale their data and use more sane lengthscales -- computing extremely small distances and then dividing by extremely small numbers is a recipe for enormous floating point errors.

Balandat:
So I think there is a more general issue with priors not having full support (or at least numerically...) if using a simple descent algorithm that doesn't natively support constraints. If the step size is too large then the effect of the lengthscale on the MLL may make you do a gradient step that kicks you outside the support (after floating point errors), which means your ML is zero (MLL -inf). Not quite sure how to best deal with this.

Balandat:
Also, it looks like a lot of the test failures I saw were caused by using the initial guess based on the prior, even with the SmoothedBox one. This seems to suggests that the solutions are relatively sensitive to the initial parameter values, which is usually not a good sign... I'll remove the initial guess logic, try to fix the outstanding issues and then update the PR. So far it's not using the Priors for the variational Likelihoods, I'll leave that to you guys.

jacobrgardner:
Tests failing based on initial hyperparameter settings is not good. I will
investigate.

Balandat:
This removes the initial guesses from the PR and just uses the standard zero values. Most issues should be resolved and most test should be passing the ones that don't are related to KissGP tests and things that you are in a better place to investigate. So I'll leave this to you for now and turn to make #113 work. 

To be more concrete about the observed issues: previously the prior would set an initial guess for the parameters. E.g. for the SmoothBox that was the center of the box. If I did that fo the same "bounds" (i.e. range of the smoothed prior) the optimization would fail b/c of Lapack errors (off diagonal elements don't converge), and looking at the optimization path for the parameters it seemed to diverge. Keeping the full prior logic but not re-initializing the parameter to the midpoint and instead using the default zero solved that, which was a little disconcerting.

Balandat:
Sorry for the diffspam, but I'm juggling changes between a git and a mercurial version of this with different testing systems so things get confusing sometimes.

Anyway, this should be in relatively good shape expect for a few test failures with KissGP / SpectralMixture. @jacobrgardner, @gpleiss, could you have a look what's going on there? I'd like to merge this into master sometime soon if possible.

I'll add some more unit tests for the other priors as well as more docstrings down the road.

jacobrgardner:
I'll try to get the last few errors fixed today.

jacobrgardner:
Okay, two things:

1. Even on master, the derivatives for the log outputscale are absurd when doing the variational version of KISS-GP regression: tensor([-8399.4590]). The bounds are the only thing stopping that test from failing on iteration 2 on master.

2. Either way, the `log_prob` function of the prior is never called by any of the variational tests (including simple GP classification).

jacobrgardner:
Alright, I fixed most of the problems that were still popping up. I also renamed the prior files to match the classes, did a quick rebase on to master and squashed all the WIP commits together.

We can probably plan on merging this tomorrow or this weekend at the latest. The remaining tasks are: 

- Update the remaining example notebooks (I did a couple of them so far)
- Maybe give us a chance to fix the variational inference + SGD issues that arose when bounds disappeared. There was a reason we swapped out Adam for SGD originally when doing variational inference, so swapping it back in might not be ideal.

@gpleiss @Balandat

Balandat:
Great, thanks for cleaning up. No rush to land this right now, better to properly fix the variational inference + SGD issues first. I'll work on improving unit test coverage for the existing priors, as well as add some documentation.

jacobrgardner:
Do we think that, since this is a fairly large breaking change, we should leave `bounds` in as a keyword argument for a while that (1) uses a smooth box prior, and (2) throws a deprecation warning?

Balandat:
@jacobrgardner I think that makes a lot of sense. I'm pretty swamped until mid next week, but I can get to it then. Happy if you want to commandeer this PR though.

jacobrgardner:
@Balandat I finally had a chance to sit down with this and largely found the main SGD issue we were having -- basically in the tests previously, the mean constant bounds were set to such a tiny range that even the initial 0 was so close to the bound we were getting absurdly gigantic gradients from the log prior.

There's still a few tests to get working and the example notebooks to update, but I'll jump on this PR next week and see if we can't finally get it merged in by the end of the week.

Balandat:
Great. I have a few local changes that I'll push later (mainly removing the NotImplementedError from the IndexKernel. I guess it would be nice to allow priors over psd matrices for that, but that can come at a later stage.

Balandat:
How do you feel we should handle things like the IndexKernel in this PR? We could get fancy and add a prior that can be used on the covariance matrix itself (e.g. a Wishart distribution for some provided scale matrix V). Or we could ignore it, but then we'd need some other more ad-hoc way of ensuring numerical stability.

jacobrgardner:
Yeah I was considering the possibility of an inverse-Wishart prior as well. I know you guys mentioned multi task as a requirement in the design doc for BayesOpt -- do you guys use priors over the task kernels? If so, it might be good to implement now. If not, I could take it or leave it. I'd like to refactor IndexKernel to use a RootLazyVariable for a pretty big speed boost like we did internally in the SKIP paper in the nearish future once priors and sampling are implemented, so we could make an "Improve IndexKernel" issue and this could happen then.

PS - sorry I haven't gotten around to this PR yet this week, I've been swamped with NIPS reviews and a grant deadline tomorrow/today. It is still definitely next on my list to help ready this up :-)

Balandat:
We do use a prior over the task kernels, but as far as I can tell it's not a Wishart (or inverse Wishart, depending whether you want to work with covariance or precision matrix), but something more ad-hoc. @bletham will know the details.

Also, refactoring the IndexKernel using RootLazyVariable down the road would be nice. The reason I'm brining this up here is that we need to do something smarter than the "replace bounds with SmoothboxPrior" approach I did for most other parameters.

jacobrgardner:
Two options make sense to me:

1. Make bounds a special case no-op for the the IndexKernel, and have neither bounds nor prior support for this kernel at the moment.

2. Implement an inverse Wishart prior.

Balandat:
I think implementing the inverse Wishart prior makes a lot of sense and won't be hard to do. The thing to consider is that it doesn't exactly fit into the current setup, where a prior is married to a parameter (vector).

Instead, for the index kernel, we have two parameters: the covariance factor matrix `W` and the variance vector `sigma`, but the prior is on the function `W*W' + sigma*I`.

So we'd need to modify how priors work, e.g. by separating registration of a prior from the parameters and allowing to optionally pass in a callable on which the prior should be. E.g. something like this:

Blueprint for registering a prior (in `module.py`):
```
def register_prior(self, parameters, prior, func=None):
    if func == None:
        # deal with this as before
    else:
        # register callable such that we add prior.log_prob(func(*parameters)) to the the MLL
```

Register parameters and prior (in `index_kernel.py`)
```
self.register_parameter("covar_factor", nn.Parameter(torch.randn(n_tasks, rank)))
self.register_parameter("var", nn.Parameter(torch.randn(n_tasks)))

def genSigma(W, var):  # this could be reused in forward
    return W.matmul(W.transpose(-1, -2)) + var.diag()

self.register_prior([self.covar_factor, self.var], InverseWishartPrior(nu, K), func=genSigma)
```

Not sure how to best deal with the issue of backward compatibility in this case though. I guess we could allow passing `None` as the bounds, in which case we don't auto-set a `SmoothedBoxPrior`, and hope that people who want to do fancy stuff with priors read the docs and actually do pass in `None`. I guess this raises the point that we should error out when a parameter appears in multiple priors.

Does this make sense?

cc @darbour, @bletham, @bkarrer

Balandat:
Ideally we'd have support for Wishart / inverse Wishart distributions in pytorch proper.

But to use it as a prior we just need to evaluate the log pdf, which is straightforward. Related: 
Related: https://github.com/pytorch/pytorch/issues/678#issuecomment-404352305

jacobrgardner:
What do you think of treating the squaring of W (e.g. W -> W*W') as a transformation, similar to how we exponentiate the log noise / log lengthscale?

Would this let us still register the prior with the parameter? Basically, could we redefine priors to instead of taking `log_transform` as an argument, we can handle arbitrary closures that transform the parameter in some way before calling `log_probability`? This might let us maintain backwards compatability with the deprecation warning.

jacobrgardner:
I guess the challenge is we can't place the inverse-Wishart prior on `W*W'` directly, because the inverse-Wishart prior has support only over positive definite matrices, which requires the added diagonal to achieve if `W` is low rank?

Balandat:
Overall this sounds good to me.

> I am not entirely convinced that we should make priors mandatory [...]

Yeah, not specifying a prior (and hence effectively using a uniform prior) is fine and shouldn't be mandatory from an external API perspective. The nice thing about having differentiable uninformative priors (i.e. those with continuous pdf) is that this solves a lot of the issues that the ad-hoc parameter bounds dealt with in a clean way without a ton of numerical issues.

> [...] since they are overwhelmingly used in Bayesopt compared to other GP applications as far as I know.

Really? I think it's pretty common (and useful) to put priors on hyperparameters, at least if you have reasonable ones from previous modeling similar processes.

andrewgordonwilson:
> Really? I think it's pretty common (and useful) to put priors on hyperparameters, at least if you have reasonable ones from previous modeling similar processes.

Priors over kernel hypers, particularly for simple kernels, is mostly just used for BayesOpt. If one is using an RBF kernel (for example), the posterior over hypers will be quite sharply peaked for any moderately sized dataset, such that hyper marginalization and maximum marginal likelihood hypers will give you similar predictive distributions, but marginalization has a lot of extra overhead (and forces us to go to approximate inference). Rasmussen (1996) did an extensive empirical study for his PhD thesis and came to a similar conclusion. Hyper marginalization will be most useful in cases where both (1) there is very little data; (2) a highly accurate representation of uncertainty is crucial. BayesOpt typically fits these criteria.  

But I would be curious to see non-BayesOpt examples where hyper priors have been shown to give much better performance than type II maximum likelihood.

Balandat:
Just to be clear, this PR doesn't aim to do hyper marginalization, in its current state it just allows to put (non-uniform) priors over the hyperparamters. Inference is still done by maximizing MLL (now involving the hyperparameter priors).

That said, a lot of our everyday use cases do indeed fit both (1) and (2), and so in practice we often do use MCMC to perform hyper marginalization.

Balandat:
Added Wishart and Inverse Wishart priors (to be put e.g. over the ICM kernel). Ideally those will be moved into torch.distributions, for now they just evaluate the pdf (no sampling).

Added support for derived priors (haven't really tested that part). 

Changed priors to have a shape rather than a size property.

jacobrgardner:
I think I've fixed the last issue in the unit tests. I have no idea how or when, but the test function for the variational regression test was sin(x*2*pi) but the function the GP was trained on was sin(x*4*pi), so it is perhaps unsurprising that it never worked..........

gpytorch/kernels/_utils.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import unicode_literals
+
+import logging
+import torch
+from gpytorch.priors import SmoothedBoxPrior
+
+logger = logging.getLogger()
+
+
+def _bounds_to_prior(
```
jacobrgardner: Can we delete this file since this function appears to be duplicated in `priors._compatability`?

Balandat: good catch, will kill the whole _utils.py file.

gpytorch/priors/prior.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @abstractproperty
+    def shape(self):
+        raise NotImplementedError()
+
+    def log_prob(self, parameter):
+        """Returns the log-probability of the parameter value under the prior."""
+        return self._log_prob(parameter.exp() if self.log_transform else parameter)
+
+    def _initialize_distributions(self):
```
gpleiss: What is this method supposed to do?

gpytorch/priors/prior.py
```diff
@@ -0,0 +1,66 @@
...
+    @property
+    def log_transform(self):
+        return self._log_transform
+
+    @abstractmethod
+    def _log_prob(self, parameter):
+        raise NotImplementedError()
+
+    @abstractproperty
+    def shape(self):
```
gpleiss: Could we use "size" as the method name, to be consistent with other method names?

gpytorch/priors/prior.py
gpleiss: Oh I see... it's necessary for the apply function. maybe Priors should have a default constructor that calls this method?

gpleiss:
I left a few tiny comments. Overall this looks awesome and I'm super pumped for it! (Especially softbox priors)

Two other comments (I feel strong about the first comment, less so about the other):

1. *Remove custom `__getattr__` and `__setattr__` in modules*. We wrote custom versions of these functions in the first place to ensure that all parameters had associated bounds. Since this is no longer the case, I think that it should be possible to register a parameter by setting `self.<param_name> = ...`, in addition to `self.register_parameter("<param_name>", ...)`.

1. *Shouldn't differentiate between parameter prior and derived prior.* As far as I can tell, a `derived_prior` is a strict generalization of a `param_prior`. Rather than differentiating between the two, I think that we should just have `derived_prior`s, and when you call `register_parameter` with a prior option, we create a derived prior where `params` just contains the newly registered parameter, and transform is just an identity function.

This would simplify the code, but I realize this PR has been open for a long time so I'd also be in favor of punting on this in favor of merging this in. (I also can do this merging if others are on board with this.)

gpytorch/priors/prior.py
Balandat: sure, I can add this

gpytorch/priors/prior.py
Balandat: I changed it from size to shape, but ok :)

Balandat:
1. I agree; opened #180 to track this.
2. I had thought about this as well, but currently we're really only using this for the prior over the IndexKernel, so not sure how pressing it is. Opened #181 to track this.

gpytorch/priors/prior.py
jacobrgardner: I think we should maintain shape. PyTorch implements a size *method* and a shape *property* that function the same way which I think we should be consistent with.

gpytorch/priors/prior.py
Balandat: Actually, let's not do this. We need the Module init call before we can assign buffers, but some of the `_initialize_distributions` calls will require some buffers to be set. So it'd be quite awkward doing this if the base class constructor called `_initialize_distributions`.

gpytorch/priors/prior.py
Balandat: Sounds good to me - I won't touch the LazyVariable part for now, opened #182 for this.

jacobrgardner:
Alright, it seems like we're all on board to merge this in!

Balandat:
Oh I had a few updates pending with size(), shape, and improved docstrings. Will create a new PR for those changes

jacobrgardner:
My bad -- I'd assumed that making the issues to track them meant we were
going to handle those in separate PRs.

Balandat:
Yeah that does mean it, but there were some outstanding fixes not tracked by the issues. No worries will put up PR soon, should be quite simple.

Balandat:
#184

Issue #125, opened 2018-05-09
LazyVariables should define a get_row


jacobrgardner:
Right now, `gpytorch.utils.pivoted_cholesky` uses `LazyVariable.__getitem__` to access individual rows of `LazyVariables`. This is usually overkill, since `__getitem__` is usually much more general than a `get_row` method would be, and therefore much more complicated.

For example, we cannot use preconditioners with SGPR right now because `InvQuadLazyVariable` uses the default `__getitem__`. 

Having `LazyVariables` define a `get_row` method that defaults to doing something like `return self[i, :]` (or `self[:, i, :]` in batch mode) would be a better solution, because writing special cases of `get_row` will usually be much easier than writing special cases of  `__getitem__`.

jacobrgardner:
I'm closing this, because we ended up just writing smarter `__getitem__` implementations for most of the lazy variables that matter here, and it's not clear we could do substantially better by special casing to getting a row anymore.

PR #140, opened 2018-05-22
Pivoted cholesky fixes

gpleiss:
- The `_preconditioner` function now also defines the log det correction (rather than the LazyVariable itself)
- The preconditioner works on batched matrices
- The preconditioner uses an approx_diag function (which is exact for most lazy variables, only an approximation for InterpolatedLazyVariables)


jacobrgardner:
I think with this we could probably make a rank 5 preconditioner on by default, as well as close #125

gpleiss:
Just made the preconditioner a default option

gpytorch/lazy/interpolated_lazy_variable.py
```diff
@@ -75,6 +75,17 @@ def __init__(
         self.right_interp_indices = right_interp_indices
         self.right_interp_values = right_interp_values
 
+    def _approx_diag(self):
```
jacobrgardner: Two things:

1. For this to actually work in a lot of the real models, we'll need `ConstantMulLazyVariable` to also define `_approx_diag` and do the obvious thing there.

2. `ConstantMulLazyVariable` is the worst.

jacobrgardner: (Other than point 1, looks good :-)

Issue #135, opened 2018-05-11
Exact GPs should compute test and train kernel matrices separately at test time


jacobrgardner:
Persuant to the discussion in #113, Exact GPs should compute the various covariance matrices needed for the predictive distribution separately. This could save a lot of time and space for applications like BayesOpt, and also make the white noise kernel implementation a little easier since we wouldn't have to pad the diagonal with `t` zeros.

@gpleiss and I talked, and I think a great solution is to only create matrices as they are needed. Basically `exact_gp.py` would create some object like a `LazyKernelEvaluationVariable`, whose `__getitem__` and `diag` methods would explicitly call the kernel.  E.g., `my_lazy_evaled_kernel[:n_train, :n_train]` would actually call the kernel on `k(x_train, x_train)`.

In addition to space and time savings, with this kind of construct, something like the code as written by @Balandat in #113 for the white noise kernel would work.

Balandat:
This would be awesome. And it would make it straightforward to cache `k(x_train, x_train)` for eval mode.

jacobrgardner:
I am going to work on this unless someone would prefer that #134 gets handled first. I have no preference on an order for these two, but this issue seems more directly problematic for using exact GPs for BayesOpt.

Balandat:
makes total sense, I think this is higher pri as it will make #113 much easier

PR #141, opened 2018-05-23
Lazily evaluate kernel functions

jacobrgardner:
Working on fixing #135. So far, simple GPs and the 1D KISS-GP tests pass.

My initial solution to this is to make the `__call__` method of `Kernel` directly create the new `LazyEvaluatedKernelVariable`, which basically supports a `__getitem__` method (so we can subindex in to a kernel matrix before evaluating the kernel, which is the point of this PR) and an `evaluate()` method. This is a meta lazy variable, in that the `evaluate()` method can return another `LazyVariable` if the kernel call does.

The reason I am obviously unhappy with this is that if the user does anything in forward other than call a kernel and return the `LazyEvaluatedKernelVariable`, we now call `evaluate()` in the model, which is obviously a Bad Thing if the covariance is, for example, already an `InterpolatedLazyVariable` or something.

I have a decent solution to this that I'll work on either tomorrow or after graduation when I have time.

(P.S.  I think the "obvious" solution of just refactoring so that we call forward on the different sets of data directly would actually require an even larger internals rework because of how we handle predictions via `LazyVariable` methods.)

### DIFF
--- a/gpytorch/kernels/grid_kernel.py
+++ b/gpytorch/kernels/grid_kernel.py
@@ -6,7 +6,7 @@ from __future__ import unicode_literals
 import torch
 from torch.autograd import Variable
 from .kernel import Kernel
-from ..lazy import ToeplitzLazyVariable, KroneckerProductLazyVariable, NonLazyVariable
+from ..lazy import ToeplitzLazyVariable, KroneckerProductLazyVariable
 from .. import settings
 
 
@@ -43,15 +43,15 @@ class GridKernel(Kernel):
 
             if settings.use_toeplitz.on():
                 first_item = grid_var[:, 0:1].contiguous()
-                covar_columns = self.base_kernel_module(first_item, grid_var, **kwargs)
+                covar_columns = self.base_kernel_module(first_item, grid_var, **kwargs).evaluate()
                 covars = [
                     ToeplitzLazyVariable(covar_columns[i : i + 1].squeeze(-2))
                     for i in range(n_dim)
                 ]
             else:
                 grid_var = grid_var.view(n_dim, -1, 1)
-                covars = self.base_kernel_module(grid_var, grid_var, **kwargs)
-                covars = [NonLazyVariable(covars[i : i + 1]) for i in range(n_dim)]
+                covars = self.base_kernel_module(grid_var, grid_var, **kwargs).evaluate()
+                covars = [covars[i : i + 1] for i in range(n_dim)]
 
             if n_dim > 1:
                 covar = KroneckerProductLazyVariable(*covars)
--- a/gpytorch/kernels/grid_kernel.py
+++ b/gpytorch/kernels/grid_kernel.py
@@ -50,7 +50,7 @@ class GridKernel(Kernel):
                 ]
             else:
                 grid_var = grid_var.view(n_dim, -1, 1)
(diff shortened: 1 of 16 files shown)


jacobrgardner:
Basically, what I'm thinking is that the new object isn't a LazyVariable at all, and just "evaluates" the kernel directly in the `__getitem__` method. I was originally not doing this because I was going to have it define a diag() method so that we could get variances more easily for exact GPs. However after working on it for a bit I think a better solution is just to define an `LazyVariable_exact_predictive_var(...)` that by default does `self._exact_predictive_covar(...).diag()`

gpleiss:
It might make sense to have this object not be a LazyVariable. We could create a new class of Lazy objects.

GPyTorch... the world's laziest library 🤣

jacobrgardner:
GPyTorch follows the __aggressively lazy__ design pattern. We refuse to actually compute anything until you actually ask us at least twice :-)

jacobrgardner:
Ok, this PR is essentially done now. It fixes #135 and should make @Balandat and @darbour's solution to #113 work.

## Internal changes
- The `Kernel.__call__` method now returns a new `LazyEvaluatedKernelVariable` rather than calling forward directly.
- `LazyEvaluatedKernelVariable` supports sub indexing, and returns a new `LazyEvaluatedKernelVariable` from `__getitem__`. It defines an `evaluate_kernel` method that actually calls the kernel. Thus, kernel calls happen after subindexing, not before. 
- `LazyEvaluatedKernelVariable` overwrites the `representation` and `representation_tree` functions to masquerade as the `LazyVariable` that results from calling `evaluate_kernel` rather than as itself. Reconstructing a `LazyVariable` from its representation tree therefore evaluates all kernel calls. Because all functions that operate on `LazyVariables` reconstruct it, these functions will work even with `LazyEvaluatedKernelVariables` in the tree.
- __For a special case__, the base `LazyVariable` class now defines an `evaluate_kernel` method that just rebuilds self. This is because `pivoted_cholesky` is not a function yet, so it (and future things like it) may need a quick way to do this.

## User level changes
- All kernels are now lazily evaluated. Therefore, even basic kernels like `RBFKernel` need to be followed with an evaluate() call if a tensor is desired. For example, `RBFKernel(x1, x2).evaluate()`.
- Nothing changes at the actual "using a GP" level, because all of this is handled in `exact_predictive_mean` and `exact_predictive_covar`.
- Nothing changes for sparse GP methods, because those kernels return `LazyVariables` that would need to be evaluated to get tensors anyways.

## Future work
- Once `pivoted_cholesky` is a `Function`, the last remaining calls to `LazyVariable.evaluate_kernel` can be removed from the code base. This is a PR in its own right, however.
- `LazyEvaluatedKernelVariable` defines a `diag` function that right now just calls `self.evaluate_kernel().diag()`, but could do this more intelligently using batch indexing. This should allow us to somewhat easily define a "diagonal only" prediction mode that gets variances instead of covariances. This is not necessarily useful for the sparse GP methods when using fast predictive variances, but is useful for exact GPs.

jacobrgardner:
My plan is to ideally get this merged early this week. I'll then move on to implementing the "variances only" prediction mode for exact GPs, and after that I'll see what issues are up for grabs.

PR #144, opened 2018-05-29
Implement variance only predictions for exact GPs

jacobrgardner:
This turned out to be really easy with the new `LazyEvaluatedKernelVariable` from #141. Our predictive covariance calculation already did __almost__ the right thing, in the sense that it handled the `covar_correction` part of the variance calculations correctly with a MatmulLazyVariable, we just were computing the full `test_test_covar` before computing the variances:
https://github.com/cornellius-gp/gpytorch/blob/56d18a4a93acbc68b3b2cc4f382ebc0c212a5062/gpytorch/lazy/lazy_variable.py#L392-L394

Now, `test_test_covar` is a `LazyEvaluatedKernelVariable` even in the exact GP setting, so all I had to do was implement the smarter `diag` function for it.

## Internal changes
- `LazyEvaluatedKernelVariable` now implements an intelligent `diag()` method that directly calls the kernel with transposed data to get only the diagonal entries.
- Doing something like the following:
```python
preds = model(test_x)
variances = preds.var()
```
now gets the variances only, while avoiding computing the full `test_test_covar` kernel. This should make things like EI calculations much faster when using exact GPs.

@Balandat @gpleiss

### DIFF
--- a/gpytorch/lazy/lazy_evaluated_kernel_variable.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_variable.py
@@ -26,11 +26,28 @@ class LazyEvaluatedKernelVariable(LazyVariable):
 
     def diag(self):
         """
-        TODO: Can be handled by calling the kernel after creating some new batch dimensions and transposing.
-
-        Hopefully, this can be easily used to add a 'variance_only' prediction mode.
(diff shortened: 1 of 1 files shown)


Balandat:
awesome, thanks!

Issue #146, opened 2018-05-29
Lapack error in TestKissGPVariationalRegression


Balandat:
Getting this error when running on current master:

```
RuntimeError: Lapack Error syev : 14 off-diagonal elements didn't converge to zero at caffe2/aten/src/TH/generic/THTensorLapack.cpp:388
  File "/usr/local/fbcode/gcc-5-glibc-2.23/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/fbcode/gcc-5-glibc-2.23/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/pytorch/gpytorch/test_gpytorch_examples#binary,link-tree/test/examples/test_kissgp_variational_regression.py", line 92, in test_kissgp_gp_mean_abs_error
    loss = -mll(output, y_batch)
  File "/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/pytorch/gpytorch/test_gpytorch_examples#binary,link-tree/gpytorch/module.py", line 184, in __call__
    outputs = self.forward(*inputs, **kwargs)
  File "/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/pytorch/gpytorch/test_gpytorch_examples#binary,link-tree/gpytorch/mlls/variational_marginal_log_likelihood.py", line 33, in forward
    for variational_strategy in self.model.variational_strategies()
  File "/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/pytorch/gpytorch/test_gpytorch_examples#binary,link-tree/gpytorch/mlls/variational_marginal_log_likelihood.py", line 33, in <genexpr>
    for variational_strategy in self.model.variational_strategies()
  File "/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/pytorch/gpytorch/test_gpytorch_examples#binary,link-tree/gpytorch/variational/mvn_variational_strategy.py", line 28, in kl_divergence
    inv_quad_rhs=inv_quad_rhs, log_det=True
  File "/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/pytorch/gpytorch/test_gpytorch_examples#binary,link-tree/gpytorch/lazy/lazy_variable.py", line 497, in inv_quad_log_det
    *args
[...]

PR #147, opened 2018-05-29
Fix seed for TestKissGPVariationalRegression

Balandat:
We currently have test failures on pytorch master (#146). This fixes those for me.


Balandat:
argh

jacobrgardner:
Adding a very small amount of noise to the training labels so that the log noise isn't ridiculously small made the test more stable.

PR #152, opened 2018-05-31
Implement white noise kernel

jacobrgardner:
Here's my take on implementing the white noise kernel (#113), and two simple unit tests for it that show it works in at least the simple GP regression case and the 1D KISS-GP regression case.

I changed things a bit so that kernels can return `None`, and `LazyVariables` treat adding `None` as a no-op. If we think this is unsafe or would make debugging more difficult in some cases, we could equivalently make a `ZeroLazyVariable` that behaves the same way.

With that in mind, the white noise kernel returns a `DiagLazyVariable` containing `self.variances` if it believes it is dealing with training data, and returns `None` otherwise. 

There are some changes to `AddedDiagLazyVariable` I'd like to make to make things even more efficient when dealing with this kind of kernel, but that can be a separate PR.

@Balandat

### DIFF
--- a/gpytorch/lazy/lazy_variable.py
+++ b/gpytorch/lazy/lazy_variable.py
@@ -818,13 +818,18 @@ class LazyVariable(object):
 
     def __add__(self, other):
         from .sum_lazy_variable import SumLazyVariable
-
+        if other is None:
+            return self
         return SumLazyVariable(self, other)
 
     def __div__(self, other):
+        if other is None:
+            return self
         return self.mul(1. / other)
 
     def __mul__(self, other):
+        if other is None:
+            return self
         return self.mul(other)
 
     def __getitem__(self, index):
--- a/gpytorch/lazy/lazy_variable.py
+++ b/gpytorch/lazy/lazy_variable.py
@@ -818,18 +818,24 @@ class LazyVariable(object):
 
     def __add__(self, other):
         from .sum_lazy_variable import SumLazyVariable
-        if other is None:
+        from .zero_lazy_variable import ZeroLazyVariable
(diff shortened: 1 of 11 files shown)


jacobrgardner:
Replacing `None` in this PR with a thin `NoneLazyVariable` might be safer.

gpytorch/lazy/lazy_variable.py
```diff
@@ -818,13 +818,18 @@ def zero_mean_mvn_samples(self, n_samples):
...
         return SumLazyVariable(self, other)
 
     def __div__(self, other):
+        if other is None:
+            return self
         return self.mul(1. / other)
 
     def __mul__(self, other):
+        if other is None:
+            return self
```
Balandat: should we raise an error when multiplying / dividing with what essentially is a zero kernel?

Balandat:
Yeah having a `ZeroLazyVariable` would make a lot of sense. Then we could have it work more generally in combining with others (+ as no-op, * returning a ZeroLazyVariable...)

jacobrgardner:
Alright, I implemented a quick `ZeroLazyVariable`. 

I flushed out a bunch of the functions that `LazyVariable` defines by default in this case to throw useful errors for easier debugging (e.g., bad things would silently happen when trying to invert a `ZeroLazyVariable`).

jacobrgardner:
Last thing to consider is size checking in `ZeroLazyVariable` (e.g., `mul` could check that `self.size() == other.size()`, `matmul` could check that `self.size(-1) == tensor.size(0)` (or `self.size(-1) == tensor.size(-2)` for batch mode).

PR #157, opened 2018-06-07
Don't allow mean of GaussianRandomVariable to be None

Balandat:
This doesn't seem to be used anywhere. It also confusing and does some undocumented magic that we should avoid.


jacobrgardner:
Hmm. Maybe an accidental hold over from when I was initially letting things return None before I implemented ZeroLazyVariable? Good catch!

PR #158, opened 2018-06-08
Support batch GP predictions (Fixes #134)

### LINKED ISSUE #134: GPyTorch should handle batched predictions for independent GPs
I'm spinning off a separate issue from the discussion in #130 . GPyTorch should handle `k x n x d`  training data, `k x n` labels, and `k x t x d` test points **at test time** as though we have `k` separate Gaussian processes, each with a training set and a set of test points. 

I think handling this at training time would be much more challenging since we'd need to call backward on a list of `k` losses, but the use case @Balandat  mentioned has fixed hyperparameters, so we only need test time support.

I am marking this as a bug, because I believe @samuelstanton has a fix for this, although it may involve some for loops so we may want to fix it a different way. I think Sam mentioned he actually had a batched solution?

jacobrgardner:
Exact GPs support batch predictions, which addresses #134. I'll open up another issue for batch training, which shouldn't be that hard (it looks like someone wanted to support this at some point: the `ConstantMean` class has an optional `batch_mode` argument that reshapes the constant parameter appropriately). Now, you can do something like:

```python
batch_train_x = train_x.repeat(3, 1, 1)
batch_train_y = train_y.repeat(3, 1)
batch_test_x = test_x.repeat(3, 1, 1)
model.set_train_data(batch_train_x, batch_train_y, strict=False)
batch_preds = model(batch_test_x)
```

and you'll get batch predictions on each batch of test data, using the corresponding train input and label batches as the input data for that GP.

## Internal Changes
Basically, this just involved bringing prior and posterior means up to speed with what was already done for kernels.
- `Mean` now reshapes `x` to be 3 dimensional, and therefore all means now operate in batch mode by default.
- Fixed `LazyVariable.exact_predictive_mean` to handle batches.
- Internally, exact GPs reshape all input data to be at least two dimensional (e.g., 11 training examples in 1 dimension is now handled internally as an `11 x 1` tensor, just as 2 dimensional data is a `11 x 2` tensor). This is necessary to avoid ambiguity about whether a `10 x 3` tensor means 10 batches of 3 1D training examples or 10 training examples in 3 dimensions.

## Interface changes
- `set_train_data` now has an optional `strict` argument which defaults to true. When false, the shape of the data is allowed to change.

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -46,9 +46,9 @@ class Kernel(Module):
             x2 = x1
 
         # Give x1 and x2 a last dimension, if necessary
-        if x1.data.ndimension() == 1:
+        if x1.ndimension() == 1:
             x1 = x1.unsqueeze(1)
-        if x2.data.ndimension() == 1:
+        if x2.ndimension() == 1:
             x2 = x2.unsqueeze(1)
         if not x1.size(-1) == x2.size(-1):
             raise RuntimeError("x1 and x2 must have the same number of dimensions!")
--- a/gpytorch/lazy/lazy_variable.py
+++ b/gpytorch/lazy/lazy_variable.py
@@ -288,15 +288,28 @@ class LazyVariable(object):
         Returns:
         - (t) - the predictive posterior mean of the test points
         """
-        n_train = train_labels.size(0)
+        n_train = train_labels.size(-1)
         if precomputed_cache is None:
-            train_mean = full_mean[:n_train]
-            train_train_covar = self[:n_train, :n_train].add_diag(noise)
-            precomputed_cache = train_train_covar.inv_matmul(train_labels - train_mean)
+            train_mean = full_mean.narrow(-1, 0, n_train)
+            if self.ndimension() == 3:
+                train_train_covar = self[:, :n_train, :n_train].add_diag(noise)
+            else:
+                train_train_covar = self[:n_train, :n_train].add_diag(noise)
 
-        test_mean = full_mean[n_train:]
-        test_train_covar = self[n_train:, :n_train]
-        res = test_train_covar.matmul(precomputed_cache) + test_mean
+            train_labels_offset = train_labels - train_mean
+            if self.ndimension() == 3:
+                train_labels_offset = train_labels_offset.unsqueeze(-1)
+            precomputed_cache = train_train_covar.inv_matmul(train_labels_offset)
+
+        test_mean = full_mean.narrow(-1, n_train, full_mean.size(-1) - n_train)
+        if self.ndimension() == 3:
+            test_train_covar = self[:, n_train:, :n_train]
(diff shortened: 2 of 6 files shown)


jacobrgardner:
I'll write a couple quick unit tests for this shortly.

Balandat:
This looks great! Let's wait for those unit tests before merging.

fwiw, I wrote a little thingy that would be useful for implementing some basic CV for small data sets that fit into memory:
```
#!/usr/bin/env python3

from typing import Tuple

import torch
from torch.utils.data import TensorDataset


class KFoldTensorDataset(TensorDataset):
    def __init__(self, *tensors: torch.Tensor, k: int, shuffle: bool = False) -> None:
        super().__init__(*tensors)
        N = super().__len__()
        if k > N:
            raise ValueError("k cannot be larger than the size of the input tensors")
        self._k = k
        self._fold_size = N // k
        self._idcs = torch.randperm(N) if shuffle else torch.arange(N)
        self._mask = self.tensors[0].new_zeros(N, dtype=torch.uint8)

    def __len__(self) -> int:
        return self._k

    def __getitem__(self, i: int) -> Tuple[torch.Tensor, ...]:
        if i > len(self) - 1:
            raise IndexError("Dataset has only {} folds".format(len(self)))
        start = self._fold_size * i
        end = start + self._fold_size if i < len(self) - 1 else self._mask.shape[0]
        try:
            self._mask[self._idcs[start:end]] = 1
            return (
                tuple(tensor[~self._mask] for tensor in self.tensors),
                tuple(tensor[self._mask] for tensor in self.tensors),
            )
        finally:
            self._mask.zero_()
```

Then you can do stuff like this to make the full batched set of test data:
```
import torch
from torch.utils.data import DataLoader

def make_cv_data(kfold_loader):
    n_tensors = len(kfold_loader.dataset.tensors)
    cat_c = [[] for _ in range(n_tensors)]
    cat = [[] for _ in range(n_tensors)]
    for ts_c, ts in kfold_loader:
        for t_c, t, c_c, c in zip(ts_c, ts, cat_c, cat):
            c_c.append(t_c)
            c.append(t)
    return (
        tuple(torch.cat(c) for c in cat_c),
        tuple(torch.cat(c) for c in cat),
    )

X = torch.rand(100, 10)
Y = torch.rand(100)

kfold_dataset = KFoldTensorDataset(X, Y, k=5, shuffle=False)
kfold_loader = DataLoader(kfold_dataset)
(train_x, train_y), (test_x, test_y) = make_cv_data(kfold_loader)
```

jacobrgardner:
I added a simple unit test that tests out making predictions on a sine dataset and a cosine dataset simultaneously using exact GPs.

I spent a bit of time trying to get things working for KISS-GP as well, but encountered #159, which blocks it from working.

I am gonna merge this in once Travis is done for now, as #159 is definitely a whole other issue.

PR #178, opened 2018-07-18
Refactoring to consolidate the two methods of defining acquisition fu…

darbour:
Refactoring to consolidate the two methods of defining acquisition functions. Modules now call functions from the  directory where appropriate. Also some renaming of subfolders for consistency, and updates to imports to remove failures from import issues.


gpytorch/gpytorchopt/bayesian_optimization_example2_oracle.py
```diff
@@ -34,7 +34,7 @@ def update_model(self, refit_hyper=True, lr=0.1, optim_steps=300, verbose=False,
         oracle_kernel_list = []
         for i in range(self.n_dims):
             oracle_kernel_list.append(gpytorch.kernels.RBFKernel(log_lengthscale_bounds=(-10, 10), active_dims=[i]))
-            oracle_kernel = gpytorch.kernels.AdditiveKernel(*oracle_kernel_list)
+            oracle_kernel = gpytorch.kernels.kernel.AdditiveKernel(*oracle_kernel_list)
```
Balandat: Was AdditiveKernel missing in the `__init__.py` of the kernels module? Or are you saying this shouldn't be part of the user-facing API? Similar comment applies in a bunch of other places.

gpytorch/gpytorchopt/gpytorchopt/acquisition/__init__.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python3
```
Balandat: Should we maintain py2 compatibility as long as pytorch does? If so, we need to convert the typing into comment-style type hints. More generally, we should add types / type hints throughout the code early on, as this will simplify development.

I would actually be fine restricting the bayesopt functionality to py3 only...

@jacobrgardner, @gpleiss, what's your take on this?

gpytorch/gpytorchopt/gpytorchopt/acquisition/functional/acquisition.py
```diff
@@ -0,0 +1,132 @@
...
+    var = pred_rv.covar().diag() * beta
+    return mean + var.sqrt()
+
+
+def max_value_entropy_search(
+    model: Module,
+    candidate_set: torch.Tensor,
+    num_samples: int,
+) -> torch.Tensor:
+    model.eval()
```
Balandat: Looks like in the other functions you expect the model to already be in eval mode? We should be consistent across functions.

gpytorch/gpytorchopt/gpytorchopt/acquisition/functional/acquisition.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/env python3
+
+from gpytorch import Module
+import torch
+from torch.distributions import Normal
+
+
+def expected_improvement(
```
Balandat: Is there a reason to use a single `acquisition.py` file for these acquisition functions, but to keep KG separate?

gpytorch/gpytorchopt/gpytorchopt/acquisition/modules.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class AcquisitionFunction(Module):
+    def __init__(self, gp_model: Module) -> None:
+        super(AcquisitionFunction, self).__init__()
+        self.gp_model = gp_model
+
+    def forward(self, candidate_set: torch.Tensor) -> torch.tensor:
+        # takes in an n*d candidate_set tensor and return an n*1 tensor
+        pass
```
Balandat: let's raise a NotImplementedError here (or make AcquisitionFunction an abstract base class)

gpytorch/gpytorchopt/gpytorchopt/acquisition/__init__.py
```diff
@@ -0,0 +1,7 @@
...
+
+from .modules import AcquisitionFunction
```
Balandat: use a single import statement instead?

gpytorch/gpytorchopt/bayesian_optimization_example2_oracle.py
darbour: Yes, it was missing. I'll add it in, this fix was lazy

gpytorch/gpytorchopt/gpytorchopt/acquisition/functional/acquisition.py
darbour: IIRC, Brian mentioned that KG wasn't trustworthy at the moment. I kept it separate so we had some marker of what was stable. In general, I think it's smarter to remove KG and add it on a branch that gets merged when it's stable. Brian is away, I wanted to wait for him to come back to do that.

Issue #180, opened 2018-07-19
Remove custom __getattr__ and __setattr__ in module.py


Balandat:
We wrote custom versions of these functions in the first place to ensure that all parameters had associated bounds. Since after #139 goes in this will no longer be the case, it should be possible to register a parameter by setting self.<param_name> = ..., in addition to self.register_parameter("<param_name>", ...).

jacobrgardner:
8c92618e66b57319435a7312665a331c49f3b9b6 forks off of the `priors` branch and seems to do this successfully. I modified one of the unit tests to test adding a parameter without self.register_parameter. Once `priors` is merged in I'll open a quick PR.

PR #183, opened 2018-07-19
Remove custom __getattr__ and __setattr__ from gpytorch.Module (addresses #180)

jacobrgardner:
It seems like this was easy enough -- I think these really only existed to stop people from creating parameters without bounds. Since bounds are now handled exclusively as priors, none of the functionality in those methods accomplished anything necessary.

Travis, what do you think?

#180

### DIFF
--- a/test/examples/test_kissgp_gp_classification.py
+++ b/test/examples/test_kissgp_gp_classification.py
@@ -28,11 +28,7 @@ class GPClassificationModel(gpytorch.models.GridInducingVariationalGP):
         self.covar_module = RBFKernel(
             log_lengthscale_prior=SmoothedBoxPrior(exp(-5), exp(6), sigma=0.1, log_transform=True)
         )
-        self.register_parameter(
-            name="log_outputscale",
-            parameter=torch.nn.Parameter(torch.Tensor([0])),
-            prior=SmoothedBoxPrior(exp(-5), exp(6), sigma=0.1, log_transform=True),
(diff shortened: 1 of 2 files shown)


test/examples/test_kissgp_gp_classification.py
```diff
@@ -28,11 +28,7 @@ def __init__(self):
         self.covar_module = RBFKernel(
             log_lengthscale_prior=SmoothedBoxPrior(exp(-5), exp(6), sigma=0.1, log_transform=True)
         )
-        self.register_parameter(
-            name="log_outputscale",
-            parameter=torch.nn.Parameter(torch.Tensor([0])),
-            prior=SmoothedBoxPrior(exp(-5), exp(6), sigma=0.1, log_transform=True),
```
Balandat: Did you mean to get rid of the prior here?

jacobrgardner: Can we keep the prior and still use `self.my_param = ...` to create a parameter?

Balandat: Well you can, but you'd have to manually call `set_parameter_priors` to set the prior. The `self.my_param = ...` syntax doesn't really allow for a clean way to set a prior I think.

jacobrgardner: Yeah, I changed it here so that we had one unittest that did things that way

PR #184, opened 2018-07-19
Some fixes that didn't make it in the priors PR

Balandat:


jacobrgardner:
Going to assume the py2.7 test failure is transient

Balandat:
yeah can't see why it wouldn't be...

PR #187, opened 2018-07-20
Decrease test randomness

jacobrgardner:
So far on this branch I've done two things:

1. Uniformly locked down seeds for all tests using not just `torch.manual_seed` but also `torch.cuda.manual_seed_all` and `random.seed`.
2. For one of the historically flakier tests (the SM kernel), I saved a set of known working hyperparameter values to a file as a state dict, which I then load after optimization. Now for this test, the optimization loop essentially exists to test whether it errors or not, but the actual test error check should not depend on the random initialization.

Point 2 above might be a good thing for us to consider doing for more of the tests if they are flaky particularly often: we run the optimization loop, but the actual test error is checked using a set of known good hyperparameter values.


jacobrgardner:
I guess Travis has issues with the pth files

PR #188, opened 2018-07-20
Speed up eval of log_prob for InverseWishartPrior

Balandat:
With `A`, `B` positive definite we have `trace(AB^-1) = trace(B^-1A)`.

For 3x3 matrices:
```
%timeit torch.trace(A.matmul(B.inverse()))
35.5 µs ± 1.03 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit torch.trace(torch.gesv(A, B)[0])
24.9 µs ± 436 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```


jacobrgardner:
I'll merge this in. If we ever have larger or structured matrices (e.g., low rank for ICM kernel), we'll probably want to consider making use of a `LazyVariable.inv_matmul` to do this with CG.

PR #190, opened 2018-07-23
Cache covar inv root [WIP]

jacobrgardner:
This is the first PR on the road to efficient handling of fantasy inputs.

1. Exact GP models now cache not just the `covar_cache` when doing LOVE, but also the `train_train_inv_covar_root` computed as part of this cache (they are identical for exact GPs, but different for SKI).
2. The cached `train_train_inv_covar_root` is now used in part of a preconditioner for the computation of `mean_cache` in `exact_predictive_mean`. In this way, we often get the exact solve immediately without many extra iterations of CG.
    1. Predictions should be a fair bit faster overall when using LOVE now, since less work is done overall to compute the `mean_cache` at no additional cost.
3. In a future PR, we will efficiently update the `train_train_inv_covar_root`, and therefore both the actual `mean_cache` and `covar_cache`.

With this done, the roadmap for efficient fantasy inputs looks like:
1. Implement a basic interface for fantasy points that for now just recomputes `mean_cache` and `covar_cache` from scratch.
2. Implement efficient updates to `train_train_inv_covar_root` to account for the fantasy examples.
3. Use the updated `train_train_inv_covar_root` to efficiently update the actual `mean_cache` and `covar_cache` values.


gpleiss:
Looks good so far!

PR #192, opened 2018-07-24
Index kernel returns a lazy variable

gpleiss:
This should allow the use of index kernels (for multitask gps) with any of the scalable methods.

I'll also update the multitask notebook (or create a new notebook) that shows how to do scalable kronecker methods.


jacobrgardner:
Thanks Geoff! With this and a fix for the non lazy variable case, multi task should at least be working again.

Issue #200, opened 2018-07-25
test_spectral_mixture_gp_mean_abs_error numerically sensitive


colesbury:
Should TestSpectralMixtureGPRegression.test_spectral_mixture_gp_mean_abs_error use double-precision tensors?

I'm seeing failures in `TestSpectralMixtureGPRegression.test_spectral_mixture_gp_mean_abs_error` on machines that don't have the FMA3 instruction set. If you're using MKL, you can reproduce this by setting the environment variable `MKL_ENABLE_INSTRUCTIONS=AVX` to disable AVX2 and FMA3 instructions in MKL.

The difference is how this line is computed in `adam.py`, specifically the `add_` call:

`exp_avg.mul_(beta1).add_(1 - beta1, grad)`
https://github.com/pytorch/pytorch/blob/4b0098f3ae1a445760f1bf743e1accbf7ae70b56/torch/optim/adam.py#L92

In this case we have `beta1 = 0.9` and effectively compute:

```
exp_avg = exp_avg + 0.1 * grad
```

With FMA3 the multiply and add are computed with full precision and then rounded. Without the FMA3 instructions, the intermediate multiplication is rounded.

I'm also seeing failures with other seeds even with FMA3. For example, if you replace `4` with `7846462` in `TestSpectralMixtureGPRegression.setUp`.

```
ERROR: test_spectral_mixture_gp_mean_abs_error (__main__.TestSpectralMixtureGPRegression)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/examples/test_spectral_mixture_gp_regression.py", line 93, in test_spectral_mixture_gp_mean_abs_error
    loss = -mll(output, train_y)
  File "/data/users/sgross/gpytorch/gpytorch/module.py", line 192, in __call__
    outputs = self.forward(*inputs, **kwargs)
  File "/data/users/sgross/gpytorch/gpytorch/mlls/exact_marginal_log_likelihood.py", line 35, in forward
    inv_quad, log_det = covar.inv_quad_log_det(inv_quad_rhs=(target - mean).unsqueeze(-1), log_det=True)
  File "/data/users/sgross/gpytorch/gpytorch/lazy/lazy_variable.py", line 583, in inv_quad_log_det
    )(*args)
  File "/data/users/sgross/gpytorch/gpytorch/functions/_inv_quad_log_det.py", line 129, in forward
    eigenvalues, eigenvectors = lanczos_tridiag_to_diag(t_mat)
[...]

PR #201, opened 2018-07-25
Use SGD in test_spectral_mixture_gp_regression.py

colesbury:
I tried a few different seeds and the test passes more reliably. It also passes with the default seed (4) on machines that don't have FMA3 instructions.

See #200

### DIFF
--- a/test/examples/test_spectral_mixture_gp_regression.py
+++ b/test/examples/test_spectral_mixture_gp_regression.py
@@ -19,8 +19,6 @@ from gpytorch.priors import SmoothedBoxPrior
 from gpytorch.random_variables import GaussianRandomVariable
 from collections import OrderedDict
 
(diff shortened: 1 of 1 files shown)


jacobrgardner:
It looks like Travis is trying to run all tests in double precision with that change.

I've been playing around with this for a few minutes, and it also looks like swapping out Adam for SGD also makes the test a great deal more stable across a number of seeds, including with FMA3 instructions disabled.

colesbury:
I've updated the PR to use SGD instead of Adam.

For reference, I'm trying to figure out a fix because https://github.com/pytorch/pytorch/pull/8919 uses a PyTorch kernel for in-place addition instead of BLAS's saxpy_ (provided by MKL). The test fails when PyTorch is not compiled with AVX2/FMA support.

PR #205, opened 2018-07-27
Get priors/variational strategies also for children of torch Modules

### LINKED ISSUE #204: named_parameter_priors won't get priors for children of torch Modules
I think the PR #196  I did fixed things like DKL to at least run, but was a bad fix because we'll fail to get priors for parameters in gpytorch.Modules that are children of torch.nn.Modules. For example, consider AdditiveKernel, which places a bunch of Kernels in a ModuleList. Since named_parameter_priors is a recursive call that won't actually run on the ModuleList because it doesn't exist, we won't actually get priors for the component Kernels even if they are defined.

Ordinarily I'd just implement a fix, but I'm actually not sure how to deal with this one. Even if we implement an in house ModuleList extension, it seems dangerous and insidious to leave this failure case in, since the intention is that GPyTorch Modules are interoperable with torch ones. Therefore, a recursive version of named_parameter_priors seems not possible -- maybe we have to expand each child of a normal torch Module in the method directly?

@Balandat

Balandat:
Fixes #204 by moving to functional recursion instead of object method recursion.

### DIFF
--- a/gpytorch/module.py
+++ b/gpytorch/module.py
@@ -3,11 +3,13 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import torch
 from collections import OrderedDict
+
+import torch
 from torch import nn
-from .random_variables import RandomVariable
+
 from .lazy import LazyVariable
+from .random_variables import RandomVariable
 from .variational import VariationalStrategy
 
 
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Thanks for fixing this @Balandat!

I confirmed manually that this works with DKL/AdditiveKernel. This afternoon I'm planning to write up some DKL regression example notebooks, at which time I'll also add a DKL unit test that covers these things.

Balandat:
unit tests are good :)

PR #208, opened 2018-07-29
Kronecker multitask example notebook

gpleiss:
This replaced `multitask_gp_regression` notebook with two notebooks:
- `multitask_gp_regression` performs Kronecker multitask gps (e.g. all tasks per input)
- `hadamard_multitask_gp_regression` performs Hadamard multitask gps (e.g. one task per input)

We need to set up some syntactic sugar for Kronecker multitask GPs, because right now they're ugly. I'll open an issue for that.

For right now, multitask GPs can be accomplished with this notebook.


Balandat:
So the Kronecker example notebook still seems to fail to produce meaningful mean/covariance estimates. Not sure what's going on there...

jacobrgardner:
@Balandat Do you mean the notebook fails on the toy data provided or on something internal? For me, the notebook works fine in terms of the final plots

Balandat:
@jacobrgardner nvm the test notebook seems to work, I must have either been mistakenly using an old notebook, or the optimization got stuck. Will see if I can repro, but for now this seems to be working. Thanks!

PR #212, opened 2018-08-02
Fix root decomposition handling of batch dimensions when none exist.

jacobrgardner:
Our root decomposition function was keeping around batch dimensions even when the underlying base lazy variable had none, which was causing issues when backpropagating through samples.

@Balandat


Balandat:
woohoo!

PR #218, opened 2018-08-04
Better multitask GP interface, more efficient multitask GP predictions

### LINKED ISSUE #69: Multi dimension Target
Hi,
Thank you for amazing Repo. Is there an example which shows how to work with given code when the input and Target has dimensions like [NxM] where N is features and M number of samples. I checked the ones with multiple dimension inputs but it still has only one dimensional target.

Best,
Monica.

### LINKED ISSUE #143: AddedDiagLazyVariable should have a smarter add_diag function
This is something that will take me 5-10 minutes tops to fix when I actually go back to work tomorrow, but `AddedDiagLazyVariable` should just add the diagonal component passed to `add_diag` to the existing diagonal component in the lazy variable, and should support non constant diagonals.

This is super minor, but right now adding diagonal components other than the log noise (e.g. when #113 is implemented) makes the preconditioner do something kind of silly. Not broken, just silly :smiley:

### LINKED ISSUE #161: LazyEvaluatedKernelVariable diag runs out of memory for SKI
I pushed a temporary fix to this (a40392a8ce9c3c65a9679da21ec971783c9b6b81) that just special cases this specific issue, but the way we now handle lazily evaluating variances when we don't need the full covariance blows up for SKI.

This shouldn't be hard to fix, because the 'fix' implemented in the commit above does the right thing for SKI, anyways. It's really only exact GPs that benefit from the new `LazyEvaluatedKernel.diag()`.

Once I have a better solution, I'll revert the temporary fix, but it was a big enough problem that I wanted to get it solved immediately.

### LINKED ISSUE #209: Multitask GP syntactic sugar
Multitask GPs can be written in GPyTorch as is (see #208), but right now it is extremely ugly.

**Proposed interface**:
- Introduce a `MultitaskKernel` model, which wraps a standard kernel function and returns the Kronecker product between an index (task) kernel and the supplied input kernel.
- Mean functions take in an `n_tasks` argument, which defaults as None. If `n_tasks=d`, means will return an `n x d` output.
- Multiple outputs (for train_y) should be expressed in an `n x d` matrix.

**Necessary updates**:
- We introduce a syntactic sugar for `GaussianRandomVariables`. If the mean is an `n x d` matrix and the covariance is an `nd x nd` LazyVariable, we will figure things out 
  - `var`, `std`, etc. will return `n x d` matrices
  - `sample` will return an `n x d x s` tensor
  - Explicitly evaluating the covariance might return an `n x d x n x d` tensor, or maybe a `nd x nd` tensor? **Thoughts?** (Probably won't happen too often, but probably a good corner case to catch.)
- `inv_quad_log_det` is updated to handle this new syntactic sugar.

The hadamard multitask case (one task per input) will remain essentially the same.

cc @Balandat @darbour @jacobrgardner

### LINKED ISSUE #213: Multi-task likelihoods
Multi-Task models currently do not allow for different noise levels for the different outputs. This is clearly something that we should support.

One way of doing this would be to implement a MultiTaskLikelihood Module that operates on the possibly generalized outputs of MultiTask models discussed in #209. 

IMHO, this suggests that in order to make things clean we should treat multi-output models (and hece the RandomVariable interface) as first-class citizens. @gpleiss, @jacobrgardner, @darbour, any thoughts?

jacobrgardner:
Alright this one is pretty involved and touches a lot of open issues, but I'm pretty happy with how the new example notebooks turned out.

## Issues addressed
Fixes #213
Fixes #209
Fixes #161
Fixes #69 
Fixes #143 
## Breaking changes
- Batch GPs **MUST** now use 3 dimensional labels, e.g. `train_y` must be `b x n x 1`, and not `b x n`. The final singleton dimension was previously optional.
## Non-breaking internal changes
- `exact_predictive_mean` now takes an `n_train` argument.
- `exact_predictive_mean` and `exact_predictive_covar` take the `likelihood` rather than a `noise` parameter. The training covariance matrix (`train_train_covar`) is now augmented by passing it through the likelihood rather than with an `add_diag` call. 
- Although no likelihoods modify the mean currently, `train_mean` is also passed through the likelihood in `exact_predictive_mean`
- Kernels now define a `size(x1, x2)` method that returns the size of the covariance matrix that will result if the kernel is called with `x1` and `x2`. The default behavior for this is `x1.size(-2) x x2.size(-2)`, but may be overriden if the kernel expects to return a larger matrix. This is used by the new `MultitaskKernel`.
- Kernels now implement a `forward_diag` method that gets the diagonal of the covariance matrix only. The default behavior for this is the same as before; however, this is special cased by both `GridInterpolationKernel` and `MultitaskKernel`. This fixes #161 
- `ExactGP` and `ExactMarginalLogLikelihood` now have some small custom logic to handle `MultitaskGaussianRandomVariables` (explained below).
## New user-side objects
- Implemented **`MultitaskKernel`**. Given a base data kernel and a number of tasks, this kernel computes the multi task kernel using the data kernel and an index kernel.
- Implemented **`MultitaskMean`**. This mean function can take either a list of mean functions (of size `n_tasks`) or a single mean function. Calling this mean function returns an `n x t` mean matrix, the result of calling each of the base mean functions (or the single mean function `t` times) on the supplied input.
- Implemented **`MultitaskGaussianRandomVariable`**. This is syntactic sugar that handles `n x t` mean / variances and sampling. See #209
- Implemented **`MultitaskGaussianLikelihood`**. This allows users to, in addition to using a shared log noise (which we keep around to be able to use preconditioning still), use a different log_noise for each task.
## Performance improvements
- Compared to the existing Kronecker implementation, the `MultitaskKernel` makes more effective use of lazy kernel evaluation, which allows us to avoid computing the full `(n_train + n_test) x (n_train + n_test)` covariance matrix in the exact multitask GP setting by computing the Kronecker product in the forward method (which results in small rectangular matrices) instead of after it (which results in large square matrices). As a result, predictions on large test sets should be much faster.
## New example notebooks
- `multitask_gp_regression.ipynb` Demonstrates the basic new interface for multitask GPs
- `multitask_ski_gp_regression.ipynb` Demonstrates this interface working with SKI.
## New unit tests
- Added unit tests for both the Kronecker multitask exact GP and multitask SKI setting that use all four of the new multitask objects.


jacobrgardner:
cc @Balandat @gpleiss

gpytorch/lazy/lazy_variable.py
```diff
@@ -400,20 +400,26 @@ def exact_predictive_mean(self, full_mean, train_labels, noise, precomputed_cach
...
             :obj:`torch.tensor`: The predictive posterior mean of the test points
         """
-        n_train = train_labels.size(-1)
+        from ..random_variables import GaussianRandomVariable
+
         if precomputed_cache is None:
             train_mean = full_mean.narrow(-1, 0, n_train)
             if self.ndimension() == 3:
-                train_train_covar = self[:, :n_train, :n_train].add_diag(noise)
+                train_train_covar = self[:, :n_train, :n_train]
```
Balandat: enough to assign train_train_covar in the if/else clauses, no need to repeat the rest of the code

gpytorch/lazy/lazy_variable.py
```diff
@@ -469,12 +475,14 @@ def exact_predictive_covar(self, n_train, noise, precomputed_cache=None):
             :obj:`gpytorch.lazy.LazyVariable`: A LazyVariable representing the predictive posterior covariance of the
                                                test points
         """
+        from ..random_variables import GaussianRandomVariable
+
         if self.ndimension() == 3:
-            train_train_covar = self[:, :n_train, :n_train].add_diag(noise)
+            train_train_covar = likelihood(GaussianRandomVariable(torch.zeros(1), self[:, :n_train, :n_train])).covar()
             test_train_covar = self[:, n_train:, :n_train]
```
Balandat: again no need to repeat the same code here, move outside if/else

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -22,7 +22,7 @@ def forward(self, input):
         assert isinstance(input, GaussianRandomVariable)
         mean, covar = input.representation()
         noise = add_diag(covar, self.log_noise.exp())
-        return GaussianRandomVariable(mean, noise)
+        return input.__class__(mean, noise)
```
Balandat: this seems unnecessary general, given that we explicitly check that `input` is an GaussianRandomVariable above

gpytorch/likelihoods/multitask_gaussian_likelihood.py
```diff
@@ -0,0 +1,31 @@
...
+from gpytorch.functions import add_diag
+from gpytorch.likelihoods import GaussianLikelihood
+from gpytorch.lazy import DiagLazyVariable, KroneckerProductLazyVariable
+
+
+class MultitaskGaussianLikelihood(GaussianLikelihood):
+    def __init__(self, n_tasks, log_noise_prior=None, log_task_noises_prior=None):
+        # TODO: Remove deprecated log_noise_bounds kwarg
+        super(MultitaskGaussianLikelihood, self).__init__(log_noise_prior=log_noise_prior)
+        self.register_parameter(name="log_task_noises",
```
Balandat: formatting nit

```
self.register_parameter(
    name="log_task_noises",
    parameter=torch.nn.Parameter(torch.zeros(n_tasks)),
    prior=log_task_noises_prior,
)
```

gpytorch/likelihoods/multitask_gaussian_likelihood.py
```diff
@@ -0,0 +1,31 @@
...
+                                parameter=torch.nn.Parameter(torch.zeros(n_tasks)),
+                                prior=log_task_noises_prior)
+        self.n_tasks = n_tasks
+
+    def forward(self, input):
+        mean, covar = input.representation()
+        eye_lv = DiagLazyVariable(torch.ones(int(covar.size(-1) / self.n_tasks), device=self.log_noise.device))
```
Balandat: use `covar.size(-1) // self.n_tasks` instead?

gpytorch/likelihoods/multitask_gaussian_likelihood.py
```diff
@@ -0,0 +1,31 @@
...
+from __future__ import unicode_literals
+
+import torch
+from gpytorch.functions import add_diag
+from gpytorch.likelihoods import GaussianLikelihood
+from gpytorch.lazy import DiagLazyVariable, KroneckerProductLazyVariable
+
+
+class MultitaskGaussianLikelihood(GaussianLikelihood):
+    def __init__(self, n_tasks, log_noise_prior=None, log_task_noises_prior=None):
```
Balandat: Great that we have separate noise levels for the tasks. But if we also have a global log_noise prior don't we end up with an identifiability problem? That is, by scaling the noise by `alpha` and each task_noise by `1/alpha` we get a continuum of solutions for the MLL MAP estimate. 

Can we just get rid of the noise prior? To keep the API simple, we could still allow to pass in a single prior and then apply that to each of the task noises internally. 

I guess more generally, I wonder whether this works with a WhiteNoiseKernel when we provide variances for the observations (each task may have different observed variances)?

gpytorch/means/multitask_mean.py
```diff
@@ -0,0 +1,27 @@
...
+        super(MultitaskMean, self).__init__()
+
+        if isinstance(base_means, Mean):
+            base_means = [base_means]
+
+        if not isinstance(base_means, list) or (len(base_means) != 1 and len(base_means) != n_tasks):
+            raise RuntimeError("base_means should be a list of means of length either 1 or n_tasks")
+
+        if len(base_means) == 1:
+            base_means = base_means * n_tasks
```
Balandat: It's a trap! This will not actually instantiate new mean objects, but put references to the same object into a list, so the different tasks would mechanically share the same parameters. Instead we;d have do something like this:

```
base_means = base_means + [deepcopy(base_means[0] for i in range(n_tasks - 1)]
```
I guess we'd have to put the means into a ModuleList then.

test/examples/test_kronecker_multitask_gp_regression.py
```diff
@@ -0,0 +1,97 @@
...
+import os
+import random
+import torch
+import unittest
+import gpytorch
+from gpytorch.kernels import RBFKernel, MultitaskKernel
+from gpytorch.means import ConstantMean, MultitaskMean
+from gpytorch.likelihoods import MultitaskGaussianLikelihood
+from gpytorch.random_variables import MultitaskGaussianRandomVariable
+# Simple training data: let's try to learn a sine function
```
Balandat: nit: add 2 blank lines

Balandat:
cc @darbour, @rajkumarkarthik, @bletham

gpytorch/likelihoods/gaussian_likelihood.py
jacobrgardner: Right now, `MultitaskGaussianRandomVariable` is a `GaussianRandomVariable`. I was doing this to preserve that (this way, `GaussianLikelihood` can still be used with multitask if you don't need different task variances

jacobrgardner:
3d4c59e13380cf99f54508c588751f7365c912dd also fixes #143 and lets us potentially remove the shared log noise in `MultitaskGaussianLikelihood` in the future.

gpytorch/means/multitask_mean.py
jacobrgardner: I was thinking the default behavior was to have a constant mean shared across tasks if the user only specifies one mean, otherwise they pass in a list of different ones.

Should the default behavior instead be to have one mean per task? I guess that makes sense. We'll also need to use a ModuleList either way, though.

gpytorch/lazy/interpolated_lazy_variable.py
```diff
@@ -344,10 +344,10 @@ def diag(self):
                 res = res.view(batch_size, n_data, -1).sum(-1)
             return res
 
-    def exact_predictive_mean(self, full_mean, train_labels, noise, precomputed_cache=None):
-        n_train = train_labels.size(-1)
+    def exact_predictive_mean(self, full_mean, train_labels, n_train, likelihood, precomputed_cache=None):
+        from ..random_variables import GaussianRandomVariable
```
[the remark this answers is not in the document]
jacobrgardner: `GaussianRandomVariable` imports lazy variables, so everything breaks if this is a file level import.

gpytorch/kernels/multitask_kernel.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def _eval_covar_matrix(covar_factor, log_var):
+    return covar_factor.matmul(covar_factor.transpose(-1, -2)) + log_var.exp().diag()
+
+
+class MultitaskKernel(Kernel):
+    def __init__(self, data_covar_module, n_tasks, rank=1, task_covar_prior=None):
+        super(MultitaskKernel, self).__init__()
+        self.task_covar_module = IndexKernel(n_tasks=n_tasks, rank=rank, prior=task_covar_prior)
```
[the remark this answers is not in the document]
jacobrgardner: Done!

jacobrgardner:
The IndexKernel log_var term is something that will get applied to the covariance between any two points that share a task, including test points or points where `x_i != x_j`. I think it therefore serves a different purpose than the `log_task_noises` in the new likelihood, which only gets added when `x_i = x_j` and only to the variances of training examples.

gpytorch/kernels/multitask_kernel.py
```diff
@@ -0,0 +1,50 @@
...
+    def forward_diag(self, x1, x2):
+        x1 = x1.transpose(-2, -3)
+        x2 = x2.transpose(-2, -3)
+        task_indices = torch.range(0, self.n_tasks - 1, device=x1.device).long()
+        task_indices = task_indices.unsqueeze(-1).unsqueeze(-1)
+        covar_i = self.task_covar_module(task_indices).evaluate_kernel()
+        covar_x = self.data_covar_module.forward(x1, x2)
+        if not isinstance(covar_x, LazyVariable):
+            covar_x = NonLazyVariable(covar_x)
+        res = KroneckerProductLazyVariable(covar_i, covar_x).matmul(torch.eye(1, device=x1.device)).unsqueeze(-1)
```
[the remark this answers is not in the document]
jacobrgardner: Rewrote the method to be a bit more clear.

gpytorch/kernels/index_kernel.py
```diff
@@ -14,6 +14,8 @@ def _eval_covar_matrix(covar_factor, log_var):
 
 class IndexKernel(Kernel):
     def __init__(self, n_tasks, rank=1, prior=None, active_dims=None):
+        if rank > n_tasks:
+            raise RuntimeError("Cannot create a task covariance matrix larger than the number of tasks")
```
Balandat: I guess that's not exactly what it would be doing - it would just make the covar_factor fat, and so introduce unnecessary parameters - the matrix will still be of the correct size.

@rajkumarkarthik is pretty much done with an implementation of an LKJ prior (evaluate either over the full correlation matrix, or over its Cholesky factor). 

How do you feel about, in the case when `n_tasks = rank`, to parameterize the task kernel via its Cholesky factor? I guess the challenge then would be to enforce that during inference we stay in the subspace of Cholesky factors: `{L in R^{nxn} : L_ij = 0 for j>1, L_ii >0, and sum_j(Lij) = 1}`

jacobrgardner: We parameterize our variational distributions by the Cholesky decomposition, so we could borrow code from that. Maybe we can open another issue for an IndexKernel refactor and push that to another PR since this one is pretty big already? I think the last issue to settle for this one is how the MultitaskMean should function (and then to actually use a ModuleList either way we go with it).

gpytorch/means/multitask_mean.py
Balandat: Yeah I think there should be one mean per task, in the basic case of a constant mean it seems that this would effectively amount to a bias kernel. With fancier mean functions I guess it's up to the user to do something reasonable.

gpytorch/kernels/index_kernel.py
Balandat: Yeah sg I'll open up a new issue.

Also just commented on the MultitaskMean. I'll try to play around with this tomorrow a bit to make sure everything runs as expected for the things I've set up.

jacobrgardner: No worries. It is the weekend after all -- I just had nothing better to do today than finish this up :-)

jacobrgardner:
Just made a small change to `WhiteNoiseKernel` that should let it support multitask with something as simple as (assuming `training_noises` are `n x t`):
```python
covar_x = MultitaskKernel(RBFKernel(), n_tasks=2, rank=1) + WhiteNoiseKernel(training_noises)
```
Assuming that all works out, we can also modify it to handle both batched data and multitask by allowing variances to be 3 dimensional.

Balandat:
Note that the MultiTaskMean will fail when using a ZeroMean - I guess that's due to the inconsistency between the forward outputs of ConstantMean and ZeroMean?

gpytorch/kernels/multitask_kernel.py
```diff
@@ -38,21 +38,24 @@ def forward_diag(self, x1, x2):
         in :class:`gpytorch.kernels.Kernel` because we need to take the Kronecker product of the diagonals of the
         base data kernel and the task kernel.
         """
-        x1 = x1.transpose(-2, -3)
-        x2 = x2.transpose(-2, -3)
         task_indices = torch.arange(self.n_tasks, device=x1.device).long()
```
Balandat: `long()` shouldn't be necessary here

gpytorch/kernels/multitask_kernel.py
```diff
@@ -38,21 +38,24 @@ def forward_diag(self, x1, x2):
...
         task_indices = task_indices.unsqueeze(-1).unsqueeze(-1)
 
         # These are small because they are vectors, therefore it is safe to evaluate them
-        covar_i = self.task_covar_module(task_indices).evaluate().squeeze(-1)
-        covar_x = self.data_covar_module.forward(x1, x2)
-        if isinstance(covar_x, LazyVariable):
-            covar_x = covar_x.evaluate()
-        covar_x = covar_x.squeeze(-1)
+        covar_i_diag = self.task_covar_module.forward_diag(task_indices)
```
Balandat: This raises a `TypeError: forward_diag() missing 1 required positional argument: 'x2'`

Balandat: however, if we replace this by `forward_diag(task_indices, task_indices)` we run into size incompatibilities in `observed_pred.confidence_region()` in the example nb.

jacobrgardner: Sorry about that, should be fixed

gpytorch/kernels/multitask_kernel.py
jacobrgardner: torch.arange doesn't produce a long tensor by default:
```python
In [2]: torch.arange(5).dtype
Out[2]: torch.float32
```

Balandat: huh. here's my output:
```
In [2]: torch.arange(5).dtype
Out[2]: torch.int64
```
I guess there must have been a change on pytorch master... let's be safe for now

gpytorch/kernels/multitask_kernel.py
```diff
@@ -39,10 +39,10 @@ def forward_diag(self, x1, x2):
         base data kernel and the task kernel.
         """
         task_indices = torch.arange(self.n_tasks, device=x1.device).long()
-        task_indices = task_indices.unsqueeze(-1).unsqueeze(-1)
```
Balandat: both this change and the one below have non-unix line endings

Balandat: Actually, the whole `multitask_kernel.py` file uses non-unix line endings throughout.

jacobrgardner: I actually have no idea what is causing this -- both atom and git are setup to stop this, and it obviously works a majority of the time. 😕

gpytorch/kernels/multitask_kernel.py
```diff
@@ -0,0 +1,80 @@
...
+        covar_i = self.task_covar_module(task_indices).evaluate_kernel()
+        covar_x = self.data_covar_module.forward(x1, x2)
+        if covar_x.size(0) == 1:
+            covar_x = covar_x[0]
+        if not isinstance(covar_x, LazyVariable):
+            covar_x = NonLazyVariable(covar_x)
+        res = KroneckerProductLazyVariable(covar_i, covar_x)
+        return res
+
+    def size(self, x1, x2):
```
gpleiss: This is a nice solution 👍

gpleiss:
@Balandat I think that's a bug in zero mean. I can make a separate PR.

Balandat:
This looks awesome! 

I've run into some issues with bad model fits when messing around with this over the weekend, so let me see if I can repro those issues before we merge this in.

Balandat:
So I haven't been able to repro the issues I saw, so from my end this seems good to go in. If you think this is ready go ahead and merge!

gpytorch/means/multitask_mean.py
```diff
@@ -0,0 +1,43 @@
...
+        super(MultitaskMean, self).__init__()
+
+        if isinstance(base_means, Mean):
+            base_means = [base_means]
+
+        if not isinstance(base_means, list) or (len(base_means) != 1 and len(base_means) != n_tasks):
+            raise RuntimeError("base_means should be a list of means of length either 1 or n_tasks")
+
+        if len(base_means) == 1:
+            base_means = base_means + deepcopy([base_means[i] for i in range(n_tasks - 1)])
```
Balandat: this should be `[deepcopy(base_means[0]) for i in range(n_tasks - 1)]` instead

Balandat:
about to push the fix to master

PR #221, opened 2018-08-06
Speed up Root/Matmul LV get_indices

gpleiss:
- When calling `get_indices` with many indices, explicitly evaluate the LazyVariable.
- Remove old Variable references

`_get_indices` is called during the diag method for variational inference. If we want `m` indices, and the RootLazyVariable is a `n x r` matrix, then the standard `_get_indices` call requires `m * r` memory. If `m` is large (which is usually the case for variational), then this will be extremely memory intensive.

Fix: do whatever requires the least amount of memory:
- If `m * r` > `n x n`, then explicitly evaluate the LazyVariable and then call `_get_indices`.
- Otherwise, use the original `_get_indices` routine.

### DIFF
--- a/gpytorch/lazy/matmul_lazy_variable.py
+++ b/gpytorch/lazy/matmul_lazy_variable.py
@@ -4,7 +4,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import torch
-from torch.autograd import Variable
 from .lazy_variable import LazyVariable
 from .non_lazy_variable import NonLazyVariable
 
@@ -27,6 +26,18 @@ class MatmulLazyVariable(LazyVariable):
         super(MatmulLazyVariable, self).__init__(lhs, rhs)
         self.lhs = lhs
         self.rhs = rhs
+
+    @property
+    def _evaluated_lhs(self):
+        if not hasattr(self, '_evaluated_lhs_memo'):
+            self._evaluated_lhs_memo = self.lhs.evaluate()
+        return self._evaluated_lhs_memo
+
+    @property
+    def _evaluated_rhs(self):
+        if not hasattr(self, '_evaluated_rhs_memo'):
+            self._evaluated_rhs_memo = self.rhs.evaluate()
+        return self._evaluated_rhs_memo
 
     def _matmul(self, rhs):
         return self.lhs._matmul(self.rhs._matmul(rhs))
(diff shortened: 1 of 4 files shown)


gpytorch/lazy/matmul_lazy_variable.py
```diff
@@ -54,38 +65,48 @@ def _transpose_nonbatch(self, *args):
...
+            torch.arange(0, inner_size, out=inner_indices.data)
 
-        left_vals = self.lhs._get_indices(
-            _outer_repeat(left_indices, inner_size), _inner_repeat(inner_indices, outer_size)
-        )
-        right_vals = self.rhs._get_indices(
-            _inner_repeat(inner_indices, outer_size), _outer_repeat(right_indices, inner_size)
-        )
+            left_vals = self.lhs._get_indices(
+                _outer_repeat(left_indices, inner_size), _inner_repeat(inner_indices, outer_size)
```
Balandat: nit: you're re-using `_inner_repeat(inner_indices, outer_size)` here, no need to compute it twice, just compute once and assign a variable

gpytorch/lazy/root_lazy_variable.py
```diff
@@ -56,36 +61,46 @@ def _transpose_nonbatch(self):
...
+            return self._evaluated[batch_indices, left_indices, right_indices]
 
-        return (left_vals.view(-1, inner_size) * right_vals.view(-1, inner_size)).sum(-1)
+        else:
+            outer_size = batch_indices.size(0)
+            inner_size = self.root.size(-1)
+            inner_indices = right_indices.new(inner_size)
+            torch.arange(0, inner_size, out=inner_indices.data)
+
+            left_vals = self.root._batch_get_indices(
```
Balandat: again you can avoid re-doing computations here

gpytorch/lazy/root_lazy_variable.py
```diff
@@ -56,36 +61,46 @@ def _transpose_nonbatch(self):
...
+            return self._evaluated[left_indices, right_indices]
+
+        else:
+            outer_size = left_indices.size(0)
+            inner_size = self.root.size(-1)
+            inner_indices = right_indices.new(inner_size)
+            torch.arange(0, inner_size, out=inner_indices.data)
+
+            left_vals = self.root._get_indices(
+                _outer_repeat(left_indices, inner_size), _inner_repeat(inner_indices, outer_size)
```
Balandat: and here...

test/lazy/test_matmul_lazy_variable.py
```diff
@@ -5,42 +5,44 @@
...
-from torch.autograd import Variable
 from gpytorch.lazy import MatmulLazyVariable
 from gpytorch.utils import approx_equal
 
 
 class TestMatmulLazyVariable(unittest.TestCase):
     def test_matmul(self):
-        lhs = Variable(torch.randn(5, 3), requires_grad=True)
-        rhs = Variable(torch.randn(3, 4), requires_grad=True)
+        lhs = torch.randn(5, 3, requires_grad=True)
```
Balandat: nice!

gpytorch/lazy/matmul_lazy_variable.py
gpleiss: Oops thanks for catching this!

gpleiss:
@Balandat removed those repeated computations

PR #222, opened 2018-08-06
Zero mean should produce a 1D output

gpleiss:
This addresses an issue pointed out in #218 

cc/ @Balandat @jacobrgardner

### DIFF
--- a/test/examples/test_kissgp_additive_regression.py
+++ b/test/examples/test_kissgp_additive_regression.py
@@ -14,7 +14,7 @@ from torch import optim
 from torch.autograd import Variable
 from gpytorch.kernels import RBFKernel, AdditiveGridInterpolationKernel
 from gpytorch.likelihoods import GaussianLikelihood
-from gpytorch.means import ConstantMean
+from gpytorch.means import ZeroMean
 from gpytorch.priors import SmoothedBoxPrior
 from gpytorch.random_variables import GaussianRandomVariable
 
@@ -27,7 +27,9 @@ for i in range(n):
         train_x[i * n + j][0] = float(i) / (n - 1)
         train_x[i * n + j][1] = float(j) / (n - 1)
 train_x = Variable(train_x)
-train_y = Variable((torch.sin(train_x.data[:, 0]) + torch.cos(train_x.data[:, 1])) * (2 * pi))
+train_y = Variable(
+    (torch.sin(train_x.data[:, 0]) + torch.cos(train_x.data[:, 1])) * (2 * pi)
+)
 
 m = 10
 test_x = torch.zeros(pow(m, 2), 2)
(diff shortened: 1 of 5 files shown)


test/examples/test_kissgp_additive_regression.py
```diff
@@ -14,7 +14,7 @@
 from torch.autograd import Variable
 from gpytorch.kernels import RBFKernel, AdditiveGridInterpolationKernel
 from gpytorch.likelihoods import GaussianLikelihood
-from gpytorch.means import ConstantMean
+from gpytorch.means import ZeroMean
```
Balandat: should we run the test against both ZeroMean and ConstantMean?

test/examples/test_kissgp_additive_regression.py
```diff
@@ -14,7 +14,7 @@
 from torch.autograd import Variable
```
Balandat: Variable is deprecated - would you mind mind removing it from this example while you're working on this?

test/examples/test_kissgp_additive_regression.py
gpleiss: All of our other tests use ConstantMean - I just wanted there to be one test with ZeroMean.

gpleiss:
@Balandat added unit tests, removed variable

test/examples/test_kissgp_additive_regression.py
```diff
@@ -75,12 +77,16 @@ def test_kissgp_gp_mean_abs_error(self):
         gp_model = GPRegressionModel(train_x.data, train_y.data, likelihood)
         mll = gpytorch.mlls.ExactMarginalLogLikelihood(likelihood, gp_model)
 
-        with gpytorch.settings.max_preconditioner_size(10), gpytorch.settings.max_cg_iterations(30):
+        with gpytorch.settings.max_preconditioner_size(
```
Balandat: this is pretty ugly - is this what black suggests?

jacobrgardner:
I've been running `black -l 120` for what it's worth, I much prefer 120 chars per line to 80

test/means/test_multitask_mean.py
```diff
@@ -0,0 +1,52 @@
...
+        a = torch.Tensor([[[1, 2], [1, 2], [2, 4]], [[2, 3], [2, 3], [1, 3]]])
+        res = self.mean(a)
+        self.assertEqual(tuple(res.size()), (2, 3, 4))
+        self.assertTrue(res[:, :, 0].eq(5).all())
+        self.assertTrue(res[:, :, 1].eq(0).all())
+        self.assertTrue(res[:, :, 2].eq(0).all())
+        self.assertTrue(res[:, :, 3].eq(7).all())
+
+
+class TestMultitaskMeanSameMean(unittest.TestCase):
```
Balandat: Can we add another test function in here that makes sure we're actually deepcopying the mean? That is, make sure "same mean" here means that this is a notational convenience rather than actually having the same mean for all tasks. Basically, just fill the data values for the means of a couple of different tasks with different values, and make sure the outputs are correct and distinct.

Balandat:
yeah `black -l 120` makes sense, this with clause just comes out looking pretty weird...

Balandat:
test failure seems unrelated, but can you still lock the seed to a different value so this passes?

gpleiss:
@Balandat changed the seed - tests pass.

PR #223, opened 2018-08-08
Prior mode for exact GPs

### LINKED ISSUE #173: Support predictions in prior mode
Right now, it's essentially impossible to use an `ExactGP` to get predictions from the prior, because (1) we require training data in exact GP models, (2) in `train` mode we ensure that the supplied data is the same as the training data, and (3) in `eval` mode we compute the predictive posterior.

Some applications (like BayesOpt) don't necessarily start with training data at all, so getting predictions from the prior can be useful for choosing initial points, for example.

gpleiss:
Can now construct an exact GP with `train_inputs=None, train_targets=None`.
That GP in `eval()` mode will return the prior
That GP in `train()` mode will return an error

Also introduces a debug mode, which suppresses warnings in exact GPs.
Cleans up some old pytorch 0.3 references in tests as well

Closes #173

### DIFF
--- a/gpytorch/functions/__init__.py
+++ b/gpytorch/functions/__init__.py
@@ -81,6 +81,9 @@ def exact_predictive_mean(full_covar, full_mean, train_labels, n_train, likeliho
     Returns:
     - (t) - the predictive posterior mean of the test points
     """
+    if not n_train:
+        return full_mean, None
+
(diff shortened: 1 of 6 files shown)


Balandat:
lgtm, thanks! Merging this in to move ahead.

Let's try in the future to improve unit test coverage in lighter-weight tests outside of the examples though: #224

Issue #224, opened 2018-08-08
Improve unit test coverage outside of examples


Balandat:
The examples take quite a while to run, let's try to get 100% unit test coverage with lightweight tests outside of the examples.

gpleiss:
LazyTensors are now in a good state. We should make similar base test cases for our other objects:

- [ ] Kernels
- [ ] Models (we have a base case in place - we now just need to use it)
- [ ] Likelihoods (maybe?)
- [ ] Priors (maybe?)
- [ ] Mean functions (maybe?)

jacobrgardner:
This is now done to enough of an extent that any specific uncovered code should be raised in more specific issues.

PR #225, opened 2018-08-08
Add suite of unit tests for WhiteNoiseKernel and ZeroLazyVariable

jacobrgardner:
This also fixes some issues that turned up with them in writing the unit tests for the batch setting. 

This PR begins work on #224. I think it is best to make PRs for unit tests as I make them rather than trying to do a big push for them all in one PR, since I may also be able to fix minor issues as I make them as in this PR.

cc/ @gpleiss @Balandat

### DIFF
--- a/gpytorch/kernels/white_noise_kernel.py
+++ b/gpytorch/kernels/white_noise_kernel.py
@@ -13,16 +13,25 @@ class WhiteNoiseKernel(Kernel):
         super(WhiteNoiseKernel, self).__init__()
         self.register_buffer("variances", variances)
 
+        # Enforce that variances are batch_size x data_size x task_size
+        # where batch_size and task_size are both 1 in the simplest case.
+        if variances.ndimension() == 3:
+            self.is_batch = True
+            self.variances = variances
+        elif variances.ndimension() == 2:
+            # By default, assume 2 dimensional data means data_size x task_size, and add a batch_size of 1
+            self.is_batch = False
+            self.variances = variances.unsqueeze(0)
+        else:
+            self.is_batch = False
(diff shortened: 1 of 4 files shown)


jacobrgardner:
Actually, this WhiteNoiseKernel will still do the wrong thing in the multitask setting, which is what I'm doing unit tests for now. 

I am going to propose that we start enforcing only 3 dimensional ever means batch mode.

gpytorch/kernels/white_noise_kernel.py
```diff
@@ -13,16 +13,24 @@ def __init__(self, variances):
         super(WhiteNoiseKernel, self).__init__()
         self.register_buffer("variances", variances)
 
+        # Enforce that variances are batch_size x data_size x task_size
```
jacobrgardner: @Balandat @gpleiss What do you guys think of this pattern? For any parameter we define internally, we always reshape it to have 3 dimensions that denote `batch_size x data_size x task_size`. If a user initializes with a 3 dimensional tensor, we keep it as is. If they pass in a 2 dimensional tensor, we assume it is `data_size x task_size` and unsqueeze a singleton batch dimension. If they pass in a  1 dimensional tensor, we assume it is `data_size` and unsqueeze it to `1 x data_size x 1`.

Balandat: Yeah I like this pattern, I think I had suggested that in an earlier discussion.

jacobrgardner: Great. In that case, I think these are all ready to go. I'll work on some multitask unit tests next.

PR #226, opened 2018-08-08
Added LKJCovariancePrior implementation

rajkumarkarthik:
LKJCovariancePrior is added as prior for ICM kernel. It uses an LKJPrior over correlation matrices along with a prior for marginal standard deviations.

There is also an implementation of LKJCholeskyFactorPrior, which is basically the LKJPrior but over the Cholesky decompoosition, and it is unused currently.

### DIFF
--- a/gpytorch/priors/lkj_prior.py
+++ b/gpytorch/priors/lkj_prior.py
@@ -30,17 +30,16 @@ class LKJPrior(Prior):
         # Normalization constant
         # Reference: Bayesian Data Analysis, 3rd ed., Gelman et al., p. 576
         i = torch.arange(n).float()
-        C = (
-            torch.sum((2 * eta - 2 + i) * i) * math.log(2) +
-            n * torch.sum(2 * torch.lgamma(i / 2 + 1) - torch.lgamma(i + 2))
-        )
+        C = (torch.sum((2 * eta - 2 + i) * i) * math.log(2) +
+             n * torch.sum(2 * torch.lgamma(i / 2 + 1) - torch.lgamma(i + 2)))
         self.register_buffer("C", C)
         self._log_transform = False
 
     def _log_prob(self, parameter):
         if not is_valid_correlation_matrix(parameter):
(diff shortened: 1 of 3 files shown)


gpytorch/priors/lkj_prior.py
```diff
@@ -0,0 +1,121 @@
...
+            torch.sum((2 * eta - 2 + i) * i) * math.log(2) +
+            n * torch.sum(2 * torch.lgamma(i / 2 + 1) - torch.lgamma(i + 2))
+        )
+        self.register_buffer("C", C)
+        self._log_transform = False
+
+    def _log_prob(self, parameter):
+        if not is_valid_correlation_matrix(parameter):
+            raise ValueError("Input is not a valid correlation matrix")
+        return self.C + (self.eta - 1) * torch.log(torch.det(parameter))
```
jacobrgardner: `torch.log(torch.det(parameter))` is probably fine for very small matrices, but in general is not a good way to compute log determinants for positive definite matrices. `logdet = parameter.potrf().diag().log().sum() * 2` is much better. Example:

```python
mat_root = torch.randn(50, 50)
covar_mat = mat_root.t().matmul(mat_root)
log_det_bad = torch.log(torch.det(covar_mat))
log_det_good = covar_mat.potrf().diag().log().sum() * 2
print(log_det_bad) # Outputs tensor(inf)
print(log_det_good) # Outputs log | covar_mat |
```

rajkumarkarthik:
Didn't realize that. Thanks for pointing it out! I updated the way the log determinants are computed now.

Balandat:
this lgtm. I will add unit tests for these priors once this PR has gone in.

@gpleiss, @jacobrgardner I'll leave you to take another look and merge this.

jacobrgardner:
Other than the log det issue (which we should also fix for the Wishart prior) I didn't see anything problematic. I am happy to merge this in now.

Balandat:
sg I'll fix for Wishart and add unit tests in #227

PR #228, opened 2018-08-09
Detach mean and covar caches at test time

### LINKED ISSUE #175: Possible extra derivative calculations at test time
See the discussion in #174. When calling backward to get the derivative of the mean function with respect to a test input, we enter the backward pass of `InvMatmul`. There doesn't seem to be an obvious reason we should be doing this, since none of the linear solves involved in the predictive __mean__ depend on the test point. 

This suggests we might also he computing derivatives with respect to model hyperparameters, which would be wasted computation.

@Balandat

jacobrgardner:
This PR detaches the mean and covariance caches computed at test time. This has two effects:

First and most importantly, backpropagating through GP predictions is bugged on master for things like BayesOpt. Specifically, a pattern like:

```python
from batch_acquisition import batch_expected_improvement
x = torch.tensor([0.25], requires_grad=True)
optimizer = torch.optim.Adam([x], lr=0.1)
for i in range(50):
    optimizer.zero_grad()
    loss = -batch_expected_improvement(x, model, 0)
    print('Iter {}, loss = {}'.format(i, loss))
    loss.backward()
    optimizer.step()
```

throws a runtime error on master (technically the gpytorchopt branch), because the caches require grad due to the model parameters but aren't created by the call to the acquisition function. Therefore, when `backward()` is called, intermediate things are deleted that aren't ever recreated in this loop, so calling `backward()` a second time is not possible without specifying `retain_graph=True`.

Second, this should be a performance improvement for derivatives through predictions, as we no longer enter backwards calls of expensive functions like `InvMatmul`.

Fixes #175

### DIFF
--- a/gpytorch/lazy/lazy_variable.py
+++ b/gpytorch/lazy/lazy_variable.py
@@ -427,7 +427,7 @@ class LazyVariable(object):
         if res.ndimension() == 3:
             res = res.squeeze(-1)
         res = res + test_mean
-        return res, precomputed_cache
+        return res, precomputed_cache.detach()
 
     def _exact_predictive_covar_inv_quad_form_cache(self, train_train_covar_inv_root, test_train_covar):
         """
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Worth noting that we were already doing things this way in the SKI case:

https://github.com/cornellius-gp/gpytorch/blob/bab8c0469b7ff7d4cbc944e436cf8a379bae8840/gpytorch/lazy/interpolated_lazy_variable.py#L377

Courtesy of 8fe38e41

gpleiss:
Small note, but we could make those `.detach_()` for slightly more efficiency

jacobrgardner:
I tried. Since these are occasionally views, we'd have to do `.contiguous()` before the inplace operation, which would eliminate the efficiency savings anyways 😢

PR #230, opened 2018-08-15
fix conditional in gpytorch.kernels.Kernel.__call__ when active_dims is set

bdecost:
I think the requested changes are relatively straightforward. I ran into this when trying to evaluate individual components of an `AdditiveKernel` instance.

In the current master branch, if `x2_` is `None` when `active_dims` is set, `x2` never gets set, so the conditional at [line 112](https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/kernels/kernel.py#L112) gets skipped over.

`AdditiveKernel` doesn't trigger this behavior because it passes `x1` and `x2` explicitly.

My proposed solution is to move the default case up to the front of the method, before the `active_dims` check.

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -101,15 +101,12 @@ class Kernel(Module):
         raise NotImplementedError()
 
     def __call__(self, x1_, x2_=None, **params):
+        x1, x2 = x1_, x2_
+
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Looks good to me. Thanks for the fix!

PR #231, opened 2018-08-15
ConstantMulLazyVariable now uses a scalar when appropriate

jacobrgardner:
`ConstantMulLazyVariable` now uses a torch scalar for its constant if there is only one element in the constant (e.g., not batch mode).

This fixes some bugs when back-propagating through samples on pytorch master. 

cc @Balandat

### DIFF
--- a/gpytorch/lazy/constant_mul_lazy_variable.py
+++ b/gpytorch/lazy/constant_mul_lazy_variable.py
@@ -10,9 +10,12 @@ from .lazy_variable import LazyVariable
 class ConstantMulLazyVariable(LazyVariable):
     def __init__(self, lazy_var, constant):
         if torch.is_tensor(constant):
+            if constant.numel() == 1:
+                constant = constant.squeeze()
             constant = constant
         else:
-            constant = torch.tensor([constant], device=lazy_var.device, dtype=torch.float32)
+            constant = torch.tensor(constant, device=lazy_var.device, dtype=torch.float32)
+
         super(ConstantMulLazyVariable, self).__init__(lazy_var, constant)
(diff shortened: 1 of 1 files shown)


gpytorch/lazy/constant_mul_lazy_variable.py
```diff
@@ -10,9 +10,11 @@
 class ConstantMulLazyVariable(LazyVariable):
     def __init__(self, lazy_var, constant):
         if torch.is_tensor(constant):
-            constant = constant
+            if constant.numel() == 1:
+                constant = constant.squeeze()
         else:
-            constant = torch.tensor([constant], device=lazy_var.device, dtype=torch.float32)
+            constant = torch.tensor(constant, device=lazy_var.device, dtype=torch.float32)
```
gpleiss: This doesn't create a scalar - it creates a 1D tensor of size `constant` (expects constant to be an int, not a float)

gpleiss: In addition, we probably shouldn't be constraining the `dtype` to be a float - we probably want it to have the same `dype` as the lazy variable (which could be a float or a double).

jacobrgardner: What do you mean? `torch.tensor(3.2)` makes a scalar

gpleiss: Whoops my b

Balandat: I suggest we give LazyVariable a `dtype` property that we can use in situations like this. What do you think?

PR #232, opened 2018-08-16
Scale kernel (for getting kernels with outputscales)

### LINKED ISSUE #229: Implement ScaleKernel
`ScaleKernel` would decorate an existing kernel and add an output scale.

Then, we wouldn't need to register a `log_outputscale` on all the existing GP modules

gpleiss:
`ScaleKernel(base_kernel)` now adds an outputscale to kernels. This will simplify what we're currently doing.

Added some docs to the files I touched.

Will update all the feature tests and example notebooks in a separate commit (could be part of this PR, or part of a second one).

Closes #229


Balandat:
Awesome, thanks a lot for adding all the docstrings! I haven't had time to look through in more detail- will this work for multi-output GPs with possibly different output scales per output?

gpleiss:
@Balandat I don't think multi-output GPs will need ScaleKernel, since the task kernel already performs a scaling operation.

gpleiss:
All feature tests now have outputscales via `ScaleKernel` (except for multitask/spectral mixture, which already have implicit scaling functionality)

Also:
- No more Variable/.data in `test/examples`
- Some fixes to LazyKernelEvaluatedVariable - it was adding extra dimensions when evaluating a batched row or column
- Had to turn down some learning rates (e.g. KISS-GP classification) - I guess this has to do with the order of the outputscale being applied (e.g. inside a grid kernel vs. outside)
- No more warnings from the test set (debug mode is set to true)

jacobrgardner:
The rest of this all looks good to me. I do think we should also update the simple GP regression tutorial notebook to have an output scale since we want that to be in the basic GP model people work with.

PR #240, opened 2018-08-20
Initialize low rank noise covar to nonzero values

jacobrgardner:
This implements the lessons learned by `IndexKernel`: `task_noise_covar_factor` can't be initialized with zeros and get gradients.

I also renamed the actual unit test function to describe the test.

@Balandat @rajkumarkarthik


rajkumarkarthik:
My bad on the unit test name. Thanks so much for fixing this, @jacobrgardner!

PR #241, opened 2018-08-20
Fix batch mode testing, improve batch mode training

### LINKED ISSUE #237: Batch predictions fail on master
Suppose you fit a GP in batch mode using `1 x n x 1` train_x and `1 x n` train_y. Now I want to predict a `1 x n' x 1` test_x, but that fails with a size mismatch in non_lazy_variable:

```
RuntimeError: size mismatch, m1: [1 x 1], m2: [100 x 1] at caffe2/aten/src/TH/generic/THTensorMath.cpp:932
```

### LINKED ISSUE #246: Multitask KISS-GP error
I am trying to run the code from Multitask_GP_Regression_Scalable_With_KISSGP.ipynb example from the master branch and get the following error at `observed_pred.confidence_region()`:

```
<ipython-input-7-4cd5b24c7024> in <module>()
     14     mean = observed_pred.mean()
     15     # Get lower and upper confidence bounds
---> 16     lower, upper = observed_pred.confidence_region()
     17 # This contains predictions for both tasks, flattened out
     18 # The first half of the predictions is for the first task

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/random_variables/random_variable.py in confidence_region(self)
     24         variable is the upper end
     25         """
---> 26         std2 = self.std().mul_(2)
     27         mean = self.mean()
     28         return mean.sub(std2), mean.add(std2)

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/random_variables/random_variable.py in std(self)
    108         batch size and d is the dimensionality of the random variable
    109         """
--> 110         return self.var().sqrt()
    111 
    112     def var(self):

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/random_variables/multitask_gaussian_random_variable.py in var(self)
     66 
     67     def var(self):
---> 68         return self._covar.diag().view(self._n, self._t)

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/lazy/sum_lazy_variable.py in diag(self)
     84 
     85     def diag(self):
---> 86         return sum(lazy_var.diag() for lazy_var in self.lazy_vars)
     87 
     88     def sum_batch(self, sum_batch_size=None):

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/lazy/sum_lazy_variable.py in <genexpr>(.0)
     84 
     85     def diag(self):
---> 86         return sum(lazy_var.diag() for lazy_var in self.lazy_vars)
     87 
     88     def sum_batch(self, sum_batch_size=None):

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/lazy/sum_lazy_variable.py in diag(self)
     84 
     85     def diag(self):
---> 86         return sum(lazy_var.diag() for lazy_var in self.lazy_vars)
     87 
     88     def sum_batch(self, sum_batch_size=None):

RuntimeError: The size of tensor a (2) must match the size of tensor b (102) at non-singleton dimension 1
```
Single-task KISS-GP and multitask without kernel interpolation works ok.

jacobrgardner:
- Fixes #237 
- Allows for batch mode training with different training sets by summing over losses (learns one parameter per batch)
- Adds batch mode tutorial notebook `simple_batch_mode_gp_regression.ipynb` that trains and tests 4 separate GPs simultaneously
- Implement better `forward_diag` for `RBFKernel` and `MaternKernel` that handles batch lengthscales correctly.
- Fixed a bug where `LazyEvaluatedKernelVariable` would transpose the batch and data dimensions even for kernels whose `forward_diag` has non-default behavior.
- `ExactGP` now only assumes we are in the multitask setting if a `MultitaskGaussianRandomVariable` is supplied. This lets `a x b` training labels mean `a` batches in the non multitask setting, but `b` tasks in the multitask setting.
- Fixes #246


gpytorch/kernels/matern_kernel.py
```diff
@@ -25,11 +26,31 @@ def __init__(
             ard_num_dims=ard_num_dims,
             log_lengthscale_prior=log_lengthscale_prior,
             active_dims=active_dims,
+            batch_size=batch_size,
             log_lengthscale_bounds=log_lengthscale_bounds,
         )
         self.nu = nu
         self.eps = eps
 
+    def forward_diag(self, x1, x2):
```
gpleiss: Could we add a unit test for this method?

gpytorch/kernels/rbf_kernel.py
```diff
@@ -87,6 +87,12 @@ def __init__(
         )
         self.eps = eps
 
+    def forward_diag(self, x1, x2):
```
gpleiss: Unit test here as well.

gpytorch/kernels/kernel.py
```diff
@@ -94,7 +95,11 @@ def size(self, x1, x2):
 
     @abstractmethod
     def forward_diag(self, x1, x2, **params):
```
gpleiss: The batch issues will happen for any lengthscale kernel (e.g. Periodic, Linear). We should probably remove the default `forward_diag` behavior.

gpytorch/kernels/matern_kernel.py
jacobrgardner: Going to open a separate issue for this because we need to make unit tests for matern kernel in general, and this PR fixes #246

gpytorch/kernels/rbf_kernel.py
jacobrgardner: Done!

PR #253, opened 2018-09-10
Register eta paramerter in LKJ prior as float

Balandat:
This fixes the issue reported by @rajkumarkarthik offline. Not sure why this behavior before this change was different between CPU and GPU. I tried to dig a little to see whether the `torch.tensor` constructor works differently for the two, or if there are weird interactions between the device and registering buffers, but didn't find anything.

### DIFF
--- a/gpytorch/priors/lkj_prior.py
+++ b/gpytorch/priors/lkj_prior.py
@@ -18,14 +18,14 @@ class LKJPrior(Prior):
     correlation matrix.
     """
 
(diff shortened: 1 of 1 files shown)


rajkumarkarthik:
Thanks for the fix, Max @Balandat!

jacobrgardner:
I'm perfectly happy with this being in since it's obviously safer. Still a bit confused why I got type errors about it and @rajkumarkarthik didn't.

Issue #239, opened 2018-08-19
Allow passing a fixed set of base_samples into sample


Balandat:
In certain situations we'd like to be able to fix the base samples in the `zero_mean_mvn_samples` functions. One use case for this is using MC-based batch acquisition functions for Bayesian Optimization, in which a large number of samples are drawn from the posterior to MC-approximate the integral involved in the acquisition function. This sampling results in stochastic function and gradient evaluations, which we can't use directly with non-stochastic (quasi-) higher order optimization algorithms. Now if we fix the underlying base_samples throughout the different iterations of the optimization, then we can use these optimizers (the result will be biased, but that may be good enough or at least a good starting point for other optimization).

To simplify this, I suggest we add a version of `sample` to the random variables that allows to pass in a set of base samples (i.i.d. N(0,1)). 

@jacobrgardner, @gpleiss any thoughts on this?

cc @darbour, @bletham, @bkarrer

Balandat:
I have a draft for this, about to put up a PR.

> Do they have an interface for this already?

@jacobrgardner What do you mean exactly? An interface for torch.distributions? Or for passing in base samples for a MVN?

jacobrgardner:
Yeah I meant for passing in a base samples. Just checking if whatever we do for this now will still be possible when we are constrained to the `torch.distributions` interface

Balandat:
So I don't see an api for passing in base_samples into distributions from `torch.distribution`, but it wouldn't be hard at all to add this in a subclassed version (we'll want to subclass all of the rvs anyway so things work with LazyTensors). Let me start prototyping random_variables that use `torch.distributions` under the hood.

PR #259, opened 2018-09-10
Allow to pass in base_samples to MVN sampling

Balandat:
see #239

### DIFF
--- a/gpytorch/lazy/lazy_variable.py
+++ b/gpytorch/lazy/lazy_variable.py
@@ -912,6 +912,13 @@ class LazyVariable(object):
         samples = covar_root.matmul(base_samples)
         return samples
 
+    def correlate_mvn_base_samples(self, base_samples):
+        if self.size()[-2:] == torch.Size([1, 1]):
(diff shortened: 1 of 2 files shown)


gpleiss:
This seems good to me! To be consistent with  PyTorch distributions, we'll eventually need to edit the shape of the output so that the `n_samples` dimension is first. But what you have right now is consistent with the rest of GaussianRandomVariable so that works.

Also, rather than having a separate method -- what do you think about `base_samples` be an optional argument that's passed into `.sample()` instead?

Balandat:
> To be consistent with PyTorch distributions, we'll eventually need to edit the shape of the output so that the n_samples dimension is first.

Yeah, I think it makes more sense to work out using torch.distributions before getting this in, as this will bring with it a bunch of changes.

> Also, rather than having a separate method -- what do you think about base_samples be an optional argument that's passed into .sample() instead?

Sure, for this we can just overwrite the sample method of the torch distribution in the gpytorch subclass.

Balandat:
Now part of #262

Issue #254, opened 2018-09-10
Rename LazyVariable to LazyTensor


jacobrgardner:
Given that (1) `LazyVariable` was intended to mimic `Variable` as closely as possible, but (2) `Variable` no longer exists in PyTorch, we should deprecate `LazyVariable` and rename it `LazyTensor`.

Issue #126, opened 2018-05-09
Migrate RandomVariable to use PyTorch distributions


gpleiss:
GPyTorch `RandomVariables` are essentially doing the same thing as PyTorch `Distributions`. We should try to leverage the PyTorch interface as much as possible.

Similarly to Pyro, we probably want to subclass each of the distributions (or at least subclassing the MultivariateNormal distribution). This will make it possible to work with LazyVariable covariances, and for us to do our custom fast sampling code.

Related to #123

Balandat:
I'm working on some early draft code to include priors. At least for the basic purpose of automatically adding the log_prob term of each parameter's prior (if specified) to the marginal log likelihood, the API implemented by the basic `Distribution` class seems adequate.

gpleiss:
Yeah, I think for the most part we'll be able to use mostly PyTorch distributions. I think the only custom distribution we need is a MultivariateNormal distribution (no surprises here...)

Balandat:
pytorch has a MultivariateNormal distribution: https://pytorch.org/docs/stable/distributions.html?highlight=multivariate#torch.distributions.multivariate_normal.MultivariateNormal

gpleiss:
Our MultivariateNormal needs some custom functionality. Mostly it needs to be able to handle LazyVaraible covariance matrices without explicitly constructing the covariance matrix.

Balandat:
So I was toying around with this a little, and I'm not sure how to best handle this. 

Regarding the MVN:
When there are no LazyTensors are involved, we can basically use the torch.distributions implementation (resulting in some changes to the api if we don't want to wrap the torch.distribution one).
If there are LazyTensors passed as args, we have to re-implement some things, but we'd want to re-use a lot of the basic functionality of torch.distributions. 

Not sure what the implications of this for autograd etc. are -- we'll have to make sure setting things up based on torch.distributions creates issues in this regard.

What do you think about doing something like the following (we can also have a class for Distributions specifying the generic API if we want):

```
import torch
from torch.distributions import MultivariateNormal as TMultivariateNormal
from torch.distributions.multivariate_normal import _batch_mv

from ..lazy import LazyTensor


class MultivariateNormal(TMultivariateNormal):
    """
    Constructs a multivariate Normal random variable, based on mean and covariance
    Can be multivariate, or a batch of multivariate gaussians

    Passing a vector mean corresponds to a multivariate Gaussian
[...]

Issue #214, opened 2018-08-03
Sampling and its documentation is inconsistent


gbohner:
Hi All, I'm relatively new to large-scale collaborative open source development, so please PM me if something should be done differently. I've been using pytorch for a project in the past few weeks, and it's great, but of course being in alpha there are quite a few inconsistencies. 

One of them is sampling of random variables:

1. Dimensions of sampled variables are inconsistent.
According to documentation in random_variables.py it should be ```n_samples * batch * dimension```, however in the most used case of GaussianRandomVariable it is ```batch * dimension * n_samples```, see in more detail in [point1](#point1).

2. Some types of random variables don't have batch mode. I think instead of adding batch mode for each, it would be reasonable to create a ```BatchRandomVariable```, see an example implementation in [point2](#point2), note that the sampling here conforms to the documented rather than the actual one for the moment.

3. Sampling is not implemented for some of the random variables, even though it should be easy, and it is a very generic way of using them. An example implementation for MixtureRandomVariable is in [point3](#point3).

#### Point1

```python
def sample(self, n_samples=1):
        """
        Draw samples from the random variable.
        n_samples - number of samples to draw (default 1)
        Returns: tensor of samples
        If the random variable is in batch mode, result will be
        s x b x d, where s is the number of samples, and b is the batch size
        and d is the dimensionality of the random variable
        If the random variable is not in batch mode, result will be
        s x ..., where s is the number of samples
        and d is the dimensionality of the random variable
        """
        raise NotImplementedError
```

Tests:

```python
b = 10 # Number of batch variables
d = 5 # Number of dimensions
n_samples = 20 # Number of samples
gauss_simple = GaussianRandomVariable(torch.randn(d,), (torch.rand(d,)+1e-6).diag())
gauss_batch = GaussianRandomVariable(torch.randn(b,d), torch.cat(b*[(torch.rand(d,)+1e-6).diag().unsqueeze(0)]))

gauss_simple.sample(n_samples).shape
 > torch.Size([5, 20])

gauss_batch.sample(n_samples).shape
 > torch.Size([10, 5, 20])
```

#### Point2

```python
class BatchRandomVariable(gpytorch.random_variables.RandomVariable):
    def __init__(self, *rand_vars, **kwargs):
        """
        Batch of random variables
        Params:
        - rand_vars (iterable of RandomVariables with b elements)
        """
        
        super(BatchRandomVariable, self).__init__(*rand_vars, **kwargs)
[...]

PR #262, opened 2018-09-12
Use torch.distributions instead of RandomVariables

Balandat:
This is a very early draft of how we could use the `torch.distributions` module instead of the currently implemented `RandomVariable`, as suggested in #126. This will break basically everything, but it's still useful as a basis for seeing how this could look like while supporting LazyVariables (or LazyTensors by #254). Lots of work will need to be done to update callsites throughout. We also should think about how this would be used in conjunction with all the multi-task stuff.

@jacobrgardner, @gpleiss, would love to hear your thoughts on this.


gpytorch/distributions/multitask_multivariate_normal.py
```diff
@@ -0,0 +1,66 @@
...
+    def covar(self):
+        return self._covar
+
+    def mean(self):
+        return self.loc
+
+    def representation(self):
+        return self.loc, self._covar
+
+    def sample(self, n_samples):
```
gpleiss: Should this be `rsample`?

gpytorch/distributions/multitask_multivariate_normal.py
```diff
@@ -0,0 +1,66 @@
...
+        self._covar = covar
+        self._nbatch = mean.shape[0] if mean.ndimension() == 3 else None
+        self._n = mean.shape[-2]
+        self._num_tasks = mean.shape[-1]
+
+    @property
+    def n_tasks(self):
+        return self._num_tasks
+
+    def covar(self):
```
gpleiss: Though this would be a breaking change, it might be good to have this be `covariance_matrix` to match the PyTorch MVN

gpytorch/distributions/multitask_multivariate_normal.py
```diff
@@ -0,0 +1,66 @@
...
+        samples = (
+            self._covar.zero_mean_mvn_samples(n_samples)
+            .view(self._num_tasks, self._n, n_samples)
+            .transpose(0, 1)
+            .contiguous()
+            .add(self.loc.unsqueeze(-1))
+        )
+        return samples
+
+    def var(self):
```
gpleiss: Probably should change this to `variance`

gpytorch/distributions/multitask_multivariate_normal.py
```diff
@@ -0,0 +1,66 @@
...
+        self._num_tasks = mean.shape[-1]
+
+    @property
+    def n_tasks(self):
+        return self._num_tasks
+
+    def covar(self):
+        return self._covar
+
+    def mean(self):
```
gpleiss: Probably should deprecate `mean` in favor of `loc`

gpleiss:
My biggest concern is sampling. As pointed out in #214, the shape of our MVN samples is different than what PyTorch does. (Our samples appear as the last dimension, in PyTorch they appear as the first dimension).

I'm not sure what the best way to address this is - other than a breaking change. @jacobrgardner thoughts?

gpytorch/distributions/multitask_multivariate_normal.py
jacobrgardner: Torch MVN defines `mean` as a property

gpytorch/distributions/multitask_multivariate_normal.py
jacobrgardner: I think Distributions have both methods -- one calls the other I believe?

jacobrgardner:
How about we keep the RandomVariables around with no breaking changes, but Distributions introduce breaking changes. The MVN RV calls Distribution sample, which returns the same shape as torch distributions, transposes the result, and throws a deprecation warning that includes a note about transposing.

Balandat:
Sorry for missing the old code in `multitask_multivariate_normal.py`, I will update this - mostly by calling the relevant methods of the `MultivariateNormal` superclass and do the appropriate reshaping.

As far as the breaking changes in the shape of samples returned: I would advocate for making that breaking change now rather than later - it will get increasingly harder to do this as the package matures. 

> How about we keep the RandomVariables around with no breaking changes, but Distributions introduce breaking changes. The MVN RV calls Distribution sample, which returns the same shape as torch distributions, transposes the result, and throws a deprecation warning that includes a note about transposing.

@jacobrgardner, is you point that we should keep the user-facing `RandomVariables` around for compatibility, but update all the internal call sites? Not entirely sure how we can do this elegantly in a way that wouldn't require a lot of special casing. Maybe one way is to write a dummy sample wrapper that depending on whether a `RandomVariable` or `Distribution` is passed issues the appropriate call (note that the torch `sample` method takes a `torch.Size` object, not an `int`), and transposes the samples if necessary. Thoughts?

gpleiss:
@Balandat @jacobrgardner perhaps what we should do is address the sampling inconsistency mentioned in #214 - and make the sampling dimension the first dimension. This would be a "breaking change," but it would also be fixing a bug/inconsistency. We could even push this as a separate PR, and we can print out a warning whenever we call `GaussianRandomVariable.sample` that the sampling code has changed.

After we do this, then I think the interface between RandomVariables and Distributions will be the same, except for the method names. We could then switch over to distributions and deprecate the old method names.

Thoughts?

Balandat:
@gpleiss that makes total sense. Would you mind working on the PR for addressing the issue in #214?That way I can focus on this PR for making distributions work.

gpleiss:
@Balandat sounds good can do!

Balandat:
@jacobrgardner, @gpleiss, 16e5cb5 significantly simplifies things by re-utilizing a lot of stuff from torch.distributions. I think this makes things much more elegant, let me know what you think. I'll work on cleaning up a test/demo notebook soon.

gpleiss:
This is looking really good! We should probably keep RVs around and deprecate them, but other than that I think we should merge this in soon.

Balandat:
Working on this now, adding tests, etc. One issue with replacing GaussianRandomVariable the way I do here is that sometimes Lanczos fails for the root_decomposition (see #268), which is used to provide the `_unbroadcasted_scale_tril` property. We should make root_decomposition more robust (and subclass the diagonal special case) before merging this in.

Balandat:
Ok, a bunch more updates and added unit tests. There are a number of test failures in the kiss_gp examples that will need to be addressed, these are likely due to using the torch.distribution batch mode (@gpleiss, @jacobrgardner, comments on this are much appreciated). Other than that things seem to work pretty well.

Balandat:
It probably makes sense to modify the likelihoods to be based on torch.distributions as well, e.g. change `log_probability` of `GaussianLikelihood` so as to hat having to worry about being consistent w.r.t. how batches are handled

Balandat:
Abandoning in favor of #288

PR #265, opened 2018-09-12
[Codemod] Rename LazyVariable to LazyTensor

Balandat:
This used both Bowler and grep/sed (for whatever Bowler didn't catch) for codemod. All tests run, but there might be some stragglers leftover.


Balandat:
hmm looks like .gitignore got hit in the process, will undo those changes

gpleiss:
Yaaaaay! Thanks @Balandat

PR #270, opened 2018-09-14
Fix inconsistency with MVN random variable sampling

### LINKED ISSUE #214: Sampling and its documentation is inconsistent

gpleiss:
Now GaussianRandomVariables and MultitaskGaussianRandomVariables are consistent with the documentation and other random variables: given an `n` dimensional Gaussian, calling `.sample(n_samples=s)` will return an `s x n` tensor (rather than an `n x s` tensor).

Calling `.sample` outputs a warning to users so that they are aware of this change.

Fixes #214


gpleiss:
cc @Balandat - after this I think we can go ahead with #262

Balandat:
Thanks for making these changes, This looks good to me.

Before we land this, can we add unit tests for the sampling though? I feel like we should improve unit test coverage quite a bit before the beta release.

gpleiss:
Sure I'll add unit tests

gpleiss:
@Balandat - added sampling tests for all the LazyTensors, back-filled some other tests/docs, and refactored a bit

Balandat:
Awesome, thanks!

Issue #255, opened 2018-09-10
Add example scripts for initialing hyperparameters for SKI


jacobrgardner:

gpleiss:
Closed by #276, #308

PR #276, opened 2018-09-18
Automatically determine grid bounds for SKI regression

gpleiss:
For `GridInterpolationKernel` (and its variants), it is no longer necessary to pass in `grid_bounds`. They can be automatically determined by the data.

```python
grid_size = gpytorch.utils.grid.choose_grid_size(x, ratio=1.5)  # Simple helper function
kernel = GridInterpolationKernel(rbf_kernel, grid_size=grid_size, num_dim=1)
```

**Note** this probably will require some back-filled unit tests and some updated examples. I'll open a separate issue for those if it looks like we like this new interface

### DIFF
--- a/gpytorch/kernels/grid_interpolation_kernel.py
+++ b/gpytorch/kernels/grid_interpolation_kernel.py
@@ -11,23 +11,116 @@ from ..utils import Interpolation
 
 
 class GridInterpolationKernel(GridKernel):
-    def __init__(self, base_kernel_module, grid_size, grid_bounds, active_dims=None):
-        grid = torch.zeros(len(grid_bounds), grid_size)
-        for i in range(len(grid_bounds)):
-            grid_diff = float(grid_bounds[i][1] - grid_bounds[i][0]) / (grid_size - 2)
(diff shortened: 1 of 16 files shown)


gpleiss:
First step in addressing #255

gpytorch/kernels/grid_interpolation_kernel.py
```diff
@@ -49,6 +145,33 @@ def forward_diag(self, x1, x2, **kwargs):
...
+            update_grid = (not self.has_initialized_grid.item()) or any(
+                x_min < bound[0] or x_max > bound[1]
+                for x_min, x_max, bound in zip(x_mins, x_maxs, self._tight_grid_bounds)
+            )
+
+            # Update the grid if needed
+            if update_grid:
+                grid_spacings = tuple((x_max - x_min) / (self.grid_size - 4.02) for x_min, x_max in zip(x_mins, x_maxs))
+                self.grid_bounds = tuple(
+                    (x_min - 2.01 * spacing, x_max + 2.01 * spacing)
```
jacobrgardner: This will be a big quality of life improvement i think, to automatically scale the grid so that the bounds are past the min and max values so that points don't end up exactly on the grid bounds.

Issue #271, opened 2018-09-17
Lapack Error in TestSimpleGPRegression


Balandat:
I'm getting a reproducible `RuntimeError: Lapack Error syev : 9 off-diagonal elements didn't converge to zero at caffe2/aten/src/TH/generic/THTensorLapack.cpp:395` in `test_simple_gp_regression.py` with this.

Not sure whether related to #266 or #270. @gpleiss, could you take a look?

gpleiss:
I'm not noticing this issue - @jacobrgardner and I are compiling PyTorch master to set if that's the issue

gpleiss:
@Balandat - this might have been fixed by 2a77a6b - could you check?

jacobrgardner:
I am pretty confident this is now resolved on pytorch master.

PR #277, opened 2018-09-18
Change seed for TestSimpleGPRegression

Balandat:
This was failing with Lapack errors on pytorch master (see #271)


gpleiss:
I'm having trouble installing PyTorch master, so I couldn't play around with it. Let's keep #271 open for right now and I'll try to diagnose the issue later this week.

Issue #261, opened 2018-09-11
Implement LCM Kernel


Balandat:
We can provide a simple ICM constructor that just wraps this for convenience/clarity. I'll have to think a little bit how we deal with the priors on the different index kernels involved - it's probably fine to have the correlation prior over all index kernels and specify that once in the constructor. Though it's not quite clear how these would be evaluated then in the current setup. Maybe the easiest is just to clone the prior and attach it to each IndexKernel.

Balandat:
this is done, thanks to @rajkumarkarthik

PR #278, opened 2018-09-19
Basic implementation of an LCM kernel to address #261

rajkumarkarthik:
This code creates a simple wrapper for an LCM kernel by adding multiple `MultitaskKernel` objects. This addresses the issue raised in #261. However, the code throws a `RuntimeError` that I don't fully understand. Does anyone know what is going on?

cc @Balandat (who contributed to this PR)


jacobrgardner:
Do you have any example code or an exact error message?

rajkumarkarthik:
Yes, here is a notebook that shows the error. (Github won't let me upload a .ipynb file, so just remove the .txt extension when you use it.)

[Multitask_GP_Regression.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/2398746/Multitask_GP_Regression.ipynb.txt)

gpytorch/kernels/lcm_kernel.py
```diff
@@ -0,0 +1,46 @@
...
+from gpytorch.kernels.multitask_kernel import MultitaskKernel
+
+
+class LCMKernel(Kernel):
+    """
+    This kernel supports the LCM kernel.
+    """
+    def __init__(self, base_kernels, n_tasks, rank=1, task_covar_prior=None):
+        """
+        Args:
```
eytan: a little light on the docstrings

jacobrgardner:
Fixed the issue you were having. Part of the multitask refactor we did to add them as first class objects was to require kernels to implement a `size(x1, x2)` method if they anticipate returning covariance matrices that have a size different from `x1.size(1) x x2.size(1)` (where x1, x2 are `b x n x d`).

Since all multitask kernels return lazy tensors that represent matrices of size `nt x nt` rather than just `n x n`, our internals need to know about that.

jacobrgardner:
Can we change the base branch for this to beta or do you need this in master?

rajkumarkarthik:
I've been working on the branch `create_lcm_wrapper`. Did this push to master?

jacobrgardner:
Nope you're good! I meant can you modify the PR so that it merges in to beta instead of master

jacobrgardner:
I went ahead and updated it -- main thing is just to make sure everything is compatible with everything else going in to beta.

rajkumarkarthik:
Thanks! What is the difference between master and beta again? Is beta for features we don't want to merge into master right away?

jacobrgardner:
`beta` is the base branch for the beta release we are hoping to do shortly -- once everything is ready, everything we're doing for the beta will get merged in to master in one go. 

Since `beta` is quite a bit ahead of master, I just figured it would be useful to make sure everything you're doing is compatible with everything that has already gone on the `beta` branch. Looks like the checks did pass, so should be all good on that aspect.

Balandat:
This looks good to me - would you mind supplying some basic unit tests for this?

rajkumarkarthik:
Done! This test does a simple sanity check if LCM and ICM provide the same results in the case where only one base kernel is provided for LCM.

test/examples/test_lcm_kernel_regression.py
```diff
@@ -0,0 +1,107 @@
...
+from gpytorch.means import ConstantMean, MultitaskMean
+from gpytorch.likelihoods import MultitaskGaussianLikelihood
+from gpytorch.random_variables import MultitaskGaussianRandomVariable
+
+
+class MultitaskGPModel(gpytorch.models.ExactGP):
+    def __init__(self, train_x, train_y, likelihood):
+        super(MultitaskGPModel, self).__init__(train_x, train_y, likelihood)
+        self.mean_module = MultitaskMean(ConstantMean(), n_tasks=2)
+        self.base_kernel_list = [RBFKernel()]
```
Balandat: Why make this an attribute of the instance? This shouldn't be necessary.

test/examples/test_lcm_kernel_regression.py
```diff
@@ -0,0 +1,107 @@
...
+        mean_x = self.mean_module(x)
+        covar_x = self.covar_module(x)
+        return MultitaskGaussianRandomVariable(mean_x, covar_x)
+
+
+class MultitaskGPModel_ICM(gpytorch.models.ExactGP):
+    def __init__(self, train_x, train_y, likelihood):
+        super(MultitaskGPModel_ICM, self).__init__(train_x, train_y, likelihood)
+        self.mean_module = MultitaskMean(ConstantMean(), n_tasks=2)
+        self.base_kernel = RBFKernel()
```
Balandat: same here, no need to add this, this should be chained together automatically.

Balandat:
Thanks for working on this!

Issue #249, opened 2018-09-01
KISS-GP + RBF kernel separate lengthscales error


tzoiker:
In KISS-GP setting with RBF kernel, for example, in 2D case (notebook attached), setting `ard_num_dims=2` gives the following error during loss calculation:

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-144-12b763efd692> in <module>()
     19     output = model(train_x)
     20     # TODO: Fix this view call!!
---> 21     loss = -mll(output, train_y)
     22     loss.backward()
     23     print('Iter %d/%d - Loss: %.3f' % (i + 1, n_iter, loss.item()))

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/module.py in __call__(self, *inputs, **kwargs)
    160 
    161     def __call__(self, *inputs, **kwargs):
--> 162         outputs = self.forward(*inputs, **kwargs)
    163         if torch.is_tensor(outputs) or isinstance(outputs, RandomVariable) or isinstance(outputs, LazyVariable):
    164             return outputs

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/mlls/exact_marginal_log_likelihood.py in forward(self, output, target)
     49 
     50         # Get log determininat and first part of quadratic form
---> 51         inv_quad, log_det = covar.inv_quad_log_det(inv_quad_rhs=(target - mean).unsqueeze(-1), log_det=True)
     52 
     53         # Add terms for SGPR / when inducing points are learned

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/lazy/lazy_variable.py in inv_quad_log_det(self, inv_quad_rhs, log_det)
    577         matrix_size = self.size(-1)
    578         batch_size = self.size(0) if self.ndimension() == 3 else None
--> 579         tensor_cls = self.tensor_cls
    580 
    581         args = lazy_var.representation()

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/lazy/lazy_variable.py in tensor_cls(self)
    882     def tensor_cls(self):
    883         if not hasattr(self, "_tensor_cls"):
--> 884             first_item = self.representation()[0]
    885             if isinstance(first_item, Variable):
    886                 first_item = first_item.data

~/anaconda/envs/py3/lib/python3.6/site-packages/gpytorch/lazy/lazy_variable.py in representation(self)
    733                 representation.append(arg)
[...]

PR #280, opened 2018-09-20
Fix diag calls to ScaleKernel in batch mode

jacobrgardner:
On master, ScaleKernel's behavior in batch mode when getting only the diagonal of the kernel is broken for the same reason that other kernels are (see e.g. #249).

Here's some simple code that reproduces the error this PR fixes:
```python
covar_module = ScaleKernel(RBFKernel(batch_size=5), batch_size=5)
covar = covar_module(torch.randn(5, 20, 1))
diag = covar.diag() # fails because it returns a torch.Size(5, 20, 5) tensor
```

While we wait for the kernel refactor @gpleiss is working on, this PR fixes this issue by implementing a `forward_diag` method for ScaleKernel. An alternative fix would be to let outputscales be `torch.Size(5, 1, 1)` and allow `ConstantMulLazyTensors` to take batches of constants for batch lazy tensors.


gpleiss:
Good catch.

Issue #182, opened 2018-07-19
Make shape a property calling size() throughout the library


Balandat:
To make things consistent with pytorch, see discussion here: https://github.com/cornellius-gp/gpytorch/pull/139#discussion_r203881586

gpleiss:
Closed by #293

Issue #292, opened 2018-09-26
Inconsistent behavior of .diag() method for different Lazy Tensors in batch mode


Balandat:
E.g: 

```
> cv.lazy_vars
[<gpytorch.lazy.lazy_evaluated_kernel_tensor.LazyEvaluatedKernelTensor at 0x7f6139087278>,
 <gpytorch.lazy.matmul_lazy_tensor.MatmulLazyTensor at 0x7f6139087780>]

> [lv.size() for lv in cv.lazy_vars]
[torch.Size([2, 1, 1]), torch.Size([2, 1, 1])]

>diags = [lazy_var.diag().contiguous() for lazy_var in cv.lazy_vars]
>[diag.size() for diag in diags]
[torch.Size([2]), torch.Size([2, 1])]
```

This causes inconsistencies when using batch mode. My suggestion is to ensure that all `.diag()` calls reduce the last two dimensions of the tensor, i.e. for a `b1 x b2 x n x n` LazyTensor this would return a Tensor of Size `b1 x b2 x n`. This is the same as the behavior of `torch.diagonal(X, dim1=-2, dim2=-1)`.

gpleiss:
This will probably be fixed with the kernel refactor I'm working on

gpleiss:
I'm going to call this closed by #293

PR #293, opened 2018-09-26
Make LazyEvaluatedKernelVariable return diag of proper size

Balandat:
Should fix (at least part of) #292. Also introduces a `.shape` property for LazyTensors


gpleiss:
This looks great! And it also closes #182 (since all of our objects now have a `.shape` method).

PR #295, opened 2018-09-27
Fix add diag

### LINKED ISSUE #294: multitask with many tasks, Function AddDiagLegacyBackward
Hello, 

I am trying to do a multitask gp with 36 tasks. However, I'm getting an error in training: 

Function AddDiagLegacyBackward returned an invalid gradient at index 1 - expected shape [1] but got []

I wanted to check a few things about setting up this model. 

Following the multitask example:https://github.com/cornellius-gp/gpytorch/blob/master/examples/03_Multitask_GP_Regression/Hadamard_Multitask_GP_Regression.ipynb

it seems that all of the data is flattened into training x and y tensors with an index tensor. So I have: 

x = [x1,....,x696], y=[y1,...,y696] and then y_indices = [0,0,0,1,1,1,1,1,1,...,35,35,35]. 

Is that the correct way to set up the indices? This seems to be the same as in the example. 

The tasks are not of the same dimension. Is that a problem? 

The model is straight from the example:

class  MultitaskGPModel(gpytorch.models.ExactGP):
        super(MultitaskGPModel, self).__init__(train_x, train_y, likelihood)
        # Default bounds on mean are (-1e10, 1e10)
        self.mean_module = ConstantMean()
        # We use the very common RBF kernel
        self.covar_module = RBFKernel()
        # We learn an IndexKernel for 2 tasks
        # (so we'll actually learn 2x2=4 tasks with correlations)
        **self.task_covar_module = IndexKernel(n_tasks=36, rank=1)**

    def forward(self,x,i):
        # Get predictive mean
        mean_x = self.mean_module(x)
        # Get all covariances, we'll look up the task-speicific ones
        # # Get the covariance for task i
        covar_i = self.task_covar_module(i)
        covar_xi = covar_x.mul(covar_i)
        return GaussianRandomVariable(mean_x, covar_xi)

Thanks for your help!

gpleiss:
- LazyTensor add_diag now handles lots of different cases (batch mode, scalars, etc.) and unit tests/arg checks for each of them
- `gpytorch.add_diag` now uses the LazyTensor method (no custom torch.tensor method)

Fixes #294


gpleiss:
Lol accidentally deleted this branch... fixed now

Issue #236, opened 2018-08-17
Allow batch-predictions from non-batch GP


Balandat:
In some cases we'd like to get independent posteriors for separate batches of test points. E.g. say we train a single GP on `n x d` train_x and `n` (single-dim) train_y data. Then we'd like to be able to pass in test data of shape `b x k x d` to produce `b` independent posteriors, where the `i`-th one is the posterior of the `k` test points in the `i`-th batch.

Currently, this raises the following error: 
```
RuntimeError: invalid argument 0: Sizes of tensors must match except in dimension 0
```

We should be able to do some reasonable broadcasting / inferring of batch size 1 here.

Pertinent comment by @jacobrgardner:
> I think it would be good if everything in GPyTorch was in batch mode by default. That should automatically solve this issue: the original model would be a model on `x` in `1 x n x d`, `y` in `1 x n x 1`.  Then, for batch predictions, you'd get broadcasting.

Balandat:
Any progress on this? I think standardizing this throughout the codebase will be necessary to make the torch distributions (and other stuff like the HeteroschedasticLikelihood) work as expected.

gpleiss:
We have a feature test that sort of covers this, but right now it expects that you update the training data to include batch data. I'll try to tackle this today or tomorrow

Balandat:
ok cool - let's consider this in the context of #262 and #286, with which we're hopefully converging to a standardized way of handling batches everywhere

PR #296, opened 2018-09-28
Fix LazyTensor.__getitem__

### LINKED ISSUE #159: Batch mode InterpolatedLazyVariable __getitem__ doesn't behave like torch.Tensor.__getitem__
`InterpolatedLazyVariable`'s `__getitem__` method has different behavior than `torch.Tensor` in batch mode. Here's a simple example:

```python
matrix = InterpolatedLazyVariable(...) # Some batch mode ILV, say 2 x 10 x 10
eval_matrix = matrix.evaluate() # Get a torch.Tensor of the same matrix.

 # Returns a 2 x 11 tensor containing eval_matrix[0, 0, :] and eval_matrix[1, 2, :]
res1 = eval_matrix[[0, 1], [0, 2], :]

# Returns a 2 x 2 x 11 tensor containing res1 replicated twice.
res2 = matrix[[0, 1], [0, 2], :].evaluate()
```

@gpleiss

gpleiss:
This fixes #159, and is a step towards #236.

__getitem__ would fail for batch LazyTensors if you supplied two or more tensor-based indices. E.g. `lazy_tensor[torch.tensor([0, 0, 1]), torch.tensor([2, 1, 2]), :]`.

This fixes the default LT.__getitem__, all custom LT.__getitem__s, and adds a lot of unit tests.

### DIFF
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -74,7 +74,7 @@ class BlockDiagLazyTensor(BlockLazyTensor):
         right_batch_indices = left_indices.div(block_size).long()
         batch_indices = batch_indices * block_size + left_batch_indices
         left_indices = left_indices.fmod(block_size)
-        right_indices = left_indices.fmod(block_size)
+        right_indices = right_indices.fmod(block_size)
 
         res = self.base_lazy_tensor._batch_get_indices(batch_indices, left_indices, right_indices)
         res = res * torch.eq(left_batch_indices, right_batch_indices).type_as(res)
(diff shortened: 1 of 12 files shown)


Balandat:
And many thanks for the unit tests - Is there a straightforward way in github to get unit test coverage stats through travis?

jacobrgardner:
Unfortunately, the `__getitem__`s are pretty necessarily complicated in a lot of cases -- some of them might be able to be simplified, but definitely not all. Turns out being able to index in to matrices that are stored in really weird ways for efficiency in all the same ways you can index in to normal tensors is challenging :P

gpleiss:
Originally the idea was that LazyTensor would handle most of complex rules (is the index an int, slice, tensor, or some combination) and that other LTs would only have to define simple rules.

Unfortunately InterpolatedLazyTesnor needs its own rule (so that you don't end up with nested InterpolatedLazyTensors), BlockDiagLazyTensor/SumBatchLazyTensor need a custom rule to handle the nested batch dimensions, and ZeroLazyTensor has a custom one for efficiency.

Balandat:
Yeah fair enough, it's a hard problem...

PR #298, opened 2018-09-28
[WIP] Refactor LazyTensor unit tests

gpleiss:
We've seen a number of bugs recently in LazyTensor behaviors. Some LTs have comprehensive unit tests, covering lots of edge cases. Others do not.

I propose creating super classes for all LT unit tests that contain lots of test cases. For each LT unit test, we will only have to write a method that constructs/evaluates the LT.

I started on a couple LTs.

### DIFF
--- a/gpytorch/lazy/constant_mul_lazy_tensor.py
+++ b/gpytorch/lazy/constant_mul_lazy_tensor.py
@@ -19,13 +19,13 @@ class ConstantMulLazyTensor(LazyTensor):
     To element-wise multiply two lazy tensors, see :class:`gpytorch.lazy.MulLazyTensor`
 
     Args:
-        lazy_var (LazyTensor) or (b x n x m)): The base_lazy tensor
+        base_lazy_tensor (LazyTensor) or (b x n x m)): The base_lazy tensor
         constant (Tensor): The constant
 
-    If `lazy_var` represents a matrix (non-batch), then `constant` must be a
+    If `base_lazy_tensor` represents a matrix (non-batch), then `constant` must be a
     0D tensor, or a 1D tensor with one element.
 
-    If `lazy_var` represents a batch of matrices (b x m x n), then `constant` can be
+    If `base_lazy_tensor` represents a batch of matrices (b x m x n), then `constant` can be
     either:
     - A 0D tensor - the same constant is applied to all matrices in the batch
     - A 1D tensor with one element - the same constant is applied to all matrices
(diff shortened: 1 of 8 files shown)


gpytorch/lazy/constant_mul_lazy_tensor.py
```diff
@@ -55,50 +55,50 @@ class ConstantMulLazyTensor(LazyTensor):
...
                         "Got a %dD lazy var. with size %s"
-                        % (numel, numel, lazy_var.ndimension(), repr(lazy_var.size()))
+                        % (numel, numel, base_lazy_tensor.ndimension(), repr(base_lazy_tensor.size()))
                     )
 
             elif constant.numel() == 1:
                 constant = constant.squeeze()
         else:
-            constant = torch.tensor(constant, device=lazy_var.device, dtype=torch.float32)
+            constant = torch.tensor(constant, device=base_lazy_tensor.device, dtype=torch.float32)
```
Balandat: seems like this should be using the dtype of the base_lazy_tensor as well

test/lazy/_lazy_tensor_test_case.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def test_matmul_vec(self):
+        lazy_tensor = self.create_lazy_tensor()
+        lazy_tensor_copy = lazy_tensor.clone()
+        evaluated = self.evaluate_lazy_tensor(lazy_tensor_copy)
+
+        test_vector = torch.randn(lazy_tensor.size(-1))
+        res = lazy_tensor.matmul(test_vector)
+        actual = evaluated.matmul(test_vector)
+        assert(approx_equal(res, actual))
```
Balandat: lets make this a `self.assertTrue(approx_equal(res, actual))` (and similarly in other places)

gpleiss:
This PR is now ready to be merged. There's a couple of bug fixes:

- Fix logdet calculations for batch LazyTensors
- Fix __getitem__ for KroneckerProductLazyTensors

Balandat:
@gpleiss, what about my comment on hard-coding the dtype to `float32` in `constant_mul_lazy_tensor.py`? Looks like that would break things when running as double?

gpleiss:
@Balandat sorry I missed that! Fixing it now

gpytorch/lazy/constant_mul_lazy_tensor.py
gpleiss: Updated

Balandat:
I'm seeing a test failure in `TestMulLazyTensorMulti`, not sure why this didn't show up in Travis. 

Also, can we add cuda versions for all the LazyTensor test cases? We've had plenty of bugs due to device incompatibilities that would be easy to avoid with tests. Should be simple to do by following the pattern I used e.g. in `test_lkj_prior.py`

gpleiss:
I'll take a look at the test failure and open another PR for the cuda tests

gpleiss:
Test failure should be fixed now. Adding an issue for cuda tests.

Issue #227, opened 2018-08-08
Add unit tests for Wishart & LKJ priors, fix logdet computation in Wishart


Balandat:
See discussion here: #226

Balandat:
logdet computation handed differently in the new Wishart implementation #290.

Unit tests for Wishart, InverseWishare and LKJCholeskyFactorPrior, LKJCovariancePrior still need to be done.

Balandat:
LKJ tests were added by #305, leaving this open for the Wishart ones

PR #299, opened 2018-09-28
Batch mode for (exact) regression model

### LINKED ISSUE #236: Allow batch-predictions from non-batch GP

gpleiss:
The following models work for ExactGPs and MultitaskGPs:

- Train on non-batch, test on batch
- Train on batch, test on batch (share hyperparameters across all batches)
- Train on batch, test on batch (different hyperparameters across batches)

This includes the following updates:
- IndexKernel and MultitaskKernel are now designed to work on batched data
- GaussianLikelihood now takes in a `batch_size` argument, which will (optionally) use different noise values for different batches
- MultitaskGaussianLikelihood also takes a `batch_size` argument, which also uses different noise values

Fixes #236

### DIFF
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -13,14 +13,44 @@ def _eval_covar_matrix(covar_factor, log_var):
 
 
 class IndexKernel(Kernel):
-    def __init__(self, n_tasks, rank=1, prior=None, active_dims=None):
+    r"""
+    A kernel for discrete indices. Kernel is defined by a lookup table.
+
+    .. math::
+
+        \begin{equation}
+            k(i, j) = \left(BB^\top + \text{diag}(\mathbf v) \right)_{i, j}
+        \end{equation}
+
+    where :math:`B` is a low-rank matrix, and :math:`\mathbf v` is a  non-negative vector.
+    These parameters are learned.
+
+    Args:
+        n_tasks (int):
+            Total number of indices.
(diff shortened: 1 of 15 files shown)


Balandat:
Yeah 'll take a look - thanks for the changes!

Balandat:
yeah so the issue is that the `LKJCovariancePrior` doesn't work correctly in batch mode. Will fix and address #227 in the process

h-jia:
Could batch in train and test be set in different?

For example, train_x `k x b x n` and train_y `k x b x 1`, where `k` is the batch. In test, test_x 'm x b x n' and test_y 'm x b x 1'

PR #304, opened 2018-09-30
Reorganize utils

gpleiss:
I wanted to clean up `gpytorch.utils` before beta:

- Move all functions out of __init__ and into appropriate submodule
- Move `approx_equals` into test/_utils.py (since we only use it for
  tests)
- Rename test/util to test/utils

BREAKING CHANGE: (for users)
- `gpytorch.utils.scale_to_bounds` is now
  `gpytorch.utils.grid.scale_to_bounds`

Important for developers:
- `gpytorch.utils.approx_equals` is moved into a test-specific folder

Will update the examples to fix breaking change tomorrow (along with fixing all other breaking changes)


gpleiss:
Yeah - I think LuaTorch had an `approx_equal` function. I'll open an issue in pytorch

PR #305, opened 2018-09-30
Make LKJCovariancePrior work properly in batch mode

Balandat:
Makes sure LKJCovariancePrior works properly in batch mode.
Also adds the ability to have heterogeneous priors over the different
marginal variances (e.g. can use a sd_prior with event_shape that is
either 1 or n if the covariance matrices are n x n)


gpleiss:
Thanks for the unit tests :)

PR #307, opened 2018-10-01
Kernels now take `diag, dim_groups` as arguments (fixes issues with ARD kernels)

gpleiss:
This should fix several issues we have with ARD batch kernels.

TLDR:
- `Kernel.forward` now takes 4 arguments: `x1`, `x2`, `diag`, and `dim_groups`
  - `diag` tells the kernel to be computed in diagonal mode (variances only, no covariances)
  - `dim_groups=<k>` computes k batches of kernels: each operating on one of `k` subsets of dimensions. (This is to be used with additive kernels, SKI, etc.)
- For most kernel implementations - you won't have to worry about these arguments much. They just need to be passed into `_create_input_grid`

The motivation for this: for kernel diagonals and additive kernels, we have been relying on hacking the batch dimension. For example, computing the kernel diagonal of a `n x d` input can be done easily if we make the input `n x 1 x d`. Similarly, additive kernels can be efficiently computed if we make the `n x d` input a `d x n x 1` input and then sum along the batch dimension. This works well if we don't have ARD or batch-specific parameters, but now we do and it causes breaking issues. (See #249)

To make sure that this isn't an issue, but also keep everything easily programmable:
- We won't do batch-dimension hacking anymore
- Computing kernel diagonals/additive kernels is now done explicitly by passing in the `dim`/`dim_groups` arguments
- For most kernels, this won't require any extra code. Most kernels now use the `_create_input_grid` helper, which automatically will do the correct thing for the diag/dim_groups modes.

I've included some documentation in the Kernel base_class to describe this for people who want to code up their own kernels.

This is currently a WIP - hopefully will be done in ~4 hours.


Balandat:
So I'm all for being explicit and not doing a bunch of magic, my only worry is that by adding additional args we are making it harder / less accessible for people to write their own kernels. From reading the docstring, I'm not really clear what `dim_groups` actually does... Is there a way to simplify this slightly? Not sure if defining this as `forward(slef, x1, x2, **kwargs)` is any better in that regard...

jacobrgardner:
@Balandat Geoff and I have been talking about this exact concern in the past couple of hours. I generally had the same concern. Geoff pointed out that in most cases those args just get passed directly on to a helper method. I do agree we can make it clearer exactly what the args do in the docstrings though

gpleiss:
At this point, this PR passes all tests - and all kernels have been refactored.
I still want to do the following:

- Replace AdditiveGridInterpolationKernel/MultiplicativeGridInterpolationKernel with more generic AdditiveStructureKernel/ProductStructureKernel
- Add feature tests that see if we solve #249 

This PR could either be merged as is, or we could wait for these last two changes.

cc/ @Balandat @jacobrgardner

gpleiss:
With the latest updates to the kronecker/additive SKI feature tests, I can confirm that this fixes #249

gpleiss:
BREAKING CHANGES:
- GridInterpolationKernel, GridKernel, InducingPointKernel - the attribute `base_kernel_module` has become `base_kernel` (for consistency)
- AdditiveGridInterpolationKernel no longer exists. Now use `AdditiveStructureKernel(GridInterpolationKernel(...))`
- MultiplicativeGridInterpolationKernel no longer exists. Now use `ProductStructureKernel(GridInterpolationKernel(...))`

gpleiss:
At this point I think this PR is ready to be reviewed/merged @jacobrgardner @Balandat

jacobrgardner:
Been going through this for the past 30 minutes to me. This tentatively looks good to me. I'm going to merge it in to beta, and go through more carefully when we review all of the beta changes. Let's open an issue for handling multiple batch dimensions later on down the road.

gpytorch/kernels/kernel.py
```diff
@@ -183,13 +183,28 @@ def _create_input_grid(self, x1, x2):
             The shape depends on the kernel's mode
 
             * `full_covar`: (`b x n x 1 x d` and `b x 1 x m x d`)
-            * `full_covar` with `dim_groups=k`: (`b x k x n x 1 x (d//k)` and `b x k x 1 x m x (d//k)`)
+            * `full_covar` with `batch_dims=(0, 2)`: (`b x k x n x 1 x (d//k)` and `b x k x 1 x m x (d//k)`)
```
jacobrgardner: Remove `d//k` reference

gpytorch/kernels/kernel.py
```diff
@@ -210,7 +225,78 @@ def __call__(self, x1, x2=None, **params):
...
+            return res
+
+        else:
+            res = LazyEvaluatedKernelTensor(self, x1_, x2_, batch_dims=batch_dims, **params)
+
+            # Now we'll make sure that the shape we're getting makes sense
+            if settings.debug.on():
+                # If we used batch_dims...
+                shape = self.size(x1_, x2_)
+                if batch_dims == (0, 2):
```
jacobrgardner: Even though we don't support multiple batch dimensions yet, it might be good to handle this correctly. Technically as coded, this breaks the contract for what batch_dims is supposed to mean. We can handle that in the issue that adds support for more than one batch dim, though.

gpytorch/kernels/linear_kernel.py
```diff
@@ -47,25 +47,32 @@ class LinearKernel(Kernel):
...
+    def __init__(self, num_dimensions, variance_prior=None, offset_prior=None, active_dims=None):
         super(LinearKernel, self).__init__(active_dims=active_dims)
         self.register_parameter(name="variance", parameter=torch.nn.Parameter(torch.zeros(1)), prior=variance_prior)
         self.register_parameter(
             name="offset", parameter=torch.nn.Parameter(torch.zeros(1, 1, num_dimensions)), prior=offset_prior
         )
 
-    def forward(self, x1, x2):
+    def forward(self, x1, x2, batch_dims=None, **params):
+        x1_ = x1 - self.offset
```
jacobrgardner: Given that we are hardcoding in checks for `batch_dims=(0, 2)` maybe it would at least make sense to raise a runtime error if it's something other than `(0)` or `(0, 2)`

gpytorch/kernels/multitask_kernel.py
```diff
@@ -41,44 +41,28 @@ def __init__(self, data_covar_module, n_tasks, rank=1, batch_size=1, task_covar_
...
+        if diag:
+            return res.diag()
+        else:
+            return res
 
     def size(self, x1, x2):
         """
         Given `n` data points `x1` and `m` datapoints `x2`, this multitask kernel returns an `(n*n_tasks) x (m*n_tasks)`
-        covariance matrix.
+        covariancn matrix.
```
jacobrgardner: spelling

gpytorch/kernels/linear_kernel.py
gpleiss: We do. It's in Kernel.__call__

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -103,13 +141,16 @@ def evaluate_kernel(self):
                 x1 = self.x1
                 x2 = self.x2
 
-            self._cached_kernel_eval = super(Kernel, self.kernel).__call__(x1, x2, **self.params)
+            self._cached_kernel_eval = super(Kernel, self.kernel).__call__(
+                x1, x2, diag=False, batch_dims=self.batch_dims, **self.params
+            )
             if self.squeeze_row:
-                self._cached_kernel_eval.squeeze_(-2)
+                self._cached_kernel_eval = self._cached_kernel_eval.squeeze(-2)
```
jacobrgardner: Did the in-place operations cause issues with autograd or something?

gpytorch/kernels/kernel.py
gpleiss: I don't understand what you mean.

jacobrgardner: Forget I said anything. As long as we do the check against `(0,2)` I am happy for now.

jacobrgardner:
I'm good with this after the comments above

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
gpleiss: Oh I think this may have been left over from some crazy debugging... I can revert it

gpleiss:
Cool the fixes are in

gpytorch/kernels/grid_interpolation_kernel.py
```diff
@@ -130,21 +129,26 @@ def _tight_grid_bounds(self):
...
 
-    def _compute_grid(self, inputs):
+    def _compute_grid(self, inputs, batch_dims):
         batch_size, n_data, n_dimensions = inputs.size()
+        if batch_dims == (0, 2):
+            inputs = inputs.view(inputs.size(0), inputs.size(1), -1, 1)
+            inputs = inputs.transpose(1, 2).contiguous()
+            batch_size = batch_size * inputs.size(1)
+            n_dimensions = n_dimensions // inputs.size(1)
+
```
Balandat: Should we raise a NotImplementedError here if that check is False and `batch_dims` is something other than `None`? What's the desired behavior for other arguments?

gpytorch/kernels/kernel.py
```diff
@@ -183,13 +183,28 @@ def _create_input_grid(self, x1, x2):
...
             * `full_covar`: (`b x n x 1 x d` and `b x 1 x m x d`)
-            * `full_covar` with `dim_groups=k`: (`b x k x n x 1 x (d//k)` and `b x k x 1 x m x (d//k)`)
+            * `full_covar` with `batch_dims=(0, 2)`: (`b x k x n x 1 x (d//k)` and `b x k x 1 x m x (d//k)`)
             * `diag`: (`b x n x d` and `b x n x d`)
-            * `diag` with `dim_groups=k`: (`b x k x n x (d//k)` and `b x k x n x (d//k)`)
+            * `diag` with `batch_dims=(0, 2)`: (`b x k x n x (d//k)` and `b x k x n x (d//k)`)
         """
-        return x1.unsqueeze(-2), x2.unsqueeze(-3)
+        x1_, x2_ = x1, x2
+        if batch_dims == (0, 2):
```
Balandat: again should we raise a NotImplementedError here if we get values that we don't yet support?

gpytorch/kernels/kernel.py
```diff
@@ -210,7 +225,78 @@ def __call__(self, x1, x2=None, **params):
...
+                # If we used batch_dims...
+                shape = self.size(x1_, x2_)
+                if batch_dims == (0, 2):
+                    if len(shape) == 2:
+                        expected_shape = torch.Size((x1_.size(-1), shape[0]))
+                    else:
+                        expected_shape = torch.Size((shape[0] * x1_.size(-1), shape[1]))
+                    if res.shape != expected_shape:
+                        raise RuntimeError(
+                            "The kernel {} is not equipped to handle batch_dims=(0, 2) "
```
Balandat: nit: name the different args in the string (being explicit is an upfront investment but will avoid bugs down the road...)

Balandat: same applies in other places

test/examples/test_kissgp_additive_regression.py
```diff
@@ -87,12 +90,12 @@ def test_kissgp_gp_mean_abs_error(self):
                     optimizer.n_iter += 1
                     optimizer.step()
 
-                for param in gp_model.parameters():
-                    self.assertTrue(param.grad is not None)
-                    self.assertGreater(param.grad.norm().item(), 0)
-                for param in likelihood.parameters():
-                    self.assertTrue(param.grad is not None)
-                    self.assertGreater(param.grad.norm().item(), 0)
+                    for param in gp_model.parameters():
```
Balandat: what's the reason you're checking this in every iteration?

gpytorch/kernels/kernel.py
jacobrgardner: At the moment the check in __call__ will raise an error if it's not something we support

Balandat:
no worries, those were minor points

PR #315, opened 2018-10-04
Fix train test split

konstantinklemmer:
Change train test split to 80 / 20 as given in the description (was 90 / 10 in the code)


gpleiss:
Thanks for catching that!

PR #322, opened 2018-10-12
[WIP] LazyTensors can handle multiple batches

gpleiss:
**Note** this PR does not break any existing code, and it passes all tests.

Proposal:
- LazyTensors (and all their operations) can handle arbitrarily many batch dimensions
- LazyTensors have a `.batch_shape` property and a `.matrix_shape` property - similar to `Distributions` (automatically determined from the `LazyTensor.shape` property)
- Merge LazyTensor._get_indices and LazyTensor._batch_get_indices 

LazyTensors right now are hard-coded to either have no batch dimension (just a matrix) or 1 batch dimension. We have lots of `if... else...` statement to deal with these cases. This PR would make it possible to make LazyTensors with arbitrarily many batch dimensions.

## What's done so far
- All core `gpytorch.functions` can handle arbitrarily many batch dimensions
- The core iterative methods (`linear_cg`, `lanczos_tridiag`, and `pivoted_cholesky`) all handle many batch dimensions
- LazyTensors have the new properties (`.batch_shape` and `.matrix_shape`).

Everything so far is mostly just refactors, but comments are welcome!
(Also this PR could be merged in as-is since it passes all tests, or it could wait until everything is done.)


gpytorch/utils/cholesky.py
```diff
@@ -10,19 +10,21 @@
...
-    potrf_list = []
-    for i in range(mat.size(0)):
-        potrf_list.append(mat[i].potrf().unsqueeze(0))
-    return torch.cat(potrf_list, 0)
+    potrf_list = [sub_mat.potrf() for sub_mat in mat.view(-1, *mat.shape[-2:])]
+    res = torch.cat(potrf_list, 0)
+    return res.view_as(mat)
 
 
 def batch_potrs(mat, chol):
```
Balandat: Can you add a TODO here for using batch cholesky decomp / solve once https://github.com/pytorch/pytorch/pull/11796 is done?

Balandat: (same up for potrf of course)

jacobrgardner:
Merging this now because Geoff and I need it for something we are working on.

gpytorch/utils/cholesky.py
Balandat: woohoo! Any rough estimates on how much faster this is compared to looping through?

vishwakftw: I am seeing at least a 2x speed improvement when compared with the implementation here. These are CPU times though:

| mat_dim | chol_dim | looping time | native_op_time |
| ----------- | ------------ | --------------- | ------------------- |
| 1000 x 3 x 3 | 1000 x 3 x 3 | 11.9 ms ± 53.8 µs per loop | 667 µs ± 10.7 µs per loop |
| 1000 x 5 x 5 | 1000 x 5 x 5 | 12.3 ms ± 264 µs per loop | 1.32 ms ± 84.4 µs per loop |
| 1000 x 20 x 20 | 1000 x 20 x 20 | 19.5 ms ± 162 µs per loop | 7.61 ms ± 501 µs per loop |
| 1000 x 50 x 50 | 1000 x 50 x 50 | 42.6 ms ± 1.96 ms per loop | 21.2 ms ± 1.87 ms per loop |

PR #329, opened 2018-10-18
[WIP] Default LazyTensor._quad_form_derivative implementation

jacobrgardner:
Alright, this is going to be a fun one. 

This PR implements a default implementation of `LazyTensor._quad_form_derivative`. After playing around with it, this will have two huge advantages:

1. **Simplicity.** By far the most difficult method to implement when implementing a LazyTensor is the derivative portion. See for example the giant mess that is things like https://github.com/cornellius-gp/gpytorch/blob/9ebbf05e8ea70132dbcce25671bb06fa1721eb81/gpytorch/lazy/kronecker_product_lazy_tensor.py#L97

With this approach, implementing this from scratch becomes optional. Thus, we truly only need the end user to define a matmul if they want to implement a new LazyTensor.

2. **Speed.** I've been playing around with this for a bit, and initial results suggest that relying on `torch.autograd.grad` is crazy fast. We get a **2x** speed up in some of the initial things I've been testing. For example:

```python
c1 = torch.randn(100)
c2 = torch.randn(100)
# "Ensure" positive definite
c1[0] = 10
c2[0] = 10

tlv1 = ToeplitzLazyTensor(c1)
tlv2 = ToeplitzLazyTensor(c2)
klv = KroneckerProductLazyTensor(tlv1, tlv2)

p = torch.randn(10000, 2)
q = torch.randn(10000, 2)

# Wall time: 64 ms
%time klv._quad_form_derivative(p, q)

# Wall time: 31 ms
%time super(KroneckerProductLazyTensor, klv)._quad_form_derivative(p, q)
```

### Things to do
- Ensure compatibility with all of the various batch modes.
- Run more timing results to see whether we should make this default call the standard behavior. If we aren't comfortable entirely scrapping the custom derivative methods, one option here would be to introduce a new `gpytorch.settings` flag for whether to use the super call or the custom call.
- Let's add a unit test to the LazyTensorUnitTest that checks custom derivative calls against the super call.

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -105,9 +105,17 @@ class LazyTensor(object):
             :obj:`torch.tensor`: derivative with respect to the arguments that are actually used to represent this
                                    this LazyTensor.
         """
-        raise NotImplementedError(
-            "The class {} requires a _quad_form_derivative function!".format(self.__class__.__name__)
-        )
+        args = self.representation()
+        for arg in args:
+            arg.requires_grad = True
+
+        loss = (left_vecs * self._matmul(right_vecs)).sum()
+        grads = torch.autograd.grad(loss, args)
+
+        for arg in args:
+            arg.requires_grad = False
+
+        return grads
 
     def _size(self):
         """
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -106,14 +106,20 @@ class LazyTensor(object):
                                    this LazyTensor.
         """
         args = self.representation()
-        for arg in args:
-            arg.requires_grad = True
+
+        toggled = [False] * len(args)
+
+        for i, arg in enumerate(args):
+            if not arg.requires_grad:
+                arg.requires_grad = True
+                toggled[i] = True
 
         loss = (left_vecs * self._matmul(right_vecs)).sum()
         grads = torch.autograd.grad(loss, args)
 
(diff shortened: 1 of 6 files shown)


Balandat:
autograd ftw!

gpleiss:
We should make sure that we’re doing careful profiling. I could imagine that our custom grads could be good when left_vecs/right_vecs is large, or when our LT is very simple - the extra matmul might be a bit of overhead.

jacobrgardner:
Yeah for sure. Achieving simplicity is quite simple since once we verify correctness in batch mode, etc we could merge as-is and then at least quad_form_derivative is optional.

For speed, we'll definitely want to carefully analyze on a LazyTensor by LazyTensor basis whether using the default or a custom implementation makes sense. NonLazyTensor is the obvious example where we want custom behavior, but there could be others.

jacobrgardner:
@gpleiss This should be all set when tests pass. I added test cases for batch and non batch `_quad_form_derivative` to the lazy tensor test suite.

One added benefit of this is that in addition to verifying the default derivative works for all of the lazy tensors, we also get to check all of our derivative implementations against autograd :-)

jacobrgardner:
(Closed and re-opened because Travis didn't want to run the tests after a force push following a rebase for some reason)

jacobrgardner:
Actually hold up on merging this -- just realized we shouldn't set args to `requires_grad=False` if they were set to True when entering the method.

jacobrgardner:
@gpleiss all set!

PR #330, opened 2018-10-18
feature: add rational quadratic kernel

samuelstanton:
Implement rational quadratic kernel

add `log_alpha` hyperparameter, otherwise implement similarly to `RBFKernel`


gpleiss:
Can you add a unit test? See the other kernels for examples.

samuelstanton:
@jacobrgardner @gpleiss I updated the PR and added a unit test for the kernel. In my own experiments on Mujoco dynamics regression, I've found that and RQ-ARD kernel can reduce predictive mean MSE by as much as 20% over a RBF-ARD kernel

jacobrgardner:
Wow what an old PR to finally get in :-)

PR #331, opened 2018-10-20
Repeat lazy tensor

gpleiss:
This adds a default repeat to LazyTensors

### DIFF
--- a/gpytorch/functions/__init__.py
+++ b/gpytorch/functions/__init__.py
@@ -30,7 +30,7 @@ def add_diag(input, diag):
         return input.add_diag(diag)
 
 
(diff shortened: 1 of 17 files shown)


jacobrgardner:
Looks good to me!

Minor point: the matmul could be made more efficient in the setting where the right hand side is also repeated (e.g. an `r x n x n` tensor against an `r x n` tensor where `r` denotes a repeat batch). 

Obviously, this is a stupid pattern since what you'd actually want to do is repeat the result of the multiplication, not repeat the matrices themselves. However, I do worry that this may come up when we want to compute e.g. the log det of a repeated tensor (using the same probe vectors for each repeat).

PR #332, opened 2018-10-22
Use custom KL div and log_prob with gpytorch.distributions.MultivarateNormal

gpleiss:
This moves the fast `log_prob` (previously in `ExactMarginalLogLikelihood`) and `kl_divergence` (previously in `MVNVariationalStrategy`) into `gpytorch.distributions.MultivariateNormal`.

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -1,8 +1,10 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import torch
+import math
 from torch.distributions import MultivariateNormal as TMultivariateNormal
 from torch.distributions.utils import _standard_normal, lazy_property
+from torch.distributions.kl import register_kl
 
 from ..lazy import LazyTensor, NonLazyTensor
 from .distribution import Distribution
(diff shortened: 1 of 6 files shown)


test/distributions/test_multivariate_normal.py
```diff
@@ -186,6 +201,47 @@ def test_multivariate_normal_batch_correlated_sampels_cuda(self):
         if torch.cuda.is_available():
             self.test_multivariate_normal_batch_correlated_sampels(cuda=True)
 
+    def test_log_prob(self):
```
Balandat: Can you actually test this against the log_prob call on `torch.distributions.MultivariateNormal` to ensure that this remains fully compatible with that?

jacobrgardner: Just pushed a commit to do this

gpytorch/distributions/multivariate_normal.py
```diff
@@ -167,3 +190,31 @@ def __mul__(self, other):
 
     def __truediv__(self, other):
         return self.__mul__(1. / other)
+
+
+@register_kl(MultivariateNormal, MultivariateNormal)
```
jacobrgardner: @gpleiss Does this mean that this will only work between two gpytorch MultivariateNormals? If so, in order to work with the pending variational refactor (specifically the variational distribution) we need to either:

1. Update this to do KL between a gpytorch MVN and any other MVN
2. Update gpytorch MultivariateNormal to accept `scale_tril` as a valid way of defining the covariance matrix.

jacobrgardner: Actually, I guess making a gpytorch MVN with a `CholLazyTensor` will work at least as well. I'll merge this!

PR #333, opened 2018-10-23
Throw error if ard_num_dims doesn't match the dimensionality

### LINKED ISSUE #328: ard_num_dims handled inconsistently across kernels
According to this comment(https://github.com/cornellius-gp/gpytorch/issues/314#issuecomment-427047286), it seems like ard_num_dims should always be equal to the dimension of the input dimension, and if it isn't, it should throw an error. However, SpectralMixtureModel is the only model that does so, the rest (RBFKernel and MaternKernel) fail in unexpected ways.

gpleiss:
Fixes #328

### DIFF
--- a/test/examples/test_kissgp_additive_classification.py
+++ b/test/examples/test_kissgp_additive_classification.py
@@ -32,7 +32,7 @@ class GPClassificationModel(gpytorch.models.AdditiveGridInducingVariationalGP):
         self.mean_module = ConstantMean(prior=SmoothedBoxPrior(-1e-5, 1e-5))
         self.covar_module = ScaleKernel(
             RBFKernel(
-                ard_num_dims=2, log_lengthscale_prior=SmoothedBoxPrior(exp(-5), exp(6), sigma=0.1, log_transform=True)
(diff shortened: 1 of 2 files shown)


test/examples/test_kissgp_additive_classification.py
```diff
@@ -32,7 +32,7 @@ def __init__(self):
         self.mean_module = ConstantMean(prior=SmoothedBoxPrior(-1e-5, 1e-5))
         self.covar_module = ScaleKernel(
             RBFKernel(
-                ard_num_dims=2, log_lengthscale_prior=SmoothedBoxPrior(exp(-5), exp(6), sigma=0.1, log_transform=True)
+                ard_num_dims=1, log_lengthscale_prior=SmoothedBoxPrior(exp(-5), exp(6), sigma=0.1, log_transform=True)
```
jacobrgardner: Looks like raising the error loudly is paying off already :-)

Maybe we should be explicit about how we intend for people to use ARD with additive structure in documentation

gpleiss: @jacobrgardner this actually only affects variational additive structure, which we're going to deprecate (or significantly change) anyways.

For additive regression ard_num_dims works in the expected way (=2 for 2D additive)

PR #335, opened 2018-10-29
Refactor Variational Inference

jacobrgardner:
This PR refactors our variational inference interface in a non-breaking way with the primary goal that the definition of the variational distribution `q(u)` should be decoupled from the strategy we use to get `q(f)` from `q(u)`. This will allow us to more easily play with approximate or factorized versions of `q(u)`.

As a result of this goal, this PR also introduces a few new features for variational inference.

## User Interface Changes
- `VariationalGP`, `GridInducingVariationalGP`, `AdditiveGridInducingVariationalGP`, `VariationalMarginalLogLikelihood` still function as expected but are all deprecated and will be removed in a future release.
- The preferred method of creating variational GPs is now to directly extend `AbstractVariationalGP` and use the new `VariationalDistribution` and `VariationalStrategy` objects to define how variational inference should be done.
- `VariationalDistribution` objects define the form of the variational distribution, `q(u)`. For now, this simply takes the form it did previously in `AbstractVariationalGP.variational_output`: we learn a variational mean `m` and a variational covariance matrix `S` parameterized by its Cholesky factorization.
- `VariationalStrategy` determines how to marginalize out the inducing point values to get `q(f|x) = \int p(f|x, u)q(u)du`. We implement two forms of this currently. For learned or non grid inducing points this follows Hensman et al., 2015. For grid inducing points this follows Wilson et al. 2016.
- `VariationalELBO` replaces `VariationalMarginalLogLikelihood`

## New features
The added flexibility from this refactor allows us to implement a few features we have been eyeing for a while:

- GPyTorch now supports SVGP style learned inducing points via the base `VariationalStrategy` with `learn_inducing_locations=True`. This is useful because grid structure is not obviously advantageous with the default `VariationalDistribution` due to the `O(m^2)` space and time cost of the variational covariance matrix `S`. [Here](https://github.com/cornellius-gp/gpytorch/blob/variational_refactor/examples/05_Scalable_GP_Regression_Multidimensional/SVDKL_Regression_SVGP_CUDA.ipynb) is an example notebook using this.
- **(Experimental)** GPyTorch now supports integration in to larger probabilistic models via Pyro if Pyro is installed. 
    - New [`gpytorch.models.PyroVariationalGP`](https://github.com/cornellius-gp/gpytorch/blob/variational_refactor/gpytorch/models/pyro_variational_gp.py) defines `model` and `guide` methods that function with any `VariationalStrategy` and `VariationalDistribution` combo.
    - `Likelihood` classes must define a `pyro_sample_y` method that takes in `q(f)` and computes then calls `pyro.sample` on `q(y)` to use this.
    - Examples of this integration are in progress and can be found [Here](https://github.com/cornellius-gp/gpytorch/tree/variational_refactor/examples/09_Pyro_Integration) (cc: @fritzo / https://github.com/uber/pyro/issues/1265)
    - **Caveat:** Using Pyro's SVI is _currently_ a performance decrease because we currently use the default pyro ELBO implementation which performs Monte Carlo estimation of the KL divergence. Our ELBO implementation (`VariationalELBO`) uses analytical KL divergence computation. The integration does work, however, and should allow for complex probabilistic modelling with fast GPs.

## Internal Changes
- The old `VariationalStrategy` and `MVNVariationalStrategy` no longer exist.
- To handle the added trace term when doing SGPR with `InducingPointKernel`, I created a new `AddedLossTerm` object that replaces the old variational strategy. `AddedLossTerm`s can be registered with `gpytorch.Modules` like old variational strategies, and add terms to the exact GP marginal log likelihood loss.


gpytorch/likelihoods/bernoulli_likelihood.py
```diff
@@ -56,3 +56,11 @@ def variational_log_probability(self, latent_func, target):
...
         target = target.unsqueeze(0).repeat(num_samples, 1).view(-1)
         return log_normal_cdf(samples.mul(target)).sum().div(num_samples)
+
+    def pyro_sample_y(self, variational_dist_f, y_obs, sample_shape, name_prefix=""):
+        import pyro
+
+        f_samples = variational_dist_f(sample_shape)
+        y_prob_samples = torch.distributions.Normal(0, 1).cdf(f_samples)
+        y_dist = pyro.distributions.Bernoulli(y_prob_samples)
+        pyro.sample(name_prefix + "._training_labels", y_dist.independent(1), obs=y_obs)
```
gpleiss: There's probably a way that we could consolidate `pyro_sample_y` and `variational_log_prob`. For example, lines 63-65 are the same for both. At this point, the use has to define 2 very similar functions for each likelihood.

Perhaps the user could define a `forward_samples` function that takes in samples from the MVN and returns a batch of likelihood samples?

gpleiss: Then `variatioanl_log_prob` and `pyro_sample_y` would both call these functions.

gpytorch/mlls/exact_marginal_log_likelihood.py
```diff
@@ -32,12 +32,15 @@ def forward(self, output, target):
...
         # Add terms for SGPR / when inducing points are learned
-        trace_diff = torch.zeros_like(res)
-        for variational_strategy in self.model.variational_strategies():
-            if isinstance(variational_strategy, MVNVariationalStrategy):
-                trace_diff = trace_diff.add(variational_strategy.trace_diff())
-        trace_diff = trace_diff / self.likelihood.log_noise.exp()
-        res = res.add(0.5, trace_diff)
+        added_loss = torch.zeros_like(res)
+        for added_loss_term in self.model.added_loss_terms():
+            if isinstance(added_loss_term, InducingPointKernelAddedLossTerm):
```
gpleiss: This is really hacky. I think `InducingPointKernel` should take in the GaussianLikelihood in the constructor. This way, the added loss term can deal with the noise itself.

(I realize what we had before was also pretty hacky, but if we're fixing it we should do something a bit better)

gpytorch/kernels/__init__.py
```diff
@@ -9,7 +9,7 @@
 from .grid_kernel import GridKernel
 from .grid_interpolation_kernel import GridInterpolationKernel
 from .index_kernel import IndexKernel
-from .inducing_point_kernel import InducingPointKernel
+from .inducing_point_kernel import InducingPointKernel, InducingPointKernelAddedLossTerm
```
gpleiss: I'm confused - is this defined in both `.mlls` and in `.inducing_point_kernel`?

gpytorch/variational/variational_distribution.py
```diff
@@ -0,0 +1,82 @@
...
+        mean_init += torch.randn_like(mean_init).mul(1e-1)
+        covar_init += torch.randn_like(covar_init).mul(1e-1)
+
+        self.register_parameter(name="variational_mean", parameter=torch.nn.Parameter(mean_init))
+        self.register_parameter(name="chol_variational_covar", parameter=torch.nn.Parameter(covar_init))
+
+        self.register_buffer("variational_params_initialized", torch.tensor(0))
+
+    @property
+    def variational_distribution(self):
```
gpleiss: Suuuuper minor point, but could this be a 'CholeskyVariationalStrategy' - which uses the super class 'VariationalStrategy' which has an abstract method for forward?

gpytorch/likelihoods/bernoulli_likelihood.py
jacobrgardner: We just discussed this via slack, and it's not so simple since the two methods actually diverge right after sampling from the latent function distribution.

gpytorch/kernels/__init__.py
jacobrgardner: Oops! I'd originally defined it in the kernel, but it's not there anymore and I forgot to remove the import

gpytorch/mlls/exact_marginal_log_likelihood.py
jacobrgardner: Will fix this, good point

jacobrgardner:
@gpleiss Should be all set!

PR #337, opened 2018-10-29
Heteroskedastic likelihoods and log-noise models

Balandat:
Allows to specify generic (log-) noise models that are used to obtain out-of-sample noise estimates. This allows e.g. to stick a GP to be fit on the (log-) measured standard errors of observed data into the GaussianLikelihood, and then jointly fit that together with the GP to be fit on the data.


gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -1,42 +1,56 @@
...
-            name="log_noise", parameter=torch.nn.Parameter(torch.zeros(batch_size, 1)), prior=log_noise_prior
-        )
+DEPRECATION_WARNING = "'GaussianLikelihood' was renamed to 'HomoskedasticGaussianLikelihood'"
 
-    @property
-    def noise(self):
-        return self.log_noise.exp()
 
-    def forward(self, input):
+class GaussianLikelihood(Likelihood):
```
jacobrgardner: I'd prefer if `GaussianLikelihood` became `_GaussianLikelihoodBase` and `HomoskedasticGaussianLikelihood` became `GaussianLikelihood` (with `HeteroskedasticGaussianLikelihood` remaining the same). I think homoskedastic is the common case, and this avoids making this a breaking change

Balandat: that's fine by me

Balandat:
This now supports also multi-task likelihoods. I subclassed the `MultitaskGaussianLikelihood` from the `_MultitaskGaussianLikelihoodBase` as with the non-MT one, using a trivial `MultitaskHomoskedasticNoise` model. Interestingly, this both simplifies the code and reduces wall time compared to the Kronecker implementation (see attached nb [remove .txt file extension]). 

[test_MultitaskHeteroskedasticLikelihood.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/2527594/test_MultitaskHeteroskedasticLikelihood.ipynb.txt)

@gpleiss, @jacobrgardner, lmk what you think - if this looks generally ok, I'll start working on fixing the test failures / updating the examples. 

Having these expressive heteroskedastic noise models is of high practical relevance in cases when there are (joint) observations of the error covariance available.

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -1,42 +1,45 @@
...
         if not isinstance(input, MultivariateNormal):
-            raise ValueError("GaussianLikelihood requires a MultivariateNormal input")
+            raise ValueError("Gaussian Likelihoods require a MultivariateNormal input")
         mean, covar = input.mean, input.lazy_covariance_matrix
-        noise = self.noise
-        if covar.ndimension() == 2:
-            if settings.debug.on() and noise.size(0) > 1:
-                raise RuntimeError("With batch_size > 1, expected a batched MultivariateNormal distribution.")
-            noise = noise.squeeze(0)
+        log_noise_covar = self.log_noise_covar(*params)
```
jacobrgardner: I would imagine we may want `log_noise_covar` modules to actually be just `noise_covar` modules, because most forms other than diagonal will probably have some other strategy to ensure positive definiteness (e.g., maybe it returns a `CholLazyTensor`). The standard one can just exponentiate by itself.

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -1,42 +1,45 @@
...
+        if isinstance(log_noise_covar, DiagLazyTensor):
```
jacobrgardner: I actually think we should modify `LazyTensor.__add__` to return an `AddedDiagLazyTensor` if we are adding a `DiagLazyTensor`. This removes this if statement, and I think we should always prefer to use `AddedDiagLazyTensors` so that preconditioning is possible.

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -1,42 +1,45 @@
...
+            full_covar = AddedDiagLazyTensor(covar, log_noise_covar.exp())
+        else:
+            # TODO: Poperly deal with non-diagonal noise covariance models
+            full_covar = covar + log_noise_covar.exp()
```
jacobrgardner: We may want to do `full_covar = full_covar.add_jitter()` after this call so that we can use preconditioning.

gpytorch/likelihoods/homoskedastic_noise.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
```
jacobrgardner: I assume this file should be deleted since it now lives in `noise_models.py`

gpytorch/models/exact_gp.py
```diff
@@ -18,8 +18,8 @@ def __init__(self, train_inputs, train_targets, likelihood):
             train_inputs = (train_inputs,)
         if train_inputs is not None and not all(torch.is_tensor(train_input) for train_input in train_inputs):
             raise RuntimeError("Train inputs must be a tensor, or a list/tuple of tensors")
-        if not isinstance(likelihood, GaussianLikelihood):
-            raise RuntimeError("ExactGP can only handle GaussianLikelihood")
+        if not isinstance(likelihood, _GaussianLikelihoodBase):
+            raise RuntimeError("ExactGP can only handle Gaussian Likelihoods")
```
jacobrgardner:
```suggestion
            raise RuntimeError("ExactGP can only handle Gaussian likelihoods")
```

gpytorch/likelihoods/multitask_gaussian_likelihood.py
```diff
@@ -14,7 +23,142 @@ def _eval_covar_matrix(task_noise_covar_factor, log_noise):
...
+        super(MultitaskGaussianLikelihood, self).__init__(
+            num_tasks=num_tasks,
+            log_noise_covar=log_noise_covar,
+            rank=rank,
+            task_correlation_prior=task_correlation_prior,
+            batch_size=batch_size,
+        )
+
+
+class MultitaskGaussianLikelihood_Kronecker(_MultitaskGaussianLikelihoodBase):
```
jacobrgardner:
```suggestion
class MultitaskGaussianLikelihoodKronecker(_MultitaskGaussianLikelihoodBase):
```

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -1,42 +1,45 @@
...
+class _GaussianLikelihoodBase(Likelihood):
+    def __init__(self, log_noise_covar):
+        super(_GaussianLikelihoodBase, self).__init__()
+        self.log_noise_covar = log_noise_covar
 
-    def forward(self, input):
+    def forward(self, input, *params):
         if not isinstance(input, MultivariateNormal):
-            raise ValueError("GaussianLikelihood requires a MultivariateNormal input")
+            raise ValueError("Gaussian Likelihoods require a MultivariateNormal input")
```
jacobrgardner:
```suggestion
            raise ValueError("Gaussian likelihoods require a MultivariateNormal input")
```

gpytorch/likelihoods/__init__.py
```diff
@@ -1,18 +1,25 @@
...
-    "Likelihood",
+    "_GaussianLikelihoodBase",
+    "_MultitaskGaussianLikelihoodBase",
+    "BernoulliLikelihood",
     "GaussianLikelihood",
+    "HeteroskedasticNoise",
+    "Likelihood",
     "MultitaskGaussianLikelihood",
-    "BernoulliLikelihood",
+    "MultitaskGaussianLikelihood_Kronecker",
```
jacobrgardner:
```suggestion
    "MultitaskGaussianLikelihoodKronecker",
```

gpytorch/likelihoods/__init__.py
```diff
@@ -1,18 +1,25 @@
...
-from .likelihood import Likelihood
-from .gaussian_likelihood import GaussianLikelihood
-from .multitask_gaussian_likelihood import MultitaskGaussianLikelihood
 from .bernoulli_likelihood import BernoulliLikelihood
+from .gaussian_likelihood import _GaussianLikelihoodBase, GaussianLikelihood
+from .likelihood import Likelihood
+from .multitask_gaussian_likelihood import (
+    _MultitaskGaussianLikelihoodBase,
+    MultitaskGaussianLikelihood,
+    MultitaskGaussianLikelihood_Kronecker,
```
jacobrgardner:
```suggestion
    MultitaskGaussianLikelihoodKronecker,
```

gpytorch/lazy/lazy_tensor.py
```diff
@@ -428,14 +428,15 @@ def evaluate_kernel(self):
...
-    def exact_predictive_mean(self, full_mean, train_labels, num_train, likelihood, precomputed_cache=None):
+    def exact_predictive_mean(self, full_mean, train_inputs, train_labels, num_train, likelihood, precomputed_cache=None):
         """
         Computes the posterior predictive covariance of a GP
         Assumes that self is the block prior covariance matrix of training and testing points
         [ K_XX, K_XX*; K_X*X, K_X*X* ]
 
         Args:
             full_mean (:obj:`torch.tensor`): the training and test prior means, stacked on top of each other
+            train_inputs TODO
```
jacobrgardner:
```suggestion
            train_inputs (:obj:`torch.tensor`): The training data inputs
```

gpytorch/functions/__init__.py
```diff
@@ -88,16 +89,17 @@ def exact_predictive_mean(full_covar, full_mean, train_labels, num_train, likeli
...
 
-def exact_predictive_covar(full_covar, num_train, likelihood, precomputed_cache=None):
+def exact_predictive_covar(full_covar, train_inputs, num_train, likelihood, precomputed_cache=None):
     """
     Computes the posterior predictive covariance of a GP
 
     Args:
         - full_covar ( (n+t) x (n+t) ) - the block prior covariance matrix of training and testing points
             [ K_XX, K_XX*; K_X*X, K_X*X* ]
+        - train_inputs TODO
```
jacobrgardner:
```suggestion
        - train_inputs (:obj:`torch.tensor`): The training data input
```

gpytorch/functions/__init__.py
```diff
@@ -66,14 +66,15 @@ def dsmm(sparse_mat, dense_mat):
...
-def exact_predictive_mean(full_covar, full_mean, train_labels, num_train, likelihood, precomputed_cache=None):
+def exact_predictive_mean(full_covar, full_mean, train_inputs, train_labels, num_train, likelihood, precomputed_cache=None):
     """
     Computes the posterior predictive mean of a GP
 
     Args:
         - full_covar ( (n+t) x (n+t) ) - the block prior covariance matrix of training and testing points
             [ K_XX, K_XX*; K_X*X, K_X*X* ]
         - full_mean (n + t) - the training and test prior means, stacked on top of each other
+        - train_inputs TODO
```
jacobrgardner:
```suggestion
        - train_inputs (:obj:`torch.tensor`): The training data input
```

Balandat:
So I made some good progress on this. 

The following nb shows how this can be used to build a joint model of both the actual tasks as well the outcomes and correlations between, using a Hadamard MTGP model, with a separate model of the same type stacked into the Heteroskedastic Likelihood module and trained on the observed log variances, providing  heteroskedastic OOS measurement error estimates. Since we typically have measurement errors for all observations within each task, it would be nice if there was a way to exploit the Kronecker structure within each task, while allowing different covariate values across the tasks.

[test_HadamardMultitaskMultiOutputHeteroskedasticLikelihood_universal.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/2566260/test_HadamardMultitaskMultiOutputHeteroskedasticLikelihood_universal.ipynb.txt)

Making this work in batch mode for MTPGs will require changes to LazyTensors (#361) if we want to avoid a lot of hacking/special casing.

Some of the handling of the input params in order to support the different modes is not very elegant (see https://github.com/cornellius-gp/gpytorch/pull/337/files#diff-52a9ec2599d1eaabfccb825b25c6b1eaR20). Welcoming ideas for how to make this cleaner.

Balandat:
So this is in relatively good shape except for the case of batched Multi-Task GPs. Here I'm running into some serious problems making this work b/c the lazies are not able to handle multi-dimensional batches properly. See `_MultitaskGaussianLikelihoodBase` for how I try to hack around this doing highly questionable things like calling `.evaluate()` unnecessarily. But even then, there are shape consistencies that pop up at runtime deep inside the LazyTensors.

The proper way to fix this would be to support general batch shapes throughout all LazyTensors (as well as support broadcasting). This seems like like it would be the right thing to do if we want the LazyTensor framework to be more generally useful.

jacobrgardner:
@Balandat We're all definitely in agreement that `LazyTensor` should handle arbitrary batch shapes. Do you think the two PRs you have for `DiagLazyTensor` (#362) and `KroneckerProductLazyTensor` (#369) will be sufficient for this setting? 

We have plans to go back and do a larger refactor of the lazy tensors to make them support arbitrary batch dimensions, so the question is what to do in the meantime.

Balandat:
> Do you think the two PRs you have for DiagLazyTensor (#362) and KroneckerProductLazyTensor (#369) will be sufficient for this setting?

@jacobrgardner No, unfortunately not. It turns out I actually need to do something smarter than just a Kronecker product to account for the different observation noise levels at the different data points. This requires `BlockDiagLazyTensor` and `MatmulLazyTensor` to work properly. I'm not entirely sure what's going on in the batch MT tests, maybe you can take a look to see if there is a quick and dirty short-term fix?

> We have plans to go back and do a larger refactor of the lazy tensors to make them support arbitrary batch dimensions.

Is there a timeline for this? I'd be happy to contribute to this if it helps getting it off the ground.

Balandat:
@gpleiss, @jacobrgardner I had to make some changes to the log_ deprecation to support the heteroskedastic likelihood setup, in the process I streamlined that getting rid of copy-pasted code.

This works except for the proper handling of batches in all Multitask scenarios (train batch eval single, train single eval batch etc.) - I hope to address this right after Thanksgiving. I may even just have those edge cases error out for now with an information warning to use the previous implementation that I still retained in the code.

jacobrgardner:
@Balandat Sounds good! The deprecation code does look much cleaner to me, thanks.

I would personally be fine with merging this in in a state that works for all models except batch multitask if that's where we are. The LT batch refactor is a larger thing that needs to happen anyways, and I think this already covers most common use cases (e.g. replacing WhiteNoiseKernel)

Balandat:
Ok great, so I update the docstrings to indicate that `MultitaskGaussianLikelihoodKronecker` is scheduled to be deprecated and should only be used for batched MTGP for the time being. Everything else can use the standard likelihoods as is, which are now implemented based on the generalized Likelihood setup using a trivial noise model. 

Except form the multi-batch MTGP these are all non-breaking changes.

I have some cool example notebooks for the heteroskedastic noise models and will put those up in a separate PR early next week.

jacobrgardner:
Oh I see, haha. I was doing my code review while you posted that. If you already have notebooks you are planning on putting up, then I’m okay with this getting merged now for sure.

adam-erickson:
Any update on the notebook for this?

Balandat:
I haven't gotten to it (clearly). Will try to put something up in the next couple of weeks.

t-vi:
@Balandat Is there something I can help to get an example for this?

Balandat:
@t-vi , sorry this fell off the radar.

Here is a simple nb that uses BoTorch's [`HeteroskedasticSingleTaskGP`](https://github.com/pytorch/botorch/blob/master/botorch/models/gp_regression.py#L273) for this: 
[demo_HeteroskedasticSingleTaskGP.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/4134688/demo_HeteroskedasticSingleTaskGP.ipynb.txt)

Here's how the model compares (visually) against the standard homoskedastic one:
heteroskedastic_example

This model requires known observation noise levels. There is also work on implementing what's called the "most likely heteroskedastic GP" that does not, but that needs to be cleaned up a little bit (I have some local changes that hopefully I can push out some time in the near futuer): https://github.com/pytorch/botorch/pull/250

t-vi:
Thank you for the quick reply! After I removed a few of the context managers (which I don't feel bad about) and the `log_scale` keyword argument (which is more scary), the heteroskedastic multitask notebook you posted here earlier seemed to work reasonably well, too.

Balandat:
Great. I haven't looked at that one in a while. One thing to note about `HeteroskedasticSingleTaskGP` in the nb I just posted is that it uses the `AddedLossTerm` abstraction to make sure that the likelihood term of the noise model is properly accounted for when computing the mll of the full model: https://github.com/pytorch/botorch/blob/master/botorch/models/gp_regression.py#L329-L332

I don't think this is properly done in the earlier nb, though I'd have to go back to it and check in more detail. Depending on the setting, this might not be a big deal in practice if things work empirically, but I'm sure you could come up with examples where this would fail.

salrm8:
Thank you @Balandat  for the np's. 
Before seeing your new np for `HeteroskedasticSingleTaskGP`, I tried to rewrite your previous np `HeteroskedasticSingleTaskGP` for a single task case. However, when trying to optimize the model hyperparameters, this message shows up: `AttributeError: 'ConstantMean' object has no attribute 'log_prob'`. Is it possible to resolve the issue? Thanks a lot!

Balandat:
@salrm8 Hmm something sounds off there, `log_prob ` should never be called on `ConstantMean`. Do you have a full repro (i.e. notebook) for this that you can share?

salrm8:
@Balandat , here is the code:
https://github.com/salrm8/gprTests/blob/master/gpr_hetero_SingleMulti.py
At the bottom of the code, we call both multi- and single-task GPs. The latter does not work. Thank you for your help.

Balandat:
@salrm8 your issue is the following block in the single task constructor: 
```
self.mean_module = gpytorch.means.ConstantMean(
    gpytorch.means.ConstantMean()
)
```

The first positional arg of `ConstantMean` is `prior`, so you register a mean module as a prior here, which messes things up in the MLL computation. If you remove this and just do 
```
```
things work.

salrm8:
@Balandat Thanks a lot! It works now.

PR #338, opened 2018-10-29
Fix bug in MatmulLazyTensor derivative return

jacobrgardner:
On master, `MatmulLazyTensor.quad_form_derivative` fails if the sub lazy tensors return derivatives with mutliple components. This fixes the return value to merge the two derivative components in to a single tuple.

cc @Balandat


Balandat:
Nice catch. Is there ever a case where non-tuples or nested tuples should be returned upstream? Otherwise, if everything consistently returned flat tuples, then you could just always do `left_grad + right_grad`.

jacobrgardner:
In the specific failure case you sent me one of the LTs was returning a tensor as the grad directly so that wouldn't work I think.

Maybe we could shorten the fix by doing something like

`left_grad = (left_grad,) if not isinstance(left_grad, tuple)`

And then add them

Balandat:
sg I can make that change

Issue #340, opened 2018-11-02
Issues with switching between batch and non-batch predictions


Balandat:
tl;dr: We need to better handle the precomputed cache for train-train covariance

Say you fit a GP in standard (non-batch) mode (`n x p` input).

- if you evaluate it in standard mode (`m x p` input), you're fine
- if you evaluate in batch mode (`b x m x p` input) **without having previously evaluated in standard mode**, you're fine
- if you first evaluate in standard mode, then in batch mode (or vice versa), an error is raised because of incompatible sizes of the train_train cache.

This is a bug, going back between the different modes should be possible.

Further, the following should be supported:
- fitting a GP in batch mode (`b x n x p` input) and then evaluating in "standard" mode (`b x m x p` input)
- fitting a GP in batch mode (`b x n x p` input)  and evaluating in batch-batch mode (`b' x b x n x p` input) 

The only thing that should not be supported is evaluating a GP fitted in batch mode (`b x n x p` input) on an incompatible size (`b' x n x p` input, where `b' != b`)

If the GP is fit in standard mode but evaluated in batched mode, we expand the train inputs here:
https://github.com/cornellius-gp/gpytorch/blob/2a96a881b77cb0f45d562e9146e89abbe4397591/gpytorch/models/exact_gp.py#L99

We then compute the same train-train covariance for each batch repeatedly during evaluation (b/c we use expand and Lazy tensors this is probably not terrible perf-wise, but unnecessary). Instead, what I would suggest is computing the train-train covariance cache once in standard mode, and then during the call of `exact_predictive_mean` do proper broadcasting here if necessary: https://github.com/cornellius-gp/gpytorch/blob/1a91ca523a24cfd9b6ad406c7994b0667b09177d/gpytorch/lazy/lazy_tensor.py#L483

This would likely require changes to how we deal with `full_mean` in the exact GP.

jacobrgardner:
The actual bug parts of this issue are fixed via #342 -- the only remaining thing left to do is make it so that in the multitask setting we don't duplicate work if possible.

Balandat:
So while this fixes forward passes, there are still issues with the backward pass, see https://github.com/cornellius-gp/gpytorch/pull/342#issuecomment-436725975

PR #342, opened 2018-11-03
Fix batch vs standard prediction inconsistencies

jacobrgardner:
Mostly solves #340, with the one exception that I couldn't get it so that we don't have to replicate the caches for batch multitask. Specifically, batch multitask fails without: https://github.com/cornellius-gp/gpytorch/blob/85c4907cc9a8c49cc440d9c19bf7e788fd00ea1c/gpytorch/models/exact_gp.py#L132

Basically this PR makes two changes:
1. Two of the error checks in LazyTensor.matmul were outright wrong, in the sense that matmul with those sizes works perfectly fine on both standard tensors and lazy tensors. For example this works just fine for both tensors and lazy tensors, despite what the error check said:
    ```python
    mat1 = torch.randn(3, 51, 100)
    mat2 = torch.randn(100)
    res = mat1.matmul(mat2) # Works fine
    ```
2. `exact_predictive_*` now just takes in `non_batch_train` if the train inputs are not batched. This lets us compute the caches without repeating work. The reason I fixed it this way is because I don't think there is a better way to do it without making wasted kernel calls.

**Note:** This PR doesn't implement the batch-batch mode `b' x b x n x d` that Max wanted because that is way way beyond the scope of just fixing a couple bugs here.

### DIFF
--- a/gpytorch/functions/__init__.py
+++ b/gpytorch/functions/__init__.py
@@ -66,7 +66,9 @@ def dsmm(sparse_mat, dense_mat):
     return DSMM(sparse_mat)(dense_mat)
 
 
-def exact_predictive_mean(full_covar, full_mean, train_labels, num_train, likelihood, precomputed_cache=None):
+def exact_predictive_mean(
+    full_covar, full_mean, train_labels, num_train, likelihood, precomputed_cache=None, non_batch_train=False
+):
     """
     Computes the posterior predictive mean of a GP
 
@@ -88,10 +90,12 @@ def exact_predictive_mean(full_covar, full_mean, train_labels, num_train, likeli
         from ..lazy.non_lazy_tensor import NonLazyTensor
 
         full_covar = NonLazyTensor(full_covar)
-    return full_covar.exact_predictive_mean(full_mean, train_labels, num_train, likelihood, precomputed_cache)
+    return full_covar.exact_predictive_mean(
+        full_mean, train_labels, num_train, likelihood, precomputed_cache, non_batch_train
+    )
 
 
-def exact_predictive_covar(full_covar, num_train, likelihood, precomputed_cache=None):
+def exact_predictive_covar(full_covar, num_train, likelihood, precomputed_cache=None, non_batch_train=False):
     """
     Computes the posterior predictive covariance of a GP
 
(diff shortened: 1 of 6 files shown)


Balandat:
Many thanks. I agree that the `b' x b x n x d` batch mode is a separate ask and outside the scope of this. Just thought I'd bring it up, given the recent changes @gpleiss made to support arbitrary batch sizes.

Balandat:
@jacobrgardner, so it turns out that while this fixes the forward pass, running a backward pass on a function of a batch evaluation of a non-batch model errors out. 

Say you trained a GP `model` in non-batch mode (using `train_x`), and then evaluate in batch mode (this PR ensures you can do this): `posterior = model(test_x_batch)`.

Now `posterior.mean.sum().backward()` will error out with a shape error:
```
<ipython-input-26-eb46631e8e66> in <module>()
      1 f_post = model(test_x_batch)
----> 2 f_post.mean.sum().backward()
      3 test_x.grad

/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/experimental/ae/bento_kernel_ae_experimental#link-tree/torch/tensor.py in backward(self, gradient, retain_graph, create_graph)
    100                 products. Defaults to ``False``.
    101         """
--> 102         torch.autograd.backward(self, gradient, retain_graph, create_graph)
    103 
    104     def register_hook(self, hook):

/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/experimental/ae/bento_kernel_ae_experimental#link-tree/torch/autograd/__init__.py in backward(tensors, grad_tensors, retain_graph, create_graph, grad_variables)
     88     Variable._execution_engine.run_backward(
     89         tensors, grad_tensors, retain_graph, create_graph,
---> 90         allow_unreachable=True)  # allow_unreachable flag
     91 
     92 

/data/users/balandat/fbsource/fbcode/buck-out/dev-nosan/gen/experimental/ae/bento_kernel_ae_experimental#link-tree/gpytorch/functions/_matmul.py in backward(self, grad_output)
     55                 lazy_tsr = self.representation_tree(*matrix_args)
     56             rhs_grad = lazy_tsr._t_matmul(grad_output)
---> 57             rhs_grad = rhs_grad.view(rhs_shape)
     58 
     59         return tuple([rhs_grad] + list(arg_grads))

RuntimeError: shape '[100, 1]' is invalid for input of size 200
```

jacobrgardner:
@Balandat I'll take a look

PR #346, opened 2018-11-04
GridKernel now functions stand-alone.

jacobrgardner:
`GridKernel` can now correctly be used as a scalable GP kernel in the case where the training data lies on a regularly spaced grid. See the new `Grid_GP_Regression.ipynb` for an example of usage.

This maintains the efficiency updates made in #345 for GridInterpolationKernel in the interpolation setting, but forming the full grid seems practically unavoidable in the "your training data actually lies on a grid" setting.


gpleiss:
This looks good, but we should probably also add something to `test/examples`.

jacobrgardner:
Test added! I'll put up another PR that allows `GridKernel` to support different sized grids in each dimension later.

Issue #291, opened 2018-09-26
Document various ways GPs can be used in batch mode


Balandat:
Batch mode (fitting and predicting) is currently a "hidden" feature as it's not very well documented and does not have a tutorial. 

We should add documentation and also unit tests to make sure this works as expected (this was the cause for a lot of bugs in the past).

Balandat:
I guess this is particularly important for batch-training of independent GPs, b/c the standard pattern 
```
loss = -mll(output, train_y)
loss.backward()
```
will fail. If none of the hyperparameters are shared, then a straightforward way to do this is to use `loss= -mll(output, train_y).sum()`, but things get murkier if some parameters are shared across batches since the losses of the individual batches will not be independent w.r.t. to the model parameters.

Balandat:
This is related to #130

gpleiss:
I'll work on an example notebook when I go through and fix all the other notebooks.

gpleiss:
I'm also working on better batch mode tests r.n. as part of the batch mode PR

Balandat:
Thanks a lot for the comprehensive tests in your PR. We should probably make sure to write up some documentation about this - maybe even a little feature matrix showing which of the following can be combined, and what the expected train/test input dimensions and resulting output dimensions are:
- batch training (shared hyperparameters)
- batch training (separate hyperparameters)
- batch evaluation (arbitrary input batch size)
- single-task vs. multi-task
- approximate GPs

gpleiss:
I'll work on it!

h-jia:
Also problem with me. Hope this can be fixed. Thanks

gpleiss:
There are now some docs about batch GPs: https://gpytorch.readthedocs.io/en/latest/batch_gps.html

Closing this for now, though I'm happy for any suggestions on how to improve the docs!

PR #351, opened 2018-11-07
bug fixes for examples

dhpollack:
A few of the examples did not run correctly.  Specifically, the classification example expected at least a 2D input (batch x features) and the Kroenecker Classification seems to use an older import nomenclature.

### DIFF
--- a/examples/02_Simple_GP_Classification/Simple_GP_Classification.ipynb
+++ b/examples/02_Simple_GP_Classification/Simple_GP_Classification.ipynb
@@ -16,7 +16,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -43,8 +43,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "train_x = torch.linspace(0, 1, 10).unsqueeze(1)\n",
-    "train_y = torch.sign(torch.cos(train_x * (4 * math.pi))).squeeze()"
+    "train_x = torch.linspace(0, 1, 10)\n",
+    "train_y = torch.sign(torch.cos(train_x * (4 * math.pi)))"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


gpleiss:
Thanks for the PR! For the simple classification example, it should be possible to pass in a 1D input (the model should automatically take care of unsqueezing the input). It's possible that there may be a bug in the variational model that prevents this from happening. I'll take a look.

dhpollack:
Here is the exact error

```
<ipython-input-5-f47a90b795e1> in <module>
      7     test_x = torch.linspace(0, 1, 101)
      8     # Get classification predictions
----> 9     observed_pred = likelihood(model(test_x))
     10 
     11     # Initialize fig and axes for plot

~/miniconda3/envs/pytorch-nightly/lib/python3.7/site-packages/gpytorch/models/abstract_variational_gp.py in __call__(self, inputs, **kwargs)
     17 
     18     def __call__(self, inputs, **kwargs):
---> 19         return self.variational_strategy(inputs)

~/miniconda3/envs/pytorch-nightly/lib/python3.7/site-packages/gpytorch/module.py in __call__(self, *inputs, **kwargs)
    180 
    181     def __call__(self, *inputs, **kwargs):
--> 182         outputs = self.forward(*inputs, **kwargs)
    183 
    184         if isinstance(outputs, tuple):

~/miniconda3/envs/pytorch-nightly/lib/python3.7/site-packages/gpytorch/variational/variational_strategy.py in forward(self, x)
     82             return variational_dist
     83         else:
---> 84             n_induc = self.inducing_points.size(-2)
     85             full_inputs = torch.cat([self.inducing_points, x])
     86             full_output = self.model.forward(full_inputs)

RuntimeError: Dimension out of range (expected to be in range of [-1, 0], but got -2)
```

dhpollack:
Are you trying to get the size of the batch?  If so, `size(0)` should work, no?  I haven't looked at the internals of your code, but naively fixed example so it would run.

jacobrgardner:
I think both of these are on me in the VI refactor #335 . I must have missed a notebook when cleaning up the imports, and I think maybe the dimension check got scrapped from AbstractVariationalGP / VariationalStrategy. 

Let me push a commit fixing that and then I'll merge this.

gpleiss:
@dhpollack `size(0)` won't always work, since all of these models are designed to function on batches of inputs (which would correspond to a batch of independent GPs). GPyTorch can handle the following types of inputs:

1D tensor (size `n`) - assumed to be a vector of single dimensional inputs
2D tensor (size `n x d`) - assumed to be a matrix of `d`-dimensional inputs
3D tensor (size `b x n x d`) - assumed to be a `b`-batch of matrices of `d`-dimensional inputs. If you want a batch of single dimensional inputs, then you have to explicitly form a `b x n x 1` tensor.

This is the pattern we're following in regression, and what we thought we were also doing for regression 😬. The regression models have a check that does the unsqueezing automatically. I think @jacobrgardner might have forgotten those when he refactored the variational models.

(Sorry @jacobrgardner, not trying to throw you under the bus or anything 😁)

gpleiss:
I have an open issue (#291) to add this explanation to the docs, so hopefully that should be there soon!

dhpollack:
@gpleiss thanks for the explanation.  I just started using this library, but quickly going through the documentation, I saw something that I found a bit confusing.  

You use a parameter called `batch_size`, but I would say this is actually the number of batches.  If one has an `b x n x d`, then the batch size is `n` and the number of batches is `b`.  Could be different for you guys, but just thought I'd let you know.  

I love that you guys are super responsive.  :+1:

dhpollack:
... but actually, that's a totally different issue from this.

jacobrgardner:
@dhpollack Yeah, what we currently refer to as `batch_size` in code is the number of batches, so maybe `num_batches` is a better name for it globally. That change would be pretty widely breaking in terms of the user interface, however...

PR #353, opened 2018-11-07
root_decomposition returns a RootLazyTensor

jacobrgardner:
This builds off of #352 to make `root_decomposition` return a `RootLazyTensor` instead of an evaluated root directly. This has the advantage that lazy tensors with efficient root decompositions can build their own. 

Right now this improves the efficiency of `DiagLazyTensor` and `BatchRepeatLazyTensor`, and the stability of `DiagLazyTensor`.

In addition, `MulLazyTensor` special cases `DiagLazyTensor` roots to not evaluate the root.

@gpleiss

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -826,7 +826,7 @@ class LazyTensor(object):
         if self.size(0) == 1:
             return self.sum_batch()
 
-        roots = self.root_decomposition()
+        roots = self.root_decomposition().root.evaluate()
         n_batch = roots.size(0) if mul_batch_size is None else mul_batch_size
         true_batch_size = roots.size(0) // mul_batch_size if mul_batch_size is not None else 1
 
@@ -858,7 +858,7 @@ class LazyTensor(object):
                 break
             else:
                 res = MulLazyTensor(RootLazyTensor(part1), RootLazyTensor(part2)).evaluate_kernel()
-                roots = res.root_decomposition()
+                roots = res.root_decomposition().root.evaluate()
                 n_batch = n_batch // 2
 
         return res
@@ -931,6 +931,7 @@ class LazyTensor(object):
         This can be used for sampling from a Gaussian distribution, or for obtaining a
         low-rank version of a matrix
         """
+        from .root_lazy_tensor import RootLazyTensor
         if not self.is_square:
             raise RuntimeError(
                 "root_decomposition only operates on (batches of) square (symmetric) LazyTensors. "
@@ -945,7 +946,7 @@ class LazyTensor(object):
             batch_shape=self.batch_shape,
             matrix_shape=self.matrix_shape,
         )(*self.representation())
-        return res
+        return RootLazyTensor(res)
 
     def root_inv_decomposition(self, initial_vectors=None, test_vectors=None):
         """
@@ -1155,7 +1156,7 @@ class LazyTensor(object):
         if self.size()[-2:] == torch.Size([1, 1]):
             covar_root = self.evaluate().sqrt()
         else:
-            covar_root = self.root_decomposition()
+            covar_root = self.root_decomposition().root
 
         if self.ndimension() == 3:
             base_samples = torch.randn(
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1317,9 +1317,12 @@ class LazyTensor(object):
         used.
         """
         from .zero_lazy_tensor import ZeroLazyTensor
+        from .diag_lazy_tensor import DiagLazyTensor
 
         if isinstance(other, ZeroLazyTensor):
             return other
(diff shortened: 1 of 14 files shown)


gpytorch/lazy/mul_lazy_tensor.py
```diff
@@ -119,12 +120,17 @@ def _matmul(self, rhs):
 
         # Here we have a root decomposition
         if isinstance(self.left_lazy_tensor, RootLazyTensor):
-            left_root = self.left_lazy_tensor.root.evaluate()
+            if isinstance(self.left_lazy_tensor.root, DiagLazyTensor):
```
gpleiss: I would want to see this tested out in practice. From the way that `MulLazyTensor` is recursively built, the size of the left/right vectors for `_quad_form_derivative` can grow exponentially unless you use a small root decomposition for all tensors. This is why we do a root decomposition on both the left and right tensors, rather than just one of them.

test/lazy/test_batch_repeat_lazy_tensor.py
```diff
@@ -23,6 +26,7 @@ def evaluate_lazy_tensor(self, lazy_tensor):
 
 class TestBatchRepeatLazyTensorBatch(BatchLazyTensorTestCase, unittest.TestCase):
     seed = 0
+    should_test_sample = True
 
     def create_lazy_tensor(self):
         toeplitz_column = torch.tensor([[4, 0, 0, 1], [3, 0, -0.5, -1]], dtype=torch.float, requires_grad=True)
```
gpleiss: Right now I don't think `root_decomposition` and `root_inv_decomposition` are tested in `LazyTensorTestCase`. It would probably be good to add those in.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -899,7 +900,7 @@ def root_decomposition(self):
             batch_shape=self.batch_shape,
             matrix_shape=self.matrix_shape,
         )(*self.representation())
-        return res
+        return RootLazyTensor(res)
```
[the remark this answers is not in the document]
jacobrgardner: In the old interpretation of `root_decomposition`, yes. My thought was to reinterpret `root_decomposition` now as "return a `RootLazyTensor` that approximates `self`. This is somewhat convenient because in several places the first thing we did after calling `root_decomposition` was create a `RootLazyTensor` anyways.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1109,7 +1110,7 @@ def zero_mean_mvn_samples(self, num_samples):
         if self.size()[-2:] == torch.Size([1, 1]):
             covar_root = self.evaluate().sqrt()
         else:
-            covar_root = self.root_decomposition()
+            covar_root = self.root_decomposition().root
```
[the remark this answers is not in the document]
jacobrgardner: Definitely! Just wanted to get your input on whether having `root_decomposition` work this way was a good idea before I handled `root_inv_decomposition` too :-).

gpytorch/lazy/mul_lazy_tensor.py
jacobrgardner: I Reverted `MulLazyTensor` back to normal and special cased `__mul__` against `DiagLazyTensor` as well as `DiagLazyTensor.__mul__`.

test/lazy/_lazy_tensor_test_case.py
```diff
@@ -143,6 +143,26 @@ def test_add_diag(self):
         actual = evaluated + other_diag.diag()
         self.assertTrue(approx_equal(res, actual))
 
+    def test_root_decomposition(self):
```
Balandat: Hmm before looking at this I guess I wasn't really aware that RootLazyTensor matmul is actually multiplying the LazyTensor with the rhs, not the root. I feel this is slightly confusing but I guess reasonable....

test/priors/test_multivariate_normal_prior.py
```diff
@@ -16,12 +16,15 @@ def test_multivariate_normal_prior_to_gpu(self):
             self.assertEqual(prior.scale_tril.device.type, "cuda")
             self.assertEqual(prior.precision_matrix.device.type, "cuda")
 
-    def test_multivariate_normal_prior_validate_args(self):
-        # This should be a ValueError in pytorch, see pytorch Github issue #11997
-        with self.assertRaises(ValueError):
-            mean = torch.tensor([0.0, 1.0])
-            cov = torch.tensor([[1.0, 2.0], [2.0, 0.5]])
-            MultivariateNormalPrior(mean, covariance_matrix=cov, validate_args=True)
+    # def test_multivariate_normal_prior_validate_args(self):
```
Balandat: Hmm I do correctly get a ValueError here. Maybe it was a bad short-lived commit upstream? Let.s add this back in (and remove the comment)

gpleiss:
Fixed the conflicts, rebased, and will merge once Travis passes

PR #355, opened 2018-11-08
VariationalStrategy now supports batch mode

jacobrgardner:
- You can now use VGP / SVGP in batch mode to train and test multiple independent variational GPs simultaneously (this already functioned for grid interpolation based variational GPs).
- Currently functions for GaussianLikelihood, will add BernoulliLikelihood support in a follow-up PR.
- Adds an SVGP Batch Regression example.


gpytorch/mlls/variational_elbo.py
```diff
@@ -29,8 +29,14 @@ def forward(self, variational_dist_f, target, **kwargs):
 
         log_likelihood = self.likelihood.variational_log_probability(variational_dist_f, target, **kwargs).div(
             num_batch
-        )
-        kl_divergence = torch.distributions.kl.kl_divergence(variational_dist_u, prior_dist).sum().div(self.num_data)
+        ).sum()
```
gpleiss: I think we need to divide by `num_batch` here?

gpleiss: Or rather divide by the batch size

gpytorch/variational/variational_strategy.py
```diff
@@ -85,24 +85,36 @@ def forward(self, x):
...
             full_output = self.model.forward(full_inputs)
             full_mean, full_covar = full_output.mean, full_output.lazy_covariance_matrix
 
-            test_mean = full_mean[n_induc:]
-            induc_mean = full_mean[:n_induc]
-            induc_induc_covar = full_covar[:n_induc, :n_induc].add_jitter()
-            induc_data_covar = full_covar[:n_induc, n_induc:]
-            data_induc_covar = full_covar[n_induc:, :n_induc]
-            data_data_covar = full_covar[n_induc:, n_induc:]
+            if full_covar.dim() == 3:
```
gpleiss: Eventually, we can collapse this using ellipses. However, ATM I don't think that our __getitem__ supports ellipses, so that's for a later date.

gpytorch/mlls/variational_elbo.py
gpleiss: Sorry - the git diff threw me off here. This is totally fine.

gpleiss:
RE: the `.sum()` inside the `VariationalELBO` - this differs from what is currently in `ExactMarginalLogLikelihood`. For batch GP mode, you have to [manually sum the batches together for the loss](https://gpytorch.readthedocs.io/en/latest/examples/01_Simple_GP_Regression/Simple_Batch_Mode_GP_Regression.html#Training-the-model)

We should be consistent between exact and variational. I think I'm generally in favor of not doing the sum inside the MLL because 1) this allows the user to have more fine-grained control over each GP in the batch and 2) it would be a breaking change to change what we have in ExactMLL

gpleiss:
Actually, I guess it's not necessarily a breaking change for ExactMLL, since calling `.sum()` on an already-summed tensor won't hurt. But I think 1) still applies.

gpytorch/mlls/variational_elbo.py
jacobrgardner: `num_data` is correct!

jacobrgardner:
Yeah the `.sum()` on the log_likelihood is just a leftover mistake -- fixing it now

gpytorch/mlls/variational_elbo.py
```diff
@@ -30,7 +30,13 @@ def forward(self, variational_dist_f, target, **kwargs):
         log_likelihood = self.likelihood.variational_log_probability(variational_dist_f, target, **kwargs).div(
             num_batch
         )
-        kl_divergence = torch.distributions.kl.kl_divergence(variational_dist_u, prior_dist).sum().div(self.num_data)
+
+        kl_divergence = torch.distributions.kl.kl_divergence(variational_dist_u, prior_dist).unsqueeze(-1)
+
+        if log_likelihood.numel() == 1:
+            kl_divergence = kl_divergence.sum()
```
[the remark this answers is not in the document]
jacobrgardner: No (e.g. the DKL CIFAR100 stuff still works) -- in the batch additive case `log_likelihood` will be a scalar so we'll be summing across all dimensions for the `kl_divergence` anyways.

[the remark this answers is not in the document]
gpleiss: Let's say you're fitting `b` GPs, each with `l` additive terms. `log_likelihood` will be a vector of length `b`, and (I'm assuming) the kl_divergence will be a `b x l` matrix? This will fail in this case.

PR #356, opened 2018-11-08
Fix test modules

jacobrgardner:
**Merge #355 first.**

I was creating unit tests for #355 and decided to fix a ton of issues with our unit tests. This PR:

1. Creates a couple simple unit tests for batch mode VI (as well as a `models` folder in `tests` that I'll start filling out. This was what I originally set out to do before I decided to fix the next three issues.
1. Creates and populates `__init__.py` files for all of the test subfolders. This allows us to run specific tests for example by doing `python -m unittest test.lazy.TestCholLazyTensor` or by running the corresponding `.py` file (which previously only worked if it didn't try to import anything from the test folder like `test._utils`)
1. Fix undiscovered tests. Before this PR, running the unittests locally ran 689 tests. After, we run nearly **1400** tests. I suspect that some files with multiple test classes were only getting the first one run (like all the `LazyTensor` test files that define both a standard and a batch test class). Fortunately everything passes.
1. Fix large numbers of spelling errors and naming inconsistencies in test class names (e.g. `KissGP` vs `KISSGP`, duplicate class names from multiple files, etc).


test/models/_model_test_case.py
```diff
@@ -0,0 +1,56 @@
+from __future__ import absolute_import
```
gpleiss: Love this!!!

test/models/_model_test_case.py
```diff
@@ -0,0 +1,56 @@
...
+        raise NotImplementedError()
+
+    def test_forward_train(self):
+        data = self.create_test_data()
+        model = self.create_model(data)
+        model.train()
+        output = model(data)
+        self.assertTrue(output.lazy_covariance_matrix.dim() == 2)
+        self.assertTrue(output.lazy_covariance_matrix.size(-1) == data.size(-2))
+        self.assertTrue(output.lazy_covariance_matrix.size(-2) == data.size(-2))
```
gpleiss: We should probably also add checks here about the gradients of the hyperparameters not being zero.

test/models/_model_test_case.py
gpleiss: It might be good to see if we could make this the superclass for most of our tests in `test/examples`. We could add a training loop in these tests, and maybe we can add a flag that optionally turns off the training loop tests if we just want a fast run of the test suite.

gpleiss:
@jacobrgardner there may have been some undiscovered tests, but now running `python -m unittest` runs every test twice. You can confirm this by adding `self.assertTrue(False)` to one of the unittests, and seeing that you get two test failures.

Additionally, you can run a single test file by running `python -m unittest test.lazy.TestCholLazyTensor` (like you said), or if you want to run all LazyTensor tests you can run `python -m unittest discover -s test/lazy`.

jacobrgardner:
> @jacobrgardner there may have been some undiscovered tests, but now running `python -m unittest` runs every test twice. You can confirm this by adding `self.assertTrue(False)` to one of the unittests, and seeing that you get two test failures.

Whoops! I'll fix this. I think I misunderstood how this was supposed to work :stuck_out_tongue:

gpleiss:
@jacobrgardner looks good to me! I added some tests to check the gradients of models as well. Why don't you take a look at that and merge it in if it looks good to you?

PR #362, opened 2018-11-09
Support for general batch shapes in DiagLazyTensor

Balandat:
This is WIP and a lot of it is untested, but I wanted to give an early preview to solicit feedback on this.


gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -3,159 +3,131 @@
...
+                with the last two dimensions represents the indices as lists of row and column indices,
+                that is, selecting the `b1 x ... x bk x t(0, l) x t(1, l)`-th element.
+                The batch indices follow standard broadcasting rules.
+
+        """
+        batch_shape = _mul_broadcast_shape(indices.shape[:-2], self._diag.shape[:-1])
+        row_indices = indices[..., 0, :].expand(batch_shape + indices.shape[-1:])
+        diag = self._diag.expand(batch_shape + self._diag.shape[-1:])
+
+        diag_elements = torch.stack(
```
Balandat: yeah I know I said things would be easier... but supporting arbitrary batch shapes in both the indices as well as the underlying `DiagLazyTensor` as well as broadcasting across the batch shapes requires some ... finesse.

gpytorch/utils/broadcasting.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import torch
+
+
+def _mul_broadcast_shape(shape_a, shape_b):
```
Balandat: It really hurts me to write this kind of code in python. This should live in C++ in ATen. Ideally, these shape checks would be factored out of the respective tensor operations there, and get an API so it can be called standalone.

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -3,159 +3,131 @@
...
-                )
-
-        res = tensor.div(self._diag.unsqueeze(-1))
-        if is_vec:
-            res = res.squeeze(-1)
-        return res
+        return self.inverse()._matmul(tensor)
 
     def inv_quad_log_det(self, inv_quad_rhs=None, log_det=False, reduce_inv_quad=True):
-        if inv_quad_rhs is not None:
```
gpleiss: I think all these checks are still necessary, and should work for multi-batch mode?

gpytorch/lazy/diag_lazy_tensor.py
gpleiss: Alternatively, we could do `(*batch_indices, row_indices, col_indices)`. And since this is an internal method, it makes sense that we you would strictly need to supply one index per tensor dimension.

Balandat: So the proposed interface supports selecting different elements for each batch, as well as broadcasting over arbitrary batch dimensions between the input indices and the actual batch size of the DiagLazyTensor. 

@gpleiss, sounds like you're suggesting to keep this simple and just get the same indices for each batch. In this case we can default to returning elements of all batches if no batch_indices are provided, and returning those of the specific batch if they are.

This would be quite different and less flexible than what I have currently, but on the other hand, it might be good not to overcomplicate things for now.

jacobrgardner: I've been playing with this a bit, and in general, the `_batch_get_indices` and `_get_indices` methods have inconsistent behavior on master, especially compared to what happens with torch tensors. For example:

```python
import torch
from gpytorch.lazy import NonLazyTensor, DiagLazyTensor, ToeplitzLazyTensor
torch_mat = torch.randn(3, 5, 5)
non_lt = NonLazyTensor(torch_mat)
diag_lt = DiagLazyTensor(torch.randn(3, 5))
toep_lt = ToeplitzLazyTensor(torch.randn(3, 5))

print(torch_mat.shape, non_lt.shape, diag_lt.shape, toep_lt.shape) # All are torch.Size([3, 5, 5])

index = ([0, 1], [0, 0], [0, 1])
print(torch_mat[index].shape)  # torch.Size([2])
print(non_lt[index].shape)  # torch.Size([2])
print(diag_lt[index].shape)  # torch.Size([2, 2, 2])
print(toep_lt[index].shape)  # torch.Size([2, 2, 2])
```
(Also, `diag_lt.shape` returns a tuple instead of a `torch.Size`).

I think we should standardize these, and in my opinion the clearly correct choice is to mirror what happens with the torch tensor: `mat[[0, 0], [0, 0], [0, 1]]` should return a tensor containing `mat[0, 0, 0]` and `mat[0, 0, 1]`.

jacobrgardner: Actually, my point appears to be a more minor bug dealing with lists vs tensors as indexes:
```python
print(toep_lt[[0, 1], [0, 0], [0, 1]].shape)  # torch.Size([2, 2, 2]) -- Bad!
a = torch.tensor([0, 1])
b = torch.tensor([0, 0])
print(toep_lt[a, b, a].shape)  # torch.Size([2]) -- Good!
```

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -3,159 +3,131 @@
...
+        diag_elements = torch.stack(
+            [diag[batch_idx][row_indices[batch_idx]] for batch_idx in product(*(range(i) for i in batch_shape))]
+        ).view(batch_shape + indices.shape[-1:])
+
+        equal_indices = (indices[..., -1, :] == indices[..., -2, :]).type_as(self._diag)
+        return diag_elements * equal_indices
+
+    def _matmul(self, rhs):
+        # this are trivial if we multiply two DiagLazyTensors
+        if isinstance(rhs, DiagLazyTensor):
```
jacobrgardner: I'd prefer if this check appeared in an overwritten `matmul` instead of in `_matmul` directly (I feel that `_matmul` should always return tensors). On the subject, I think that `LazyTensor.matmul` should by default return a `MatmulLazyTensor` if it gets called with a `LazyTensor` right hand side instead of erroring.

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -3,159 +3,131 @@
...
+        return DiagLazyTensor(self._diag.sqrt())
 
     def root_inv_decomposition(self, initial_vectors=None, test_vectors=None):
-        if self.exact_root_decomposition:
-            return self.__class__(self._diag.sqrt().reciprocal()).evaluate()
-        else:
-            return super(DiagLazyTensor, self).root_decomposition()
+        return self.root_decomposition().inverse()
+
+    def sum_batch(self, sum_batch_size=None):
```
jacobrgardner: This should be easy to define / shouldn't need to raise a `NotImplementedError`. Just sum `self._diag` along all batch dimensions and return a new `DiagLazyTensor`.

Balandat: yeah I just didn't get to this thus far and so `NotImplementedError` seemed to be the right error to raise :)

Balandat: So I assume the reason `SumBatchLazyTensor` exists at all is because gpytorch uses a single batch to represent a batch of blocks. Is there a particular reason this was done (apart form the lack of support for multi-dimensional batch shapes)? I feel this should be moved to a proper multi-shape representation (e.g. `b x k x n x n` instead of `bk x n x n`)

jacobrgardner: It's how we handle additive structure efficiently in the setting where the kernels we are adding can be coalesced in to a single batch tensor or lazy tensor (see AdditveStructureKernel)

jacobrgardner: Specifically, SumBatchLazyTensor's matmul is far more efficient than SumLazyTensor's

Balandat: Sure that makes sense. I guess what I was asking about is the internal representation as a multi-dim batch (vs. a single-dim batch with batch size `bk`). That shouldn't really affect perf much, but avoid some size arithmetic. Not important at this point though, just a side note

jacobrgardner: Oh, yeah. I think moving forward that `b x k x n x n` makes the most sense.

gpytorch/lazy/diag_lazy_tensor.py
Balandat: moved to a new `.matmul` method, which just calls the superclass's method if `rhs` is not a `DiagLazyTensor`

gpytorch/distributions/multivariate_normal.py
```diff
@@ -228,6 +228,10 @@ def kl_mvn_mvn(p_dist, q_dist):
     root_p_covar = p_covar.root_decomposition()
 
     mean_diffs = q_mean - p_mean
+    if isinstance(root_p_covar, LazyTensor):
```
jacobrgardner: Some of these changes I think are incompatible or unnecessary with the way I changed `root_decomposition` to work in https://github.com/cornellius-gp/gpytorch/pull/353. 

Maybe it would make sense to rebase this branch on top of the `lazy_root_decomp` branch so that instead of `DiagLazyTensor.root_decomposition` returning a `DiagLazyTensor` it just uses the new `root_decomposition` from that PR which accomplishes the same goal?

jacobrgardner:
@Balandat Other than my comment above about rebasing on `lazy_root_decomp` this all looks good to me!

gpytorch/distributions/multivariate_normal.py
Balandat: @jacobrgardner is there anything blocking #353 from going in?

jacobrgardner: @Balandat not in my opinion, no

Balandat: Do you want @gpleiss's review, or would you be happy with mine?  :)

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -3,159 +3,164 @@
...
+                one for each batch dimension. If smaller than the number of batch dimensions of the
+                DiagLazyTensor, select all elements of the omitted batches.
+
+        Returns:
+            An `i`-dim tensor containing the requested elements if all batch index tensors were
+            passed in, otherwise a `bl x ... bk x i` batch of such tensors, where `l = len(batch_indices) + 1`.
+
+        """
+        batch_dims = len(batch_indices)
+        if batch_dims > self._batch_dim:
```
gpleiss: Since `_get_indices` is an internally used method, I don't know if this check is necessary.  Additionally, I think it should always be the case `len(batch_indices) == self._batch_dim`, unless this changed somehow with broadcasting.

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -3,159 +3,164 @@
...
 
-    def _batch_get_indices(self, batch_indices, left_indices, right_indices):
-        equal_indices = left_indices.eq(right_indices).type_as(self._diag)
-        return self._diag[batch_indices, left_indices] * equal_indices
-
-    def _get_indices(self, left_indices, right_indices):
-        equal_indices = left_indices.eq(right_indices).type_as(self._diag)
-        return self._diag[left_indices] * equal_indices
+    def abs(self):
+        return DiagLazyTensor(self._diag.abs())
```
gpleiss: This would also be an interesting method to chuck onto LazyTensors, but I'm not quite sure how yet. Perhaps LazyTensor subclasses should define a method for performing arbitrary element-wise operations. Then the base LazyTensor class can call that subclass method, and say that we want to perform the `abs` operation.

Not for this PR, just thinking out loud :)

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -3,159 +3,164 @@
...
-                Set to True only if the represented diagonal is relatively small.
+            :attr:`diag` (Tensor):
+                A `b1 x ... x bk x n` Tensor, representing a `b1 x ... x bk`-sized batch
+                of `n x n` diagonal matrices
         """
         super(DiagLazyTensor, self).__init__(diag)
         self._diag = diag
-        self.exact_root_decomposition = exact_root_decomposition
+        self._batch_shape = diag.shape[:-1]
+        self._batch_dim = len(self._batch_shape)
```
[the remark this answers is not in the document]
Balandat: I agree, will add that in. 



For instance, I believe that `.shape` of a `DiagLazyTensor` should return the actual shape of the full (batched) tensor.

gpytorch/lazy/diag_lazy_tensor.py
Balandat: yeah this method implements "broadcasting" in the sense described in the docstring: 

> `:attr:batch_indices (LongTensor)`: A variable number of `i`-dim tensors of batch indices, one for each batch dimension. If smaller than the number of batch dimensions of the DiagLazyTensor, select all elements of the omitted batches.

Balandat: Also, even though it's internal, having good error messages makes things much easier to debug...

PR #368, opened 2018-11-13
Variational improvements

gpleiss:
# Improvements to variational inference

## Pyro integration
- MVN.expand now uses BatchRepeatLazyTensor (faster and less hacky)
- Use DiagLazyTensor for independence in the observation sample
- Make Pyro stable for regression (this happens by initializing the variational distribution as the prior)

## Standard variatioanl inference
- Introduce an Empirical ELBO (this isn't actually that useful, but helps to debug Pyro models)
- Initialize the variational distribution as the prior
- Simply/speed up VariationalStrategy (for training)

## Random changes
- Replace `torch.potrf` with `torch.cholesky`


test/priors/test_lkj_prior.py
```diff
@@ -78,10 +78,10 @@ def test_lkj_cholesky_factor_prior_log_prob(self, cuda=False):
         prior = LKJCholeskyFactorPrior(2, torch.tensor(0.5, device=device))
         self.assertFalse(prior.log_transform)
         S = torch.eye(2, device=device)
-        S_chol = torch.potrf(S, upper=False)
+        S_chol = torch.cholesky(S, upper=False)
         self.assertAlmostEqual(prior.log_prob(S_chol).item(), -1.86942, places=4)
         S = torch.stack([S, torch.tensor([[1.0, 0.5], [0.5, 1]], device=S_chol.device)])
-        S_chol = torch.stack([torch.potrf(Si, upper=False) for Si in S])
+        S_chol = torch.stack([torch.cholesky(Si, upper=False) for Si in S])
```
jacobrgardner:
```suggestion
        S_chol = torch.stack([torch.cholesky(Si) for Si in S])
```

test/priors/test_lkj_prior.py
```diff
@@ -100,10 +100,10 @@ def test_lkj_cholesky_factor_prior_batch_log_prob(self, cuda=False):
 
         self.assertFalse(prior.log_transform)
         S = torch.eye(2, device=device)
-        S_chol = torch.potrf(S, upper=False)
+        S_chol = torch.cholesky(S, upper=False)
```
jacobrgardner:
```suggestion
        S_chol = torch.cholesky(S)
```

jacobrgardner:
@gpleiss All good now! Just needed to update the training procedures for the new initialization in some cases.

gpleiss:
Almost ready - need to fix the KL scale issue for pyro models, and update the pyro notebooks. Every other notebook has been re-run tho

@jacobrgardner - for all the notebooks, I bumped down the learning rate and used a random (or evenly spaced) initialization for the inducing points. This makes things much more stable and requires fewer iterations to converge.

gpytorch/mlls/variational_elbo.py
```diff
@@ -23,7 +24,7 @@ def __init__(self, likelihood, model, num_data, combine_terms=True):
         self.num_data = num_data
 
     def forward(self, variational_dist_f, target, **kwargs):
-        num_batch = target.size(0)
+        num_batch = variational_dist_f.event_shape[0]
```
martinjankowiak: @gpleiss is this really correct?

martinjankowiak: hmm in my code, which has multiple GPs batched together, it doesn't seem to return the size of the mini-batch. seems to always be equal to 1. although maybe there's some other issue on my end...

gpleiss: Oh I think I found the issue - I think we have a mean that's not squeezed properly

gpleiss: Just pushed to master, should be fixed.

martinjankowiak: ah thanks. that introduced other errors for me but maybe that's on me...

martinjankowiak: @gpleiss update: fixed my problem, thanks!

gpleiss: Awesome!

Issue #370, opened 2018-11-14
Robust model fitting (RuntimeError when learning rate is too high)


bletham:
Thanks for the continued development on GPyTorch. It's fantastic for constructing models but one area that we've recently been struggling with is robustly fitting models, due to numerical issues during optimization. Even a simple ExactGP model errors out during fitting with a RuntimeError if the learning rate is too large.

This makes model fitting a real challenge: too small of a learning rate and it takes forever, too large and it errors. To get around learning rate tuning, we've been using L-BFGS (both from https://github.com/hjmshi/PyTorch-LBFGS and from Scipy), but even this can fail when the L-BFGS takes too large of a step. We want to run this for a large, heterogeneous collection of problems and so really need robust model fitting that doesn't require hyperparameter optimization.

Attached is a notebook with a simple GP example that fails to fit with default L-BFGS setup, or when the Adam lengthscale is too long. It also includes a plot showing the unfortunate hyperparameter sensitivity of Adam.

I've been in touch with @hjmshi about this. Ideally, the L-BFGS would be given a chance to backtrack from the overly large step by getting a NaN for the function return and leaving the model in a useable state. Is there anything that can be done along these lines to make fitting more robust?

[gpytorch_optim_test5.zip](https://github.com/cornellius-gp/gpytorch/files/2578640/gpytorch_optim_test5.zip)

jacobrgardner:
@Balandat totally. It may actually be easier than catching the errors: most of the errors that pop up happen downstream from getting a NaN in a result from CG or the preconditioner calculation. We could probably simply detect a NaN in one of those places, and return NaN for the overall loss if we do.

Balandat:
You guys probably know best where to do these checks and how to do them efficiently - do you mind putting up a PR for this? This would be super helpful.

hjmshi:
@Balandat Yes, this should be sufficient for L-BFGS to know to backtrack.

bletham:
@jacobrgardner this is great, thanks a bunch! I'll let you know if I can break it again :-)

PR #371, opened 2018-11-14
Add positive_nonlinearity option to likelihoods and kernels

jacobrgardner:
This PR adds support for using transformations other than `torch.exp` to ensure positiveness of hyperparameters by passing in a different nonlinearity to most built in modules that have parameters that must be positive. For example:

```python
softplus = torch.functional.F.softplus

        self.covar_module = gpytorch.kernels.ScaleKernel(
            gpytorch.kernels.RBFKernel(positive_nonlinearity=softplus),
            positive_nonlinearity=softplus
        )
    
    def forward(self, x):
        mean_x = self.mean_module(x)

likelihood = GaussianLikelihood(positive_nonlinearity=sofplus)
model = ExactGPModel(train_x, train_y, likelihood)
```

It is likely in the future that `torch.functional.F.softplus` will become the default choice in the near future because it circumvents issues that arise when large steps during optimization lead to large parameter value changes.


jacobrgardner:
@bletham Using softplus instead of exp partially solves #370 in the sense that Adam in your notebook now works up to learning rates of 6. 

@samuelstanton Using softplus may actually make training substantially more stable for your application as well, since you deal with particularly extreme hyperparameter values.

jacobrgardner:
I've been playing around with using `softplus` instead of `exp` in some of our VI notebooks where we've previously had some trouble with stability issues, and the results are pretty amazing actually. 

In the SVDKL regression examples, replacing `exp` with `softplus` and no other change results in a 10% relative decrease in mean absolute error, and makes convergence much less sensitive to the exact learning rate chosen.

Let's give it some time to play around with the option and make sure we understand all of the ramifications, but I'm definitely in favor of sunsetting use of `exp` as the default option here ASAP.

jacobrgardner:
Regarding dropping `log`: ideally yes, we would drop these. There are concerns about this being breaking, however (e.g. saved models would no longer load properly). 

We could do that change now; however my preference is to change the names once we are confident in using softplus as the default option so that all of the breaking changes for that happen at once.

Balandat:
Seems reasonable. What about the prior kwargs? Those shouldn't affect model loading if the priors themselves are still stored with the `log_` prefix. But I guess that's not really worth the effort...

jacobrgardner:
I think ideally we rename `log_*_prior` to `*_prior`, and the `transform` (we can change the name to this, agreed) gets passed along to the prior.

gpleiss:
@Balandat @jacobrgardner - since this is sort of a beta feature, I'm okay with keeping the `log_*` names for right now. We can add another issue/PR for doing a soft deprecation of the names.

gpleiss:
I'm okay with merging this if @Balandat is!

PR #372, opened 2018-11-14
Replace batch_potrs with native batched potrs

vishwakftw:
Just as specified in the title.

### DIFF
--- a/gpytorch/utils/cholesky.py
+++ b/gpytorch/utils/cholesky.py
@@ -16,19 +16,6 @@ def batch_potrf(mat):
     return res.view_as(mat)
 
 
-def batch_potrs(mat, chol):
-    """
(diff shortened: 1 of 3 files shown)


jacobrgardner:
Sweet thanks! I think we were just waiting over the weekend for this to make it from `master` in to the actual pytorch-nightly build :-)

vishwakftw:
Batch Cholesky Decomposition is merged in PyTorch master. I’ll send in a PR tomorrow to replace usage of batch_potrf with torch.cholesky (the nightly builds should be ready by then).

jacobrgardner:
Awesome!

@gpleiss we can probably finally investigate the impact that preconditioning will have on DKL models since it shouldn't be terribly expensive anymore once this lands. Although it may still be too slow until the advanced indexing fix lands too. We'll see!

PR #373, opened 2018-11-14
Handle NaN gracefully in loss computation.

### LINKED ISSUE #370: Robust model fitting (RuntimeError when learning rate is too high)

jacobrgardner:
This PR makes the GPyTorch internals handle NaNs that occur due to round-off from terrible hyperparameter values more gracefully. In particular:

- If the preconditioner contains NaNs, we throw a warning and attempt to continue without preconditioning. 
    - In testing these NaNs appear somewhat intermittently such that even if one iteration fails due to it the next will often succeed. In @bletham's notebook this avoids crashing with learning rates well up in to the thousands (when combined with `param_transform=softplus` of course).
- If NaNs are encountered from CG outputs, these are handled more gracefully. 
    - The linear solves already handle these gracefully in the sense that we don't try to do anything more than basic operations, so the inv quad term will already show up as NaN. All that was needed was a check for the tridiagonal matrices used for log det calculations.

In my opinion, this closes #370 in combination with #371.

### DIFF
--- a/gpytorch/functions/_inv_quad_log_det.py
+++ b/gpytorch/functions/_inv_quad_log_det.py
@@ -123,15 +123,18 @@ class InvQuadLogDet(Function):
 
         # Compute log_det from tridiagonalization
         if self.log_det:
-            if self.batch_shape is None:
-                t_mat = t_mat.unsqueeze(1)
-            eigenvalues, eigenvectors = lanczos_tridiag_to_diag(t_mat)
-            slq = StochasticLQ()
-            log_det_term, = slq.evaluate(self.matrix_shape, eigenvalues, eigenvectors, [lambda x: x.log()])
+            if torch.any(torch.isnan(t_mat)).item():
+                log_det_term = float("nan")
+            else:
+                if self.batch_shape is None:
+                    t_mat = t_mat.unsqueeze(1)
+                eigenvalues, eigenvectors = lanczos_tridiag_to_diag(t_mat)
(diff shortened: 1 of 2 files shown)


Balandat:
great, thanks!

Issue #374, opened 2018-11-14
Allow priors to take a param_transform


Balandat:
#371 added support for passing a `param_transform` into many modules to swap out the default exponential transform (i.e. internal parameters in log scale). 

The priors currently still only support a `log_scale` boolean flag, so there is no easy way to put an existing prior on the transformed parameter (without implementing a new prior). We should change this and support passing in a `param_transform` to the prior constructor as well.

Balandat:
This is a good opportunity to also tackle #181 and #339

Balandat:
In line with #181, instead of passing the transform into the prior, we could instead just use what's currently a "derived prior" and register the transform there. The only restriction of that would be that we couldn't have different transforms for parameters in the evaluation vs. the prior. I guess while that's not a typical use case, it would allow to do things like simply put a log-normal prior on a parameter that internally is parameterized using the (inverse) softplus function.

One way to allow for that would be to handle transforms in two places:
1. Always feed the transformed parameter into the prior by doing what #181 suggests, using the construct of the derived_prior (employing `param_transform`)
2. Allow priors themselves to take an optional transform arg, which could be used to transform whatever input it gets. 

That way, if you wanted to put a log-normal prior on the actual lengthscales (which could be transformed via softplus internally), you could use `log` as the transform in a NormalPrior.

@jacobrgardner, @gpleiss any thoughts on this?

gpleiss:
I think I'm fine with either. If I'm understanding what you're suggesting, 2 could be an optional syntactic sugar for 1, right?

Balandat:
Hmm so working on this it seems it would make sense internally to associate the transform with the parameter itself, not the module the parameter is registered with. The constructors for the user-facing api could remain the same, but the transform would be registered with the parameter, similar to what's currently done for the derived prior transforms. 

E.g. for kernels that would look like the following:
```
self.register_parameter(
    name="log_lengthscale",
    parameter=torch.nn.Parameter(torch.zeros(batch_size, 1, lengthscale_num_dims)),
    param_transform=param_transform,
    prior=log_lengthscale_prior,
)

@property
def lengthscale(self):
    if self.has_lengthscale:
        return self._transform_param("log_lengthscale", self.log_lengthscale).clamp(self.eps, 1e5)
    else:
        return None
```

where `_transform_param` is defined on the module level and returns the previously registered transform of the param, or just the param in case there is none registered. 

[...]

Issue #181, opened 2018-07-19
Do not differentiate between parameter prior and derived prior


Balandat:
> As far as I can tell, a derived_prior is a strict generalization of a param_prior. Rather than differentiating between the two, I think that we should just have derived_priors, and when you call register_parameter with a prior option, we create a derived prior where params just contains the newly registered parameter, and transform is just an identity function.

Balandat:
Done in #376

PR #376, opened 2018-11-14
Revamp priors

Balandat:
As discussed in #374.

This 
1. registers parameter transforms with the parameters
2. makes sure priors are always evaluated on the transformed parameters
3. unifies `parameter_prior` and `derived_prior` into simply `prior` (addressing #181)
4. allows general transforms also for the priors. For instance, by passing in `torch.log` to the `transform` constructor of the `NormalPrior`, you get a log-normal prior.


Balandat:
This causes some `AttributeError`s on `@property`-decorated functions that I don't really understand. Maybe I'm missing something obvious here?

gpytorch/kernels/scale_kernel.py
```diff
@@ -49,15 +49,18 @@ class ScaleKernel(Kernel):
...
+            name="log_outputscale",
+            parameter=torch.nn.Parameter(torch.zeros(batch_size)),
+            prior=log_outputscale_prior,
+            transform=param_transform,
         )
 
     @property
     def outputscale(self):
-        return self.param_transform(self.log_outputscale)
+        return self.transfom_parameter("log_outputscale", self.log_outputscale)
```
Balandat: This fails with an `AttributeError` even when you do something trivial like the following (I have no idea why): 

```
from gpytorch.kernels import ScaleKernel, RBFKernel

sk = ScaleKernel(RBFKernel())
sk.outputscale
``

Comment (Balandat) [reply gpytorch/kernels/scale_kernel.py]: to clarify: The `AttributeError` is w.r.t to `outputscale`, so it doesn't even enter this function here.

Comment (Balandat) [reply gpytorch/kernels/scale_kernel.py]: lol this is totally silly, you can see the typo inside the @property definition above: `transfom_parameter`. The error message in this case is just totally bogus, why doesn't it throw the AttributeError for the correct issue?

Comment (Balandat) [reply gpytorch/kernels/scale_kernel.py]: also, having the `Property` github handle must be the worst thing (or any name that is a common decorator for that matter)

**Comment by Balandat** [comment]
So apart from fixing the docs failure this is in good shape and ready for review.

(Some of the highlighted apparent non-changes were due to making sure to use unix instead of windows line breaks)

File: gpytorch/priors/prior.py [hunk 2]
```diff
@@ -15,10 +15,9 @@ class Prior(Distribution, Module, ABC):
     respective prior in computing the Marginal Log-Likelihood.
     """
 
-    @property
-    def log_transform(self):
-        return self._log_transform
+    def transform(self, parameter):
+        return self._transform(parameter) if self._transform is not None else parameter
 
     def log_prob(self, parameter):
```
gpleiss: Nit pick: could we change the argument name from `parameter`, since priors now operate on transformed parameters?

gpytorch/module.py
```diff
@@ -106,68 +101,122 @@ def named_added_loss_terms(self):
...
             raise AttributeError("Cannot assign parameter before Module.__init__() call")
         super(Module, self).register_parameter(name, parameter)
         if prior is not None:
-            self.set_parameter_priors(**{name: prior})
+            self.register_prior("_".join([name, "prior"]), prior, name)
+        if transform is not None:
+            self.register_parameter_transform(name, transform)
 
-    def register_derived_prior(self, name, prior, parameter_names, transform):
+    def register_parameter_transform(self, parameter_name, transform):
```
gpleiss: Is this something that we have to register?

gpleiss: For example - for `lengthscale` - if we register a prior on the `lengthscale` attribute, then calling `getattr(self, lengthscale)` will return the already-transformed lengthscale. So the module wouldn't need to know about the parameter transform.

gpytorch/module.py
```diff
@@ -106,68 +101,122 @@ def named_added_loss_terms(self):
...
+        if parameter_name not in self._parameters:
+            raise AttributeError(
+                "Unknown parameter {name} for {module}".format(name=parameter_name, module=self.__class__.__name__)
+            )
+        self._parameter_transforms[parameter_name] = transform
+
+    def register_prior(self, name, prior, *parameter_names, transform=None):
```
gpleiss: Here - what if `register_prior` took in `name`, `prior`, and then `parameter`, which would either be a string (a name for an attribute) or a callable (some function to produce the parameters)?

For example:

- If you wanted to register a prior on `log_lengthscale` (in the log space) you could do `register_prior('lengthscale', SmoothedBoxPrior(-3, 3), "log_lengthscale")`.
- If you wanted to register a prior on `lengthscale` (not in log space), you could do `register_prior('lengthscale', SmoothedBoxPrior(0.001, 1000), "lengthscale")` - since `lengthscale` is a property on the kernel.
- You could also do `register_prior('lengthscale', SmoothedBoxPrior(0.001, 1000), lambda: return self.log_lengthscale().exp())`.

I think this would cover all the possible cases, unless I'm missing something?

gpytorch/kernels/cosine_kernel.py
```diff
@@ -57,19 +57,18 @@ class CosineKernel(Kernel):
...
-        super(CosineKernel, self).__init__(
-            has_lengthscale=False, active_dims=active_dims, param_transform=param_transform
-        )
+        super(CosineKernel, self).__init__(has_lengthscale=False, active_dims=active_dims)
         self.eps = eps
         self.register_parameter(
             name="log_period_length",
             parameter=torch.nn.Parameter(torch.zeros(batch_size, 1, 1)),
             prior=log_period_length_prior,
+            transform=param_transform,
```
gpleiss: I guess I'm not a huge fan of putting the transform on the `register_parameter` call, because now our GPyTorch params are getting further removed from PyTorch params.

It's also a little weird with the naming. For example, `self.log_period_length` returns the log of the period length, even though it knows about the parameter transform.

gpytorch/likelihoods/multitask_gaussian_likelihood.py
```diff
@@ -31,31 +31,34 @@ def __init__(
...
             if task_prior is not None:
-                self.register_derived_prior(
-                    name="MultitaskErrorCovariancePrior",
-                    prior=task_prior,
-                    parameter_names=("task_noise_covar_factor", "log_noise"),
+                self.register_prior(
+                    "MultitaskErrorCovariancePrior",
+                    task_prior,
+                    "task_noise_covar_factor",
+                    "log_noise",
```
gpleiss: With the new suggested interface, I think you could do `register_prior("MultitaskErrorCovariancePrior", task_prior, "_eval_covar_matrix")`.

`_eval_covar_matrix` would probably be re-written so that it takes no arguments, which should be fine since `task_noise_covar_factor` and `log_noise` are parameters that belong to the likelihood anyways.

gpytorch/mlls/variational_elbo.py
```diff
@@ -1,94 +1,93 @@
...
+                params = [p if ptf is None else ptf(p) for p, ptf in params_and_tfs]
+                lp_arg = params[0] if prior_tf is None else prior_tf(*params)
+                res.add_(prior.log_prob(lp_arg).sum().div(self.num_data))
+            return res
+        else:
+            log_prior = torch.zeros_like(log_likelihood)
+            for _, prior, params_and_tfs, prior_tf in self.named_priors():
+                params = [p if ptf is None else ptf(p) for p, ptf in params_and_tfs]
+                lp_arg = params[0] if prior_tf is None else prior_tf(*params)
+                log_prior.add_(prior.log_prob(lp_arg).sum())
```
gpleiss: Here, with the new proposed interface, you'd get `lp_arg = param() if callable(param) else getattr(module, param)`.

gpytorch/mlls/variational_elbo.py
```diff
@@ -1,94 +1,93 @@
...
+        num_samples = settings.num_likelihood_samples.value()
+        variational_samples = variational_dist_u.rsample(torch.Size([num_samples]))
+        kl_divergence = (
+            variational_dist_u.log_prob(variational_samples) - prior_dist.log_prob(variational_samples)
+        ).mean(0)
+        kl_divergence = kl_divergence.div(self.num_data)
+
+        res = log_likelihood - kl_divergence
+        for _, prior, params, tf in self.named_priors():
+            lp_arg = params[0] if tf is None else tf(*params)
```
gpleiss: This bit of code is different than the one for VariationalELBO

gpytorch/priors/prior.py
```diff
@@ -15,10 +15,9 @@ class Prior(Distribution, Module, ABC):
     respective prior in computing the Marginal Log-Likelihood.
     """
 
-    @property
-    def log_transform(self):
-        return self._log_transform
+    def transform(self, parameter):
+        return self._transform(parameter) if self._transform is not None else parameter
```
gpleiss: Priors no longer need a `transform` function, right? Since the transform is being handled when you register a prior.

gpytorch/kernels/kernel.py
```diff
@@ -102,14 +102,14 @@ def __init__(
...
         self.batch_size = batch_size
         self.__has_lengthscale = has_lengthscale
-        self.param_transform = param_transform
         if has_lengthscale:
             self.eps = eps
             lengthscale_num_dims = 1 if ard_num_dims is None else ard_num_dims
             self.register_parameter(
                 name="log_lengthscale",
                 parameter=torch.nn.Parameter(torch.zeros(batch_size, 1, lengthscale_num_dims)),
                 prior=log_lengthscale_prior,
```
gpleiss: As mentioned in a previous PR, we'll probably want to soft-deprecate `log_lengthscale_prior` in favor of `lengthscale_prior`.

gpytorch/priors/prior.py
Balandat: sure I can use sth neutral like `tensor`

gpytorch/module.py
Balandat: Hmm yeah that sounds like a reasonable approach. let me think through this a little more but I think I can make that change

gpytorch/kernels/kernel.py
Balandat: yeah happy to do this as part of this PR

gpytorch/kernels/cosine_kernel.py
Balandat: I can see why we wouldn't want to make parameters different from torch ones

gpytorch/mlls/variational_elbo.py
Balandat: so this should implement the same functionality as before - are you saying there is something weird unrelated to this PR?

gpytorch/priors/prior.py
Balandat: So the thinking is that this transform can be independent of the parameter transform. The idea is that using this you can easily instantiate, say, a log-normal prior, independent of registering it anywhere. This should help separate concerns and compartmentalize the design. If you think this is an unnecessary complication this could be removed, then in order to do this the transform would have to be included in the transform registered with this prior (and you couldn't have a standalone log-normal `Prior` object without implementing a new class)

jacobrgardner:
With regards to how we could deprecate `log_` from the parameter names, we could just make properties, e.g.
```python
@property
def log_lengthscale(self):
    warning('...deprecated...')
    return self._param_transform(self.lengthscale).log()
```
If we could easily access the inverse transform, we could also define a setter for `log_lengthscale` that exponentiates the input and then applies the inverse transform.

gpytorch/kernels/index_kernel.py
```diff
@@ -50,24 +50,25 @@ def __init__(self, num_tasks, rank=1, batch_size=1, prior=None, param_transform=
...
-        var = self.param_transform(log_var)
-        return covar_factor.matmul(covar_factor.transpose(-1, -2)) + var.diag()
+    @property
+    def var(self):
+        return self._param_transform(self.log_var)
+
+    def _eval_covar_matrix(self):
+        covar_factor, variance = self.covar_factor, self.var
+        return covar_factor.matmul(covar_factor.transpose(-1, -2)) + variance.diag()
+        # D = var * torch.eye(var.shape[-1], dtype=var.dytpe, device=var.device)
```
jacobrgardner:
```suggestion
```

gpytorch/kernels/index_kernel.py
```diff
@@ -50,24 +50,25 @@ def __init__(self, num_tasks, rank=1, batch_size=1, prior=None, param_transform=
...
+        # return self.covar_factor.matmul(self.covar_factor.transpose(-1, -2)) + D
```
jacobrgardner:
```suggestion
```

Issue #378, opened 2018-11-15
Add unit tests for batch shapes and broadcasting


jacobrgardner:
(Related to #291, #361, etc, builds on work in #356 )

Our unit test coverage of basic batch mode functionality could be substantially improved. We should add basic shape unit tests for the following:

- [x] All kernels should accept `n x d` inputs and return `n x n` covariance matrices (or another appropriate shape that is kernel specific).
- [x] All kernels should accept `b x n x d` inputs and return `b x n x n` covariance matrices (or another appropriate shape).
- [ ] All models extending `gpytorch.models.GP` should be callable on  `n x d` inputs and return appropriately sized MultivariateNormal distributions (`(mu, Sigma) = (n, n x n)`).
- [ ] All models extending `gpytorch.models.GP` should be callable on  `b x n x d` inputs and return appropriately sized MultivariateNormal distributions (`(mu, Sigma) = (b x n, b x n x n)`).
- [ ] All models extending `gpytorch.models.GP` initialized with `n x d` training data (or inducing points) should be (repeatedly) callable on both `m x d` inputs and `b x m x d` inputs, returning MVN distributions with the appropriate sizes.
- [x] All `LazyTensors` should broadcast `matmul` calls in all situations that normal `torch.Tensors` do. This may require more work on the general goal of "`LazyTensors` support arbitrary batch shapes that @Balandat has been working on.

gpleiss:
I can tackle this!

Balandat:
Note also the related issue #384 (though they're sufficiently separate that I can take care of that)

PR #379, opened 2018-11-15
Simplify batch handling, fix batch -> non_batch evaluation and backward calls

jacobrgardner:
Fixes an issue @sdaulton raised in #340  where testing first in batch mode and then in non batch mode would fail. Also fixes an issue raised by @Balandat in #342 where running a backward pass after switching modes would fail. 

This was caused by the mean cache being a different size (`n` vs `n x 1`) depending on whether the model was first called on batch test data or non-batch test data. This PR simplifies things by just always having the mean cache not have the trailing `1` in the size to match how we handle means and labels.

I will codify this (and all related batch vs non-batch settings) working in unit tests by the end of next week when I tackle #378.


Balandat:
lgtm, thanks for the fix

Issue #385, opened 2018-11-18
Drop py2 support


Balandat:
It looks like we haven't really been supporting py2 for a little while now, see https://github.com/cornellius-gp/gpytorch/pull/383#pullrequestreview-176066680

Let's formally drop support for it, put that into the readme, and update the code accordingly (get rid of `from future ...` imports, add `#!/usr/bin/env python3` to all files).

More controversial: If we feel ambitious and want to require >3.6, we could also start typing things....

jacobrgardner:
I think if we are going to drop py2, then I am certainly also in favor of requiring py36 or higher since I'm not sure you can even install py34 anymore.

Balandat:
Yeah I guess the ambitious referred to starting to type things :). We can do that organically step-by-step in new PRs though, no need to rush this.

Balandat:
I guess the readme already stated py>=3.6 as a requirement...

gpleiss:
CLosed by #387

PR #387, opened 2018-11-18
Cleanup code to be py3 only

Balandat:
The readme already stated py >= 3.6 as a requriment, this is just some cleanup of files so the linter stops complaining about py2 stuff.

Addresses #385


Balandat:
hey I got some practice in directory tree walking and string replacement :)

PR #388, opened 2018-11-19
Replace batch_potrf with native batched cholesky

vishwakftw:
Just as specified in the title.

Had to wait for so long for the new binaries to be uploaded on conda.


gpleiss:
Thanks @vishwakftw!

PR #393, opened 2018-11-25
Pyro can handle batch models

gpleiss:
There is a better fix for batch VGP models with Pyro, but this is a simple solution for right now.

All batch Gaussian distributions touched by Pyro (`p_u`, `q_u`, `q_f`) are all converted into non-batch variational distributions using a BlockDiagLazyTensor. This has no effect on the `log_prob` or `kl_divergence` terms in the equations.

### DIFF
--- a/gpytorch/variational/cholesky_variational_distribution.py
+++ b/gpytorch/variational/cholesky_variational_distribution.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 import torch
-from ..lazy import CholLazyTensor
+from ..lazy import CholLazyTensor, DiagLazyTensor, NonLazyTensor
 from ..distributions import MultivariateNormal
 from .variational_distribution import VariationalDistribution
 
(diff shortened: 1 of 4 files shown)


gpytorch/variational/cholesky_variational_distribution.py
```diff
@@ -47,33 +47,16 @@ def variational_distribution(self):
         that the diagonal remains positive.
         """
         chol_variational_covar = self.chol_variational_covar
+        dtype = chol_variational_covar.dtype
+        device = chol_variational_covar.device
+        mask = torch.ones(self.chol_variational_covar.shape[-2:], dtype=dtype, device=device).triu()
+        chol_variational_covar = chol_variational_covar.mul(mask)
```
jacobrgardner: I assume that this happened because `triu()` doesn’t operate in batch mode?

gpleiss: Yeah

Issue #369, opened 2018-11-13
Make all LazyTensors work with any batch size


Balandat:
#362 extended `DiagLazyTensor` to work with general batch sizes. We need to do the same for the KroneckerProductLazyTensor. This is required to make the new Heteroskedastic Likelihoods work correctly in the multi-batch multi-task case. 

Currently, the behavior is
```
T = NonLazyTensor(torch.rand(2, 3, 4))
KLT = KroneckerProductLazyTensor(T, T)
KLT.shape, KLT.evaluate().shape
# torch.Size([2, 9, 16]), torch.Size([2, 9, 16]) <- great
```

```
T = NonLazyTensor(torch.rand(1, 2, 3, 4))
KLT = KroneckerProductLazyTensor(T, T)
KLT.shape, KLT.evaluate().shape
# <- (torch.Size([9, 16]), torch.Size([9, 64])) <- not great
```

Balandat:
@jacobrgardner, @gpleiss, how do you feel about broadcasting the batch sizes here? That is, you could do `KroneckerProductLazyTensor(T1, T2)` for Tensors `T1` and `T2` as long as `T1.shape[:-2]` and `T2.shape[:-2]` are broadcastable. This seems like reasonable behavior.

gpleiss:
That seems super reasonable!

gpleiss:
I'm going to make this a more general issue, about all LazyTensors supporting arbitrary batch dimensions. Expect a flurry of small PRs about this :)

PR #394, opened 2018-11-25
Remove _batch_get_indices function from LazyTensors

gpleiss:
`_get_indices` now (in theory) supports an arbitrary number of batch dimensions. This PR uses `_get_indices` everywhere now, and remove the `_batch_get_indices` method from LazyTensors.

In practice, most LazyTensors only support 0 or 1 batch dimensions. However, this sets the infrastructure up for future LazyTensors that use multiple batch dimensions.

A step towards achieving #369

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1296,7 +1296,6 @@ class LazyTensor(object):
             return new_lazy_tensor
 
         # Special case: if both row and col are tensor indexed, then we use _get_indices
-        # or _get_indices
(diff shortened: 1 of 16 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -1296,7 +1296,7 @@ def __getitem__(self, index):
             return new_lazy_tensor
 
         # Special case: if both row and col are tensor indexed, then we use _get_indices
-        # or _batch_get_indices
+        # or _get_indices
```
jacobrgardner:
```suggestion
```
`_get_indices` already referenced on the previous line comment

Issue #398, opened 2018-11-26
Size mismatch in batch mode GaussianLikelihood


jacobrgardner:
@Balandat just ran in to an issue that I tracked back to this line:
https://github.com/cornellius-gp/gpytorch/blob/00ddfc844820c7682a50919fcdb6eaf4e9c50ab3/gpytorch/likelihoods/gaussian_likelihood.py#L31

Should this be:
```python
shape = mean.shape
```
Haven't we been letting batch means be `b x n`, meaning shape would be `[b]` and then as written `noise_covar` will be `b x b x b` instead of `b x n x n`?

jacobrgardner:
Hm... I don't think `shape = mean.shape` is exactly right either. Maybe I can try to come up with a minimal example of the failure I saw

jacobrgardner:
I see what you're saying. Yeah, I think checking the type of the input makes sense or maybe a shape property. Either way I think it's broken on master for batch mode non Multitask GPs

jacobrgardner:
As an addendum, I'm surprised the batch mode unit test passed...

jacobrgardner:
This is fixed via #400

PR #400, opened 2018-11-26
Fix shaping of noise_covar in GaussianLikelihood

jacobrgardner:
My attempt at addressing #398, plus adding a check in AddedDiagLazyTensor that will fail if this ever happens again.

I think the issue is multi-task vs not: in Multitask the mean is `b x n x t` but in standard GP regression the mean is `b x n`.

cc/ @Balandat if you have a better solution, feel free to close this or add on to it.


Balandat:
I'm ok with this - I thought I had a specific reason for doing this, but I just ran through my test cases and did not run into any issues, so maybe it was an attempt at getting the multi-task GPs to work with single batch fit / multi-batch eval or vice versa, which ultimately I did not make part of my earlier PR.

PR #403, opened 2018-11-27
Fixed uses of log_transform

Chillee:
This PR broke a lot of the examples, including DKL: https://github.com/cornellius-gp/gpytorch/pull/376

This PR fixes the issue in the same way that was done in that PR, as in https://github.com/cornellius-gp/gpytorch/pull/376/files#diff-d7c2cf3e486598ffaf365fcf5b61a2b0L44


jacobrgardner:
Thanks for catching this!

PR #404, opened 2018-11-29
Improve LazyTensor.__getitem__

gpleiss:
A more stable version of yesterday's hotfix.

`LazyTensor.__getitem__` now calls a `LazyTensor._getitem` helper (similar to how `LazyTensor.size` calls `LazyTensor._size`). When `LazyTensor._getitem` is called, there is:
- One index per dimension (unsupplied indices are filled with `slice(None, None, None)`.
- Each index is an int, a slice, or a LongTensor (all lists, Ellipsis have been processed out).

LTs that had custom `__getitem__`s now have custom `_getitem`s. This is a pure refactor - no new functionality (except for `ConstantMulLazyTensor`, which is now a bit more stable). This significantly reduces repeated code.


gpleiss:
@jacobrgardner this is now ready

PR #407, opened 2018-11-30
Option to skip LazyTensor.logdet forward pass

### LINKED ISSUE #396: Flag to turn off logdet computations for MLLs
For exact and variational GPs, the marginal log likelihood requires computing the log determinant of the prior covariance matrix. This isn't a super expensive operation (since we get most of the terms for free from a single call to `linear_cg`). However, we still have to make a few calls to `torch.symeig`, which can add a bit of a performance hit.

If we are simply using the MLL as a loss function to optimize hyperparameters (and not for Gibbs sampling, model selection, etc.), then we don't actually care about the logdet value (just its derivatives, which we get for free from `linear_cg`). Therefore, we could avoid these `torch.symeig` calls.

I propose a `gpytorch.beta_features.skip_logdet` feature flag that skips the logdet computation (but still computes the correct gradient, obviously). It will be off by default.

gpleiss:
This is going to be a not-highly-advertised feature.

This PR introduces `gpytorch.settings.skip_logdet_forward`. Setting this to True makes `LazyTensor.logdet` return a zero tensor, but one that does the correct thing on the backward pass.

This allows advanced users (like us 😁) to get much faster performance for hyperparameter optimization. If we're using a single-order method - we don't actually care about the logdet value - we just care about its derivatives. This settings will allow us to skip unnecessary forward-pass computation.

Fixes #396

### DIFF
--- a/gpytorch/functions/_inv_quad_log_det.py
+++ b/gpytorch/functions/_inv_quad_log_det.py
@@ -95,7 +95,7 @@ class InvQuadLogDet(Function):
         # Perform solves (for inv_quad) and tridiagonalization (for estimating log_det)
         rhs = torch.cat(rhs_list, -1)
         t_mat = None
-        if self.log_det:
+        if self.log_det and not settings.skip_logdet_forward.on():
             solves, t_mat = linear_cg(
                 matmul_closure,
                 rhs,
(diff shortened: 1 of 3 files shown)


jacobrgardner:
Looks good. Something to keep in mind: it's easy enough to still print the actual loss e.g. every 5 minibatches instead of every 1 with this flag. Then you still get accurate progress estimates for optimization but effectively still cut the log det time cost by 5x.

PR #408, opened 2018-11-30
Fix issues with root decomposition when using base_samples in MultivariateNormal

Balandat:
Previously this would error with incompatible dimensions for matmul. Fixed by using root_decompositon.matmul instead of root_decomposition.root.matmul.

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python3
 
-import torch
 import math
+
+import torch
 from torch.distributions import MultivariateNormal as TMultivariateNormal
-from torch.distributions.utils import _standard_normal, lazy_property
 from torch.distributions.kl import register_kl
+from torch.distributions.utils import _standard_normal, lazy_property
 
 from ..lazy import LazyTensor, NonLazyTensor
 from .distribution import Distribution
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Doesn't `root_decomposition` now return a `LazyTensor` that represents `LL'`, so this change would make the sampling call do `\Sigma*v + mu` instead of `Lv + mu`?

Balandat:
Hmm let me take a closer look at that. Either way, this would previously error out when using the `.root` attribute of the root decomposition so something is off.

Balandat:
Hmm you're right. The issue I saw was that I passed in `torch.eye(n)` as the covariance matrix, and not `DiagLazyTensor(torch.ones(n))`. As a result the root decomposition gets messed up occasionally and returns nans, which then results in a dimension mismatch error down the road. Red herring.

Either way, we need to fix these instabilities with root_decomposition, it can't be that taking the root of the identity matrix silently errors with NaNs...

jacobrgardner:
Well, explicitly constructing a diagonal matrix and sticking it in a NonLazyTensor and then calling `root_decomposition` is a pretty extreme edge case. 

We can look in to detecting these edge cases and having a fall-back plan. However, since that fall back plan would probably look like calling `torch.cholesky` on an `n x n` diagonal matrix, that would arguably be just as foolish.

Balandat:
How often in the code do we call root decompositions on NonLazyTensors? If that's an edge case in itself, or if it only happens for tiny matrices, then we could just overwrite root_decomposition to call cholesky in NonLazyTensor

PR #411, opened 2018-12-03
Module.initialize can now initialize real parameter values

### LINKED ISSUE #405: Module.initialize should work with transformed parameters
`Module.initialize` right now only works for actual parameters. It should also work for transformed parameters.

For example, the following lines work:

```python
likelihood = gpytorch.likelihoods.GaussianLikelihood(param_transform=torch.exp)
likelihood.initialize(raw_noise=math.log(5))
print(likelihood.noise.item())  #5.
```

Similarly:
```python
likelihood = gpytorch.likelihoods.GaussianLikelihood(param_transform=torch.exp)
likelihood.initialize(raw_noise=torch.tensor(math.log(5)))
print(likelihood.noise.item())  #5.
```

The following line silently errors:
```python
likelihood = gpytorch.likelihoods.GaussianLikelihood(param_transform=torch.exp)
likelihood.initialize(noise=5)
print(likelihood.noise.item())  #1.
```

jacobrgardner:
Fixes #405 

You can now initialize transformed model parameters with their actual values. For example:
```python
kernel = RBFKernel()
kernel.initialize(lengthscale=3.14)
print(kernel.lengthscale)  # 3.14
```
More generally, in addition to parameters, `Module.initialize` now works with any keyword argument specifying a named `@property` with a defined setter.

cc/ @gpleiss

### DIFF
--- a/gpytorch/kernels/cosine_kernel.py
+++ b/gpytorch/kernels/cosine_kernel.py
@@ -91,7 +91,9 @@ class CosineKernel(Kernel):
         return self._set_period_length(value)
 
     def _set_period_length(self, value):
-        self.initialize(log_period_length=self._inv_param_transform(value))
+        if not torch.is_tensor(value):
+            value = torch.tensor(value)
+        self.initialize(raw_period_length=self._inv_param_transform(value))
 
     def forward(self, x1, x2, **params):
         x1_ = x1.div(self.period_length)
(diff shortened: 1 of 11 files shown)


jacobrgardner:
Module.initialize is also now more lenient when tensor arguments are supplied. Before failing, it will try to use both `expand_as` and `view_as` to make the input tensor match the size of the tensor being copied in to.

Balandat:
lgtm!

PR #414, opened 2018-12-05
Catch an edge case with batch-> non-batch eval

Balandat:
This fixes an edge case in which the sequence of operations `batch eval -> non-batch eval` on a non-batch model failed (the reverse sequence succeeded).

This special casing is starting to get a little hairy... Not sure if there is a good way of simplifying this. Maybe one way would be to factor out all these check into some utility that takes the `precomputed_cache` and does the appropriate things to it, so the rest of the code can remain relatively clean? 

cc @sdaulton, @bkarrer


gpleiss:
We are going to change how we do cache stuff when we add the fantasy point stuff, so that might be a good time to take a look at how we could do this better. But for now this works!

PR #415, opened 2018-12-06
Simple decorator for caching instance methods w/o args

Balandat:
Useful for things like `evaluate()` etc.

See discussion here: https://github.com/cornellius-gp/gpytorch/commit/7570ee1d758c93dd4054009fc12be6cc31b54031#r31576790

We can extend this to other cases where want to do caching, we just need to be careful about not messing with autograd by doing that.

### DIFF
--- a/gpytorch/lazy/constant_mul_lazy_tensor.py
+++ b/gpytorch/lazy/constant_mul_lazy_tensor.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
 import torch
+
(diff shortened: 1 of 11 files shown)


gpleiss:
@Balandat could you add this memoize to MatmulLazyTensor as well?

Balandat:
Sure, I can add this to a number of places. I can also make one that supports arguments that can be used in other contexts, that would be a separate PR though

Issue #422, opened 2018-12-10
root_decomposition bug, perf


Balandat:
Currently, `root_decomposition` fails for LazyTensors with a single element with an obscure index error (except for DiagLazyTensor, where it is special-cased). More generally, the iterative low-rank approximation algorithm is quite wasteful to use for small matrix sizes - performing a Cholesky decomposition on the evaluated LazyTensor can be much faster (see this nb: [root_decomposition_perf.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/2664250/root_decomposition_perf.ipynb.txt))

We should dig some more into this and do a more thorough analysis. I suggest once we have done this we can just choose whether to perform a cholesky decomposition or instead use the iterative algorithm depending on the size of the tensor. 

In fact, for now I will push a small fix that does this for all cases with `n * n < 250` or so.

cc @bkarrer

gpleiss:
@Balandat this seems very reasonable. Rather than `numel`, I would do `self.matrix_shape.numel()` as the check, which would then use Cholesky on large batches of small matrices.

jacobrgardner:
Closed by #423

PR #423, opened 2018-12-10
Use cholesky for small matrices in root_decomposition

Balandat:
See #422

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1073,6 +1073,11 @@ class LazyTensor(object):
                 "Got a {} of size {}.".format(self.__class__.__name__, self.size())
             )
 
+        # when dealing with small matrices, it's usually faster to use Choleksy decomposition
+        if self.numel() < 250:
+            res = torch.cholesky(self.evaluate())
+            return RootLazyTensor(res)
+
         res, _ = RootDecomposition(
             self.representation_tree(),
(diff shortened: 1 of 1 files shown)


Balandat:
Hmm not sure about the test failure here, probably has to do with the magic done with the collapsed representation in `SumBatchLazyTensor`. Yet another reason to support general batch shapes in the `SumBatchLazyTensor` representation.

jacobrgardner:
I can take a look at the test failure

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1073,6 +1073,11 @@ def root_decomposition(self):
                 "Got a {} of size {}.".format(self.__class__.__name__, self.size())
             )
 
+        # when dealing with small matrices, it's usually faster to use Choleksy decomposition
+        if self.numel() < 250:
```
gpleiss:
```suggestion
        if self.matrix_shape.numel() < 256:
```

gpleiss: This will work for large batch matrices, and 256 for supporting powers of 2.

gpleiss: Actually, rather than using a magic number we should maybe hardcode this into a setting.

gpleiss:
The test error is just a seed issue. The sampling test requires a carefully selected seed since it inherently relies on randomness.

PR #424, opened 2018-12-11
SoftmaxLikelihood forward handles single feature

yushangdi:
SoftmaxLikelihood forward handles single feature.


jacobrgardner:
I'm happy to merge this since it is inline with what happens in `forward` but I'm pretty sure using a softmax likelihood on a single output GP is degenerate -- your logits will just be the mixing weights multiplied by your random function value, but the function value is incapable of changing which mixing weight is highest. Therefore you have a constant classifier. 

More generally, any GP with fewer than c outputs will have at least one class it can't predict as the  maximum (because you'd need to be able to produce the c columns of the  identity matrix, which isn't possible when mixing fewer than c outputs. Maybe thing to do is throw an error and say that the GP should have at least `c outputs?

PR #426, opened 2018-12-12
Move exact GP prediction methods outside of LazyTensor

jacobrgardner:
Step 1 on the road to fantasy observation handling.

This PR moves all Exact GP prediction methods outside of `LazyTensor`.

- Introduces `PredictionStrategy` class that can be subclassed with the `@register_prediction_strategy(LazyTensorType)` decorator to indicate that that subclass is to be used when the `train_train_covar` is of type `LazyTensorType`. For example: https://github.com/cornellius-gp/gpytorch/blob/d21de0ffda3edee1bd20ec9554b734f5e510a588/gpytorch/models/exact_prediction_strategies.py#L177-L178
indicates that this class should be used when `train_train_covar` is an `InterpolatedLazyTensor`.
- `gpytorch.models.exact_prediction_strategies.prediction_strategy` takes a `LazyTensor` representing `train_train_covar` and creates an appropriate `PredictionStrategy` object with the correct predictive methods.
- Kernels no longer define a `has_custom_predictions` method, and the corresponding checks in `LazyEvaluatedKernelTensor` are no longer necessary. The mapping from LazyTensor type to prediction strategy is now explicit.

The next PR will be to move `mean_cache` and `covar_cache` inside of the prediction strategy, and make `ExactGP` cache only the prediction strategy. After that, efficiently handling fantasy observations should be pretty straightforward.


gpytorch/models/exact_gp.py
```diff
@@ -147,33 +148,44 @@ def __call__(self, *args, **kwargs):
...
-            self.mean_cache = mean_cache
-            self.covar_cache = covar_cache
-            if num_tasks > 1:
-                if train_targets.ndimension() == 2:
-                    # Batch multitask
-                    predictive_mean = predictive_mean.view(train_targets.size(0), -1, num_tasks).contiguous()
-                else:
-                    # Standard multitask
-                    predictive_mean = predictive_mean.view(-1, num_tasks).contiguous()
+                train_train_covar = full_covar[..., :num_train, :num_train].evaluate_kernel()
```
gpleiss: The indenting seems off here?

gpytorch/models/exact_gp.py
```diff
@@ -147,33 +148,44 @@ def __call__(self, *args, **kwargs):
...
+                test_train_covar = full_covar[..., num_train:, :num_train]
+                test_test_covar = full_covar[..., num_train:, num_train:]
+
+                pred_strat = prediction_strategy(train_train_covar)
```
gpleiss: Rather than storing the mean_cache and covar_cache, it might make sense to cache pred_strat? This might be cleaner, since then pred_strat can make however many caches it wants.

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def _exact_predictive_covar_inv_quad_form_cache(self, train_train_covar_inv_root, test_train_covar):
+        test_train_covar = test_train_covar.evaluate_kernel()
+        if not isinstance(test_train_covar, SumLazyTensor):
+            return super(SumPredictionStrategy, self)._exact_predictive_covar_inv_quad_form_cache(
+                train_train_covar_inv_root, test_train_covar
+            )
+        else:
+            return tuple(
+                prediction_strategy(lazy_tensor)._exact_predictive_covar_inv_quad_form_cache(
```
gpleiss: If you take the suggestion I have about caching pred_strat, then you'd probably want SumPredStrat to cache the sub-prediction strategies.

gpytorch/models/exact_gp.py
jacobrgardner: Read the last paragraph of my PR description :-)

gpytorch/models/exact_gp.py
jacobrgardner: Nah it's right, ExactGP now handles the train data = None case to return the prior

PR #429, opened 2018-12-12
`gpytorch.beta_features.fast_pred_*` -> `gpytorch.settings.fast_pred_*`

gpleiss:
The LOVE variances are "established" enough at this point that they deserve to be settings and not beta_features :)

They're still off by default. Calling `gpytorch.beta_features.fast_pred_*` returns the appropriate context manager but now throws a deprecation warning.


gpytorch/__init__.py
```diff
@@ -33,6 +32,10 @@
 from .mlls import ExactMarginalLogLikelihood, VariationalMarginalLogLikelihood
 
 
+# Old deprecated stuff
```
Balandat: could this not just call the "new" one from `gpytorch.settings`? 

`with gpytorch.settings.fast_pred_var()` is pretty verbose, especially b/c there doesn't seem to be a good syntax for calling a bunch of different context managers concurrently.

Balandat: hmm I guess you can always do sth like `import gpytorch.settings as gsets` and then call it via `with gsets.fast_pred_var()`

jacobrgardner: I am strongly in favor of keeping them in `settings`. The right way to shorten the name in my opinion is to first `from gpytorch import settings` and then `with settings.context(...)`

jacobrgardner: Naming `gpytorch.settings` something even shorter like `gsets` is also fine imo @Balandat.

PR #431, opened 2018-12-12
Simplify and fix batch mode checks

jacobrgardner:
This PR dramatically simplifies the batch + multitask checks we did in `ExactGP.__call__` to three lines of code. In addition, it makes progress on handling arbitrary batch dimensions by not explicitly assuming only the first dimension is a batch dimension.

This also fixes a bug where we need to update the PredictiveStrategy objects if we expand the train objects to match batch mode test data. We also now use a BatchRepeatLazyTensor to expand `train_train_covar` if needed.

I also added unit test coverage to `test_batch_gp_regerssion` to make sure this doesn't happen again.

### DIFF
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -94,13 +94,14 @@ class ExactGP(GP):
             # Exact inference
             non_batch_train = False
 
+            batch_shape = inputs[0].shape[:-2]
+
             if all(tin.dim() == 2 for tin in self.train_inputs):
                 # Train inputs were non-batch
                 non_batch_train = True
             # If we're doing batch testing, but did std training, adjust the training inputs
             for i, (train_input, input) in enumerate(zip(train_inputs, inputs)):
-                if train_input.dim() < input.dim():
-                    train_inputs[i] = train_input.unsqueeze(0).expand(input.size(0), *train_input.size())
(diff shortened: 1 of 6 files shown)


Balandat:
So this looks great in general, but there is still some remaining issue with the cache. See this notebook: 
[batch_eval_issue.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/2674148/batch_eval_issue.ipynb.txt)

Balandat:
so it seems this happens when the batch shape changes between evaluations. There is no issue when the batch shape does not change.

jacobrgardner:
@Balandat Actually that was an intentional failure in some sense -- I feel we should semantically prefer to use "non-batch testing" when we just have n data points in d dimensions, and singleton batch dimensions don't make sense.

The reason I feel this way is that I believe that at the moment we should support only three cases:
1. train in batch, test in batch, and the batch sizes match. Otherwise what does it mean to test 5 GPs on 2 batches of data? Once multi-batch support is in, there's a case to be made that 5 GPs on 2 batches of data should return `5 x 2 x ...` predictions, but until then they should match if they are both batch.
2. train in batch, test on non-batch
3. train in non-batch, test on batch.

3 batches of training data and 1 batch of test data would violate point 1.

I fixed this to not fail by just squeezing out the first dimension, but in general I feel that singleton batch dimensions should always be dropped.

Balandat:
great, thanks!

PR #432, opened 2018-12-14
Add option to not detach test caches

jacobrgardner:
On master, if you ever need derivatives with respect to *training* inputs at test time, that will not work properly because we detach the training dependent caches we use for predictions. Derivatives with respect to test inputs work just fine.

On the other hand, not detaching the caches by default can lead to performance issues and issues with needing to retain the graph unnecessarily.

To solve this, I've introduced a `detach_test_caches` option that is enabled by default. By disabling this, it is possible to get derivatives with respect to training inputs.

To allow derivatives with respect to the training inputs, just do:
```python
with gpytorch.settings.detach_test_caches(False):
    ...
```

@Balandat This was actually broken before the predictions got moved out: we still detached the caches, but we did so *after* the first set of predictions. Thus, while we got derivatives correctly the first time predictions were made, any subsequent predictions after the first would not get appropriate derivatives. This option fixes that issue.

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -52,7 +52,10 @@ class DefaultPredictionStrategy(object):
         Returns
             - A precomputed cache
         """
-        return train_train_covar_inv_root.detach()
+        if settings.detach_test_caches.on():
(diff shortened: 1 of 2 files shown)


Balandat:
This works! I've been fantasizing about this all along.

PR #433, opened 2018-12-17
Normalize error for pivoted Cholesky

jacobrgardner:
Prevents pivoted Cholesky from terminating too soon on matrices with very small diagonals.


Balandat:
This seems to help a bunch with some stability issues we've seen, many thanks.

PR #434, opened 2018-12-17
Fix kernel memory usage

jacobrgardner:
We were trying to be too clean with recent changes to kernel computations for our own good with using `create_input_grid`, and ended up using `O(dn^2)` memory accidentally. This makes our kernels correctly use `O(n^2)` memory (for exact GPs).

An example of the difference this makes is that exact GPs can be run on the `elevators` dataset after the fix but not before it.

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -156,20 +156,11 @@ class Kernel(Module):
         Computes the covariance between x1 and x2.
         This method should be imlemented by all Kernel subclasses.
 
-        .. note::
-
-            All non-compositional kernels should use the :meth:`gpytorch.kernels.Kernel._create_input_grid`
-            method to create a meshgrid between x1 and x2 (if necessary).
-
-            Do not manually create the grid - this is inefficient and will cause erroneous behavior in certain
-            evaluation modes.
-
         Args:
             - :attr:`x1` (Tensor `n x d` or `b x n x d`)
             - :attr:`x2` (Tensor `m x d` or `b x m x d`)
             - :attr:`diag` (bool):
                 Should the Kernel compute the whole kernel, or just the diag?
-                For most Kernels, this option will be passed into `create_input_grid`
             - :attr:`batch_dims` (tuple, optional):
                 If this option is passed in, it will tell the tensor which of the
                 three dimensions are batch dimensions.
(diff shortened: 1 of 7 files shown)


gpytorch/kernels/kernel.py
```diff
@@ -187,42 +187,28 @@ def forward(self, x1, x2, diag=False, batch_dims=None, **params):
         """
         raise NotImplementedError()
 
-    def _create_input_grid(self, x1, x2, diag=False, batch_dims=None, **params):
```
gpleiss: If you're removing this function from Kernels, can you update the docstrings? (This function is reference to a lot)

test/kernels/test_grid_kernel.py
```diff
@@ -28,23 +28,23 @@ def test_grid_grid(self):
...
+        self.assertLess(torch.norm(grid_eval - actual_eval), 2e-5)
 
     def test_nongrid_nongrid(self):
         base_kernel = RBFKernel()
         data = torch.randn(5, 2)
         kernel = GridKernel(base_kernel, grid)
         grid_eval = kernel(data, data).evaluate()
         actual_eval = base_kernel(data, data).evaluate()
-        self.assertLess(torch.norm(grid_eval - actual_eval), 1e-5)
+        self.assertLess(torch.norm(grid_eval - actual_eval), 2e-5)
```
gpleiss: I think the reason you have to increase these error bounds is because of the way we're computing the kernel.

gpytorch/kernels/kernel.py
```diff
@@ -187,42 +187,28 @@ def forward(self, x1, x2, diag=False, batch_dims=None, **params):
...
+    def _covar_sq_dist(self, x1, x2, **params):
+        if params.get('batch_dims') == (0, 2):
+            x1 = x1.unsqueeze(0).transpose(0, -1)
+            x2 = x2.unsqueeze(0).transpose(0, -1)
 
-        Args:
-            :attr:`x1` (Tensor `n x d` or `b x n x d`)
-            :attr:`x2` (Tensor `m x d` or `b x m x d`) - for diag mode, these must be the same inputs
+        x1_norm = x1.pow(2).sum(dim=-1, keepdim=True)
+        x2_norm = x2.pow(2).sum(dim=-1, keepdim=True)
```
gpleiss: Why not use the built-in norm function and then square afterwards? (This should be more memory efficient, and likely more numerically stable)

gpytorch/kernels/kernel.py
```diff
@@ -187,42 +187,28 @@ def forward(self, x1, x2, diag=False, batch_dims=None, **params):
         """
         raise NotImplementedError()
 
-    def _create_input_grid(self, x1, x2, diag=False, batch_dims=None, **params):
-        """
-        This is a helper method for creating a grid of the kernel's inputs.
-        Use this helper rather than maually creating a meshgrid.
-
-        The grid dimensions depend on the kernel's evaluation mode.
+    def _covar_sq_dist(self, x1, x2, **params):
```
gpleiss: This should have a docstring

gpytorch/kernels/kernel.py
```diff
@@ -187,42 +187,28 @@ def forward(self, x1, x2, diag=False, batch_dims=None, **params):
...
+        if params.get('batch_dims') == (0, 2):
+            x1 = x1.unsqueeze(0).transpose(0, -1)
+            x2 = x2.unsqueeze(0).transpose(0, -1)
 
-        Args:
```
gpleiss: It might be good to whiten x1 and x2 beforehand? This might improve the stability.

test/kernels/test_grid_kernel.py
jacobrgardner: Centering the data in the RBF kernel prevents us from having to increase the tolerance.

jacobrgardner:
@gpleiss Done!

gpytorch/kernels/rbf_kernel.py
```diff
@@ -90,9 +90,8 @@ def __init__(
...
     def forward(self, x1, x2, **params):
-        x1_ = x1.div(self.lengthscale)
-        x2_ = x2.div(self.lengthscale)
-        x1_, x2_ = self._create_input_grid(x1_, x2_, **params)
-
-        diff = (x1_ - x2_).norm(2, dim=-1)
-        return diff.pow(2).div_(-2).exp_()
+        center = x1.mean(dim=-2, keepdim=True)
+        x1_ = (x1 - center).div(self.lengthscale)
+        x2_ = (x2 - center).div(self.lengthscale)
```
gpleiss: Sorry - last change: could we move the centering code into `_covar_sq_distance`?

jacobrgardner: Good point, fixed

PR #437, opened 2018-12-18
`inv_matmul` function takes optional argument for left multiply

gpleiss:
**TLDR** - this PR reduces the number of CG calls from 3 -> 2 for variational inference.

The function `inv_matmul` takes in an optional second argument (`left_tensor`). If it is supplied, the function compute:

`left_tensor` @ lazy_tensor^{-1} @ right_tensor`

If it is not supplied, `inv_matmul` simply computes `lazy_tensor^{-1} @ right_tensor`.

If `left_tensor` is supplied we can compute `lazy_tensor^{-1} @ [right_tensor; left_tensor]` in batch on the forward pass. This makes it possible to skip a call on CG in the backward pass.

### DIFF
--- a/gpytorch/functions/__init__.py
+++ b/gpytorch/functions/__init__.py
@@ -89,23 +89,42 @@ def matmul(mat, rhs):
     return mat.matmul(rhs)
 
 
-def inv_matmul(mat, rhs):
-    """
-    Computes a linear solve with several right hand sides.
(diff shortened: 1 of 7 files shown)


gpleiss:
Blehhhh okay I'll take a look

gpleiss:
Okay - this should be fixed in the next PR that I do (that caches the CG calls), so I'll just close this PR for now

Issue #438, opened 2018-12-19
Numerical issues under new _covar_sq_dist helper


Balandat:
From #434:

> I'm happy to switch back, but I want to be very sure that it doesn't affect numerical stability. Can you make sure that we don't have to increase any error bounds on any of the tests?

Unfortunately, this does significantly affect numerical stability, in particular when using an algorithm for fitting that uses (expanding) line search (such as L-BFGS). With this change many of our tests that previously passed reliably now fail.

I'm assuming the issue is due to the fact that now the individual components rather than their differences are squared: `x1_norm = x1.pow(2).sum(dim=-1, keepdim=True)`. This causes overflow issues especially when using `torch.float` dtype, so the resulting MVNs have `nan`-valued loc/covar parameters. Switching to `torch.double` fixes many of the instability issues, but that's of course not a good solution when running on the GPU.

jacobrgardner:
I've put up a PR to revert the kernel PR for now, and I'll reopen it until we're satisfied that it actually works.

Could you provide an example test that fails with the change but passes without it?

Balandat:
> Could you provide an example test that fails with the change but passes without it?

Will do. The actual test is pretty heavyweight and relies on other code, let me see if I can isolate what's being called during fittig that generates NaNs.

PR #440, opened 2018-12-19
Fix kernel memory usage

jacobrgardner:
See #434, #438, #439. The current kernels are significantly more numerically stable, but use too much memory. We need to achieve the best of both worlds here, as the memory problems prevent exact GPs from running on as large of datasets as they otherwise could, and the round-off problems in the memory efficient version of the kernel cause other problems.

### DIFF
--- a/gpytorch/kernels/cosine_kernel.py
+++ b/gpytorch/kernels/cosine_kernel.py
@@ -98,8 +98,6 @@ class CosineKernel(Kernel):
     def forward(self, x1, x2, **params):
         x1_ = x1.div(self.period_length)
         x2_ = x2.div(self.period_length)
-        x1_, x2_ = self._create_input_grid(x1_, x2_, **params)
-
-        diff = torch.norm((x1_ - x2_).abs(), 2, -1)
+        diff = self._covar_sq_dist(x1_, x2_, **params).sqrt_()
         res = torch.cos(diff.mul(math.pi))
         return res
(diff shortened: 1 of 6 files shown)


gpleiss:
Bleh.... okay I'll take a look

jacobrgardner:
@Balandat @gpleiss This should be all set now! Here are the changes:

- Wherever possible, we use torch.pdist instead of our own distance computation. This includes the case when `x1` and `x2` are equal and non-batch, as well as the batch setting when using pytorch-nightly.

- When we have to fall back on our own distance computation, we make things more stable by zeroing out the diagonal manually and replacing any negative distances with 0. This is a well known trick for making kernels more stable (see e.g. the sklearn kernel implementations).

- When asking for just the diagonal from `covar_sq_dist` we return all zeros if `x1` and `x2` are equal.

jacobrgardner:
@Balandat should be all set, other than the failures on pytorch master which seem unrelated. I'm looking in to this now.

I fixed the easy performance issues and the `RuntimeError` checking is smarter. 

Additionally, in cases where the kernel is called with `covar(x)` *and* the resulting `LazyEvaluatedKernelTensor` is not indexed in to, we skip the `torch.equals(x1, x2)` check. Calling `covar(x, x)` or slicing the kernel necessitates this check still. It's really not all that slow, and it's worth noting that using default settings a check of this cost is something we do every forward call already:
https://github.com/cornellius-gp/gpytorch/blob/ca8ea7ee1f7a6289f6bf16584319c86ecd6c83a1/gpytorch/models/exact_gp.py#L156

jacobrgardner:
@Balandat Actually, the failing tests seem to pass on master but not on this branch. I've no idea how they are possibly related :-/

jacobrgardner:
@Balandat I'll handle the `x1_eq_x2` case slightly smarter later

PR #442, opened 2018-12-20
Rename potrs with cholesky_solve

vishwakftw:
As stated in the title.

This change was introduced in https://github.com/pytorch/pytorch/pull/15334 .

Now the upper arguments in both cholesky_solve and cholesky are the same, so you don't have to explicitly specify upper every time unless necessary.

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -135,7 +135,7 @@ class DefaultPredictionStrategy(object):
             # TODO: Delete this part of the if statement when PyTorch implements potrs derivative.
             fant_cache_lower = torch.gesv(small_system_rhs.unsqueeze(-1), schur_complement)[0]
         else:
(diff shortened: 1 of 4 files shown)


Balandat:
Test failure on pytorch stable is expected. @jacobrgardner, @gpleiss, should we start a pytorch_current branch or sth that tracks backwards-incompatible changes like this that require the most recent pytorch?

gpleiss:
@Balandat when is the next PyTorch release going to be? I'm not a huge fan of having long-running branches. We could try/cactch this for right now?

vishwakftw:
As far as I can tell, the next release is slated for [mid-Jan](https://github.com/pytorch/pytorch/milestone/4).

Balandat:
New version has been released, should we update & merge this PR?

vishwakftw:
The new version is a bug fix only release, so this was not included in it. I guess we have to wait for 1.1.

Balandat:
Gotcha

gpleiss:
Closing in favor of #510

PR #441, opened 2018-12-19
Option to use Cholesky for root decomposition

gpleiss:
Setting `with gpytorch.settings.fast_computations(covar_root_decomposition=False):` will use Cholesky for root decompositions.

(I will expand `gpytorch.settings.fast_computations` to include other functions in a later PR - just want to see if you all like this style)

PR #456, opened 2019-01-07
Setting for exact logprob computations

gpleiss:
Extension of #441 

Setting `gpytorch.settings.fast_computations(log_prob=False)` will use the Cholesky decomposition for computing the exact marginal log likelihood of GPs.

(Not recommended for performance, but useful for debugging).

### DIFF
--- a/test/examples/test_simple_gp_regression.py
+++ b/test/examples/test_simple_gp_regression.py
@@ -299,8 +299,8 @@ class TestSimpleGPRegression(unittest.TestCase):
             self.test_posterior_latent_gp_and_likelihood_with_optimization(cuda=False)
 
     def test_posterior_with_exact_computations_cuda(self):
-        with gpytorch.settings.fast_computations(covar_root_decomposition=False, log_prob=False):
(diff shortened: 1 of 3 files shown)


test/examples/test_simple_gp_regression.py
```diff
@@ -294,6 +294,15 @@ def test_posterior_latent_gp_and_likelihood_with_optimization_cuda(self):
         if torch.cuda.is_available():
             self.test_posterior_latent_gp_and_likelihood_with_optimization(cuda=True)
 
+    def test_posterior_with_exact_computations(self):
+        with gpytorch.settings.fast_computations(covar_root_decomposition=False, log_prob=False):
+            self.test_posterior_latent_gp_and_likelihood_with_optimization(cuda=False)
+
+    def test_posterior_with_exact_computations_cuda(self):
+        with gpytorch.settings.fast_computations(covar_root_decomposition=False, log_prob=False):
```
Balandat: nit: seems more natural to check for cuda first and then context-manage

PR #458, opened 2019-01-09
log_det -> logdet (with appropriate deprecation warnings

gpleiss:
- log_det -> logdet
- inv_quad_log_det -> inv_quad_logdet
- Add some new deprecation helpers


gpleiss:
Making this change to match PyTorch's function name (they use "logdet").

PR #462, opened 2019-01-11
GP regression with derivative information

dme65:
Adding a new kernel `RBFKernelGrad` and a mean `ConstantMeanGrad` that can be used together with MultiTask to train GPs from value and derivative information.


gpytorch/kernels/scale_kernel.py
```diff
@@ -94,11 +94,12 @@ def forward(self, x1, x2, batch_dims=None, diag=False, **params):
...
-
         if torch.is_tensor(orig_output):
             outputscales = outputscales.view(-1, *([1] * (orig_output.dim() - 1)))
 
         if diag:
             return delazify(orig_output) * outputscales
-
         return orig_output.mul(outputscales)
+
+    def size(self, x1, x2):
```
gpleiss: We might want to add this to our other decorator kernels as well (GridInterpolationKernel, ProductKernel, etc.)

gpytorch/means/constant_mean_grad.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+
+import torch
+from .mean import Mean
+
+
+class ConstantMeanGrad(Mean):
```
gpleiss: Is there a way that we could make this more general - like how `MultitaskMean` wraps arbitrary means?

gpytorch/kernels/rbf_kernel_grad.py
```diff
@@ -0,0 +1,152 @@
...
+        b = 1
+        if len(x1.size()) == 2:
+            n1, d = x1.size()
+            n2, d = x2.size()
+        else:
+            b, n1, d = x1.size()
+            _, n2, _ = x2.size()
+
+        K = torch.zeros(b, n1 * (d + 1), n2 * (d + 1), device=x1.device, dtype=x1.dtype)  # batch x n1(d+1) x n2(d+1)
+        ell = self.lengthscale.squeeze()
```
gpleiss: This squeeze will cause broadcasting issues if `batch_size != 1`. Is it necessary?

dme65: I get a shape error when I call `likelihood(model(test_x)).confidence_region()` if I remove this squeeze. Is there an obvious reason why?

gpytorch/means/constant_mean_grad.py
dme65: I'm not sure if this is possible. The mean vector evaluated at one point is a vector of length d+1 with the original scalar mean function and the corresponding d partial derivatives along x_1, ..., x_d. I think this class has to be implemented for each mean, just like ConstantMean etc.

gpleiss: Okay that should be fine then.

Issue #460, opened 2019-01-11
Deploy to Anaconda Cloud


cdgreenidge:
First--wonderful work on the package!

I would like to depend on gpytorch in a project that already uses conda. Unfortunately conda doesn't allow the installation of dependencies via pip, so it would be nice if gpytorch was deployed to Anaconda Cloud (or similar). 

I'd be happy to contribute the build files. Any thoughts?

jacobrgardner:
Definitely agreeable to maintaining our releases on conda as well! Just went through the process of making a conda account for it, once we have build files I can get it uploaded.

cdgreenidge:
Great, I will start work and file a pull request when it's done.

PR #463, opened 2019-01-11
Add meta.yaml for conda build

cdgreenidge:
Addresses #460.

A couple notes on the conda build file:

- It gets package version info from `setup.py`, so most routine releases shouldn't require `meta.yaml` to be modified.
- It does not get dependencies from `setup.py`, so adding dependencies *will* require `meta.yaml` to be modified. This should happen only rarely.
- The testing is minimal; it only makes sure the package imports correctly. This speeds up the build and avoids duplicating the unit tests run on Travis. 
- The build only supports *NIX (to avoid `setup.bat` clutter) but the package is type "noarch" so it will work on all platforms. 

I didn't see any code to automatically deploy to PyPI, so I didn't add anything to automatically deploy to Anaconda Cloud. In any case the manual deployment process is easy--`conda-build` prints instructions on build success.

Once it's deployed to Anaconda Cloud, users can install with the following commands:

```
$ conda config --append channels pytorch
$ conda install -c CHANNEL_NAME pgplvm
```

where `CHANNEL_NAME` is the name of the channel on which gpytorch is hosted.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -1,0 +1,42 @@
+{% set data = load_setup_py_data(setup_file="setup.py", from_recipe_dir=True) %}
+
+package:
+  name: {{ data.get("name")|lower }}
+  version: {{ data.get("version") }}
+
+source:
+  path: ./
+
+build:
+  noarch: python
+  script: "$PYTHON setup.py install --single-version-externally-managed --record=record.txt"
+
+requirements:
+  host:
+    - python>=3.6
+
+  run:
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Thanks, this worked!

I also took care of uploading it to a channel while testing it out, so you can now install gpytorch-0.1.1 via conda: https://anaconda.org/gpytorch/gpytorch

PR #464, opened 2019-01-14
JIT for distance computation, more efficient Matern kernel, JIT scripted linear CG

jacobrgardner:
TL;DR: Reduces the memory overhead of kernels (and therefore exact GPs) by ~3x when using squared distances, reduces the memory usage of RBFKernel specifically by ~4x.

- `__slow_sq_dist` is now `__jit_sq_dist` and uses the torch JIT script functionality. This is around as fast as `torch.pdist(...).pow_(2)`, but (a) uses much less memory than `torch.pdist` surprisingly, and (b) uses around 3x less memory than `__slow_sq_dist`.

- Kernels can now pass a `jit_func` to `covar_dist`. If they do, this JIT function is called in non-diag mode instead of doing the distance computation at all. This is useful if the Kernel is just the distance plus a few steps, because the few extra steps can just be thrown in the JIT function which saves memory.

## To do
- Use torch.pdist in the `elif not square_dist` branch once `torch.pdist` is less buggy / bad.
- Figure out how to do a numerically stable `__jit_dist` for non squared distances.
- Write custom JIT functions for other kernels (right now implemented for RBF only).

cc/ @KeAWang @gpleiss

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -263,8 +263,10 @@ class Kernel(Module):
         elif jit_func is not None:
             res = jit_func(x1, x2, torch.tensor(diag), torch.tensor(x1_eq_x2))
         elif not square_dist:
-            # TODO: Use torch.pdist here when it isn't buggy.
-            res = torch.norm(x1.unsqueeze(-2) - x2.unsqueeze(-3), p=2, dim=-1)
+            if x1_eq_x2:
+                res = self.__jit_sq_dist(x1, x2, torch.tensor(diag), torch.tensor(x1_eq_x2)).clamp_min_(1e-30).sqrt_()
+            else:
+                res = torch.norm(x1.unsqueeze(-2) - x2.unsqueeze(-3), p=2, dim=-1)
         else:
             res = self.__jit_sq_dist(x1, x2, torch.tensor(diag), torch.tensor(x1_eq_x2))
 
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -10,7 +10,6 @@ from ..utils.deprecation import _deprecate_kwarg
 from ..utils.transforms import _get_inv_param_transform
 from torch.nn.functional import softplus
 from numpy import triu_indices
-import warnings
 
 
 class Kernel(Module):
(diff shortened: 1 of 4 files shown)


gpytorch/kernels/kernel.py
```diff
@@ -193,29 +192,28 @@ def __pdist_dist(self, x1):
...
         # Zero out negative values
         res.clamp_min_(0)
 
         return res
 
-    def _covar_dist(self, x1, x2, diag=False, batch_dims=None, square_dist=False, **params):
+    def _covar_dist(self, x1, x2, diag=False, batch_dims=None, square_dist=False, jit_func=None, **params):
         """
         This is a helper method for computing the Euclidean distance between
         all pairs of points in x1 and x2.
```
Balandat: maybe explain in the docstring what `jit_func` is and how it's supposed to be used?

gpytorch/kernels/kernel.py
```diff
@@ -193,29 +192,28 @@ def __pdist_dist(self, x1):
 
         return res
 
-    def __slow_sq_dist(self, x1, x2, diag, x1_eq_x2):
+    @torch.jit.script
```
Balandat: Hmm so I'm not familiar with this - if you define a `@torch.jit.script` function on a class, it doesn't take the standard `self` arg? (makes sense b/c it's scripted and can't work with arbitrary python objects, but still somewhat confusing). 

See my comment below how this could be re-used.

gpytorch/kernels/rbf_kernel.py
```diff
@@ -89,8 +90,31 @@ def __init__(
             eps=eps,
         )
 
-    def forward(self, x1, x2, **params):
+    @torch.jit.script
```
Balandat: This is quite repetitive - why not define `__jit_sq_dist` as a utility function outside the `Kernel` class, then you should be able to have the  `__jit_rbf_kernel` just call that.

test/kernels/test_grid_kernel.py
```diff
@@ -28,7 +28,7 @@ def test_grid_grid(self):
         self.assertIsInstance(grid_covar, KroneckerProductLazyTensor)
         grid_eval = kernel(grid_data, grid_data).evaluate()
         actual_eval = base_kernel(grid_data, grid_data).evaluate()
-        self.assertLess(torch.norm(grid_eval - actual_eval), 1e-5)
+        self.assertLess(torch.norm(grid_eval - actual_eval), 1.2e-5)
```
Balandat: magic!

gpytorch/kernels/rbf_kernel.py
jacobrgardner: For efficiency -- I'm not sure that if we call a scripted function (or even whether we *can*) within a scripted function whether the JIT will compile the whole trace or not. The whole point of defining the custom function for RBF is so that the exponentiation happens in the JIT script, which seems to save us a kernel matrix copy worth of memory.

gpytorch/kernels/kernel.py
jacobrgardner: Yeah, that seems to be the case -- things break if it takes self.

gpytorch/kernels/kernel.py
jacobrgardner: I don't see why we'd ever not use the JIT version. The `jit_func` isn't intended to provide a squared distance function, rather it overrides what `_covar_dist` does with custom behavior. This is done because for RBF specifically, we want the exp() and div() done in the JIT script because it saves additional memory.

jacobrgardner:
I've added two things to this PR:

1. JIT scripted linear CG internals
2. A more efficient (both time and memory) implementation of the non squared distance matrix.

Balandat:
Cool, thanks. Let me check this with some downstream code.

Balandat:
alright this seems to run fine with our tests, this is good to go from my end

gpytorch/utils/linear_cg.py
```diff
@@ -55,6 +97,9 @@ def linear_cg(
         initial_guess = torch.zeros_like(rhs)
     if preconditioner is None:
         preconditioner = _default_preconditioner
+        no_precond = True
```
Balandat: Nit: This seems a little odd with the double negative below. How about just using precond as the flag?

gpytorch/kernels/kernel.py
```diff
@@ -248,6 +270,8 @@ def _covar_dist(self, x1, x2, diag=False, batch_dims=None, square_dist=False, **
 
         res = None
 
+        distance_module = Distance(postprocess_func)
+
         if diag:
```
gpleiss: `diag` mode should also call `postprocess`. Otherwise, the kernel has a lot of repeated code, and Kernel subclasses will have to have lots of if statements that `_covar_dist` is trying to avoid.

gpytorch/kernels/rbf_kernel.py
```diff
@@ -89,8 +95,10 @@ def __init__(
...
 
-    def forward(self, x1, x2, **params):
+    def forward(self, x1, x2, diag=False, **params):
         x1_ = x1.div(self.lengthscale)
         x2_ = x2.div(self.lengthscale)
-        diff = self._covar_dist(x1_, x2_, square_dist=True, **params)
-        return diff.div_(-2).exp_()
+        diff = self._covar_dist(x1_, x2_, square_dist=True, diag=diag, postprocess_func=postprocess_rbf, **params)
+        if diag:
+            diff.div_(-2).exp_()
```
gpleiss: See comment above.

Issue #465, opened 2019-01-16
Implement cholesky-based root_decomposition for lazy tensors


Balandat:
The new `fast_computations(covar_root_decomposition=False)` setting allows to use `torch.cholesky` for computing root decompositions. Right now this lives in `LazyTensor`, so when calling this the full matrix has to be instantiated, except for a small number of lazies that overwrite `root_decomposition` (e.g. `DiagLazyTensor`).

For things like `BlockDiagLazyTensor` this is of course very inefficient. We should overwrite root_decomposition there, and if use batch Cholesky decomposition to compute the Cholesky decomposition of each sub-block, and stitch the results together in another `BlockDiagLazyTensor`.

gpleiss:
Something that I've been thinking about recently: most of our public LazyTensor methods do lots of checking (i.e. `root_decomposition` determines whether or not to do a Cholesky-based root decomposition, `tranpsose` does argument checking, `__getitem__` adds extra indices when necessary, etc.)

I think moving forward we should make it a pattern where all of our public LazyTensor methods that do any sort of checking/casing should call an underscore'd version of the method. We've started this pattern a bit: (e.g. `__getitem__` calls `_getitem` after processing all the batch dimensions).

It might be good to have `_root_decomposition`, `_inv_quad_logdet`, etc.

gpleiss:
@jacobrgardner ^^

gpleiss:
@Balandat - I have a WIP on this that I want to merge in after the multi-batch LT stuff.

gpleiss:
I think this has been in GpyTorch for a while now :) Closing

PR #466, opened 2019-01-16
Smarter root_decomposition for BlockDiagLazyTensor

Balandat:
This will speed up things significantly when called with `settings.fast_computations(covar_root_decomposition=False)`

See #465

### DIFF
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -1,8 +1,12 @@
 #!/usr/bin/env python3
 
 import torch
-from .block_lazy_tensor import BlockLazyTensor
+
 from .. import settings
+from ..utils.memoize import cached
+from .block_lazy_tensor import BlockLazyTensor
+from .non_lazy_tensor import NonLazyTensor
+from .root_lazy_tensor import RootLazyTensor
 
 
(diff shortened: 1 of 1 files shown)


gpytorch/lazy/block_diag_lazy_tensor.py
```diff
@@ -118,6 +122,14 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
             logdet_res = logdet_res.view(*self.batch_shape, -1).sum(-1)
         return inv_quad_res, logdet_res
 
+    @cached(name="root_decomposition")
+    def root_decomposition(self):
+        if settings.fast_computations.covar_root_decomposition.on():
+            return super().root_decomposition()
```
gpleiss: I imagine that what we probably should be doing for the standard case is:

`RootLazyTensor(self.__class__(self.base_lazy_tensor.root_decomposition().root, num_blocks=self.num_blocks))`

This should be faster and more accurate than what we're currently doing.

Balandat: sounds good, will update

Balandat:
@gpleiss I made the change for the fast_computations case as well, this should be good to go in

PR #470, opened 2019-01-18
Add event_shape property to MTMVN

Balandat:
Prior to this, MTMVNs would return the event shape of the underlying MVN representation, which is not the shape that (r)sample would return.


Balandat:
simple enough

PR #471, opened 2019-01-23
Multi-output GP model

Balandat:
A draft for a multi-output GP model (+ associated Likelihood + MLL). Little more than a convenient container of a list of models. In particular evaluation is done sequentially internally. The better way to do this would be to specify a structured block-diagonal kernel, with a block for each output. This however will require some more work on the internals of how inference is handled (in terms of properly sub-partitioning the full covariance matrix).

Will add some examples and unit tests for this soon.


gpytorch/module.py
```diff
@@ -34,12 +34,13 @@ def __call__(self, *inputs, **kwargs):
...
             return outputs
 
-        elif torch.is_tensor(outputs) or isinstance(outputs, Distribution) or isinstance(outputs, LazyTensor):
-            return outputs
-        else:
-            raise RuntimeError(
-                "Output must be a Distribution, torch.Tensor, or LazyTensor. Got {}".format(outputs.__class__.__name__)
-            )
+        return outputs
+        # elif torch.is_tensor(outputs) or isinstance(outputs, Distribution) or isinstance(outputs, LazyTensor):
```
Balandat: commented this out for convenience for now until multi-output api stabilizes.

Balandat:
not sure what causes the docs failure here...

Balandat:
@jacobrgardner, @gpleiss how do you feel about getting this in? I know it's not really an efficient way to do things, but it's a convenient wrapper that I'd like to use downstream.

jacobrgardner:
A few high level things before actual code review:

1. I think the notebook should clearly link to the Batch mode example notebook (maybe the batch mode notebook should also be moved to the same folder?). For independent GPs, if people *can* use batch mode, people always *should* use batch mode due to the additional parallelism.

2. We should fix the doc errors before merging, because readthedocs doesn't build *any* of the documentation if there is any failure. This is why we have Travis's docs build be extremely strict about warnings.

gpytorch/models/utils.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3
```
jacobrgardner: Either move this to `__init__.py` in `gpytorch.utils` or define in `multi_output_gp.py` directly since that's where it gets used.

jacobrgardner: Given that `_get_tuple_args` appeared directly in the likelihood, maybe let's just define it in `multi_output_gp.py` for now

gpleiss:
I can take a look at the docs error, and can do a detailed review as well.

The one high-level concern I have is naming. `MultiTask` and `MultiOutput` are similar - and the differences are that `MultiTask` has correlated outputs and is batched. Is there some better naming convention we could use?

Balandat:
> I think the notebook should clearly link to the Batch mode example notebook

Sure, will add that. 

> Is there some better naming convention we could use?

So I guess the idea is that `AbstractMultiOutputGP` isn't necessarily assuming independence between outputs, but provides a abstract (hence the name) API for multi-output models (compared to the Kronecker GP it allows more flexibility in terms of how the kernel is specified, and does not assume block design). The thing you would subclass for the independent use case is `IndependentMultiOutputGP`, which should be pretty clear.

Balandat:
Regarding the doc failure: For some reason I cannot display the file here on GitHub either. I think this has to do with the fact that I manually removed some components that our proprietary distribution of jupyter adds in - let me make sure that this doesn't corrupt the file.

Balandat:
Alright this is should be ready

Balandat:
@gpleiss I can definitely do 1. - I'm not sure about the more generic classes b/c there is some input parsing that is going on, let me check.

gpleiss:
@Balandat sounds good!

Issue #478, opened 2019-01-25
Initializing GaussianLikelihood noise broken


KeAWang:
```
>>> import gpytorch
>>> gl = gpytorch.likelihoods.GaussianLikelihood()
>>> gl.initialize(noise=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../gpytorch/gpytorch/module.py", line 89, in initialize
    setattr(self, name, val)
  File "../lib/python3.6/site-packages/torch/nn/modules/module.py", line 579, in __setattr__
    object.__setattr__(self, name, value)
  File ".../gpytorch/gpytorch/likelihoods/gaussian_likelihood.py", line 63, in noise
    self.noise_covar.initialize(value)
TypeError: initialize() takes 1 positional argument but 2 were given
```

raw_noise works though
```>>> gl.initialize(raw_noise=3)
GaussianLikelihood(
  (noise_covar): HomoskedasticNoise()
)```

gpleiss:
We're open to a PR 😄

PR #479, opened 2019-01-26
Fix GaussianLikelihood noise initialization and remove log deprecations

KeAWang:
See #478


KeAWang:
The `test_fantasy_updates_*` tests are failing. I replaced all the places where `log_*` was used instead of `raw_*` in the tests which shouldn't make a difference since `deprecate_kwargs` used to do that automatically. So I'm not sure why they're failing now... Any ideas?

jacobrgardner:
@KeAWang The initialize method was smarter than that. Instead of just replacing `log` with `raw` it exponentiated the log initialization and then did the inv param transform. See the `Module.initialize` code you deleted.

By just replacing `log` with `raw` you've changed the initial parameter values.

gpytorch/module.py
```diff
@@ -63,26 +63,16 @@ def hyperparameters(self):
...
-        from .utils.log_deprecation import MODULES_WITH_LOG_PARAMS
 
         for name, val in kwargs.items():
             if isinstance(val, int):
                 val = float(val)
-            if any(isinstance(self, mod_type) for mod_type in MODULES_WITH_LOG_PARAMS) and "log_" in name:
-                base_name = name.split("log_")[1]
-                name = "raw_" + base_name
-                if not torch.is_tensor(val):
-                    val = self._inv_param_transform(torch.tensor(val).exp()).item()
```
jacobrgardner: E.g. here

PR #480, opened 2019-01-26
Make zero_mean_mvn samples work with arbirary sample_shapes

Balandat:
Makes this consistent with the torch interface, declutters some of the special casing by using shape arithmetic. This will be useful whenutilizing QMC base samples, in which case we can't just reshape
everyting into a big tensor but have to comply with batch / sample shape (because samples are not iid anymore).

It also enables sampling from multi-dimensional batch shapes in LazyTensors (once those are fully supported)


Balandat:
The test for BatchRepeatLazyTensor fails b/c of an issue in matmul -- maybe someone who knows the implementation of BatchRepeatLazyTensor can take a look?

gpleiss:
I can take a look tomorrow

Balandat:
Ping? @gpleiss

gpleiss:
Ooops sorry checking now!

Balandat:
yeah that makes a lot of sense. Let me know how it goes. I also have some draft commits that allow sampling from degenerate MVNs that I'll stack on top of that.

gpleiss:
Works pretty well! I'll open up a new PR.

gpleiss:
Closing in favor of #481

PR #482, opened 2019-01-28
Support batch mode mvns in from_independent_mvns

sdaulton:
- Pass `num_blocks` to BlockDiagLazyTensor
- Concatenate base covariance matrices such that `covar` is `t x n x n` or `bt x n x n`, where `t` is the number blocks (and tasks) here 

Added unit tests for `from_independent_mvns`


gpytorch/distributions/multitask_multivariate_normal.py
```diff
@@ -48,13 +48,22 @@ def from_independent_mvns(cls, mvns):
...
+            raise ValueError("Multiple batch dimensions are not supported in from_independent_mvns.")
         mean = torch.stack([mvn.mean for mvn in mvns], -1)
         # TODO: To do the following efficiently, we don't want to evaluate the
         # covariance matrices. Instead, we want to use the lazies directly in the
         # BlockDiagLazyTensor. This will require implementing a new BatchLazyTensor:
         # https://github.com/cornellius-gp/gpytorch/issues/468
-        covar = torch.stack([mvn.covariance_matrix for mvn in mvns])
-        covar_lazy = BlockDiagLazyTensor(NonLazyTensor(covar))
+        batch_mode = len(mvns[0].covariance_matrix.shape) == 3
+        covar = torch.cat(
```
jacobrgardner: Is there any reason not to make a CatLazyTensor out of the lazy covariance matrices and then make a BlockDiagLazyTensor from that?

Balandat: @jacobrgardner that make sense and will work if `batch_mode = True`, but I am not sure that's straightforward to do if `batch_mode = False`, b/c of the unsqueezing that is necessary for `CatLazyTensor` to work.

Ideally, we'd have a `BlockDiagLazyTensor` that works in general batch mode (see above).

Balandat: @sdaulton can you maybe case-distinguish here and use the CatLazyTensor as jake suggested if `batch_mode = True`, provided that works?

sdaulton: Will do

PR #484, opened 2019-01-29
Support recursively initializing parameters

### LINKED ISSUE #476: Recursively initialize Module parameters
Say I have an `ExactGP` called `gp`. It would be great if I could just do `gp.initialize(kwargs)` and all the parameters will be initialized recursively. This would allow us to put all the initialization values in one place. 

Note that it would have to raise an error if there were a parameter name collision.

KeAWang:
closes #476

### DIFF
--- a/gpytorch/module.py
+++ b/gpytorch/module.py
@@ -47,14 +47,26 @@ class Module(nn.Module):
         """
         Set a value for a parameter
 
-        kwargs: (param_name, value) - parameter to initialize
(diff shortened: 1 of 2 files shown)


gpleiss:
Looks good, but can you use this in one of the example tests? (Just so we have some test coverage of it)

KeAWang:
Added a test in `test_simple_gp_regression.py`

PR #487, opened 2019-01-30
Fix InducingPointKernel memory usage and speed

jacobrgardner:
Right now, InducingPointKernel unintentionally gets the diagonal correction by calling the kernel on the full data and then calling `diag` rather than calling the kernel with `diag=True`:
https://github.com/cornellius-gp/gpytorch/blob/4e30c948055e20c35b11e7435e8eba673a7f3401/gpytorch/kernels/inducing_point_kernel.py#L63
 This has clearly unintended ramifications for memory usage and speed. While I was at it, I also changed `_inducing_inv_root` to take advantage of the fact that `torch.cholesky` now operates in batch mode.

### DIFF
--- a/gpytorch/kernels/inducing_point_kernel.py
+++ b/gpytorch/kernels/inducing_point_kernel.py
@@ -4,7 +4,7 @@ import math
 import torch
 from .kernel import Kernel
 from ..functions import add_jitter
-from ..lazy import lazify, delazify, DiagLazyTensor, MatmulLazyTensor, PsdSumLazyTensor, RootLazyTensor
+from ..lazy import delazify, DiagLazyTensor, MatmulLazyTensor, RootLazyTensor
 from ..distributions import MultivariateNormal
 from ..mlls import InducingPointKernelAddedLossTerm
 
@@ -43,9 +43,9 @@ class InducingPointKernel(Kernel):
             return self._cached_kernel_inv_root
         else:
             inv_roots_list = []
+            chols = torch.cholesky(add_jitter(self._inducing_mat), upper=True)
             for i in range(self._inducing_mat.size(0)):
-                jitter_mat = add_jitter(self._inducing_mat[i])
-                chol = torch.cholesky(jitter_mat, upper=True)
+                chol = chols[i]
                 eye = torch.eye(chol.size(-1), device=chol.device)
                 inv_roots_list.append(torch.trtrs(eye, chol)[0])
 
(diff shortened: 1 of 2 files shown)


gpytorch/kernels/inducing_point_kernel.py
```diff
@@ -60,8 +60,8 @@ def _get_covariance(self, x1, x2):
             covar = RootLazyTensor(k_ux1.matmul(self._inducing_inv_root))
 
             # Diagonal correction for predictive posterior
-            correction = (lazify(self.base_kernel(x1, x2)).diag() - covar.diag()).clamp(0, math.inf)
-            covar = PsdSumLazyTensor(DiagLazyTensor(correction), covar)
+            correction = (self.base_kernel(x1, x2, diag=True) - covar.diag()).clamp(0, math.inf)
+            covar = covar + DiagLazyTensor(correction)
```
gpleiss: Why not PsdSumLazyTensor?

gpytorch/kernels/inducing_point_kernel.py
```diff
@@ -43,9 +43,9 @@ def _inducing_inv_root(self):
             return self._cached_kernel_inv_root
         else:
             inv_roots_list = []
+            chols = torch.cholesky(add_jitter(self._inducing_mat), upper=True)
```
gpleiss: 👏

gpytorch/kernels/inducing_point_kernel.py
```diff
@@ -43,9 +43,9 @@ def _inducing_inv_root(self):
...
             for i in range(self._inducing_mat.size(0)):
-                jitter_mat = add_jitter(self._inducing_mat[i])
-                chol = torch.cholesky(jitter_mat, upper=True)
+                chol = chols[i]
                 eye = torch.eye(chol.size(-1), device=chol.device)
```
gpleiss: Minor nit: we can create this identity matrix outside the for loop

gpytorch/kernels/inducing_point_kernel.py
jacobrgardner: This keeps things as an AddedDiagLazyTensor, but I have no strong opinions about it -- is there a good reason to prefer PsdSumLazyTensor over AddedDiagLazyTensor given that we'd have to do a root decomposition of `covar` anyways?

gpleiss: `covar` is already a RootLazyTensor, so we can avoid an extra root decomposition

jacobrgardner: Okay, I put PsdSumLazyTensor back in and simplified the for loop in inducing_inv_root further.

PR #488, opened 2019-01-30
add feature flag to skip posterior covar computation

### LINKED ISSUE #477: Option to not compute the variances
In some use cases, we only ever care about the predictive mean of the GP. It would save a lot of computation time if there were an option to not compute the variances. Currently we have a hack to turn off variance computation at https://github.com/cornellius-gp/gpytorch/commit/34912d666664c34b0e420dafbd7b424445a52bc0

Maybe we can un-hackify this and make it an option for all GPs.

KeAWang:
Attempt at implementing skipping posterior covariance. 

This would close #477

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -4,7 +4,9 @@ import torch
 
 from .. import settings
 from ..distributions import MultivariateNormal
-from ..lazy import InterpolatedLazyTensor, LazyTensor, MatmulLazyTensor, RootLazyTensor, SumLazyTensor
+from ..lazy import (
+    InterpolatedLazyTensor, LazyTensor, MatmulLazyTensor, RootLazyTensor, SumLazyTensor, ZeroLazyTensor
+)
 from ..utils.interpolation import left_interp, left_t_interp
 from ..utils.memoize import cached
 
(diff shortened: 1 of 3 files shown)


KeAWang:
@gpleiss would something like this work for #477? I'm not too familiar with the predictive strategies part of gpytorch

KeAWang:
Fixed

jacobrgardner:
Can you add the feature flag to also take effect in the non LOVE setting as well?

jacobrgardner:
nit: I might also call it `skip_posterior_variances` since you're actually skipping the variance calculation as well (covariances are already not computed until they are needed since the computation is done via lazy arithmetic)

andrewgordonwilson:
We may want to use a name like `no_posterior_variances` or `posterior_mean_only` instead, to avoid overloading terminology with the "skip" method.

KeAWang:
`no_posterior_variances` would probably be better. `posterior_mean_only` might imply that the forward pass will return the posterior mean instead of a `MultivariateNormal`.

KeAWang:
I'm unable to reproduce the Travis failing unittests locally at all. I've run the unittests on all combinations of pytorch stable, pytorch master, with and without CUDA and explicitly checked that the means computed with and without `skip_posterior_variances` are in fact equal. 

For now, I've made it so that those two unittests are run only with CUDA to prevent Travis from running them and failing them.

PR #493, opened 2019-02-01
Fix cuda test in test_batch_whitened_svgp_gp_regression

Balandat:
The issue was that different number of optimization iterations were done in the cuda test. That's why there shouldn't be a bunch of copy-pasted code that makes it hard to spot issues - instead let's stick to the pattern where as much code as possible is reused.


Balandat:
wait something got mixed up here it seems

Balandat:
Wrong file was edited, will submit a new PR with the correct modifications

PR #495, opened 2019-02-04
Improve setup.py

gpleiss:
setup.py now does the following:

- Includes optional dependencies (for code formatting, examples, tests, doc building, Pyro integration) in `extra_requires`
- Reads requirements from appropriate `requirements.txt` (for less duplication)
- Does not require `torch` if `pytorch-nightly` is installed.

### DIFF
--- a/gpytorch/__init__.py
+++ b/gpytorch/__init__.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-
 from .module import Module
 from . import (
     beta_features,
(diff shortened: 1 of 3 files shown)


gpytorch/__init__.py
```diff
@@ -39,6 +40,9 @@
 # Old deprecated stuff
 fast_pred_var = beta_features._moved_beta_feature(settings.fast_pred_var, "gpytorch.settings.fast_pred_var")
 
+# Get the version
+with open(os.path.join(os.path.dirname(__file__), "..", "VERSION"), "r") as f:
```
Balandat: So doing this kind of stuff outside the `setup.py` file can bey annoying in the context of some build systems (e.g. ours...) that repackage the source into .par files or the like. This can be handled by explicit specifying the source files in the build system so it's workable, but I'd advise you to use the `os` business sparingly -  otherwise it'll make it harder to use the package in non-vanilla environments.

An alternative would be to just have a standard `version.py` python module that specifies a single `VERSION` variable - that kind of stuff is typically handled automatically by build systems. Or just specify that variable directly in `setup.py` if that't the only place where it is used.

gpleiss: Got rid of the os stuff.

jacobrgardner:
@Balandat this does less bad external file reading now, and follows the way torchvision gets their version out

PR #497, opened 2019-02-05
Fix bug in grid_size utility for additive structure

gpleiss:
`choose_grid_size` was scaling by the number of dimensions to give reasonable grid sizes for Kronecker structure. We don't need this for additive structure.
There is now an option in `choose_grid_size` that fixes this.


gpleiss:
I think this got merged in?

PR #498, opened 2019-02-05
Make zero_mean_mvn samples work with arbirary sample_shapes (onto batch_lazy_tensor PR)

gpleiss:
(Same thing as #480, but onto the batch lazy tensor PR WIP)

Makes this consistent with the torch interface, declutters some of the special casing by using shape arithmetic. This will be useful whenutilizing QMC base samples, in which case we can't just reshape
everyting into a big tensor but have to comply with batch / sample shape (because samples are not iid anymore).

It also enables sampling from multi-dimensional batch shapes in LazyTensors (once those are fully supported)


Balandat:
Awesome, thanks for rebasing this. This look mostly good, but I do get an error when sampling from the MultitaskMultivariateNormal when using the cholesky decomposition to get the root. I'll try to fix this myself.

Balandat:
I fixed the issue, this happened when the approximate root_decomposition is low rank and non-square. This should be good to go into your branch!

PR #499, opened 2019-02-05
Kernel checkpointing

gpleiss:
This PR enables implicitly computing kernel matrices through MVMs. It computes the kernel in batches, trading of time for memory.

This feature is activated by using `with gpytorch.beta_features.checkpoint_kernel(<value>)`

### DIFF
--- a/gpytorch/beta_features.py
+++ b/gpytorch/beta_features.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 import warnings
-from .settings import _feature_flag
+from .settings import _feature_flag, _value_context
 from .settings import fast_pred_var as _fast_pred_var
 from .settings import fast_pred_samples as _fast_pred_samples
 
@@ -24,6 +24,25 @@ class _moved_beta_feature(object):
 
 fast_pred_var = _moved_beta_feature(_fast_pred_var)
 fast_pred_samples = _moved_beta_feature(_fast_pred_samples)
+
+
+class checkpoint_kernel(_value_context):
+    """
+    Should the kernel be computed in chunks with checkpointing or not? (Default, no)
+
+    If `split_size = 0`:
+        The kernel is computed explicitly. During training, the kernel matrix is
+        kept in memory for the backward pass. This is the fastest option but the
+        most memory intensive.
+    If `split_size > 0`:
+        The kernel is never fully computed or stored. Instead, the kernel is only
+        accessed through matrix multiplication. The matrix multiplication is
+        computed in `segments` chunks. This is slower, but requires significantly less memory.
+
+    Default: 0
+    """
+
+    _global_value = 0
 
 
 class diagonal_correction(_feature_flag):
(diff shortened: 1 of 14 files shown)


gpleiss:
@jacobrgardner ready for review (no longer against batch_lt3)

gpytorch/beta_features.py
```diff
@@ -26,6 +26,25 @@ def __getattr__(self, name):
 fast_pred_samples = _moved_beta_feature(_fast_pred_samples)
 
 
+class checkpoint_kernel(_value_context):
```
jacobrgardner: I know we used this for the existing code, but could we change to e.g. `checkpoint_kernel_size` or `kernel_partition_size` or anything that has the word "size" in it since that's what the value represents?

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -122,101 +193,97 @@ def diag(self):
...
+            res = res.evaluate()
+        return res.view(self.shape[:-1]).contiguous()
+
+    @cached(name="kernel_eval")
     def evaluate_kernel(self):
         """
         NB: This is a meta LazyTensor, in the sense that evaluate can return
         a LazyTensor if the kernel being evaluated does so.
         """
-        if hasattr(self, "_cached_kernel_eval"):
```
jacobrgardner: Was removing this intentional? If so it seems strange to still save the kernel call in a private cache. Also, I think this has perf implications for predictions (see below).

gpytorch/models/exact_gp.py
```diff
@@ -235,7 +235,7 @@ def __call__(self, *args, **kwargs):
             train_mean = full_mean.narrow(-1, 0, train_targets.size(-1))
 
             if self.prediction_strategy is None:
-                train_train_covar = full_covar[..., :num_train, :num_train].evaluate_kernel()
+                train_train_covar = full_covar[..., :num_train, :num_train]
```
jacobrgardner: This might be dangerous perf wise. Prediction strategies cache `train_train_covar` and then use it repeatedly. Since `LazyEvaluatedKernelTensor` no longer even caches its kernel calls, won't this lead to a new kernel call every posterior variance call unless `fast_pred_var` is on?

gpytorch/variational/variational_strategy.py
```diff
@@ -63,7 +63,7 @@ def prior_distribution(self):
         """
         out = self.model.forward(self.inducing_points)
         res = MultivariateNormal(
-            out.mean, out.lazy_covariance_matrix.evaluate_kernel().add_jitter()
+            out.mean, out.lazy_covariance_matrix.add_jitter()
```
jacobrgardner: Same perf question here without the evaluation cache

jacobrgardner:
(Closing and reopening so Travis runs)

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
gpleiss: using cached annotation

gpytorch/models/exact_gp.py
gpleiss: using cached annotation

gpleiss:
@jacobrgardner fixed

PR #500, opened 2019-02-05
Wrapper around cholesky to automatically add jitter if matrix is not p.d.

Balandat:
`psd_safe_cholesky` is a simple wrapper around `torch.cholesky` that adds jitter if the matrix is not p.d.. If computing `torch.cholesky` on that still fails a `RuntimeError` is raised. This also hackfixes the  newly discovered pytorch bug: https://github.com/pytorch/pytorch/issues/16780

### DIFF
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -3,6 +3,7 @@
 import torch
 
 from .. import settings
+from ..utils.cholesky import psd_safe_cholesky
 from ..utils.memoize import cached
 from .block_lazy_tensor import BlockLazyTensor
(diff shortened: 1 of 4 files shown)


Balandat:
Some context: This is nice to have for optimizing q-batch acquisition functions in BayesOpt, where as the solution improves we end up with points close to each other, which can lead to degenerate posteriors. This currently either errors out or silently produces NaNs on the GPU (see bug).

Balandat:
Not sure about the test failures here things ran fine locally - I probably messed something up staging this.

Balandat:
hmm, why doesn't this run Travis on the recent commit?

Balandat:
Since we need this soon I rebased this on master.

PR #502, opened 2019-02-06
Use torch native triu_indices

Balandat:
Not sure if this is part of the latest stable release,let's see what Travis says...

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -8,7 +8,6 @@ from ..module import Module
 from .. import settings
 from ..utils.transforms import _get_inv_param_transform
 from torch.nn.functional import softplus
-from numpy import triu_indices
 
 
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Hm, I think the pdist function is actually unused now that I switched everything to use the JIT code. I found the pdist code to actually be quite a bit slower

Balandat:
if it's unused let's kill it

vishwakftw:
@Balandat `tril_indices` are `triu_indices` not part of the latest release. I believe this will have to wait until v1.1.

Balandat:
Argh, not a fan of bugfix-only releases.... 

Anyway, if this is unused as @jacobrgardner indicated we should just scrap it altogether.

gpleiss:
I can confirm that we no longer use `__pdist_dist`. I'll remove it.

Issue #503, opened 2019-02-07
Fast pred var? More like 1/3 as fast pred var


KeAWang:
It seems that after commit 0ff64169766bd22d1d8cae35ef0ab3f0d6f16e2e, fast_pred_var is now 3x slower.

Using the following code
```
model.eval()
likelihood.eval()
import time

for i in range(10):
    start = time.time()
    with gpytorch.settings.fast_pred_var():
        preds = model(test_x)
    print(time.time() - start)
```
I get 
```
0.05411267280578613
0.0015854835510253906
0.0012111663818359375
0.0011920928955078125
0.0011920928955078125
0.0011932849884033203
0.0012063980102539062
0.0011904239654541016
0.0011909008026123047
0.00119781494140625
```
from the previous commit a1e8bcc26a4b432776620b242023adb7acf206e3.

On the other hand, I get
```
0.05980277061462402
0.003302335739135742
0.0028259754180908203
0.0028007030487060547
0.0028228759765625
0.002866983413696289
0.0028259754180908203
0.0033342838287353516
0.0032806396484375
0.003268718719482422
```
from the breaking commit 0ff64169766bd22d1d8cae35ef0ab3f0d6f16e2e.

The code used to train the model is taken from the official fast_pred_var notebook:

```
import math
import torch
import gpytorch
from matplotlib import pyplot as plt

# Make plots inline
%matplotlib inline

import urllib.request
import os.path
from scipy.io import loadmat
from math import floor

if not os.path.isfile('skillcraft.mat'):
    print('Downloading \'skillcraft\' UCI dataset...')
    urllib.request.urlretrieve('https://drive.google.com/uc?export=download&id=1xQ1vgx_bOsLDQ3RLbJwPxMyJHW7U9Eqd', 'skillcraft.mat')
    
data = torch.Tensor(loadmat('skillcraft.mat')['data'])
X = data[:, :-1]
X = X - X.min(0)[0]
X = 2 * (X / X.max(0)[0]) - 1
y = data[:, -1]

# Use the first 80% of the data for training, and the last 20% for testing.
train_n = int(floor(0.4*len(X)))

train_x = X[:train_n, :].contiguous().cuda()
train_y = y[:train_n].contiguous().cuda()

test_x = X[train_n:, :].contiguous().cuda()
test_y = y[train_n:].contiguous().cuda()

class GPRegressionModel(gpytorch.models.ExactGP):
        def __init__(self, train_x, train_y, likelihood):
[...]

Issue #406, opened 2018-11-29
Add performance tests, memory tests


jacobrgardner:
It'd be great to have a few performance tests for exact GPs and maybe a few smoke tests for memory. 

While I don't think these could be reliably run by our CI, it'd be nice to know when running unit tests locally whether a change inadvertently affects the amount of data we can train an exact GP on or substantially slows things down.

I believe that the issues that led to #404 being necessary are sufficient evidence this should be done sooner rather than later :-).

Thoughts on how to best do this?

PR #507, opened 2019-02-07
Fix slowdown of covar_dist

KeAWang:
This commit (mostly) fixes issue #503. Now we cache the distance_module
the first time covar_dist is called. This way the JIT no longer
recompiles the module every time.

Unfortunately, it is still about 100us slower than what it was before
the breaking commit 0ff6416, with 0.0013s per forward pass instead of
0.0012s (about 8% slower).

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -157,6 +157,7 @@ class Kernel(Module):
                     "lengthscale_prior", lengthscale_prior, lambda: self.lengthscale, lambda v: self._set_lengthscale(v)
                 )
 
+        self.distance_module = None
         # TODO: Remove this on next official PyTorch release.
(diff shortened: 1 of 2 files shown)


test/examples/test_simple_gp_regression.py
```diff
@@ -374,6 +374,14 @@ def test_posterior_latent_gp_and_likelihood_fast_pred_var(self, cuda=False):
 
             self.assertLess(torch.max(var_diff / noise), 0.05)
 
+            if cuda:
```
gpleiss: Tests like these are guaranteed to be flakey. Let's remove this for now, and address it when we tackle performance tests (#406)

gpleiss:
@KeAWang remove the timing test and then I'll merge

PR #510, opened 2019-02-07
Stop cholesky_solve warnings on pytorch-nightly

gpleiss:
This works on both pytorch stable and nightly. It creates a utility function that chooses the correct function to call (`torch.cholesky_solve` if avaialable, otherwise `torch.potrs`).

Also removes some stale tridiagonal cholesky code that is no longer used (and is very slow).


gpytorch/models/exact_prediction_strategies.py
```diff
@@ -141,10 +142,10 @@ def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_
         small_system_rhs = targets - fant_mean - fant_train_covar.matmul(self.mean_cache)
         # Schur complement of a spd matrix is guaranteed to be positive definite
         if small_system_rhs.requires_grad or schur_complement.requires_grad:
-            # TODO: Delete this part of the if statement when PyTorch implements potrs derivative.
+            # TODO: Delete this part of the if statement when PyTorch implements cholesky_solve derivative.
```
Balandat: Let's make an issue for this and reference it here.

gpleiss: Will do

PR #511, opened 2019-02-09
Add context manager for selecting min mem cuda device

Balandat:
This is helpful for avoiding OOMs when running a bunch of cuda tests in parallel on a multi-GPU machine.

This is just a proof of concept, if this seems reasonable I'll add it to all of the cuda tests.

### DIFF
--- a/test/_utils.py
+++ b/test/_utils.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python3
 
+from contextlib import contextmanager
+from typing import Generator
+
 import torch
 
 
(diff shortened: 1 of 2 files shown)


test/_utils.py
```diff
@@ -17,3 +20,18 @@ def approx_equal(self, other, epsilon=1e-4):
             "Size mismatch between self ({self}) and other ({other})".format(self=self.size(), other=other.size())
         )
     return torch.max((self - other).abs()) <= epsilon
+
+
+def get_cuda_max_memory_allocations() -> int:
```
jacobrgardner: Is the typing Python 3.6 compatible? We don't yet require 3.7.

jacobrgardner: Other than that, this looks good to me!

Balandat: yeah typing is py36+

gpleiss: I think we have to be py3.5 compatible for the docs, but Travis isn't complaining...
I'll merge, but let's check that we haven't destroyed the docs in the process :)

Balandat:
Shooting from the hip there are we? This was meant to be a draft PR for visibility. Still need to update all the other cuda tests. But hey let’s test in prod :)

gpleiss:
Oooops my bad! At least it only affects the unit tests 😬

PR #492, opened 2019-01-31
Arbitrary number of batch dimensions for LazyTensors

gpleiss:
## Major refactors
- [x] Refactor `_get_indices` from all LazyTensors
- [x] Simplify `_getitem` to handle all cases - including tensor indices
- [x] Write efficient `_getitem` for (most) LazyTensors
  - [x] `CatLazyTensor`
  - [x] `BlockDiagLazyTensor`
  - [x] `ToeplitzLazyTensor`
- [x] Write efficient `_get_indices` for all LazyTensors
- [x] Add a custom `_expand_batch` method for certain LazyTensors
- [x] Add a custom `_unsqueeze_batch` method for certain LazyTensors
- [x] BlockDiagLazyTensor and SumBatchLazyTensor use an explicit batch dimension (rather than implicit one) for the block structure. Also they can sum/block along any batch dimension.
- [x] Custom `_sum_batch` and `_prod_batch` methods
  - [x] `NonLazyTensor`
  - [x] `DiagLazyTensor`
  - [x] `InterpolatedLazyTensor`
  - [x] `ZeroLazyTensor`

## New features
- [x] LazyTensors now handle multiple batch dimensions
- [x] LazyTensors have `squeeze` and `unsqueeze` methods
- [x] Replace `sum_batch` with `sum` (can accept arbitrary dimensions)
- [x] Replace `mul_batch` with `prod` (can accept arbitrary dimensions)
- [x] `LazyTensor.mul` now expects a tensor of size `*constant_size, 1, 1` for constant mul. (More consistent with the Tensor api).
- [x] Add broadcasting capabilities to remaining LazyTensors

## Tests
- [x] Add MultiBatch tests for all LazyTensors
- [x] Add unittetsts for BlockDiagLazyTensor and SumBatchLazyTensor using any batch dimension for summing/blocking
- [x] Add tests for `sum` and `prod` methods
- [x] Add tests for constant mul
- [x] Add tests for permuting dimensions
- [x] Add tests for LazyEvaluatedKernelTensor

## Miscelaneous todos (as part of the whole refactoring process)
- [x] Ensure that `InterpolatedLazyTensor.diag` didn't become more inefficient
- [x] Make CatLazyTensor work on batch dimensions
- [x] Add to LT docs that users might have to overwrite `_getitem`, `_get_indices`, `_unsqueeze_batch`, `_expand_batch`, and `transpose`.
- [x] Fix #573 

## Details

The new `__getitem__` method reduces all possible indices to two cases:
[...]

PR #517, opened 2019-02-13
Custom RBF and Matern Functions

KeAWang:
This pull request tries to reduce the memory requirements of a RBFKernel and a MaternKernel by introducing a custom backward function when the lengthscales require grad.
- [x] Implement RBFCovariance Function. Fallback to RBFKernel when using ARD or when x1, x2 require grad
- [x] Implement MaternCovariance Function. Fallback to MaternKernel when using ARD or when x1, x2 require grad
- [x] Add unittests for forward and backward

ARD is not yet implemented because I'm not sure if there's a better way to do it than to create a distance matrix for each dimension (resulting in a 3D Tensor) since we need gradients of a 2D Tensor with respect to each of the lengthscales.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ GPyTorch is primarily maintained by:
 
 <img width="300" src=https://brand.cornell.edu/assets/images/downloads/logos/cornell_logo_simple/cornell_logo_simple.svg alt="Cornell Logo" />
 
-We would like to thank our other contributors including (but not limited to) Eytan Bakshy, David Arbour, Ruihan Wu, Bram Wallace, Sam Stanton, and Jared Frank.
+We would like to thank our other contributors including (but not limited to)  David Arbour, Eytan Bakshy, Jared Frank, Sam Stanton, Bram Wallace, Ke Alexander Wang, Ruihan Wu.
 
 ## Acknowledgements
 Development of GPyTorch is supported by funding from [Facebook](https://research.fb.com/), the [Bill and Melinda Gates Foundation](https://www.gatesfoundation.org/), the [National Science Foundation](https://www.nsf.gov/), and [SAP](https://www.sap.com/index.html).
--- a/gpytorch/kernels/rbf_kernel.py
+++ b/gpytorch/kernels/rbf_kernel.py
@@ -3,6 +3,7 @@
 from .kernel import Kernel
 from torch.nn.functional import softplus
 import torch
+from ..functions import RBFCovariance
 
 
 @torch.jit.script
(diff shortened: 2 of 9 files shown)


Balandat:
Haven't really looked into this, just wanted to make a side comment here: In light of #492, we'll want to start support arbitrary batch shapes also in the kernels - so I guess this is a call to be cognizant of that, and to make sure the code will be easily adaptable to these coming changes.

KeAWang:
Fixed. Thanks for pointing that out. Also, I noticed that `MaternKernel.forward` might not actually be arbitrary batch size compatible because how it computes the mean will mess up the broadcasting. I added a commit that fixes that.

I'm worried that there might be a lot of other code that assume there's only one batch dimension but we don't know about them because our tests all assume there's only batch dimension. @jacobrgardner @gpleiss ?

gpleiss:
@KeAWang #492

gpleiss:
I'd rather not merge in this PR now - I think we should test against this PR for a while to see if we can make further memory gains.

Balandat:
> I'm worried that there might be a lot of other code that assume there's only one batch dimension but we don't know about them because our tests all assume there's only batch dimension.

Yes, I think outside of the LazyTensor refactor most other code assumes a single batch dimension, this will all need to be changed. Most of the hard work should be done in @gpleiss's refactor though.

KeAWang:
Here's a quick memory usage benchmark from the output of `torch.cuda.max_memory_allocated` on 10000 1 dimensional training points:

Using RBFKernel on master:
   after forward: 1.2 GB
   after backward: 2.1 GB
Using RBFCovariance:
   after forward: 1.2 GB
   after backward: 1.6 GB
Using MaternKernel (nu=1/2) on master:
   after forward: 2.4 GB
   after backward: 2.8 GB
Using MaternFunction (nu=1/2):
   after forward: 1.2 GB
   after backward: 1.6 GB
Using MaternKernel (nu=3/2) on master:
   after forward: 2.8 GB
   after backward: 3.6 GB
Using MaternFunction (nu=3/2):
   after forward: 1.2 GB
   after backward: 1.6 GB
Using MaternKernel (nu=5/2) on master:
   after forward: 3.2 GB
   after backward: 4.4 GB
Using MaternCovariance (nu=5/2):
   forward: 1.6 GB
   backward: 1.6 GB

README.md
```diff
@@ -92,7 +92,7 @@ GPyTorch is primarily maintained by:
 
 <img width="300" src=https://brand.cornell.edu/assets/images/downloads/logos/cornell_logo_simple/cornell_logo_simple.svg alt="Cornell Logo" />
 
-We would like to thank our other contributors including (but not limited to) Eytan Bakshy, David Arbour, Ruihan Wu, Bram Wallace, Sam Stanton, and Jared Frank.
+We would like to thank our other contributors including (but not limited to)  David Arbour, Eytan Bakshy, Jared Frank, Sam Stanton, Bram Wallace, Ke Alexaner Wang, Ruihan Wu.
```
jacobrgardner: @KeAWang At least spell your name right :-)

jacobrgardner:
@KeAWang Can you make sure to keep this branch up to date with any changes you make to the kernels on the `implicit_kernel` branch? It'll probably be cleaner to merge those features separately if possible. If not, we can do that whole branch at once I suppose.

README.md
KeAWang: Oops! This is embarrassing...

KeAWang:
Yep I've been pushing to this remote branch and then merging into the `implicit_kernel` branch

KeAWang:
This should be ready for review now

gpytorch/kernels/rbf_kernel.py
```diff
@@ -94,7 +95,18 @@ def __init__(
...
+        if (
+            x1.requires_grad
+            or x2.requires_grad
+            or (self.ard_num_dims is not None and self.ard_num_dims > 1)
+            or diag
+        ):
+            x1_ = x1.div(self.lengthscale)
+            x2_ = x2.div(self.lengthscale)
+            diff = self._covar_dist(x1_, x2_, square_dist=True, diag=diag, **params)
+            return postprocess_rbf(diff)
```
jacobrgardner: @KeAWang Why does postprocess get called out of the JIT function now? The purpose of having this method at all was  was so that the full computation would be done inside a JIT script.

If we're not going to have postprocess scripts and only do the distance code in a JIT script, then the `dist_postprocess_func` arg needs to be removed everywhere and the function inlined here.

jacobrgardner:
This looks good to me other than my comment above about the distance postprocessing functions.

KeAWang:
Hm I'm not sure why this was there. It might have been left over from my earlier implementation of RBFCovariance. I reverted this change but now `test_fantasy_updates` mysteriously fails...

KeAWang:
Fixed.

PR #519, opened 2019-02-19
Improve speed, stability of woodbury for preconditioner

gpleiss:
This PR does several things:

- Use lower-level BLAS calls for the woodbury formula. This improves the speed slightly
- Use lower-level BLAS calls for AddedDiagLazyTensor._matmul (the matmul function we use the most out of any LazyTensor)
- Move all woodbury stuff into its own file (we'll probably be reusing it for some future preconditioners and stuff)

The big thing that this PR does:

- Scale the diagonal component of the woodbury formula by its' smallest number. When the diagonal component of woodbury is very small, it's possible that dividing by the diagonal could be numerically unstable. This PR should improve stability by scaling before applying the formula and undoing the scale after.


gpleiss:
@jacobrgardner ready for review

PR #530, opened 2019-02-22
Allow arbitrary batch shapes in MTMVN

Balandat:
Now that the Lazies support it, the MTMVN should as well.


Balandat:
Tests all pass locally.

I guess Travis isn't being kicked off b/c of the following exclusion:
```
branches:
  only:
    - master
```
Should we remove this restriction?

Balandat:
cc @sdaulton

gpleiss:
Looks good to me! If it passes locally then it should be good :)

Issue #537, opened 2019-02-26
Inconsistent representation of MultitaskMultivariateNormal covariance


Balandat:
Consider a 2-task model with 3 points each. The shape of the joint covariance matrix `C` is `6x6`, but the current code seems to assume different structures in different places.  

1. `from_independent_mvns` assumes that `C` has block structure w.r.t. the tasks, i.e. that `C = [[C_a, 0], [0, C_b]]`, where `C_a` and `C_b` are the marginal covariances across the data points of task `a` and `b`, respectively.

2. `variance` and `rsample` assume that `C` has block structure w.r.t. the data points, i.e. that `C[2*i:2*(i+1), 2*i:2*(i+1)]`  is the covariance across tasks for data point `i`.

Clearly, the two are incompatible. This means that the variances and samples generated for a MTMVN generated from `from_independent_mvns` are incorrect.

Balandat:
Here's a (terribly inefficient) function for converting from 1. to 2. for full (non-lazy) tensors.
```
def convert_matrix(task_block_diag_cov, n_tasks):
    n = task_block_diag_cov.shape[-1] // n_tasks
    cout = torch.zeros_like(task_block_diag_cov)
    for t in range(n_tasks):
        for i in range(n):
            for j in range(n):
                val = task_block_diag_cov[..., t*n+i, t*n+j]
                cout[..., t+n_tasks*i, t+n_tasks*j] = val
    return cout
```

Balandat:
It seems like having the option of using both representations internally could make a lot of sense in order to exploit structure.

Balandat:
Resolved via #538

Balandat:
Actually, there is is a bug in #538 - working on a fix of the fix

PR #538, opened 2019-02-26
Add interleaved option to MTMVN

Balandat:
This adds an `interleaved` flag to the MTMVN constructor (defaults to True).

- If True, covariance matrix is interpreted as block-diagonal w.r.t. inter-task covariances for each observation. This is the status quo behavior.
- If False, it is interpreted as block-diagonal w.r.t. inter-observation covariance for each task. This makes it much easier to construct a MTMVN from independent MVNs. It also potentially allows for more efficient computations by exploiting structure more effectively. For instance the internal covariance matrix generated by `from_independent_mvns` is now block-diagonal, so it can be efficiently cholesky-decomposed in batch mode (this is handled inside the LazyTensor implementation).

This also cleans up and streamlines some code, and improves test coverage (in particular, tests cover both data types on both the GPU and CPU).

This addresses #537

### DIFF
--- a/gpytorch/distributions/multitask_multivariate_normal.py
+++ b/gpytorch/distributions/multitask_multivariate_normal.py
@@ -7,7 +7,7 @@ from .multivariate_normal import MultivariateNormal
 
 
 class MultitaskMultivariateNormal(MultivariateNormal):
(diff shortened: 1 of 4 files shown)


Balandat:
Thanks for the quick review.

PR #544, opened 2019-03-01
More CG improvements

gpleiss:
- Normalize the vectors before performing solves
- Rather than adding epsilon to all the divisions, does masking to prevent division by zero errors.


gpleiss:
Test failures should be fixed.

PR #547, opened 2019-03-02
Fix batch_symeig cuda index issue

Balandat:
`type_as` does not retain the cuda device index, so the output for small matrices (cpu computation) will be on the default cuda device rather than the one of the input matrix.


Balandat:
@gpleiss, @jacobrgardner I wouldn't be surprised if there were other places in the code where this `type_as` behavior is relevant.

PR #551, opened 2019-03-05
Add collapse_dims option to get_base_samples

Balandat:
Adds an optional `collapse_dims` arg to get_base_samples of MVN and MTMVN, used to indicate which dimensions of the base samples to collapse to size 1. This useful in order to avoid sampling variance across batches when using `rsample`.

Also prettified docstrings, and fixed an issue with the batch_shape of the MTMVN.


Balandat:
@gpleiss, @jacobrgardner does this seem ok to you (apart from the docs issue)?

Balandat:
On second thought, I don't think we need this. I'll push the fixes to the MTMVN event_shape directly.

Issue #555, opened 2019-03-08
Minor change in docs for settings.max_cg_iterations


harvineet:
As per docs, default value of settings.max_cg_iterations is 20. However, in code it is set to 100.
https://github.com/cornellius-gp/gpytorch/blob/4e671371b1305bc4609475de9579a7535c11f481/gpytorch/settings.py#L284-L291

Apologies for nitpicking :-)

gpleiss:
No worries! Thanks for the catch :)
Could you open up a PR?

harvineet:
Sure.

PR #556, opened 2019-03-08
Setting default max_cg_iterations to 20 as per doc

harvineet:
PR for issue #555

### DIFF
--- a/gpytorch/settings.py
+++ b/gpytorch/settings.py
@@ -288,7 +288,7 @@ class max_cg_iterations(_value_context):
     Default: 20
     """
 
-    _global_value = 100
+    _global_value = 20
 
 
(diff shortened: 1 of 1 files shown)


jacobrgardner:
@harvineet Actually the value is correct and the docs are wrong -- we are moving more towards a model where we set a fairly high max CG iterations and instead rely on a relative residual norm to test convergence.

harvineet:
@jacobrgardner Got it. Will close this.
Let me know if I should edit the docstring instead (optionally, with convergence explanation).

jacobrgardner:
Sure, if you want to make a pass and update the docstrings in settings.py we're open to a PR! It's probably not the only wrong one.

PR #564, opened 2019-03-11
Implement deep Gaussian processes via doubly stochastic VI

jacobrgardner:
This is an implementation and example notebook for deep Gaussian processes where inference and training are performed using the doubly stochastic variational approach introduced in @hughsalimbeni's paper.

I've found the implementation to work quite well across a variety of benchmark datasets with relatively minimal tuning, even when adding additional layers. It is also pretty fast with the CG backend I'd say. 

The only thing that is kind of gross right now is the prediction interface because of the lack of multiple batch dimension support across the package (not just for `LazyTensor` which is almost done): it requires a quick reshape after getting the predictions to be sensible dimensions.

With that being said, given that it may take some time for multi batch support to make its way through all of the variational code (I expect exact GPs will be quick), I wanted to open this now.


gpleiss:
Given that #492 is about to drop (for real this time), let's hold off from merging this until then.

hughsalimbeni:
Nice work! 

Are you including a mean function between layers? In the 2017 paper we used the identity mean function between inner layers (or linear in high dimensions), as this can help quite a bit.

jacobrgardner:
@hughsalimbeni Ah, good suggestion. One of the reasons I'd tagged you in the first place was so that you could suggest if we'd missed anything :-). I'll add an identity mean module and be sure that it's the one we use in the example notebook (assuming it helps performance on these particular datasets).

FWIW, I've been seeing solid performance even without it, in the sense that it fairly convincingly outperforms sparse and exact GPs. A deep GP with small deep kernels also convincingly outperforms a single layer deep kernel GP, which is quite nice.

examples/12_Deep_Gaussian_Processes/index.rst
```diff
@@ -0,0 +1,8 @@
+.. mdinclude:: README.md
+
+.. toctree::
+   :glob:
+   :maxdepth: 1
+   :hidden:
+
+   Deep_Gaussian_Processes.ipynb
```
gpleiss: I guess this should probably be 13_Deep_Gaussian_Processes?

gpytorch/models/deep_gps/deep_gp.py
```diff
@@ -0,0 +1,173 @@
...
+        self,
+        num_samples,
+        noise_prior=None,
+        batch_size=1,
+        param_transform=softplus,
+        inv_param_transform=None,
+        **kwargs
+    ):
+        # TODO: Rewrite to be a general DeepLikelihood wrapper once the remainder of GPyTorch supports multiple
+        # batch dimensions.
```
gpleiss: I think this could be rewritten now pretty easily :)

gpytorch/models/deep_gps/deep_gp.py
```diff
@@ -0,0 +1,173 @@
...
+            last_layer_inputs = last_layer_inputs.expand(
+                self.output_dims,
+                self.num_samples * inputs.size(-2),
+                self.input_dims
+            )
+
+            return AbstractVariationalGP.__call__(self, last_layer_inputs)
+        else:
+            inputs = self._reshape_input(inputs)
+            if inputs.dim() == 4:
```
gpleiss: This can definitely be simplified now :)
Would it make sense for the samples dimension to be the outer dimension?

gpytorch/models/deep_gps/deep_gp.py
```diff
@@ -0,0 +1,173 @@
...
+            - variational_strategy (VariationalStrategy): Strategy for changing q(u) -> q(f) (see other VI docs)
+            - input_dims (int): Dimensionality of input data expected by each GP
+            - output_dims (int): Number of GPs in this layer, equivalent to output dimensionality.
+            - last_layer (bool): True if this is to be the last layer in the deep GP, false otherwise.
+            - num_samples (int): Number of samples to draw from q(f) for returning
+        """
+        super(AbstractDeepGPLayer, self).__init__(variational_strategy)
+        self.input_dims = input_dims
+        self.output_dims = output_dims
+        self.num_samples = num_samples
```
gpleiss: I wonder if `self.num_samples` should come from one of our context managers?

gpytorch/models/deep_gps/deep_gp.py
```diff
@@ -0,0 +1,173 @@
...
+
+        return inputs
+
+    def __call__(self, inputs):
+        """
+        Forward data through this hidden GP layer.
+
+        If the input is 2 dimensional, we pass the input through each hidden GP, resulting in a `h x n` batch
+        Gaussian distribution. We then draw `s` samples from these Gaussians and reshape the result to be
+        `s x n x h` (e.g., h becomes the dimensionality of the output).
```
gpleiss: I wonder - would it make sense for all layers to return a MultivariateNormal distribution object? __call__ would be designed to handle tensor inputs (default behavior) or distribution inputs (draw a sample from them).

gpytorch/models/deep_gps/deep_gp.py
```diff
@@ -0,0 +1,173 @@
...
+
+            if reshape_output:
+                samples = torch.distributions.Normal(mean_qf, std_qf).rsample()
+                samples = samples.view(self.output_dims, num_samples, -1).permute(1, 2, 0)
+            else:
+                samples = torch.distributions.Normal(mean_qf, std_qf).rsample(torch.Size([num_samples]))
+                samples = samples.transpose(-2, -1)
+
+            loss_term = NegativeKLDivergence(self.variational_strategy)
+            self.update_added_loss_term("hidden_kl_divergence", loss_term)
```
gpleiss: Hmmm this is a little dirty... would it be possible for `AbstractDeepGP` to handle the kl terms?

PR #565, opened 2019-03-12
Fix bug in setting training data for ExactGP if training inputs are None

neighthan:
It appears valid for a GP to have its training inputs/labels set to `None`. However, if you call `set_train_data` when the current training inputs are `None`, you'll get an error at `for input, t_input in zip(inputs, self.train_inputs)` (because `None` is not iterable and thus can't be `zip`ped). This PR moves the check for `if strict` up so that, if `strict == False`, the entire checking section is skipped and it doesn't matter whether the training data was `None` or not.

When `strict == True`, the error messages have been improved a bit. Now they'll say both what the expected attribute (e.g. shape) was and what shape was found. To get a more informative error, I also ensured that if the training inputs/targets are `None`, the shape, etc. will be viewed as `None` so you'll still get the expected `RuntimeError` (instead of an error at `zip` that is harder to diagnose).

This passed `flake8` fine, and I think the only errors I got in the tests were for CUDA out of memory.

### DIFF
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -45,7 +45,16 @@ class ExactGP(GP):
         return super(ExactGP, self)._apply(fn)
 
     def set_train_data(self, inputs=None, targets=None, strict=True):
-        """Set training data (does not re-fit model hyper-parameters)"""
+        """
+        Set training data (does not re-fit model hyper-parameters).
+
+        Args:
+            - :attr:`inputs` the new training inputs
+            - :attr:`targets` the new training targets
+            - :attr:`strict`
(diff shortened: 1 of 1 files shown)


neighthan:
I also added a bit of documentation for the arguments of `set_train_data` (since I came to look at the source code to figure out what `strict` would do when I needed it). I didn't specify shapes for `inputs` and `targets` - should these be the same as in `get_fantasy_model`? E.g.

```python
"""
- :attr:`inputs` (Tensor `m x d` or `b x m x d`)
- :attr:`targets` (Tensor `m` or `b x m`)
"""
```

jacobrgardner:
This looks great! Thanks @neighthan

PR #566, opened 2019-03-12
Detach train_train_covar for slow variance computation if requested.

### LINKED ISSUE #554: Repeated posterior evaluation (and backward) with same base_samples raises second backward issue
I am encountering an second backward error when doing repeated posterior evaluations with the same set of base_samples. If I use `detach_test_caches(True)` and `fast_pred_var()`, there is no issue. However, when I do not use `fast_pred_var()` and/or set `detach_test_caches(False)`, then the second backward issue is raised.

Repro:
[gpytorch_base_samples_caching_backward_issue.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/2937770/gpytorch_base_samples_caching_backward_issue.ipynb.txt)

jacobrgardner:
This fixes #554 by correctly detaching `train_train_covar` and ensuring that the pass through the likelihood happens in a `torch.no_grad` context if `gpytorch.settings.detach_test_caches` is on.

cc/ @Balandat @sdaulton

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -995,7 +995,11 @@ class LazyTensor(object):
 
     @property
     def requires_grad(self):
-        return any(arg.requires_grad for arg in tuple(self._args) + tuple(self._kwargs.values()))
+        return any(
+            arg.requires_grad
+            for arg in tuple(self._args) + tuple(self._kwargs.values())
+            if hasattr(arg, "requires_grad")
+        )
 
     @requires_grad.setter
     def requires_grad(self, val):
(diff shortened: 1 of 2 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -995,7 +995,11 @@ def representation_tree(self):
 
     @property
     def requires_grad(self):
-        return any(arg.requires_grad for arg in tuple(self._args) + tuple(self._kwargs.values()))
+        return any(
+            arg.requires_grad
```
Balandat: nit: can do in one go using `getattr(arg, "requires_grad", False)`

Balandat: feel free to ignore this though just a side note

PR #568, opened 2019-03-13
Preconditioner tolerance setting

jacobrgardner:
Adds a setting for controlling the preconditioner convergence tolerance.

### DIFF
--- a/gpytorch/settings.py
+++ b/gpytorch/settings.py
@@ -301,6 +301,16 @@ class cg_tolerance(_value_context):
     _global_value = 0.05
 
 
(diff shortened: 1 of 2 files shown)


gpytorch/settings.py
```diff
@@ -301,6 +301,16 @@ class cg_tolerance(_value_context):
     _global_value = 0.05
 
 
+class preconditioner_tolerance(_value_context):
+    """
+    Relative residual tolerance to use for terminating CG.
+
+    Default: 0.05
```
KeAWang: The docs default doesn't match the value?

Issue #189, opened 2018-07-23
Implement Cylindrical Kernels


darbour:

Balandat:
@darbour, I remember you mentioning you had some draft version of this? Could you surface this in a PR so we know if there are any changes that would be useful to make (see Kernel refactor #266).

PR #577, opened 2019-03-16
[WIP] implement cylindrical kernel

KeAWang:
Implement the Cylindrical Kernel as defined in http://proceedings.mlr.press/v80/oh18a.html. Addresses issue #189.

- [x] Implement the Kernel
- [x] Sanity check on toy regression problem
- [ ] Write unit tests

### DIFF
--- a/gpytorch/kernels/cylindrical_kernel.py
+++ b/gpytorch/kernels/cylindrical_kernel.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+
+import torch
+from .kernel import Kernel
+from ..utils.transforms import _get_inv_param_transform
+from torch.nn.functional import softplus
+from .. import settings
+
+
+class CylindricalKernel(Kernel):
+    r"""
+    Computes a covariance matrix based on the Cylindrical Kernel between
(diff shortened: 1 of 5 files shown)


gpytorch/kernels/cylindrical_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+    def angular_weights(self, value):
+        self._set_param(value, 'angular_weights')
+
+    @property
+    def alpha(self):
+        return self._param_transform(self.raw_alpha)
+
+    @alpha.setter
+    def alpha(self, value):
+        self._set_param(value, 'alpha')
```
jacobrgardner: The setters should apply the inverse transform to the value before initializing the raw parameter in `_set_param`

gpytorch/kernels/cylindrical_kernel.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @property
+    def beta(self):
+        return self._param_transform(self.raw_beta)
+
+    @beta.setter
+    def beta(self, value):
+        self._set_param(value, 'beta')
```
jacobrgardner: same here

PR #579, opened 2019-03-17
Make the CG termination warning more useful by showing residuals

KeAWang:
Right now CG will warn the user if the tolerance is not reached. However the warning can be made more useful by also letting the user know how bad the residual is. 
Sometimes it's important to know if the residual is massive (like 100x the tolerance) or if the residual is only slightly above the tolerance.

### DIFF
--- a/gpytorch/utils/linear_cg.py
+++ b/gpytorch/utils/linear_cg.py
@@ -286,11 +286,13 @@ def linear_cg(
     # Un-normalize
     result.mul_(rhs_norm)
 
-    if not tolerance_reached:
+    if not tolerance_reached and n_iter > 0:
         warnings.warn(
-            "CG terminated before reaching the tolerance specified by gpytorch.settings.cg_tolerance. If performance"
-            " is affected, consider raising the maximum number of CG iterations by running code in"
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Currently GPyTorch is intended to support python 3.6, so we can't use `f'{varname}'` strings internally (fixed this issue in the multi GPU notebook as well).

KeAWang:
But the gpytorch README says >=3.6?

KeAWang:
Removed the f-string

Balandat:
There is part of the where I used type annotations I think? That would require 3.6+.

Given Jake's comment on #580, i don't see any reason to not require 3.6.

jacobrgardner:
Okay, let's switch to python 3.6 but before merging in f-strings we update the readthedocs conf file to build against python 3.6 and get a PR up for that first.

Given that we are using python 3.6, I also think it'd be great to see us start using type annotations more generally throughout the package. This could likely just be done e.g. the same way we got rid of `torch.autograd.Variable` and other legacy torch things, where people add them gradually when they are modifying files.

Balandat:
+1 on type annotations. botorch is 100% typed and it's quite helpful both for readability and actual type checking through mypy or pyre.

PR #581, opened 2019-03-18
Update readthedocs to use python 3.6

jacobrgardner:
I believe this is the minimal change necessary to get readthedocs to build on python 3.6. See http://blog.readthedocs.com/python-36-support/

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,0 +1,4 @@
+build:
+  image: latest
+python:
+  version: 3.6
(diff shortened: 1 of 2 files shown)


gpleiss:
This didn't launch CI? Also, can we build ReadTheDocs on this branch and confirm that it still works?

gpleiss:
Also you should update travis to run the docs build against 3.6

readthedocs.yml
```diff
@@ -0,0 +1,4 @@
+build:
+  image: latest
+python:
+  version: 3.6
```
Balandat: looks like so is complaining about no empty line at end of file

PR #584, opened 2019-03-18
Fix linear kernel parameters

### LINKED ISSUE #582: Linear kernel's variance prior is broken
Repro: 
[linear_kernel_prior_issue.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/2979717/linear_kernel_prior_issue.ipynb.txt)

sdaulton:
Fix call to `register_prior` and clean up unused arguments. Fixes #582

### DIFF
--- a/gpytorch/kernels/linear_kernel.py
+++ b/gpytorch/kernels/linear_kernel.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import torch
+import warnings
 from .kernel import Kernel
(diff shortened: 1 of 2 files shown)


jacobrgardner:
These change are all definitely correct but are breaking in the sense that models won't load and existing code using `LinearKernel` that passes a `num_dimensions` argument won't run anymore, which is the reason I didn't do this clean up when I made my hotfix.

Maybe let's soft deprecate the `num_dimensions` argument and `offset` parameter so that setting them works but raises a deprecation warning?

PR #585, opened 2019-03-19
Use Cholesky for small inv_matmul calls

jacobrgardner:
This PR changes `LazyTensor.inv_matmul` to evaluate and use Cholesky by default for matrices less than the setting size (currently 256 by default).

Here's where I'm coming from:
- On the CPU with matrices this small, Cholesky is still a bit faster than CG.
- With this little data, a GPU is still slower than a CPU (up until around 300 data points in my testing).
- This PR is independent of @gpleiss's proposed `use_cholesky` flag that will be useful for debugging. I think this is just the right thing to do _by default_ in the very small data regime.
- We should try to use the best tool for any given linear algebra task.

For training where we don't need as low tolerances, CG actually is still faster than Cholesky even in this tiny data regime. This is why I changed `inv_matmul` and not `inv_quad_logdet`.

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -794,6 +794,17 @@ class LazyTensor(ABC):
                     )
                 )
 
+        if self.size(-1) < settings.max_cholesky_numel.value() and left_tensor is None:
+            squeeze_res = False
+            if right_tensor.dim() == 1:
+                # torch.potrs doesn't let you run with vectors why
+                right_tensor = right_tensor.unsqueeze(-1)
+                squeeze_res = True
+            res = torch.potrs(right_tensor, self.evaluate().cholesky(), upper=False)
+            if squeeze_res:
+                res = res.squeeze(-1)
+            return res
+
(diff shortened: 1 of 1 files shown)


gpleiss:
So I totally agree with this, but I think I have a better interface on a branch that I'd like to propose.

I implemented a `_cholesky` and `_cholesky_solve` method for LazyTensors. It defaults to using `self.evaluate().cholesky()`, but does smart things for DiagLazyTensor and BlockDiagLazyTensor.

jacobrgardner:
@gpleiss Okay, I'll close this and let you handle a better interface for it!

gpleiss:
See #586 . Still need to implement the functionality you added here, but it creates a Cholesky infrastructure.

Balandat:
> For training where we don't need as low tolerances, CG actually is still faster than Cholesky even in this tiny data regime. This is why I changed inv_matmul and not inv_quad_logdet.

Just a note on this: If we use LBFGS or other quasi-Newton methods that perform line search, we will probably want to have low tolerances for inv_quad_logdet as well.

PR #586, opened 2019-03-19
Better Cholesky methods for use in small data regimes and for debugging

gpleiss:
This PR does a few things:

- [x] Introduces the `_cholesky`, `_cholesky_solve`, `_root_decomposition`, and `_root_inv_decomposition` private methods for LazyTensors. This makes it easy to define efficient Cholesky decompositions without having to re-write all the logic of `
- [x] Use Cholesky for `inv_matmul` (on by default for small matrices)
- [x] Use Cholesky for `inv_quad_logdet` (on by default for small matrices)
- [x] Make sure cholesky option is possible for VariationalStrategies

Cholesky is faster and more efficient than CG with `n <= 300` (see #585)

### DIFF
--- a/gpytorch/functions/_inv_matmul.py
+++ b/gpytorch/functions/_inv_matmul.py
@@ -6,7 +6,7 @@ from .. import settings
 
 
 def _solve(lazy_tsr, rhs, preconditioner):
-    if settings.fast_computations.solves.off() or lazy_tsr.matrix_shape.numel() <= settings.max_cholesky_numel.value():
+    if settings.fast_computations.solves.off() or lazy_tsr.size(-1) <= settings.max_cholesky_size.value():
         return lazy_tsr._cholesky()._cholesky_solve(rhs)
     else:
         return lazy_tsr._solve(rhs, preconditioner)
--- a/gpytorch/functions/_inv_matmul.py
+++ b/gpytorch/functions/_inv_matmul.py
@@ -5,6 +5,13 @@ from torch.autograd import Function
 from .. import settings
 
 
+def _solve(lazy_tsr, rhs, preconditioner):
+    if settings.fast_computations.solves.off() or lazy_tsr.matrix_shape.numel() <= settings.max_cholesky_numel.value():
+        return lazy_tsr._cholesky()._cholesky_solve(rhs)
+    else:
+        return lazy_tsr._solve(rhs, preconditioner)
+
+
 class InvMatmul(Function):
     def __init__(self, representation_tree, has_left=False):
         self.representation_tree = representation_tree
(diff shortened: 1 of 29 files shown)


gpytorch/functions/_inv_matmul.py
```diff
@@ -5,6 +5,13 @@
 from .. import settings
 
 
+def _solve(lazy_tsr, rhs, preconditioner):
+    if settings.fast_computations.solves.on() or lazy_tsr.matrix_shape.numel() <= settings.max_cholesky_numel.value():
```
jacobrgardner: `lazy_tsr.matrix_shape.numel()` will be 256 when `self` is 16x16 -- I think we should compare `lazy_tsr.size(-1)`

gpytorch/lazy/lazy_tensor.py
```diff
@@ -379,6 +379,33 @@ def _approx_diag(self):
...
+        """
+        (Optional) Assuming that `self` is a Cholesky factor, computes the cholesky solve
+
+        ..note::
+            This method is used as an internal helper. Calling this method directly is discouraged.
+
+        Returns:
+            (LazyTensor) Cholesky factor
+        """
+        return cholesky_solve(rhs.double(), self.evaluate().double()).to(self.dtype)
```
jacobrgardner: Unfortunately, using `cholesky_solve` is dangerous and will prevent us from getting gradients through predictions because it doesn't have a gradient implemented. We might need to manually do this with two `torch.trtrs` solves, which does support gradients.

jacobrgardner: Or better, use `trtrs` when either input requires grad, and use `potrs` otherwise (I do this e.g. for the fantasy cache updates in `ExactPredictionStrategy`)

gpleiss: @jacobrgardner the way everything is written right now - `_cholesky_solve` is never called directly, but only as part of the forward/backward methods of the `InvMatmul` function. So we do get a backward pass.

It'll be the same for `InvQuadLogDet` as well.

gpytorch/functions/_inv_matmul.py
gpleiss: Sure - I'm just using the existing `max_cholesky_numel` flag that existed for `root_decomposition`.

gpleiss:
@jacobrgardner @Balandat ready for review

test/_base_test_case.py
```diff
@@ -0,0 +1,48 @@
...
+                torch.manual_seed(seed)
+                if torch.cuda.is_available():
+                    torch.cuda.manual_seed_all(seed)
+                random.seed(seed)
+
+    def tearDown(self):
+        if hasattr(self, "rng_state"):
+            torch.set_rng_state(self.rng_state)
+
+    def assertAllClose(self, tensor1, tensor2, rtol=1e-4, atol=1e-5, equal_nan=False):
```
jacobrgardner: How does this differ materially from `torch.allclose`?

gpytorch/settings.py
```diff
@@ -325,14 +354,14 @@ class _use_eval_tolerance(_feature_flag):
     _state = False
 
 
-class max_cholesky_numel(_value_context):
+class max_cholesky_size(_value_context):
     """
-    If the number of elements of a LazyTensor is less than `max_cholesky_numel`,
+    If the size of of a LazyTensor is less than `max_cholesky_size`,
     then the `root_decomposition` of LazyTensor will use Cholesky rather than Lanczos.
```
jacobrgardner:
```suggestion
     then `root_decomposition` and `inv_matmul` of LazyTensor will use Cholesky rather than Lanczos/CG.
```

test/_base_test_case.py
gpleiss: It gives nicer error messages

Issue #531, opened 2019-02-22
Support general batch shapes throughout


Balandat:
#492 goes a long way to making LazyTensors compatible with arbitrary batch shapes. We also need to translate this to user-facing objects such as Kernels etc.

- [x] Kernels (#589)
- [x] Likelihoods
- [x] Means
- [x] Variational objects
- [x] Prediction strategies (models)
- [x] Grid interpolation
- [x] Distributions? (double check that all operations are multi-batch)

Balandat:
Here is some *very* early draft of how this could look like: bdc4277e702cf5c85243f379f42f90c3fa7032ea

Getting this to work will require substantial changes throughout the codebase. @gpleiss, @jacobrgardner do you have an estimate for a timeframe within which this could happen?

PR #589, opened 2019-03-20
Support for multiple batch dimensions in *base* kernels

jacobrgardner:
First of a few PRs to support multiple batch dimensions throughout GPyTorch. This specifically adds support to most of the base kernels.
- Simplifies kernel constructors to not repeat work done in the Kernel constructor. This includes collapsing unused args in to **kwargs. The argument for doing this is that the kernels were extremely inconsistent about which super args they explicitly took in the first place. If we want them expanded back out, there should at least be consistency.
- Nearly all base kernels + ScaleKernel now support multiple batch dimension inputs.
- `batch_size` kernel argument has been deprecated in favor of `batch_shape`, which takes a `torch.Size` and is used to make multibatch lengthscales etc.
- Added a battery of unit tests for batch and multibatch kernel calling to the `BaseKernelTestCase` class.

Todo:
- [x] Finish updating docstrings

### DIFF
--- a/gpytorch/kernels/cosine_kernel.py
+++ b/gpytorch/kernels/cosine_kernel.py
@@ -21,9 +21,9 @@ class CosineKernel(Kernel):
     where :math:`p` is the periord length parameter.
 
     Args:
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (torch.Size, optional):
             Set this if you want a separate lengthscale for each
-            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `1`
+            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `torch.Size([1])`
         :attr:`active_dims` (tuple of ints, optional):
             Set this if you want to compute the covariance of only a few input dimensions. The ints
             corresponds to the indices of the dimensions. Default: `None`.
@@ -40,7 +40,7 @@ class CosineKernel(Kernel):
 
     Attributes:
         :attr:`period_length` (Tensor):
-            The period length parameter. Size = `batch_size x 1 x 1`.
+            The period length parameter. Size = `*batch_shape x 1 x 1`.
 
     Example:
         >>> x = torch.randn(10, 5)
@@ -51,25 +51,17 @@ class CosineKernel(Kernel):
         >>> # Batch: Simple option
         >>> covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.CosineKernel())
         >>> # Batch: different lengthscale for each batch
-        >>> covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.CosineKernel(batch_size=2))
+        >>> covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.CosineKernel(batch_shape=torch.Size([2])))
         >>> covar = covar_module(x)  # Output: LazyVariable of size (2 x 10 x 10)
     """
 
-    def __init__(
-        self,
-        active_dims=None,
-        batch_size=1,
-        period_length_prior=None,
-        eps=1e-6,
-        param_transform=softplus,
-        inv_param_transform=None,
-        **kwargs
-    ):
-        super(CosineKernel, self).__init__(
-            active_dims=active_dims, param_transform=param_transform, inv_param_transform=inv_param_transform
+    def __init__(self, period_length_prior=None, **kwargs):
+        super(CosineKernel, self).__init__(**kwargs)
+
+        self.register_parameter(
+            name="raw_period_length", parameter=torch.nn.Parameter(torch.zeros(*self.batch_shape, 1, 1))
         )
-        self.eps = eps
-        self.register_parameter(name="raw_period_length", parameter=torch.nn.Parameter(torch.zeros(batch_size, 1, 1)))
+
         if period_length_prior is not None:
             self.register_prior(
                 "period_length_prior",
--- a/gpytorch/kernels/cosine_kernel.py
+++ b/gpytorch/kernels/cosine_kernel.py
@@ -3,7 +3,6 @@
 import math
 import torch
 from .kernel import Kernel
-from torch.nn.functional import softplus
 
 
 class CosineKernel(Kernel):
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -3,8 +3,6 @@
 import torch
 from .kernel import Kernel
 from ..lazy import DiagLazyTensor, InterpolatedLazyTensor, PsdSumLazyTensor, RootLazyTensor
-from torch.nn.functional import softplus
-from ..utils.deprecation import _deprecate_kwarg_with_transform
 
 
 class IndexKernel(Kernel):
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -50,7 +50,7 @@ class IndexKernel(Kernel):
     ):
         if rank > num_tasks:
             raise RuntimeError("Cannot create a task covariance matrix larger than the number of tasks")
-        super(IndexKernel, self).__init__(**kwargs)
+        super().__init__(**kwargs)
 
         self.register_parameter(
             name="covar_factor", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks, rank))
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -4,6 +4,7 @@ import torch
 from .kernel import Kernel
 from ..lazy import DiagLazyTensor, InterpolatedLazyTensor, PsdSumLazyTensor, RootLazyTensor
 from torch.nn.functional import softplus
+from ..utils.deprecation import _deprecate_kwarg_with_transform
 
 
 class IndexKernel(Kernel):
@@ -22,7 +23,7 @@ class IndexKernel(Kernel):
     Args:
         :attr:`num_tasks` (int):
             Total number of indices.
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (torch.Size, optional):
             Set if the MultitaskKernel is operating on batches of data (and you want different
             parameters for each batch)
         :attr:`rank` (int):
@@ -42,14 +43,21 @@ class IndexKernel(Kernel):
             The element-wise log of the :math:`\mathbf v` vector.
     """
 
-    def __init__(self, num_tasks, rank=1, batch_size=1, prior=None, param_transform=softplus, inv_param_transform=None):
+    def __init__(
+        self,
+        num_tasks,
+        rank=1,
+        prior=None,
+        **kwargs
+    ):
         if rank > num_tasks:
             raise RuntimeError("Cannot create a task covariance matrix larger than the number of tasks")
-        super(IndexKernel, self).__init__(param_transform=param_transform, inv_param_transform=inv_param_transform)
+        super(IndexKernel, self).__init__(**kwargs)
+
         self.register_parameter(
-            name="covar_factor", parameter=torch.nn.Parameter(torch.randn(batch_size, num_tasks, rank))
+            name="covar_factor", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks, rank))
         )
-        self.register_parameter(name="raw_var", parameter=torch.nn.Parameter(torch.randn(batch_size, num_tasks)))
+        self.register_parameter(name="raw_var", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks)))
         if prior is not None:
             self.register_prior("IndexKernelPrior", prior, self._eval_covar_matrix)
 
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -7,6 +7,7 @@ from ..lazy import lazify, delazify, LazyEvaluatedKernelTensor, ZeroLazyTensor
 from ..module import Module
 from .. import settings
 from ..utils.transforms import _get_inv_param_transform
+from ..utils.deprecation import _deprecate_kwarg_with_transform
 from torch.nn.functional import softplus
 
 
@@ -97,9 +98,9 @@ class Kernel(Module):
         :attr:`ard_num_dims` (int, optional):
             Set this if you want a separate lengthscale for each input
             dimension. It should be `d` if :attr:`x1` is a `n x d` matrix.  Default: `None`
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (torch.Size, optional):
             Set this if you want a separate lengthscale for each batch of input
-            data. It should be `b` if :attr:`x1` is a `b x n x d` tensor.  Default: `1`
+            data. It should be `b1 x ... x bk` if :attr:`x1` is a `b1 x ... x bk x n x d` tensor.  Default: `1`
         :attr:`active_dims` (tuple of ints, optional):
             Set this if you want to compute the covariance of only a few input dimensions. The ints
             corresponds to the indices of the dimensions. Default: `None`.
@@ -129,7 +130,7 @@ class Kernel(Module):
         self,
         has_lengthscale=False,
         ard_num_dims=None,
-        batch_size=1,
+        batch_shape=torch.Size([1]),
         active_dims=None,
         lengthscale_prior=None,
         param_transform=softplus,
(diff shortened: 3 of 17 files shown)


gpytorch/kernels/index_kernel.py
```diff
@@ -42,14 +41,21 @@ class IndexKernel(Kernel):
...
+        self,
+        num_tasks,
+        rank=1,
+        prior=None,
+        **kwargs
+    ):
         if rank > num_tasks:
             raise RuntimeError("Cannot create a task covariance matrix larger than the number of tasks")
-        super(IndexKernel, self).__init__(param_transform=param_transform, inv_param_transform=inv_param_transform)
+        super(IndexKernel, self).__init__(**kwargs)
```
Balandat: `super(). __init__(**kwargs)` (py3, also in other places)

gpytorch/kernels/cosine_kernel.py
```diff
@@ -21,9 +20,9 @@ class CosineKernel(Kernel):
     where :math:`p` is the periord length parameter.
 
     Args:
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (torch.Size, optional):
             Set this if you want a separate lengthscale for each
-            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `1`
+            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `torch.Size([1])`
```
gpleiss: Can the default be `torch.Size([])`?

gpytorch/kernels/kernel.py
```diff
@@ -129,7 +130,7 @@ def __init__(
         self,
         has_lengthscale=False,
         ard_num_dims=None,
-        batch_size=1,
+        batch_shape=torch.Size([1]),
```
gpleiss: Default should be `torch.Size([])` I think.

gpytorch/kernels/kernel.py
```diff
@@ -194,11 +201,8 @@ def _set_lengthscale(self, value):
         self.initialize(raw_lengthscale=self._inv_param_transform(value))
 
     def size(self, x1, x2):
-        non_batch_size = (x1.size(-2), x2.size(-2))
-        if x1.ndimension() == 3:
-            return torch.Size((x1.size(0),) + non_batch_size)
-        else:
-            return torch.Size(non_batch_size)
+        non_batch_shape = torch.Size([x1.size(-2), x2.size(-2)])
+        return x1.shape[:-2] + non_batch_shape
```
gpleiss: Just a thought - should we support broadcasting for kernel params?

For example - imagine you want to pass `n x d` data through kernels with `b` different lengthscales. You could construct a kernel with `batch_shape=torch.Size((b,))` and then pass `x` through it to get an output of size `b x n x n`.

gpleiss: In that case, you'd probably want to determine the batch shape using one of the broadcast helpers.

gpleiss: Actually, you probably should determine the batch shape using the broadcast helper anyways, in case x1 and x2 have different batch shapes?

gpytorch/kernels/matern_kernel.py
```diff
@@ -41,7 +40,7 @@ class MaternKernel(Kernel):
         :attr:`ard_num_dims` (int, optional):
             Set this if you want a separate lengthscale for each
             input dimension. It should be `d` if :attr:`x1` is a `n x d` matrix. Default: `None`
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
             Set this if you want a separate lengthscale for each
             batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `1`
```
gpleiss: Doc string needs updathing for the default.

gpytorch/kernels/periodic_kernel.py
```diff
@@ -33,7 +32,7 @@ class PeriodicKernel(Kernel):
         This kernel does not have an ARD lengthscale option.
 
     Args:
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
             Set this if you want a separate lengthscale for each
             batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `1`.
```
gpleiss: Same here.

gpytorch/kernels/rbf_kernel_grad.py
```diff
@@ -19,7 +18,7 @@ class RBFKernelGrad(RBFKernel):
         decorate this kernel with a :class:`gpytorch.kernels.ScaleKernel`.
 
     Args:
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
             Set this if you want a separate lengthscale for each
             batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `1`.
```
gpleiss: Same here.

gpytorch/kernels/rbf_kernel.py
```diff
@@ -35,7 +34,7 @@ class RBFKernel(Kernel):
         :attr:`ard_num_dims` (int, optional):
             Set this if you want a separate lengthscale for each
             input dimension. It should be `d` if :attr:`x1` is a `n x d` matrix. Default: `None`
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
             Set this if you want a separate lengthscale for each
             batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `1`.
```
gpleiss: Same here.

gpytorch/kernels/scale_kernel.py
```diff
@@ -30,7 +28,7 @@ class ScaleKernel(Kernel):
     Args:
         :attr:`base_kernel` (Kernel):
             The base kernel to be scaled.
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
             Set this if you want a separate outputscale for each batch of input data. It should be `b`
             if :attr:`x1` is a `b x n x d` tensor. Default: `1`
```
gpleiss: Same here.

gpytorch/kernels/spectral_mixture_kernel.py
```diff
@@ -19,7 +19,8 @@ class SpectralMixtureKernel(Kernel):
 
         Unlike other kernels,
         * :attr:`ard_num_dums` **must equal** the number of dimensions of the data
-        * :attr:`batch_size` **must equal** the batch size of the data (1 if the data is not batched)
+        * :attr:`batch_shape` **must equal** the batch size of the data (1 if the data is not batched)
```
gpleiss: "`torch.Size([])` if the data is not batched"

gpytorch/kernels/spectral_mixture_kernel.py
```diff
@@ -19,7 +19,8 @@ class SpectralMixtureKernel(Kernel):
...
+        * :attr:`batch_shape` **cannot** contain more than one batch dimension.
```
gpleiss: Is this necessary? We should check for it if so.

gpytorch/kernels/spectral_mixture_kernel.py
```diff
@@ -28,7 +29,7 @@ class SpectralMixtureKernel(Kernel):
         :attr:`ard_num_dims` (int, optional):
             Set this to match the dimensionality of the input.
             It should be `d` if :attr:`x1` is a `n x d` matrix. Default: `1`
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
             Set this if the data is
             batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `1`
```
gpleiss: Same here.

test/kernels/test_matern_kernel.py
```diff
@@ -11,6 +11,9 @@ class TestMaternKernel(unittest.TestCase, BaseKernelTestCase):
     def create_kernel_no_ard(self, **kwargs):
         return MaternKernel(nu=1.5, **kwargs)
 
+    def create_kernel_ard(self, num_dims, **kwargs):
+        return MaternKernel(nu=1.5, ard_num_dims=num_dims, **kwargs)
+
     def test_forward_nu_1_over_2(self):
         a = torch.tensor([4, 2, 8], dtype=torch.float).view(3, 1)
         b = torch.tensor([0, 2], dtype=torch.float).view(2, 1)
```
gpleiss: It might make more sense to have different test classes for each of the `nu` settings.

test/kernels/test_rbf_kernel.py
```diff
@@ -11,6 +11,9 @@ class TestRBFKernel(unittest.TestCase, BaseKernelTestCase):
     def create_kernel_no_ard(self, **kwargs):
         return RBFKernel(**kwargs)
 
+    def create_kernel_ard(self, num_dims, **kwargs):
+        return RBFKernel(ard_num_dims=num_dims, **kwargs)
+
     def test_ard(self):
```
gpleiss: Do we need this test if we have BaseKernelTestCase?

gpytorch/kernels/cosine_kernel.py
jacobrgardner: Not without being breaking or just special casing that default to add a `[1]` anyways.
```python
torch.nn.Parameter(torch.zeros(*self.batch_shape, 1, lengthscale_num_dims)
```
Should by default resolve to `torch.zeros(1, 1, lengthscale_num_dims)`. 

Removing the default leading 1 is a much larger refactor that involves:
1. Not expanding `x` internally to be at least 3D
2. Dealing with the breaking change to model loading.

Since the goal of getting multibatch working doesn't require that, my preference would be to do a second PR for that refactor if we want it.

gpytorch/kernels/rbf_kernel_grad.py
```diff
@@ -19,7 +18,7 @@ class RBFKernelGrad(RBFKernel):
         decorate this kernel with a :class:`gpytorch.kernels.ScaleKernel`.
 
     Args:
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
```
jacobrgardner:
```suggestion
        :attr:`batch_shape` (torch.Size, optional):
```

gpytorch/kernels/rbf_kernel.py
```diff
@@ -35,7 +34,7 @@ class RBFKernel(Kernel):
         :attr:`ard_num_dims` (int, optional):
             Set this if you want a separate lengthscale for each
             input dimension. It should be `d` if :attr:`x1` is a `n x d` matrix. Default: `None`
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
```
jacobrgardner:
```suggestion
        :attr:`batch_shape` (torch.Size, optional):
```

gpytorch/kernels/periodic_kernel.py
```diff
@@ -33,7 +32,7 @@ class PeriodicKernel(Kernel):
         This kernel does not have an ARD lengthscale option.
 
     Args:
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
```
jacobrgardner:
```suggestion
        :attr:`batch_shape` (torch.Size, optional):
```

gpytorch/kernels/matern_kernel.py
```diff
@@ -41,7 +40,7 @@ class MaternKernel(Kernel):
         :attr:`ard_num_dims` (int, optional):
             Set this if you want a separate lengthscale for each
             input dimension. It should be `d` if :attr:`x1` is a `n x d` matrix. Default: `None`
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
```
jacobrgardner:
```suggestion
        :attr:`batch_shape` (torch.Size, optional):
```

gpytorch/kernels/index_kernel.py
jacobrgardner:
```suggestion
        super().__init__(**kwargs)
```

gpytorch/kernels/periodic_kernel.py
jacobrgardner:
```suggestion
            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `torch.Size([1])`.
```

gpytorch/kernels/matern_kernel.py
jacobrgardner:
```suggestion
            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `torch.Size([1])`
```

gpytorch/kernels/rbf_kernel_grad.py
jacobrgardner:
```suggestion
            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `torch.Size([1])`.
```

gpytorch/kernels/spectral_mixture_kernel.py
jacobrgardner: Yes, I couldn't easily get the SM kernel supporting multi batch. I can do a second PR for it, but I also don't think it's mission critical to get 0.3.0 out.

gpytorch/kernels/spectral_mixture_kernel.py
```diff
@@ -28,7 +29,7 @@ class SpectralMixtureKernel(Kernel):
         :attr:`ard_num_dims` (int, optional):
             Set this to match the dimensionality of the input.
             It should be `d` if :attr:`x1` is a `n x d` matrix. Default: `1`
-        :attr:`batch_size` (int, optional):
+        :attr:`batch_shape` (int, optional):
```
jacobrgardner:
```suggestion
        :attr:`batch_shape` (torch.Size([1]), optional):
```

gpytorch/kernels/spectral_mixture_kernel.py
jacobrgardner:
```suggestion
            batch of input data. It should be `b` if :attr:`x1` is a `b x n x d` tensor. Default: `torch.Size([1])`
```

test/kernels/test_matern_kernel.py
jacobrgardner: Reasonable.

test/kernels/test_rbf_kernel.py
jacobrgardner: I think it's nice to have because it doesn't explicitly assume that at least the most basic kernel call is correct. All of the test cases we can write for `BaseKernelTestCase` are basically limited to testing complex kernel calls against simpler ones (e.g., a batch mode call vs filling in the batch kernel matrix in for loops).

gpytorch/kernels/cosine_kernel.py
gpleiss: I would say that we definitely want to do that. But you're right that it could be in a different PR.

We could do a PR that makes Likelihoods multibatch capable, and then do a PR that sets the default batch shape for both of them to `torch.Size([])` and fixes model loading.

gpytorch/kernels/spectral_mixture_kernel.py
gpleiss: Yeah that's not mission critical, but then the kernel should check for that and throw an error.

test/kernels/test_rbf_kernel.py
gpleiss: Could we rename it to something more generic then?

gpytorch/kernels/cosine_kernel.py
gpleiss: I think that the `torch.Size([])` change should be made before 0.3.0 for sure though.

jacobrgardner: Yeah we can do it before 0.3, but I think it'll make the most sense to do it all at once for likelihoods and kernels so we don't have a window on master where lengthscales are `1` by default but noises are `1 x 1 x 1` by default.

gpytorch/kernels/spectral_mixture_kernel.py
jacobrgardner:
```suggestion
        * :attr:`batch_shape` **must equal** the batch size of the data (torch.Size([1]) if the data is not batched)
```

gpytorch/kernels/cosine_kernel.py
gpleiss: Sounds like a plan.

jacobrgardner:
All addressed I think. I'll add a card to the 0.3.0 project to remove the default singleton batch dimension internally.

gpleiss:
Closes part of #531

PR #591, opened 2019-03-21
Change likelihood signature, several Pyro fixes

gpleiss:
# Likelihood updates
- [x] `Likelihood.__call__` now compute either marginal or conditional distribution
  - If the input is a Tensor, it assumes that the input is samples from f(x), and computes the conditional distribution `p(y | f(x))`
  - If the input is a gpytorch.distribution.MultivariateNormal, it assumes that the input is the distribution f(x), and computes the marginal distribution `p(y | x)`
- [x] `Likelihood.forward` now defines the conditional likelihood distributions
- [x] `Likelihood.marginal` now 
  - [x] Custom `marginal` definitions for Gaussian/MultitaskGaussian/Bernoulli likelihoods
  - [x] Default `marginal` returns a batched distribution, where the batch represents samples from the marginal
- [x] `Likelihood.expected_log_prob` is now a auto-defined function (used for variational inference)
- [x] Use Pyro distributions (if available, otherwise use Torch distributions)
- [x] `BernoulliLikelihood` now expects labels in {0, 1}, rather than {-1, 1}
  - [x] Soft deprecation for old-style labels
  - [x] Update classification tests

# Pyro fixes
- [x] Use `pyro.plate` for conditional independence of observations
- [x] Fix `pyro_sample_y` for GaussianLikelihoods

** This will be a breaking change ** for anyone who has defined a custom likelihood. However, most GPyTorch users won't notice any difference.


gpytorch/likelihoods/likelihood.py
```diff
@@ -1,60 +1,134 @@
...
+            raise RuntimeError("Likelihoods expects a MultivariateNormal input to make predictions, got a {}".format(
+                input.__class__.__name__
+            ))
+            return super().__call__(input, *params, **kwargs)
+        # Marginal
+        elif isinstance(input, MultivariateNormal):
+            return self.marginal(input, *params, **kwargs)
+        # Error
+        else:
+            raise RuntimeError(
```
jacobrgardner: This error is inconsistent -- it says we expect either a `MultivariateNormal` or a tensor, but if you call with a tensor we throw a runtime error anyways. Maybe either call forward if `input` is a tensor or fix the error message

gpytorch/likelihoods/likelihood.py
```diff
@@ -1,60 +1,134 @@
...
+            with pyro.plate(name_prefix + ".output_values_plate", function_dist.batch_shape[-1], dim=-1):
+                with pyro.poutine.block():
+                    function_samples = pyro.sample(name_prefix + ".function_values", function_dist)
+                output_dist = self(function_samples, *params, **kwargs)
+                samples = pyro.sample(name_prefix + ".output_values", output_dist, obs=observations)
+                return samples
+
+except ImportError:
+    class Likelihood(_Likelihood):
+        pass
```
jacobrgardner:
```suggestion
        def pyro_sample_outputs(self, *args, **kwargs):
            raise RuntimeError("Failed to import pyro. Is it installed correctly?")
```
More useful than the `AttributeError` this would raise otherwise.

test/examples/test_simple_gp_classification.py
```diff
@@ -16,7 +16,7 @@
 
 def train_data(cuda=False):
     train_x = torch.linspace(0, 1, 10)
-    train_y = torch.sign(torch.cos(train_x * (4 * pi)))
+    train_y = torch.sign(torch.cos(train_x * (4 * pi))).add(1).div(2)
```
jacobrgardner: Update the example notebooks too?

gpytorch/utils/quadrature.py
```diff
@@ -54,10 +55,14 @@ def forward(self, func, gaussian_dists):
...
             - Result of integrating func against each univariate Gaussian in gaussian_dists.
         """
-        mus = gaussian_dists.mean
-        vars = gaussian_dists.variance
+        means = gaussian_dists.mean
+        variances = gaussian_dists.variance
 
-        shifted_locs = torch.sqrt(2.0 * vars).unsqueeze(-1) * self.locations + mus.unsqueeze(-1)
-        res = (1 / math.sqrt(math.pi)) * (func(shifted_locs) * self.weights).sum(-1)
+        locations = _pad_with_singletons(self.locations, num_singletons_before=0, num_singletons_after=means.dim())
```
jacobrgardner: haha oops, I must have not even realized this existed :-)

gpytorch/likelihoods/bernoulli_likelihood.py
```diff
@@ -20,31 +19,27 @@ class BernoulliLikelihood(Likelihood):
...
-        import pyro
-
-        f_samples = variational_dist_f(sample_shape)
-        y_prob_samples = torch.distributions.Normal(0, 1).cdf(f_samples)
-        y_dist = pyro.distributions.Bernoulli(y_prob_samples)
-        pyro.sample(name_prefix + "._training_labels", y_dist.independent(1), obs=y_obs)
+        output_probs = base_distributions.Normal(0, 1).cdf(link)
+        return base_distributions.Bernoulli(probs=output_probs)
+
+    def expected_log_prob(self, observations, function_dist, *params, **kwargs):
```
jacobrgardner: Can we add a `variational_log_prob` method to the base `Likelihood` that calls `expected_log_prob` with a deprecation warning or is this hopeless? We have some custom variational stuff that this is super annoyingly breaking for.

gpytorch/likelihoods/likelihood.py
```diff
@@ -1,60 +1,134 @@
...
+            Distribution object (the marginal distribution, or samples from it)
+        """
+        sample_shape = torch.Size((settings.num_likelihood_samples,))
+        function_samples = function_dist.rsample(sample_shape)
+        return self.forward(function_samples)
+
+    def __call__(self, input, *params, **kwargs):
+        # Conditional
+        if torch.is_tensor(input):
+            raise RuntimeError("Likelihoods expects a MultivariateNormal input to make predictions, got a {}".format(
```
jacobrgardner: On reflection about my comment below, I'm assuming this runtime error just shouldn't be here since it's immediately followed by a call to forward?

gpytorch/likelihoods/bernoulli_likelihood.py
gpleiss: I can do that

gpytorch/likelihoods/likelihood.py
gpleiss: Yes - that was not intentional :)

gpytorch/likelihoods/likelihood.py
gpleiss: Fixed

gpytorch/likelihoods/bernoulli_likelihood.py
gpleiss: It's updated, but will the re-definition of `forward` break your custom variational stuff?

jacobrgardner: No we don't actually use forward for this likelihood, we're never interested in p(y)

gpleiss:
@jacobrgardner comments addressed

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -17,22 +17,24 @@ def __init__(self, noise_covar):
...
-            shape = mean.shape
-        noise_covar = self.noise_covar(*params, shape=shape)
-        full_covar = covar + noise_covar
-        return input.__class__(mean, full_covar)
+            shape = base_shape
+        return self.noise_covar(*params, shape=shape)
 
-    def variational_log_probability(self, input, target):
-        raise NotImplementedError
+    def forward(self, function_samples, *params, **kwargs):
```
Balandat: this would be a good place to start with type hints to make clear that function_samples are tensors and function_dist below is an instance of a gpytorch MultivariateNormal

jacobrgardner: We need to land the rtd yaml file before adding type hints

gpleiss:
So I'm a bit torn on this re-definition of forward (and I'm totally open to reverting), but here's the reasoning:

For almost every non-Gaussian likelihood, there is no closed-form solution for the marginal distribution. In most cases (e.g. for SoftmaxLikelihood, more complex regression likelihoods, etc.), the best you could do is draw samples - by first sampling the GP and then computing the conditional on those samples. So for many likelihoods, we'll only be defining the conditional distribution anyways.

Additionally, a likelihood is typically defined by its conditional distribution `p(y | f)`. So in some sense it seems natural that we should let forward define its conditional distribution (especially for more complex likelihoods) rather than its marginal.

That being said, I would be open to having `forward` remain as the marginal distribution, and introduce a new `conditional` method. `forward` would then have a default implementation.

Thoughts? @Balandat @jacobrgardner

jacobrgardner:
I don't really have strong opinions either way, because `__call__` will do the right thing, and we should discourage people from calling `forward` on modules directly unless they really really know what they want anyways.

PR #593, opened 2019-03-25
Fixed noise GaussianLikelihood

Balandat:
Draft of a super simple fixed noise Gaussian Likelihood that takes noise levels upon instantiation and does not infer any noise levels during fitting. 

Note that the typical gpytorch pattern of using `self.likelihood(model(test_x))` does not make sense in this case. One will have to define a separate likelihood for the output. This separate output likelihood could also be a fixed noise likelihood (if for some reason observation noise levels at the test data were known). It could be a GaussianLikelihood with noise level initialized to some reasonable values (e.g. to the mean of the observation noise of the test data assuming homoskedasticity).


Balandat:
@jacobrgardner this is WIP, but do you have any comments on how this may be used in conjunction with `get_fantasy_strategy` (at least without having to do bunch of instance checks on `self.likelihood`)? Right now this would error out [here](https://github.com/cornellius-gp/gpytorch/blob/f690c25de7713419e95212877be3cde1979ef6c7/gpytorch/models/exact_prediction_strategies.py#L119)

jacobrgardner:
@Balandat I would make three changes:

1. The likelihood takes an `observation_noise` kwarg that, if specified, gets added instead of the fixed noise saved in the likelihood. This gives an interface for making predictions `p(y*)` on test data that has known observation noise, and will also simply fantasy observations.
2. For fantasy observations, we make two changes:
    - Likelihood defines a `get_fantasy_likelihood` similar to `get_fantasy_strategy` and `get_fantasy_model`. The default behavior is a no-op, the behavior in this case is to add the new noise observations to the fixed noise.
     - The call you linked to simply passes along kwargs to that likelihood call, in this case the observation noise for the fantasy point(s).

Balandat:
sounds good, will work on incorporating this.

Balandat:
argh apparently I suck at git

Balandat:
Added the observation_noise kwarg. I was surprised all tests were passing. Then I checked test coverage for likelihoods and it looks pretty dire....

Anyway, lmk if this looks good so far so I can take a stab at the fantasy likelihood

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -26,14 +30,24 @@ def _shaped_noise_covar(self, base_shape, *params):
...
+            var = observation_noise
+        return base_distributions.Normal(function_samples, var)
 
-    def marginal(self, function_dist, *params, **kwargs):
+    def marginal(
+        self, function_dist: MultivariateNormal, *params: Any, observation_noise: Optional[Tensor] = None, **kwargs: Any
+    ) -> MultivariateNormal:
         mean, covar = function_dist.mean, function_dist.lazy_covariance_matrix
-        full_covar = covar + self._shaped_noise_covar(mean.shape, *params)
+        if observation_noise is None:
```
jacobrgardner: Do we actually need to check this here? Won't the `noise_covar` handle this if we just pass `**kwargs` along, or am I missing something

Balandat: Yeah so that's an option, I guess it depends on whether we want to be able to pass custom observation noise also to other likelihoods. The current implementation will make sure all likelihoods derived from `_GaussianLikelihoodBase` will handle the custom observation noise. If we instead the kwargs through then we'll get null-ops from that on the other likelihoods (e.g. those with inferred noise). That seems like an inconsistency we'd like to avoid?

Balandat: I guess this could be pushed into the noise models. Unfortunately things are a little messy down there because of the different handling of inputs in `ExactGP's` `__call__` method depending on whether we're in train or eval mode, which results in things like this: https://github.com/cornellius-gp/gpytorch/blob/ebc4459d1768136db080cf23cd00faffec78e0e7/gpytorch/likelihoods/noise_models.py#L54

Balandat:
Ok so now only FixedNoiseGaussianLikelihood uses the `observation_noise` kwarg.

I did try to get the fantasies working with the likelihood but I'm running into singularity issues inside the construction of the fast fantasies. I haven't dug into this too deeply yet. @jacobrgardner, maybe you can take a look since  you know that code well?

Here's a nb with tests and fanatasy repro failure: [dev_fixed_noise_likelihood.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3036769/dev_fixed_noise_likelihood.ipynb.txt)

Balandat:
I reduced the scope of this, removing the fantasy parts. Also did the honors of adding one of the first likelihood-specific tests...

If this looks good let's get this in, we can do the fantasy part on top of that.

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -2,42 +2,47 @@
...
+
+    def __init__(self, noise_covar: Noise, **kwargs: Any) -> None:
 
-    def __init__(self, noise_covar, **kwargs):
         super().__init__()
         param_transform = kwargs.get("param_transform")
         if param_transform is not None:
             warnings.warn("The 'param_transform' argument is now deprecated. If you want to use a different "
-                          "transformaton, specify a different 'noise_constraint' instead.")
+                          "transformaton, specify a different 'lengthscale_constraint' instead.")
```
jacobrgardner: Why are we warning about lengthscale constraints in the likelihood? `noise_constraint` seems like it'd be correct.

jacobrgardner:
@Balandat lgtm other than my one comment about `lengthscale -> noise`

gpytorch/likelihoods/gaussian_likelihood.py
Balandat: This is outdated issue. I pushed a fix to this to master to make it noise_constraint a couple of days ago. Should go away with a rebase

PR #601, opened 2019-03-29
Don't run preconditioner for cholesky solves

gpleiss:
This is unnecessary and slow.

### DIFF
--- a/gpytorch/lazy/cached_cg_lazy_tensor.py
+++ b/gpytorch/lazy/cached_cg_lazy_tensor.py
@@ -97,12 +97,12 @@ class CachedCGLazyTensor(LazyTensor):
             probe_vector_tmats=probe_vector_tmats,
         )
         self.base_lazy_tensor = base_lazy_tensor
-        self.eager_rhss = [eager_rhs.detach_().requires_grad_(False) for eager_rhs in eager_rhss]
-        self.solves = [solve.detach_().requires_grad_(False) for solve in solves]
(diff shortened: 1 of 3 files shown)


gpytorch/lazy/cached_cg_lazy_tensor.py
```diff
@@ -97,12 +97,12 @@ def __init__(
             probe_vector_tmats=probe_vector_tmats,
         )
         self.base_lazy_tensor = base_lazy_tensor
-        self.eager_rhss = [eager_rhs.requires_grad_(False) for eager_rhs in eager_rhss]
-        self.solves = [solve.requires_grad_(False) for solve in solves]
-        self.probe_vectors = probe_vectors.requires_grad_(False)
-        self.probe_vector_norms = probe_vector_norms.requires_grad_(False)
-        self.probe_vector_solves = probe_vector_solves.requires_grad_(False)
-        self.probe_vector_tmats = probe_vector_tmats.requires_grad_(False)
+        self.eager_rhss = [eager_rhs.detach_().requires_grad_(False) for eager_rhs in eager_rhss]
```
Balandat: `detach_().requires_grad_(False)` -> `detach()`, also below

jacobrgardner: I went ahead and fixed this because I need this updated for a small PR I am working on anyways that makes `InvMatmul` twice differentiable.

PR #605, opened 2019-03-31
Remove default batch dimension from kernels

gpleiss:
Right now the default batch shape for kernels is `torch.Size([1])`, with some squeezing if the data is not supplied as batch data.

This makes the default `batch_shape=torch.Size([])`. Saved models with the old default batch shape still load (with deprecation warning).

(Note that this speeds up evaluation quite a bit with small amounts of data).

### DIFF
--- a/gpytorch/kernels/additive_structure_kernel.py
+++ b/gpytorch/kernels/additive_structure_kernel.py
@@ -49,7 +49,7 @@ class AdditiveStructureKernel(Kernel):
             evaluate = True
             res = NonLazyTensor(res)
 
-        res = res.sum(-3).unsqueeze(0)
+        res = res.sum(0)
 
         if evaluate:
             res = res.evaluate()
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -2,6 +2,7 @@
 
 from abc import abstractmethod
 import torch
+import warnings
 from torch.nn import ModuleList
 from ..lazy import lazify, delazify, LazyEvaluatedKernelTensor, ZeroLazyTensor
 from ..module import Module
@@ -9,7 +10,6 @@ from .. import settings
 from ..utils.deprecation import _deprecate_kwarg_with_transform
 from ..utils import broadcasting
 from ..constraints import Positive
-import warnings
 
 
 @torch.jit.script
@@ -101,7 +101,7 @@ class Kernel(Module):
             dimension. It should be `d` if :attr:`x1` is a `n x d` matrix.  Default: `None`
         :attr:`batch_shape` (torch.Size, optional):
             Set this if you want a separate lengthscale for each batch of input
-            data. It should be `b1 x ... x bk` if :attr:`x1` is a `b1 x ... x bk x n x d` tensor.  Default: `1`
+            data. It should be `b1 x ... x bk` if :attr:`x1` is a `b1 x ... x bk x n x d` tensor.
         :attr:`active_dims` (tuple of ints, optional):
             Set this if you want to compute the covariance of only a few input dimensions. The ints
             corresponds to the indices of the dimensions. Default: `None`.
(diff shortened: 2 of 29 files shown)


Balandat:
Thanks, this looks like the right thing to do.

> Note that this speeds up evaluation quite a bit with small amounts of data

Nice.

gpytorch/kernels/additive_structure_kernel.py
```diff
@@ -49,7 +49,7 @@ def forward(self, x1, x2, batch_dims=None, **params):
             evaluate = True
             res = NonLazyTensor(res)
 
-        res = res.sum(-3).unsqueeze(0)
+        res = res.sum(0)
```
gpleiss: `res = res.sum(-3)` - so that this works for batch mode

gpytorch/kernels/kernel.py
```diff
@@ -304,11 +331,7 @@ def _covar_dist(
         if diag:
             # Special case the diagonal because we can return all zeros most of the time.
             if x1_eq_x2:
-                # We are computing distances between points and themselves, which are all zero
-                if batch_dims == (0, 2):
-                    res = torch.zeros(x1.shape[0] * x1.shape[-1], x2.shape[-2], dtype=x1.dtype, device=x1.device)
-                else:
-                    res = torch.zeros(x1.shape[0], x2.shape[-2], dtype=x1.dtype, device=x1.device)
+                res = torch.zeros(*x1.shape[:-2], x1.shape[-2], dtype=x1.dtype, device=x1.device)
```
gpleiss: Does this break the `batch_dims` case?

gpytorch/kernels/kernel.py
```diff
@@ -318,18 +341,13 @@ def _covar_dist(
...
             res = self.distance_module._jit_sq_dist(x1, x2, torch.tensor(x1_eq_x2), postprocess)
         else:
             res = self.distance_module._jit_dist(x1, x2, torch.tensor(x1_eq_x2), postprocess, torch.tensor(False))
 
-        if batch_dims == (0, 2):
-            if diag:
-                res = res.transpose(0, 1).contiguous().view(-1, res.shape[-1])
-            else:
-                res = res.transpose(0, 1).contiguous().view(-1, *res.shape[-2:])
-
```
gpleiss: Same here. I think this line is still necessary for additive structure/SKIP.

gpytorch/kernels/product_structure_kernel.py
```diff
@@ -54,7 +54,7 @@ def forward(self, x1, x2, batch_dims=None, **params):
             evaluate = True
             res = NonLazyTensor(res)
 
-        res = res.prod(-3).unsqueeze(0)
+        res = res.prod(0)
```
gpleiss: `prod(-3)`

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -44,12 +44,40 @@ def _getitem(self, row_index, col_index, *batch_indices):
...
+                    *self.x1.shape[-2:]
+                )[(*batch_indices, row_index, _noop_index)]
         x2 = self.x2
         if self.batch_dims == (0, 2):
             x2 = x2.permute(0, 2, 1).contiguous()
             x2 = x2.view(-1, x2.size(-1), 1)
-        x2 = x2[(*batch_indices, col_index, _noop_index)]
+        try:
+            x2 = x2[(*batch_indices, col_index, _noop_index)]
+        except IndexError:
```
gpleiss: I would add a comment here (which I forgot to do) explaining that the try/catch is for speed purposes.

gpytorch/kernels/additive_structure_kernel.py
jacobrgardner: Relative indexing won't work now that we support multibatch. I made batch_dims always swap the `d dimension to be the first batch dimensions, hence the sum(0).

gpytorch/kernels/kernel.py
jacobrgardner: No, by this point the d dimension is already its own batch dimension (one of the changes of this PR is to handle batch_dims with multibatch instead of combining it with another batch dimension.

gpytorch/kernels/kernel.py
jacobrgardner: Not with the new multibatch way of handling batch_dims

gpytorch/kernels/product_structure_kernel.py
jacobrgardner: See above

PR #610, opened 2019-04-01
Better psd_safe_cholesky, VI initialization

jacobrgardner:
This PR makes two improvements:
1. If `psd_safe_cholesky` cant produce a positive definite matrix by adding jitter, it now tries to add increasing amounts of jitter for a few times to see if that will work.
2.  Our variational parameter initialization now uses `psd_safe_cholesky` instead of `torch.cholesky` by specifying `scale_tril`.

### DIFF
--- a/gpytorch/utils/cholesky.py
+++ b/gpytorch/utils/cholesky.py
@@ -23,26 +23,27 @@ def psd_safe_cholesky(A, upper=False, out=None, jitter=None):
         if A.dim() > 2 and A.is_cuda:
             if torch.isnan(L if out is None else out).any():
                 raise RuntimeError
-    except RuntimeError:
+
+        return L
+    except RuntimeError as e:
         if jitter is None:
             jitter = 1e-6 if A.dtype == torch.float32 else 1e-8
-        idx = torch.arange(A.shape[-1], device=A.device)
-        Aprime = A.clone()
-        Aprime[..., idx, idx] += jitter
-        try:
-            L = torch.cholesky(Aprime, upper=upper, out=out)
-            # TODO: Remove once fixed in pytorch (#16780)
-            if A.dim() > 2 and A.is_cuda:
-                if torch.isnan(L if out is None else out).any():
-                    raise RuntimeError("singular")
-        except RuntimeError as e:
-            if "singular" in e.args[0]:
-                raise RuntimeError("Adding jitter of {} to the diagonal did not make A p.d.".format(jitter))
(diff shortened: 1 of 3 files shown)


Balandat:
cc @sdaulton

jacobrgardner:
It seems like Travis is taking a vacation from running our unit tests?

gpytorch/utils/cholesky.py
```diff
@@ -23,26 +23,27 @@ def psd_safe_cholesky(A, upper=False, out=None, jitter=None):
...
-        except RuntimeError as e:
-            if "singular" in e.args[0]:
-                raise RuntimeError("Adding jitter of {} to the diagonal did not make A p.d.".format(jitter))
-            else:
-                raise e
-        warnings.warn("A not p.d., added jitter of {} to the diagonal".format(jitter), RuntimeWarning)
+        for i in range(3):
+            jitter = jitter * (10 ** i)
+            Aprime = A.clone()
+            Aprime.diagonal(dim1=-2, dim2=-1).mul_(jitter)
```
Balandat: Let's try to avoid in-place modifications at all costs, we've seen this slow down backward passes by insane amounts

jacobrgardner: The existing code was already doing an in-place add.

I also don't think we should treat in-place operations like voodoo magic that we should avoid and just eat 2x-4x memory costs, especially for `n x n` matrices.

Balandat: Well let's at least be aware - I've seen cases where moving away from a simple in-place `mul_` ended up speeding up the backward pass by around 5X. Or getting rid of an innocuous indexing operation on the GPU resulting in similar speed differences.

gpytorch/utils/cholesky.py
```diff
@@ -23,26 +23,27 @@ def psd_safe_cholesky(A, upper=False, out=None, jitter=None):
...
+                # TODO: Remove once fixed in pytorch (#16780)
+                if A.dim() > 2 and A.is_cuda:
+                    if torch.isnan(L if out is None else out).any():
+                        raise RuntimeError("singular")
+                return L
+            except RuntimeError:
+                continue
 
-    return L
+        warnings.warn("A not p.d., added jitter of {} to the diagonal".format(jitter), RuntimeWarning)
```
Balandat: Hmm I guess this will never be raised if one of the values succeeds - will put up a fix.

PR #612, opened 2019-04-02
Less lazy kernel evaluation for small predictions

gpleiss:
This PR does a majority of kernel prediction operations not in lazy mode.
It lazily computes `[k* k**]` in batch, and then does most operations with torch.tensors.

This reduces the overhead with small predictions.


jacobrgardner:
I merged this in to #615 so that we can keep benchmarking on the same branch instead of having old time costs appear everywhere. I also fixed the failing tests resulting from this -- closing this here in favor of #615.

PR #614, opened 2019-04-03
Root inv decomp cholesky

dme65:
Fix to make `get_fantasy_model` work in small data regime


jacobrgardner:
I'm like 66% sure the test failures here are actually PyTorch's fault, not ours: https://github.com/pytorch/pytorch/issues/18825

gpleiss:
For now we could just add a backward hook that does `grad = 0.5 * (grad + grad.transpose(-1, -2)`, which will work even once the fix to pytorch is in place.

gpleiss:
Additionally - in this PR I'm also making it the case the gradient for `inv_matmul` is also symmetric.

PR #615, opened 2019-04-04
[WIP] Efficiency Improvements for Predictions on small test sets

jacobrgardner:
**Merge #605 First**

This PR is for continued work on improving our efficiency for small test sets, and builds off of #605 and #612 

- [x] Merge test_train_covar and test_test_covar kernel calls.
- [x] Remove custom `__getattr__`, or at least only use in `debug` mode potentially.
- [x] Lower overhead `__getitem__` for `NonLazyTensor`
- [x] Lower overhead `__getitem__` in majority of cases for `LazyEvaluatedKernelTensor`
- [x] Break up JIT scripts to minimize branching.
- [x] Use `torch.addmm` for distance computations in non-batch setting.


gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -251,3 +255,16 @@ def representation_tree(self):
...
+    def __getitem__(self, index):
+        """
+        Supports subindexing of the matrix this LazyTensor represents. This may return either another
+        :obj:`gpytorch.lazy.LazyTensor` or a :obj:`torch.tensor` depending on the exact implementation.
+        """
+        # Process the index
+        index = index if isinstance(index, tuple) else (index,)
+        if index[0] is Ellipsis and all([isinstance(idx, slice) for idx in index[1:]]):
+            *batch_indices, row_index, col_index = index
+            return self._getitem(row_index, col_index, *batch_indices)
```
gpleiss: This could have weird issues if `len(index) != self.dim()`.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1678,14 +1678,6 @@ def __mul__(self, other):
         """
         return self.mul(other)
 
-    def __setattr__(self, name, val):
```
gpleiss: Can we just delete this method?

gpytorch/lazy/non_lazy_tensor.py
```diff
@@ -80,6 +80,9 @@ def mul(self, other):
         else:
             return super(NonLazyTensor, self).mul(other)
 
+    def __getitem__(self, index):
+        return self.tensor.__getitem__(index)
+
```
gpleiss: This also fails - because we may need to wrap it as a NonLazyTensor. I tried this earlier.

gpytorch/kernels/kernel.py
```diff
@@ -318,6 +368,7 @@ def _covar_dist(
             x2 = x2.unsqueeze(0).transpose(0, -1)
```
gpleiss: Can we make `covar_dist` a public method? It's really confusing right now to tell what is a helper method and what's not.

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
jacobrgardner: Made it more explicitly a special case for indexing of the form `mat[..., slice, slice]`.

gpytorch/lazy/lazy_tensor.py
jacobrgardner: It's really really slow because of how often it gets called, and the checking it does really only ever comes up when we write a *brand new* LazyTensor (e.g., the error will never actually be raised in production code with existing lazy tensors).

I say we can it until we see 3rd party people (1) contributing their own new LazyTensors and (2) actually run in to this issue.

gpytorch/lazy/non_lazy_tensor.py
jacobrgardner: Implemented the fix for this that we discussed offline

jacobrgardner:
@Balandat @gpleiss This is done now I think. Not sure why we're seeing a grad failure on PyTorch stable -- we're seeing the same failure on master.

jacobrgardner:
The failure travis is getting doesn't occur for me locally with pytorch 1.0.1-post2 :-(

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
gpleiss: This could also still fail if the index is `mat[..., 0:3]`, for example.

gpytorch/module.py
```diff
@@ -74,7 +74,7 @@ def initialize(self, **kwargs):
                 module.initialize(**{name: val})
             elif not hasattr(self, name):
                 raise AttributeError("Unknown parameter {p} for {c}".format(p=name, c=self.__class__.__name__))
-            elif name not in self._parameters:
+            elif name not in self._parameters and name not in self._buffers:
```
Balandat: @gpleiss so this actually breaks some of the things we're doing to re-initalize models with new data. Say you have a model using a `FixedNoiseGaussianLikelihood` with `n` initial observations. You get another observation in a BayesOpt loop, and you want to update your model in-place (using set_train_data). For this you also need to update the new noise value. But if you go through `initialize` (which is what this change here results in) you'll end up with a runtime error b/c the `initialize` method tries to reshape a 11 element tensor to a 10 element tensor. 

The alternative is to re-register the buffer itself with the larger tensor. But that will mean that this has to be done in a model-specific way and we need to know the underlying model setup (rather than using a standard noise property).

Thoughts?

jacobrgardner: I'm not a big fan of `initialize` being able to change the shape of parameters because I feel like that's going to result in more bugs than legitimate use cases. I'd rather FixedNoiseGaussianLikelihood have functionality for this -- for example the `fantasy` method that will be used for adding noises to the vector.

Balandat: Yeah not suggesting that initialize itself should be able to change the size. I guess for now we can handle this on our end until we have the fantasy method functionality set up.

gpleiss: It seems that this will be an issue mostly related to fantasy/BayesOpt updates with FixedNoiseGaussianLikelihood. We should maybe brainstorm interfaces specifically for this case.

Issue #629, opened 2019-04-09
[Bug] Initialize parameters complying with constraints if transform=None


Balandat:
# 🐛 Bug

When passing `transform=None` into constraints, constraints are not enforced via a transform. Yet the parameter initialization in most places sets initial values to zero, disregarding the constraint. E.g. if you pass `noise_constraint=GreaterThan(1e-4, transform=None` into the `GaussianLikelihood`, then the initial value of the noise is 0. This usually results in nan errors in the first step of fitting. You can manually set the initial value to something more reasonable, but it's kind of cumbersome to have to do that manually every time when instantiating a model.

## Expected Behavior

If a non-enforced constraint is present, the initial value should be chosen in compliance with the constraint in some reasonable fashion, e.g. as the mid point of an `Interval` constraint, or as 2*lower_bound for a `GreaterThan` constraint.

gpleiss:
I think this was closed by #643 ? (Reopen if not!)

PR #643, opened 2019-04-12
[RFC] Add initial_value property to constraints.

Balandat:
This simplifies intializing parameters from a constraint object. Addresses #629


Balandat:
cool I'll add a quick unit test for this as well

PR #651, opened 2019-04-17
Alter sample data so example exemplifies the proper use case

nsfinkelstein:
The point of Hadamard multitask GP regression is that we do not observe the values of all tasks for all inputs. The data used in this example previously had all tasks observed for all inputs. This commit changes that, so as to help emphasize the use case.


jacobrgardner:
Yeah this is a good point -- it's been something that's been on the list to change but we never got to. Thanks for doing this for us :-).

nsfinkelstein:
Thanks for the quick review!

Issue #654, opened 2019-04-20
Hadamard Multioutput GP With Mean & Kernel functions that vary by task


nsfinkelstein:
# 🚀 Feature Request

At present, there does not seem to be a way to learn a multioutput GP with task-specific mean and kernel functions if not all tasks are observed for all inputs, as shown in this example (https://github.com/cornellius-gp/gpytorch/blob/master/examples/03_Multitask_GP_Regression/Hadamard_Multitask_GP_Regression.ipynb).

## Other

I'd be happy to contribute the code for this. The only question I have is about the best way to do it.  The base mean class assumes means will be called with a single input, `x`, representing the inputs. For the hadamard case, we would expect two inputs, `x` and `i`, representing the inputs, and the indices of the task that was observed at each 

It seems that there are two options. i) add an optional `i` parameter in the base class, or have the hadamard multitask mean class not descend from `Mean`. Perhaps there are other options you prefer. I assume there will be a similar issue with kernels.

What is the best way to proceed?

Thanks
Noam

nsfinkelstein:
An example implementation, similar to the multitask mean, is below.

```py

class HadamardMultitaskMean(gpytorch.module.Module):
    def __init__(self, base_means, num_tasks):
        """
        Args:
            base_means (:obj:`list` or :obj:`gpytorch.means.Mean`): If a list, each mean is applied to the data.
                If a single mean (or a list containing a single mean), that mean is copied `t` times.
            num_tasks (int): Number of tasks. If base_means is a list, this should equal its length.
        """
        super(HadamardMultitaskMean, self).__init__()

        if isinstance(base_means, gpytorch.means.Mean):
            base_means = [base_means]

        if not isinstance(base_means, list) or (len(base_means) != 1 and
                                                len(base_means) != num_tasks):
            raise RuntimeError(
                "base_means should be a list of means of length either 1 or num_tasks"
            )

        if len(base_means) == 1:
            base_means = base_means + [
                deepcopy(base_means[0]) for i in range(num_tasks - 1)
            ]

        self.base_means = ModuleList(base_means)
        self.num_tasks = num_tasks

    def forward(self, x, i):
        all_means = torch.cat(
            [sub_mean(x).unsqueeze(-1) for sub_mean in self.base_means],
            dim=-1)

        means = torch.zeros_like(x)
        for task in range(self.num_tasks):
            means[i == task] = all_means[:, task:task + 1][i == task]

        return means.squeeze()

[...]

PR #658, opened 2019-04-22
Hadamard mean

nsfinkelstein:
This pull request implements a hadamard mean as discussed in https://github.com/cornellius-gp/gpytorch/issues/654.

I left some of the tests commented out because I don't understand how batches are being handled in the Mean classes. Would it be possible to ask for a few pointers on handling dimensionality in a way that's consistent with the other Mean classes, especially around batches?


jacobrgardner:
@nsfinkelstein Sorry for the delay on this, I've been crazy busy with this other feature I'm trying to get implemented. I'll take a look at the PR tomorrow morning and try to give some direction on handling batch mode for the means!

jacobrgardner:
If it's sufficient, here's a quick overview of how batch shapes work for kernels and means:

- We assume inside the kernel/mean that data is always in tensors of the form `x` where `x.shape = [b1, b2, ..., bk x n x d]`, or generally `x.shape = [*batch_shape, n, d]`, where `*batch_shape` is possibly empty.
- When called with `last_dim_is_batch=True`, we want to treat the `d` dimension as another batch dimension, e.g. we want to transpose `x` to `b1 x ... x bk x d x n x 1`. This lets us handle things like additive structure more efficiently.
- Kernels on data of this size should generally return a matrix of size `b1 x b2 x ... x bk x n x n`.
- Means on data of thise size shold return a matrix of size `b1 x ... x bk x n`
- Many kernels and means have parameters. If you want your kernel to be able to operate in batch mode, the mean/kernel module should take a `batch_shape` argument and resize the parameter accordingly.

The interpretation of batch dimensions is training multiple GPs in parallel. E.g., having `X = b x n x d` and `y = b x n` corresponds to training `b` independent GPs. If the kernel is created with no batch shape specified, these GPs will share e.g. the lengthscale parameter. If `batch_shape` is specified, the intention is that each batch (GP) should use a different parameter (e.g. lengthscale).

Thus:
```python
x = torch.randn(3, 5, 2)
batch_shape = x.shape[:-2]
K = RBFKernel(batch_shape=batch_shape)(x)  # Gets a 3 x 5 x 5 set of RBF covariance matrices.
mu = ConstantMean()(x)  # Gets a 3 x 5 set of mean vectors where the constant is shared across batches.
```

gpytorch/means/hadamard_multitask_mean.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class HadamardMultitaskMean(MultitaskMean):
+    def forward(self, x, i):
+        """
+        Evaluate the appropriate mean in self.base_means on each input
+        """
+        means = [self.base_means[task](x[i == task]) for task in range(self.num_tasks)]
+
+        res = torch.zeros_like(x)
```
jacobrgardner: Probably:
```suggestion
        batch_shape = x.shape[:-2]
        res = torch.zeros(*batch_shape, x.size(-2))
```

gpytorch/means/hadamard_multitask_mean.py
```diff
@@ -0,0 +1,27 @@
...
+        for task in range(self.num_tasks):
+            if len(means[task]) > 0:
+                mean = means[task]
+
+                if res[i == task].dim() == means[0].dim() + 1:
```
jacobrgardner: We shouldn't have to do batch shape checks / unsqueezes here. Every mean should return a `*batch_shape x n_task` tensor, and `res` should be made `*batch_shape x n` (see above comment)

gpytorch/means/hadamard_multitask_mean.py
```diff
@@ -0,0 +1,27 @@
...
+                    mean = mean.unsqueeze(dim=-1)
+
+                res[i == task] = mean
+
+        if res.dim() == means[0].dim() + 1:
```
jacobrgardner: Same here. Did you run in to means that weren't returning `*batch_shape x n` or `*batch_shape x n x t` tensors? If so, that's something we need to fix.

nsfinkelstein:
Thanks for the explanation and review! I'll make changes this weekend.

gpytorch/means/hadamard_multitask_mean.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+
+import torch
+from .multitask_mean import MultitaskMean
+
+
+class HadamardMultitaskMean(MultitaskMean):
```
jacobrgardner: For `HadamardMultitaskMean` are you sure we want to be extending `MultitaskMean`? `MultitaskMean` returns `*batch_shape x n x t` tensors, assuming we have an observation for each data point for each task. Since in this case each data point along the `n` dimension belongs to a different task, I think we just want a `*batch_shape x n` result.

Maybe with the addition of `HadamardMultitaskMean` we should deprecate `MultitaskMean` and rename it `KroneckerMultitaskMean`.

PR #661, opened 2019-04-23
Create an LMC Kernel

nsfinkelstein:
The current state of this pull request is that the kernel seems to work, but is untested and cannot handle the batch dimension. In addition, it seems to break when kernels are evaluated lazily.


gpleiss:
@nsfinkelstein - sorry for the very slow reply here! I think that you should already be able to implement the LMC kernel in GPyTorch:

```python
def __init__():
    # ...
    rbf_kernels = gpytorch.kernels.RBFKernel(batch_shape=torch.Size(num_latent))
    task_kernels = gpytorch.kernels.IndexKernel(batch_shape=torch.Size(num_latent))
    # ...

def forward():
    # ...
    covar = (rbf_kernels(x) * task_kernels(task_indices)).sum(-3)
    # ...
```

This will have the advantage of being batched - which will be much faster than using for-loops.

nsfinkelstein:
@gpleiss Thanks!. Do I understand correctly that the approach you demonstrate will only work when all tasks are observed at all timepoints? 

My use case is one with sparse observations in continuous time, such that tasks are never observed at the same time (even approximately).

gpleiss:
@nsfinkelstein  - the approach i sketched assumes that each entry (`x`) corresponds to one output (corresponding to `task_indices`). This is essentially what is in the hadamard multitask example, but with a different kernel (corresponding to LCM)

Issue #662, opened 2019-04-23
[Bug] pip install installs torch stable if torch nightlies are present


Balandat:
Installing nightlies first and then installing gpytorch reverts torch to latest stable:
```
pip install -q torch_nightly -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
pip install -q git+https://github.com/cornellius-gp/gpytorch.git
torch.__version__  # 1.0.1.post2
```

Installing gpytorch first and then installing nightlies works properly:
```
pip install -q git+https://github.com/cornellius-gp/gpytorch.git
pip install -q torch_nightly -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
torch.__version__  # 1.1.0.dev20190423
```

Taking a step back, we do all this version parsing magic in setup.py [here](https://github.com/cornellius-gp/gpytorch/blob/master/setup.py#L30-L46).
Not really sure why this is necessary, can't we just require this to be `"torch>=1.0.1"`? Or is there some issue with the package naming?

jacobrgardner:
Before #663 was merged the reason we did version parsing was to solve exactly this issue, where we want to allow people to update gpytorch without overwriting their torch install if they have torch-nightly installed.

Do you have a solution in mind or should we revert #663? I'd rather do the version parsing than have `pip install gpytorch` do silly things for people who are on pytorch-nightly.

Balandat:
Yeah so is scrapped this because the parsing actually didn't work for me. Turns out this is an obscure bug with pip 19.0.1 in conjunction with having a `pyproject.toml` file: https://github.com/pypa/pip/issues/6163

Not quite sure what the proper solution is for this. Sledgehammer would be removing `pyproject.toml`. We could also try to check pip version at runtime. Maybe there is also a way to check whether either torch or torch-nightly is installed without actually importing them?

Balandat:
@jacobrgardner I put up #666 if you want to merge that in and go back to the way things were.

PR #663, opened 2019-04-23
Remove version parsing for torch nightlies, req. >=1.0.1 instead

Balandat:
Seems like this is unnecessary. Requiring `"torch>=1.0.1"` will not install stable if a dev version from the nightlies is present.

Addresses #662

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,23 +27,6 @@ def find_version(*file_paths):
 readme = open("README.md").read()
 version = find_version("gpytorch", "__init__.py")
 
-# See if we have the development version of pytorch installed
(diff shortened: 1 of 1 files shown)


Balandat:
Hmm actually there seems to be a difference in how this is dealt with between pip and conda. In particular, if the `torch-nightly` pip package is installed then setup will still install the stable version. I'll put up another PR to address this issue.

Balandat:
argh apparently there is a bug in pip 19.0.1 that shows up in conjunctions with having a pyproject.toml file (which we do) that always results in setup.py not being able to import torch: https://github.com/pypa/pip/issues/6163

PR #666, opened 2019-04-25
Add parsing for torch nightlies/dev back in

Balandat:
Addresses #662. Note that this does currently not work on pip 19.0.1 due to a bug: https://github.com/pypa/pip/issues/6163


Balandat:
Hmm this still doesn't seem to work, even on newer pip versions doing a pip install of gpytorch from source will install torch stable.

PR #676, opened 2019-05-04
Add LinearMean

KeAWang:
Seems like a reasonably simple parametric `Mean` for users to try out?


gpleiss:
Seems reasonable, once the tests are fixed!

PR #680, opened 2019-05-06
add an example to do NUTS sampling

fehiepsi:
This PR adds a notebook to demonstrate how to use Pyro's NUTS to do sampling for GP hyperparameters. Currently the PR contains nothing but I will make progress during upcoming days. I am still not familiar with how GPyTorch works so I hope GPyTorch devs will help me if I have some questions during the way. :) Thanks in advance!


gpleiss:
Just let us know if you have any questions :)

fehiepsi:
@gpleiss Could you show me a convenient way to change constraints of parameters? For example, I want to set lower bound for `lengthscale` but the following code
```
model.covar_module.base_kernel.register_constraint("lengthscale", gpytorch.constraints.GreaterThan(0.1))
```
triggers an error `RuntimeError: Attempting to register constraint for nonexistent parameter.` I know that `lengthscale` is not a Parameter but `raw_lengthscale` is.

In addition, does `ExactMarginalLogLikelihood` cover log_abs_det_jacobian of transforms (from unconstrained to constrained) or I have to explicitly add those terms to the final loss?

fehiepsi:
Base on the source code of `ExactMarginalLogLikelihood`, it seems that GPyTorch losses do not include jacobian of transforms in GPyTorch. I'll find a work around instead.

jacobrgardner:
@fehiepsi Sorry for the delay, I must have missed this when you first asked! Will things not work if you are using one of the Pyro optimizers / ELBOs? 

For the exact MLL case, note that `gpytorch.Module` has a `register_added_loss_term` functionality that you can use for this. See `InducingPointKernel` or my deep GP branch for example uses of this.

jacobrgardner:
To your other question: register the constraint to the raw parameter. See the Hyperparameters tutorial notebook under `examples/00_Basic_Usage`

fehiepsi:
Thanks @jacobrgardner ! I'll check them out. I think that I was facing the Cholesky issue due to `lengthscale` move to the very small area. In HMC, when we update mass matrix at the end of each window, we need to find a suitable learning rate w.r.t to that mass matrix. While finding good learning rate, Cholesky issue happens due to large value of learning rate. This is unfortunately because rather than returning `nan` value (which is equivalent to potential energy is `inf` in HMC language), PyTorch will trigger ValueError when applying Cholesky to singular matrix. Maybe we can add some try-catch exception to deal with this issue in Pyro. cc @neerajpad 

I'll need to investigate more to see where is the real problem.

jacobrgardner:
Fixing this for Cholesky might be as simple as catching the value error thrown from psd_safe_cholesky and returning nan. Things may be more challenging when you get past 128 data points when the default mode of operation switches to CG, however, which can fail to converge more than hard crash.

Looking in to the failure, it looks like you are getting raw lengthscales on the order of -10^9, which is causing the whole kernel matrix to be NaN, so obviously all methods fail.

fehiepsi:
> Looking in to the failure, it looks like you are getting raw lengthscales on the order of -10^9, which is causing the whole kernel matrix to be NaN, so obviously all methods fail.

Yup, that's why I need to change the constraint of lengthscale (to force it strictly positive). But returning `nan` is a better solution to HMC/NUTS.

gpytorch/constraints/constraints.py
```diff
@@ -94,7 +94,8 @@ def transform(self, tensor):
                     "GreaterThan or LessThan instead."
                 )
 
-        transformed_tensor = (self._transform(tensor) * self.upper_bound) + self.lower_bound
+        scale = self.upper_bound - self.lower_bound
+        transformed_tensor = (self._transform(tensor) * scale) + self.lower_bound
```
fehiepsi: This seems like a bug (or my math is wrong), so I fix it.

gpytorch/constraints/constraints.py
```diff
@@ -2,7 +2,8 @@
 
 import math
 import torch
-from torch.nn.functional import softplus, sigmoid
+from torch import sigmoid
```
fehiepsi: `torch.nn.functional.sigmoid` is depricated

gpytorch/module.py
```diff
@@ -114,6 +114,15 @@ def initialize(self, **kwargs):
...
         return self
 
+    def _load_raw_parameters(self, **kwargs):
+        for name, val in kwargs.items():
+            if '.' in name:
+                module, name = self._get_module_and_name(name)
+                module._load_raw_parameters(**{name: val})
+            else:
+                delattr(self, name)
+                setattr(self, name, val)
```
fehiepsi: this method is used to load values to model's parameters.

fehiepsi:
@gpleiss @jacobrgardner It seems that we have a working version now. If you have some requirements or further recommendations to improve the notebook, please let me know. I'll address them. Thanks! :)

gpytorch/module.py
jacobrgardner: `model.initialize(**hypers)` should accomplish this

jacobrgardner:
Thanks @fehiepsi! I'll check the notebook out soon!

gpytorch/module.py
fehiepsi: Unfortunately, `model.initialize(**hypers)` only copies data. Because I need to compute `grad` of potential function (loss) w.r.t. hypers, I need to overwrite model's attributes instead. Can I change how `initialize` method works to use overwriting instead of copying?

jacobrgardner: Yeah, I think non breaking changes to the internals to make the interface cleaner are totally fine. By the same token, the transform Jacobians can probably live directly in the constraint objects, so that getting them all of them is simplified

fehiepsi: I think it will be a breaking change due to `expand_as`, `view_as` behaviour at [those lines](https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/module.py#L91), so I keep this as is.

jacobrgardner: Do gpytorch.priors need log_det_abs_jacobian implementations or do those already inherit what they need from upstream? Maybe we could use Pyro distributions for non uniform priors?

fehiepsi: I don't think that gpytorch.priors need log_det_abs_jacobian. Pyro NUTS only needs to know how to transform from unconstrained space to constrained space; then it will adjust joint density based on that transform, regardless the prior which we put into the parameter. We'd need to make sure that the constraint for that parameter is a subset of the support of the prior though (to avoid drawing invalid values). So I think that something likes this will work well for `noise` parameter in NUTS:
```
likelihood = gpytorch.likelihoods.GaussianLikelihood(
    noise_constraint=gpytorch.constraints.GreaterThan(1e-3),
    noise_prior=gpytorch.priors.NormalPrior(0, 1)
)
```

fehiepsi: In Pyro, we use default constraint is the support of prior, so we don't need to specify `constraint` as in GPyTorch. I think that we can do it in GPyTorch too, which mean that if I specify
```
likelihood = gpytorch.likelihoods.GaussianLikelihood(
    noise_prior=gpytorch.priors.GammaPrior(0, 1)
)
```
then `noise_constraint` will be (by default) `gpytorch.constraints.Positive()`. But supporting that behaviour is not necessary though.

Balandat:
Any updates on this?

fehiepsi:
@Balandat This PR is based on pyro dev version. Recently, the interface of MCMC in pyro is also changed a bit so I would like to wait for the next pyro release to revisit this PR. :)

fehiepsi:
Closed in favor of #918!

Issue #667, opened 2019-04-26
[Feature Request] Fantasies evaluated at the same point


danielrjiang:
# 🚀 Feature Request

Suppose `get_fantasy_model` is called with inputs having shape b x m x d and targets having shape b x m. If the inputs are identical across the b batches (with differing targets), then memory can be saved by potentially using the same kernel matrix for all b fantasy models. In the case of more than one batch dimension, then it might be nice to allow for this in the outermost batch dimension(s), perhaps. E.g., we have inputs of shape b1 x b2 x ... x bk x m x d and targets of shape b1 x b2 ... x bk x m x d. Here, if the inputs could be identical across the outermost batch dimension b1.

## Motivation

This type of feature could be very useful for BO where the fantasy models are usually called at the same point X. But if many samples are used, memory becomes an issue.

## Pitch

See above.

**Describe alternatives you've considered**

Alternatives would be to sequentially evaluate the fantasy models and reduce the batching, but this is slow.

**Are you willing to open a pull request?** (We LOVE contributions!!!)

PR #687, opened 2019-05-08
[RFC] Fantasizing repeatedly at the same input points

Balandat:
If `get_fantasize_model` is called with a `m x d` input tensor and a `b x m` target tensor, this change makes sure to not automatically expand the input and compute the fantasy updates for each batch (they are the same). Instead, the fantasy update is done only once, and the resulting updated covariance caches are copied (using a `BatchRepeatLazyTensor`).

This is an early draft and requires additional tests.

Addresses #667

To test, see attached notebook:
[test_same_input_fantasies.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3159647/test_same_input_fantasies.ipynb.txt)

### DIFF
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -91,10 +91,13 @@ class ExactGP(GP):
 
         .. note::
             If `targets` is a batch (e.g. `b x m`), then the GP returned from this method will be a batch mode GP.
+            If `inputs` is of the same (or lesser) dimension as `targets`, then it is assumed that the fantasy points
+            are the same for each target batch.
 
         Args:
             - :attr:`inputs` (Tensor `m x d` or `b x m x d`): Locations of fantasy observations.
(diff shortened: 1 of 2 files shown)


jacobrgardner:
@Balandat This all looks good to me, actually. Do you have any perf results on this?

Balandat:
I haven't done detailed timings, but in the nb creating a fantasy model with 10 new points in 50 batches for gp with `n=500` points in `d=6` dimensions takes 6ms instead of 30ms. Haven't tested this in our closed-loop applications, but @danielrjiang will be looking into this. 

The other thing that I want to do on top of this is support for batched gps, i.e. batch-fantasizing an already batched GP.

gpleiss:
lgtm!

jacobrgardner:
The rough timing test seems sufficient to me since the running time should clearly be faster in this case; was more just asking for a basic sanity check which you have.

PR #690, opened 2019-05-09
Fast fantasies in batch (batch evaluation, not batch model) mode.

Balandat:
Updates #687 to work with batched inputs. 

`inputs` can now be `b1 x ... x bk x m x d` or `f x b1 x ... x bk x m x d`, and `targets` are `b1 x ... x bk x m` or `f x b1 x ... x bk x m`.

If the `f` dimension is present in the targets but not the inputs, this will assume all fantasies are evaluated at the same inputs (in the respective batch), and computation happens more (time, memory) efficiently by only doing the update once.

Test:
[test_same_input_fantasies_batched.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3164081/test_same_input_fantasies_batched.ipynb.txt)

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -84,6 +84,9 @@ class DefaultPredictionStrategy(object):
         """
         # Here the precomputed cache represents S,
         # where S S^T = (K_XX + sigma^2 I)^-1
+        if not torch.is_tensor(precomputed_cache):
+            # can't call tensor matmul on a lazy, so we'll flip things around
+            return precomputed_cache.transpose(-1, -2).matmul(test_train_covar.transpose(-1, -2)).transpose(-1, -2)
         return test_train_covar.matmul(precomputed_cache)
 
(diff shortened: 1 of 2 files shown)


gpytorch/models/exact_prediction_strategies.py
```diff
@@ -84,6 +84,9 @@ def _exact_predictive_covar_inv_quad_form_root(self, precomputed_cache, test_tra
         """
         # Here the precomputed cache represents S,
         # where S S^T = (K_XX + sigma^2 I)^-1
+        if not torch.is_tensor(precomputed_cache):
+            # can't call tensor matmul on a lazy, so we'll flip things around
+            return precomputed_cache.transpose(-1, -2).matmul(test_train_covar.transpose(-1, -2)).transpose(-1, -2)
```
Balandat: @jacobrgardner, @gpleiss I wonder if it would make more sense to use a tensor and broadcast for the matmul, rather than using the BatchRepeatLazyTensor. Thoughts?

Balandat:
Gotta get this in - going ahead since this shouldn't touch anything outside of the obscure stuff we're doing.

Issue #688, opened 2019-05-09
Replace torch.gesv with torch.solve


Balandat:
`torch.gesv` currently throws the following DeprecationWarning under torch 1.1. We should move to `torch.solve` and require 1.1 in the next release.

```
.../gpytorch/gpytorch/models/exact_prediction_strategies.py:150: UserWarning: torch.gesv is deprecated in favour of torch.solve and will be removed in the next release. Please use torch.solve instead.
  fant_cache_lower = torch.gesv(small_system_rhs, schur_complement)[0]
```

PR #693, opened 2019-05-11
Support for fantasizing batched models

Balandat:
Allows to fantasize on a model that is already in batch mode.

Also supports
- batch-fantasizing on a batched model, following proper broadcasting semantics
- efficient fantasizing for repeated points

What is still missing is support for multi-task models.

Test nb: [test_batch_model_fantasies.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3168443/test_batch_model_fantasies.ipynb.txt)

### DIFF
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -110,12 +110,9 @@ class ExactGP(GP):
                 "all test independent caches exist. Call the model on some data first!"
             )
 
-        if self.train_inputs[0].dim() > 2:
-            raise RuntimeError(
-                "Adding fantasy observations to a GP that is already batch mode is currently unsupported."
-            )
+        model_batch_shape = self.train_inputs[0].shape[:-2]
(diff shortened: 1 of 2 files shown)


Balandat:
This also addresses #688 by using `torch.solve` instead of `torch.gesv`

Balandat:
@jacobrgardner @gpleiss this fails b/c currently pytorch stable in travis is 1.0.1. and not 1.1. If you're ok requiring torch 1.1 for the next release we can just switch travis over.

jacobrgardner:
@Balandat Let's definitely switch Travis' PyTorch stable over to 1.1. I think our intention should be to pin specific releases of GPyTorch to specific releases of PyTorch, but certainly each release should be pinned to the latest release of PyTorch, at least while things are rapidly improving for us on the PyTorch side.

Balandat:
#694 switches travis over

Issue #712, opened 2019-05-30
get_fantasy_model for ModelListGP


Balandat:
Right now `IndependentModelList` does not implement the `get_fantasy_model` method. We can write a lightweight wrapper that creates a fantasy model for each of the models and returns these fantasy models in a new `IndependentModelList`.

PR #713, opened 2019-05-30
Add support for get_fantasy_model to IndependentModelList

Balandat:
Addresses #712


Balandat:
This also works with FixedNoiseGaussianLikelihood models, will add this to the test cases.

PR #718, opened 2019-06-04
Fix for bug when adding multivariate normals

darbour:
Fixes a small bug where adding two multivariate normals was broken because of a reference to self._mean rather than self.mean


Balandat:
Thanks for the fix!

Issue #730, opened 2019-06-10
Making base test cases importable?


Balandat:
We'd like to use some of the gpytorch test infrastructure outside of gpytorch. Specifically, we'd like to re-use some of `BaseKernelTestCase` for some new kernels that we're playing around with (we're planning on upstreaming these once they work). The issue is that we can't import this class b/c the `test` module isn't part of the gpytorch package (and we don't want to require the source and do some questionable path manipulations). 

Since this is just a barebones class and not an actual unittest test case, we could move this into a `pytorch.test` module and leave the actual tests in the `test` folder (the same applies to other test infra such as `BaseTestCase`, `BaseLikelihoodTestCase` etc). 

This shouldn't have any effects other than making the base test infra importable. 

@jacobrgardner, @gpleiss thoughts on this?

jacobrgardner:
Yeah if it is useful to you I'm not opposed to moving the base test cases under gpytorch.test

Balandat:
cool I'll put up a diff for this soon then

Balandat:
Done in #733

PR #733, opened 2019-06-12
Move base test cases into gpytorch.test module

Balandat:
Addresses #730. Does some cleanup of names along the way.


Balandat:
Test failure should have nothing to do with these changes here.

PR #746, opened 2019-06-21
Add new example of latent function inference

### LINKED ISSUE #745: [Examples] Example of latent function inference
Hi All!

I would like to propose a new example notebook for latent function inference.
All current examples are either Regression or Classification problems, so I thought it might be useful to extend them with a bit more exotic problem.

The example considers the following simple problem:
given inputs X and observations Y from an exponential distribution with the scale = lambda(X), find lambda(X).

It took me quite some time to figure out how to define a non-gaussian likelihood and use pyro.svi with it, so I think it might be worth sharing.
I tested my example with a variety of lambda(X) functions and it infers these functions correctly.
Nonetheless, there must be still room for improvement and I would very much appreciate any comments.

Pull request follows.

Thank you,
Nina

NinelK:
This is an example of latent function inference based on pyro.svi.
It considers the following simple problem:
given inputs X and observations Y from an exponential distribution with the scale = lambda(X), find lambda(X).

I propose this example because there is currently no examples other than regression/classification problems, and I thought latent function inference might be useful for others.

Closes #745


jacobrgardner:
This is great! We've been meaning to add better examples where the integration with pyro is useful for more than just an alternate implementation of SVI (e.g., latent variable models), so thanks a lot for this.

My one question is your implementation of `marginal` in the likelihood -- as written it is essentially a no-op (e.g., given `p(f|D, x)` it returns `p(f|D, x)`. It is intended to return the predictive distribution for `y` (e.g., an approximate or analytical solution to `\int p(y|f)p(f|D, x) df` depending on which is possible -- at the moment `model(test_x)` and `likelihood(model(test_x))` are identical operations which is not intended.

in this case, we probably want just the default implementation of `marginal` that exists on the `Likelihood` base class -- did you try making predictions through the likelihood with the default?

NinelK:
Yes, indeed, marginal is supposed to be marginal and return `p(y|x)`. Thank you very much for correcting me!

I have also rewritten sampling from p(y|x) using `likelihood(model(x))`. However, the default `marginal` method produces 10 (`settings.num_likelihood_samples`) samples of `f` for each input in order to estimate `\int p(y|f)p(f|D, x) df`, whereas I only need 1 random sample of `f` to generate a single sample from `p(y|x)`. I do not see a nice way to change the number of samples drawn from `p(f|D,x)` in the default `marginal` method for a particular call (changing settings does not look like a good option). Now I simply take one vector from the results of `likelihood(model(x))`, but then some unnecessary computations are performed.

jacobrgardner:
By default the value of `gpytorch.settings.num_likelihood_samples` is used in marginal, so you could modify this to be 1 if you only need one sample

NinelK:
Thank you very much, now sampling looks perfect!

jacobrgardner:
@NinelK Thanks again for doing this! I updated the example notebook with two things:

1. Just some new text explaining what's going on in each cell.
2. You were computing the GP predictions of the latent scale function by doing `exp(E[f])`. I changed this to do `E[exp(f)]` via:
```python
with torch.no_grad():
    output = model(test_x)
samples = output.rsample(torch.Size([2048]))
F_mean = gplink(samples).mean(0)
F_var = gplink(samples).var(0)
lower = F_mean - 1.95 * F_var.sqrt()
upper = F_mean + 1.95 * F_var.sqrt()
```

Do you want to take a look through to make sure you agree with my changes and I'll merge?

NinelK:
1. @jacobrgardner Thank you very much for adding explanations. The comment about pyro variational GP model is especially helpful.
2. I intentionally visualized `exp(E[f])`, because it is strictly positive as a valid scale parameter of exponential distribution should be. Lower bound of the confidence interval of `E[exp(f)]` may be negative, because p(gplink(f)|x) is positively skewed. I think it might be confusing and may seem like there is non-zero probability to get gplink(f)<0, which is not true.
So, I would visualize `exp(E[f])`, as it seems more informative to me. Everything else looks great, thank you! I agree with all of the other changes.

jacobrgardner:
I see what you're saying regarding the lower confidence bound. Certainly the mean in either case is positive.

I think the lower confidence bound going negative is more a consequence of plotting the latent function as though it were Gaussian in each slice. `E[exp(f)] - 2*std(exp(f))` isn't really a meaningful thing to be plotting because `exp(f)` is not Gaussian and therefore `E[exp(f)] - std(exp(f))` isn't truly a lower confidence bound in any reasonable sense of the phrase.

My concern with visualizing `exp(E[f])` and `exp(Var[f])` is that they aren't necessarily meaningful quantities from a statistical perspective; e.g., `exp(E[f])` doesn't necessarily accurately represent the expected scale parameter at any x, where as `E[exp(f)]` does.

My preference would be to plot `E[exp(f)]` because it is the actually statistically meaningful quantity. I'm not sure how to handle the confidence region. 

What do you think?

NinelK:
I think I can prove that the percentiles stay the same after gplink transformation. 
Gplink() transformation is monotonous (in this example non-decreasing). Let us consider, for example, the 5th percentile of a Gaussian distribution `p(f|x)`: f_lower.
Then, since gplink is non-decreasing:
- for all f<f_lower gplink(f)<gplink(f_lower) (there is 5% probability to get such `f`)
- for all f>f_lower gplink(f)>gplink(f_lower)  (95% probability)
which means that gplink(f_lower) is also the 5th percentile of a non-gaussian distribution `p(gplink(f)|x)`.

So when we plot `gplink(lower(f))` and `gplink(upper(f))` we get the correct confidence intervals.

jacobrgardner:
Okay, so in this case it's true that the CDFs are mappable like this. An obvious way to see this is that the distribution of interest is log-normal (thus, if `F(z) = \tau` where `F` is the CDF of the normal distribution, then `G(exp(z)) = \tau` where `G` is the CDF of the corresponding log-normal distribution).

However:

1. `E[exp(f)]` != `exp(E[f])`. E.g., `E[N(2, 0.25)] = 2`, but `E[Lognormal(2, 0.25)] = exp(2 + 0.125) != exp(2)`. Therefore, I believe the mean computation has to be corrected as I did. I don't think we need to use sampling though, as the mean of the scale function should be exactly `exp(mu(f) + \sigma^2(f)/2)` via lognormality.

2. For the plot, regardless of whether the confidence regions are numerically correct, the problem remains that we are treating the output distribution `exp(f)` as Gaussian by plotting the mean +/- two standard deviations. Maybe we should plot the marginal log Normal distribution at each x?

jacobrgardner:
Actually we do still need to do sampling for the correct mean I think because the underlying fs are generated from a correlated multivariate Normal, not independent normal distributions, so the exp(mu + sigma^2/2) formula doesn't apply

NinelK:
Yes, I understand that `E[exp(f)]` != `exp(E[f])`. However, I think that the mean of `p(gplink(f)|x)` is not required. The median and certain percentiles should work well as a visualization of a skewed distribution.

> the problem remains that we are treating the output distribution `exp(f)` as Gaussian by plotting the mean +/- two standard deviations

Mean +/- two standard deviations of a Gaussian distribution `p(f|x)` can be reinterpreted as the 2nd, 50th and 98th percentile, which will stay the same after gplink transformation.

> Maybe we should plot the marginal log Normal distribution at each x?

Yes, it is also an option. The only downside of this option is that the plot of the results in log-scale will not exactly repeat the first plot of the synthetic data, so it would not be immediately obvious that we found the `scale` function.

> underlying fs are generated from a correlated multivariate Normal, not independent normal distributions, so the `exp(mu + sigma^2/2)` formula doesn't apply

The marginals of the multivariate normal distribution are still Normal, so the formula should still apply. But again, I do not think that we necessarily need the mean of this non-gaussian distribution.

To sum up, I would simply label the blue lines in the original plot as the 2nd, 50th and 98th percentiles. Log-scale is also an option, but it requires a bit of explanation.

jacobrgardner:
@NinelK I reverted my changes to the confidence regions to compute them by warping the lower and upper bounds returned from the GP, but kept my change to plot the mean instead of the median scale function -- I'm happy to merge this as is now.

Do you think this is okay, or do you feel strongly that the median scale function should be plotted rather than the mean?

NinelK:
@jacobrgardner It looks like a reasonable solution, thank you. I am happy with this version.

PR #750, opened 2019-06-23
Remove Cholesky related TODO and cholesky_solve util function

vishwakftw:
Since PyTorch >= 1.1, I thought it would be appropriate to update / clear these.


Balandat:
Thanks for the PR. There is a number of other call sites for `cholesky_solve` that need to be updated though - do you want to give that a shot or should we take care of it?

vishwakftw:
I'll fix them, I'm sorry that I forgot about the call sites.

vishwakftw:
@Balandat this should be good to go now.

Balandat:
looks good, thanks!

Issue #744, opened 2019-06-21
[Bug] Interval constraint miscalculates bounds


MichaelDoron:
# 🐛 Bug

The transform function in the Interval constraint class allows tensors to be above the upper bound.

## To reproduce

** Code snippet to reproduce **

import torch
import gpytorch

lower_bound = 0.2
upper_bound = 0.3
tensor = torch.Tensor([lower_bound + upper_bound + 0.1])

constrained_tensor = gpytorch.constraints.Interval(lower_bound = lower_bound, upper_bound=upper_bound).transform(tensor)
assert constrained_tensor >= lower_bound, 'constrained tensor smaller than lower bound'
assert constrained_tensor <= upper_bound, 'constrained tensor larger than upper bound'

** Stack trace/error message **

AssertionError: constrained tensor larger than upper bound

## Expected Behavior

We would want the constrained tensor to be between lower bound and upper bound

## System information

**Please complete the following information:**
-  0.3.2
-  1.1.0
-  macOS

jacobrgardner:
Fixed in #751

PR #751, opened 2019-06-23
Fix Interval class upper bound transform for bug #744

MichaelDoron:
Fix Interval class upper bound transform for bug #744. The new code multiplies the tensor by (upper_bound - lower_bound) before adding lower_bound, thus keeping the tensor between bounds.

### DIFF
--- a/gpytorch/constraints/constraints.py
+++ b/gpytorch/constraints/constraints.py
@@ -94,7 +94,7 @@ class Interval(Module):
                     "GreaterThan or LessThan instead."
                 )
 
-        transformed_tensor = (self._transform(tensor) * self.upper_bound) + self.lower_bound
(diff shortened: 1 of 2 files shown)


gpleiss:
This seems reasonable, thanks for the catch! You'll also have to change the `inverse_transform` function as well - this is what's causing the errors.

gpleiss:
@MichaelDoron it also looks like the values in the two failing tests are wrong. Would you be able to fix those as well? (It should be as simple as modifying the values in lines 31, 51, and 52 in `test/constraints/test_constraints.py`._

MichaelDoron:
@gpleiss Sure, thanks for the guidance. I kept the explicit values for the upper and lower bounds in the tests for readability (e.g. `actual_value = ((5.0 - 1.0) * sigmoid(v)) + 1.0` instead of `actual_value = (4.0 * sigmoid(v)) + 1.0`)

PR #765, opened 2019-06-28
Minor cleanup of psd_safe_cholesky

Balandat:
- avoids re-cloning `A` during the jitter loop (to help perf)
- emits a "cholesky_cuda: singular U." RuntimeError (instead of an unnamed one) to be more consistent with the cpu cholesky failures (this really needs to be fixed upstream...)
- handle jitter variable in a more concise way


Balandat:
Actually, just realized this changes the functionality regarding adding jitter. Will update.

Issue #766, opened 2019-06-28
[Bug] Incorrect gradients for kernels using torch.cdist


sdaulton:
# 🐛 Bug
Auto-differientiating torch.cdist on the GPU is broken (https://github.com/pytorch/pytorch/issues/22353). This means kernels using torch.cdist have incorrect gradients on the GPU

## To reproduce
[cdist_issue_gpytorch.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3340948/cdist_issue_gpytorch.ipynb.txt)

## Expected Behavior

Numerical and analytic gradients match

## System information

**Please complete the following information:**
GPyTorch Version: master
PyTorch Version (e.g., 1.0): master
OS (e.g., Linux): CentOS Linux 7 (Core)
CUDA/cuDNN version:CUDA Version 9.2.88
Tesla M40

sdaulton:
cc @Balandat @danielrjiang

Balandat:
It seems that the straightforward solution to this is to add a check that the tensors not be cuda tensors here: https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/kernels/kernel.py#L25

Balandat:
And to make sure this gets fixed upstream before pytorch 1.2.

jacobrgardner:
The newest non cdist version of the distance computation is pretty darn fast even in low data settings. Let's either just disable cdist for GPU tensors as max suggests or disable cdist entirely for now.

@Balandat do we need to release 0.3.4 as a result of this?

Balandat:
> @Balandat do we need to release 0.3.4 as a result of this?

That would be great. Let's make sure we understand the implications of using either the custom or cdist implementation first though, in particular regarding batch mode (see my comment on the PR).

Balandat:
@jacobrgardner I didn't get to look into this much further (not sure I will anytime soon). Do you mind putting out a 0.3.4 at this point while we're waiting for cdist to get implemented properly on the pytorch end?

jacobrgardner:
@Balandat Sure thing, I'll put out 0.3.4 first thing tomorrow.

Balandat:
Great, thanks. I checked about prioritizing fixing up cdist (both gradients and perf) with the pytorch folks, let's see what they say.

PR #767, opened 2019-06-28
Only use torch.cdist on CPU

sdaulton:
Resolves https://github.com/cornellius-gp/gpytorch/issues/766, while issue is going in PyTorch upstream

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -22,7 +22,13 @@ class Distance(torch.nn.Module):
         self._postprocess = postprocess_script
 
     def _sq_dist(self, x1, x2, postprocess, x1_eq_x2=False):
-        if hasattr(torch, 'cdist') and x1.dim() < 3 and x1.size(-2) <= 512 and x2.size(-2) <= 512:
+        if (
+            hasattr(torch, 'cdist')
+            and x1.dim() < 3
+            and x1.size(-2) <= 512
+            and x2.size(-2) <= 512
+            and not (x1.is_cuda or x2.is_cuda)
+        ):
             # torch cdist doesn't support batch, also only available after pytorch 1.1
             # cdist is a bit faster on small tensors. Check again after pytorch PR #20605
             res = torch.cdist(x1, x2).pow(2)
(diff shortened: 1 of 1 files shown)


gpytorch/kernels/kernel.py
```diff
@@ -22,7 +22,14 @@ def __init__(self, postprocess_script=default_postprocess_script):
         self._postprocess = postprocess_script
 
     def _sq_dist(self, x1, x2, postprocess, x1_eq_x2=False):
-        if hasattr(torch, 'cdist') and x1.dim() < 3 and x1.size(-2) <= 512 and x2.size(-2) <= 512:
+        if (
+            hasattr(torch, 'cdist')
```
Balandat: Note that we can remove the `x1.dim() < 3` check since `torch.cdist` does support batch mode now. Or is there a reason to keep this @jacobrgardner, @gpleiss?

Also, interestingly gradients are fine for batched tensors. So we can use torch.cdist for small inputs even on the GPU if batched.

Finally how about defining the following helper so we can keep this logic in one place?

```
def _use_cdist(x1: Tensor, x2: Tensor) -> bool:
    return (
        hasattr(torch, 'cdist')
        and x1.size(-2) <= 512
        and x2.size(-2) <= 512
        # TODO: remove when https://github.com/pytorch/pytorch/issues/22353 is fixed
        and not (x1.is_cuda or x2.is_cuda)
    )
```

jacobrgardner: I'm generally inclined to relegate cdist to the very small non batch setting because of the extreme performance issues with it on larger matrices (e.g., it is still up to 40x slower than our custom distance code).

I would be okay with a helper for when to use cdist. My hope is that the PyTorch team can eventually just make cdist the superior option in all settings.

Let me know when you want me to put 0.3.4 out.

Balandat: > My hope is that the PyTorch team can eventually just make cdist the superior option in all settings.

At the very least one could take the non-cdist code here as inspiration and port the logic into ATen (possibly with additional optimizations). 

> Let me know when you want me to put 0.3.4 out.

Let me do some digging how perf compares in batch mode, then I'll put up a PR with a helper that cleans this up. After that goes in we can put out 0.3.4.

Issue #724, opened 2019-06-05
[Examples] Standardize Predictors for ARD


wjmaddox:
This isn't a major issue, but examples should that data should be standardized for ARD (and most other kernels) best practices and numerical stability. Otherwise, zero gradients for lengthscales can appear with very high variance data (e.g. skillcraft).

```python
# model definition
class GPRegressionModel(gpytorch.models.ExactGP):
        def __init__(self, train_x, train_y, likelihood):
            super(GPRegressionModel, self).__init__(train_x, train_y, likelihood)
            self.mean_module = gpytorch.means.ConstantMean()
            self.covar_module = gpytorch.kernels.RBFKernel(ard_num_dims=train_x.size(1))
        def forward(self, x):
            mean_x = self.mean_module(x)
            covar_x = self.covar_module(x)
            return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)

# generate very unbalanced data
torch.random.manual_seed(1000)
train_x = torch.cat([torch.randn(30,1)*5000, torch.randn(30,1)*0.001], dim=1)
train_y = torch.randn(30)

# and standardize it
train_x_std = train_x / train_x.std(0)
```

Then attempt to compute log probs for both non-standardized and standardized predictors:
```python
lh = gpytorch.likelihoods.GaussianLikelihood()
model = GPRegressionModel(train_x, train_y, lh)
mll = gpytorch.mlls.ExactMarginalLogLikelihood(lh, model)
loss = -mll( model(train_x), train_y)
loss.backward()
print(model.covar_module.raw_lengthscale.grad)
# returns tensor([[0., 0.]])
```

```python
lh = gpytorch.likelihoods.GaussianLikelihood()
model = GPRegressionModel(train_x, train_y, lh)
mll = gpytorch.mlls.ExactMarginalLogLikelihood(lh, model)
loss = -mll( model(train_x_std), train_y)
loss.backward()
print(model.covar_module.raw_lengthscale.grad)
# returns tensor([[-0.0136,  0.0462]])
```

Noticed from attempting to use ARD on skillcraft dataset by modifying the fast pred var [example](https://github.com/cornellius-gp/gpytorch/blob/master/examples/01_Simple_GP_Regression/Simple_GP_Regression_With_LOVE_Fast_Variances_CUDA.ipynb)

r69shabh:
I've submitted a PR to address this issue: #2751

## Root Cause

[...]

PR #781, opened 2019-07-11
Subtract mean of x1 from x1 and x2 in distance computation.

### LINKED ISSUE #780: [Bug] RBF Kernel is sometimes not positive definite
# 🐛 Bug

When I create a RBFKernel with 512 evenly spaced points between 500 and 501, and form the resulting kernel matrix, the matrix seems numerically positive definite (minimum eigenvalue is -1e-6). But, when I do the same with 513 points, the kernel matrix is very far from positive definite (minimum eigenvalue is -0.1). This behavior seems to crop up when using 513 points or more -- everything seems fine when using 512 or less; the results are also dependent on the range from which the datapoints are generated.

Is this a bug, or am I grossly misusing some feature?

## To reproduce

```python
import numpy as np
import torch
import gpytorch

upper = 500
lower = 501
covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.RBFKernel())
covar_module.base_kernel.lengthscale = 1.0
covar_module.outputscale = 1.0


# Evaluate a RBF kernel on 512 evenly spaced points between 100 and
#  101. The resulting kernel matrix is numerically PSD (smallest eigenvalue
#  is -1e-6)
X = torch.linspace(lower, upper, 512)
covarX = covar_module(X, X).numpy()
print('Min eigenvalue with 512 points', np.linalg.eigvals(covarX).min())

# Evaluate a RBF kernel on 513 evenly spaced points between 100 and
#  101. The resulting kernel matrix is *not* numerically PSD (smallest eigenvalue
#  is -0.1)
X2 = torch.linspace(lower, upper, 513)
covarX2 = covar_module(X2, X2).numpy()
print('Min eigenvalue with 513 points', np.linalg.eigvals(covarX2).min())

```

The output that I get from the above code is:
```
Min eigenvalue with 512 points -1.1961828e-06
Min eigenvalue with 513 points -0.18964253
```

## Expected Behavior

I would expect the minimum eigenvalue with 513 points to not be so negative.

## System information

-  GPyTorch Version: 0.3.3
- PyTorch Version: 1.1.0 
- Computer OS: Ubuntu 18.04

jacobrgardner:
This makes kernel computations more stable in the regime where feature values have very large magnitude but distances are still on the order of the lengthscale init.

In general, it is still best practice for us to recommend users to normalize their data because this does *not* solve the problem where the lengthscale initialization is probably wrong on very large data (e.g., this solves the issue when the data is between [500, 501] instead of [0, 1], but not when the data is between [500, 600] and the lengthscale is still initialized to `softplus(0)`.

Fixes #780

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -34,6 +34,10 @@ class Distance(torch.nn.Module):
             # cdist is a bit faster on small tensors. Check again after pytorch PR #20605
             res = torch.cdist(x1, x2).pow(2)
         else:
+            adjustment = x1.mean(-2, keepdim=True)
+            x1 = x1 - adjustment
+            x2 = x2 - adjustment  # x1 and x2 should be identical in all dims except -2 at this point
+
             # Compute squared distance matrix using quadratic expansion
             x1_norm = x1.pow(2).sum(dim=-1, keepdim=True)
             x1_pad = torch.ones_like(x1_norm)


Balandat:
> In general, it is still best practice for us to recommend users to normalize their data

This problem is particularly acute for the standard models we ship with BoTorch, as the priors assume normalization. FWIW, I plan to add some warnings on the BoTorch end if the inputs aren't (close to) normalized/standardized: https://github.com/pytorch/botorch/issues/208

KeAWang:
FWIW I think warning users if the inputs aren't normalized should be moved into gpytorch since this really a low level issue. It would also help prevent user errors and issues ex: #724

Balandat:
> I think warning users if the inputs aren't normalized should be moved into gpytorch since this is really a low level issue

So gpytorch doesn't really have a rigid API for training data and pre-specified (deliberately so). So to make that generically useful this could become a little clumsy (but maybe it will be fine).

Given that the out-of-the box BoTorch models are much more opinionated it's easier to add that check there.

PR #782, opened 2019-07-11
Add __getitem__ to kernels, which allows slicing batch dimensions

### LINKED ISSUE #758: [Bug] Batch-indexing lazy covariance matrix fails for batch-mode ExactGP
# 🐛 Bug

Evaluating a batch-mode ExactGP with training inputs of shape `b x n x d` and targets of shape `b x n` on a set of test points with shape `m x d` yields a lazy_covariance_matrix with shape `b x m x m`. Indexing along the batch dimension fails.

## To reproduce

[batch_indexing_bug_lazy_covar.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3331406/batch_indexing_bug_lazy_covar.ipynb.txt)

## Expected Behavior

Selecting a single index of the batch dimension:

```
mvn.lazy_covariance_matrix[0, :, :]
```
should yield the lazy covariance matrix for the 0th batch (with shape `m x m`)

## System information
-  0.3.2
-  1.0.0a0

jacobrgardner:
Right now on master, there's a problem where if `lazy_evaluated_kt` is a `LazyEvaluatedKernelTensor`, then doing for example `lazy_evaluated_kt[0, :, :]` will correctly slice `x1` and `x2` from `b x n x d` to `n x d`, but if the kernel itself is batch mode, the size of the sliced tensor will still be `b x n x m` because the kernel call on the `n x d` data will do broadcasting.

In other words, `lazy_evaluated_kt[0, :, :]` currently means "Take the 0th batch of data, run it through my batch of GPs, and give me a `b x n x n` set of covariance matrices." What we probably want it to mean is "Take the 0th batch of data, run it through the 0th GP, and give me an `n x n`covariance matrix," since this is more consistent with how every other `LazyTensor` works for `__getitem__`.

The only way I could see to solve this issue was to add a `__getitem__` to kernels, so that `lazy_evaluated_kt[0, :, :]` now roughly corresponds to `kernel[0](x1[0, :, :], x2[0, :, :])`. 

I'm not wild about kernels having `__getitem__` methods, so I'm happy to take suggestions on alternate mechanisms for solving this problem.

Fixes #758

### DIFF
--- a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
@@ -118,7 +118,7 @@ class LazyEvaluatedKernelTensor(LazyTensor):
                 x2 = x2[(*batch_indices, row_index, dim_index)]
 
         if len(batch_indices) == 0 or all(ind == slice(None, None, None) for ind in batch_indices):
-            new_kernel = self.kernel  # Avoid unnecessary when we aren't explicitly indexing batch dims copying
+            new_kernel = self.kernel  # Avoid unnecessary copying when we aren't explicitly indexing batch dims
         else:
             new_kernel = self.kernel.__getitem__(batch_indices)
 
--- a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
@@ -117,8 +117,13 @@ class LazyEvaluatedKernelTensor(LazyTensor):
                 x2 = x2.expand(*([1] * len(batch_indices)), *self.x2.shape[-2:])
                 x2 = x2[(*batch_indices, row_index, dim_index)]
 
+        if len(batch_indices) == 0 or all(ind == slice(None, None, None) for ind in batch_indices):
+            new_kernel = self.kernel  # Avoid unnecessary when we aren't explicitly indexing batch dims copying
+        else:
+            new_kernel = self.kernel.__getitem__(batch_indices)
+
         # Now construct a kernel with those indices
-        return self.__class__(x1, x2, kernel=self.kernel, last_dim_is_batch=self.last_dim_is_batch, **self.params)
+        return self.__class__(x1, x2, kernel=new_kernel, last_dim_is_batch=self.last_dim_is_batch, **self.params)
 
     def _matmul(self, rhs):
         # This _matmul is defined computes the kernel in chunks
--- a/gpytorch/kernels/additive_structure_kernel.py
+++ b/gpytorch/kernels/additive_structure_kernel.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 from .kernel import Kernel
+from copy import deepcopy
 
 
 class AdditiveStructureKernel(Kernel):
--- a/gpytorch/kernels/additive_structure_kernel.py
+++ b/gpytorch/kernels/additive_structure_kernel.py
@@ -50,3 +50,9 @@ class AdditiveStructureKernel(Kernel):
 
     def num_outputs_per_input(self, x1, x2):
         return self.base_kernel.num_outputs_per_input(x1, x2)
+
+    def __getitem__(self, index):
+        new_kernel = deepcopy(self)
+        new_kernel.base_kernel = self.base_kernel.__getitem__(index)
+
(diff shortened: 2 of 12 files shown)


gpleiss:
Honestly, this might be the best idea. However, we'd have to write custom getitems for a couple of our other kernels (Periodic, SpectralMixture, etc.)

The only other thing that I could think of is checking the size of the output after a getitem. If the sizes do not match, then we can assume that there was some sort of batch dimension added, and then we can do a further getitem on the output?

I'm not sure which idea I like better. The getitem on the kernel seems a bit cleaner, but does require more implementation effort. Checking the size of the result and doing additional getitem work as necessary is a bit messier, but requires less implementation effort on behalf of kernel implementers.

jacobrgardner:
@gpleiss @Balandat This should be good to go now.

I ended up writing a more general `Kernel.__getitem__` that should work for *most* base kernels. It works for all kernels currently in the repo that are not composite (i.e., that have base kernels).

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -117,8 +117,13 @@ def _getitem(self, row_index, col_index, *batch_indices):
                 x2 = x2.expand(*([1] * len(batch_indices)), *self.x2.shape[-2:])
                 x2 = x2[(*batch_indices, row_index, dim_index)]
 
+        if len(batch_indices) == 0 or all(ind == slice(None, None, None) for ind in batch_indices):
+            new_kernel = self.kernel  # Avoid unnecessary when we aren't explicitly indexing batch dims copying
```
jacobrgardner:
```suggestion
            new_kernel = self.kernel  # Avoid unnecessary copying when we aren't explicitly indexing batch dims
```

jacobrgardner:
@sdaulton Yeah, at the moment this PR fixes things for the other direction (e.g., you have a lazy evaluated tensor that you want to index `A[0, :, :]` to reduce the batch dimensionality).

How are you ending up with a LazyEvaluatedKernelTensor with more batch dimensions than the kernel has? Are you just calling a `b` batch kernel with `b1 x b` batch data? I'm surprised this doesn't currently raise an error in the kernel call itself.

In terms of solving the issue, the only way I could see to go about this would be to expand all parameters of the kernel to the appropriate size and hope that automatically gives us a kernel that does the right thing in multi batch. Either that, or Kernels now define an `expand` method...

jacobrgardner:
> 1. Can we make `Kernel.__getitem__` recursively call `__getitem__` on any of it’s modules that are Kernels? Similarly to the way that `.to` or `.apply` works on `torch.nn.Modules`? This will reduce the amount of special cases that we have in this PR.

Yeah I think this is reasonable -- with this, pretty much every kernel can use the default implementation. I'll get this done.

> 1. Can we confirm that the `deepcopy` isn’t causing any huge memory issues?

We deepcopy kernels etc in other use cases already on master (for example, fantasy observations). 

It won't get called more often than the fantasy models, so I'm not worried about repeated deepcopy calls either. A very limited number of code paths will actually call Kernel.__getitem__. At present, the only use case that I can think of is (1) calling a batch GP on a batch of data, and then (2) indexing the predictive covariance matrix `covar[0, :, :]` to get the covariance for one (or a subset) of the functions.

sdaulton:
> How are you ending up with a LazyEvaluatedKernelTensor with more batch dimensions than the kernel has? Are you just calling a `b` batch kernel with `b1 x b` batch data? I'm surprised this doesn't currently raise an error in the kernel call itself.

Yes, this is a common scenario in BoTorch using a `BatchedMultiOutputGPyTorchModel`, which is a batch (shape `b`) GP, where each batch corresponds to an independent output. This model is then evaluated on some test data with additional batch dimensions (e.g `b' x b x n x d` or `b' x 1 x n x d`).

> In terms of solving the issue, the only way I could see to go about this would be to expand all parameters of the kernel to the appropriate size and hope that automatically gives us a kernel that does the right thing in multi batch. Either that, or Kernels now define an `expand` method...

Yeah, this seems reasonable.

jacobrgardner:
@gpleiss This is probably all set.

@sdaulton I'll open up a second PR that adds something like `expand` functionality to kernels to handle the second case you mentioned. It seems orthogonal to #785, which is the issue this PR closes.

PR #790, opened 2019-07-16
Use QR decomposition for computing pseudoinverse in fantasy strategies

Balandat:
More numerically stable. Also adds jitter if the matrix is singular.

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -6,16 +6,22 @@ import torch
 
 from .. import settings
 from ..lazy import (
-    InterpolatedLazyTensor, MatmulLazyTensor, RootLazyTensor, SumLazyTensor, ZeroLazyTensor,
-    lazify, delazify, LazyEvaluatedKernelTensor, NonLazyTensor, BatchRepeatLazyTensor
+    InterpolatedLazyTensor,
+    MatmulLazyTensor,
+    RootLazyTensor,
+    SumLazyTensor,
+    ZeroLazyTensor,
+    lazify,
+    delazify,
+    LazyEvaluatedKernelTensor,
+    NonLazyTensor,
+    BatchRepeatLazyTensor,
 )
 from ..utils.interpolation import left_interp, left_t_interp
 from ..utils.memoize import cached, add_to_cache
 
 
-def prediction_strategy(
-    train_inputs, train_prior_dist, train_labels, likelihood
-):
+def prediction_strategy(train_inputs, train_prior_dist, train_labels, likelihood):
     train_train_covar = train_prior_dist.lazy_covariance_matrix
     if isinstance(train_train_covar, LazyEvaluatedKernelTensor):
         cls = train_train_covar.kernel.prediction_strategy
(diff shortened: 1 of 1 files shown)


gpytorch/models/exact_prediction_strategies.py
```diff
@@ -196,27 +199,29 @@ def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_
...
         new_root[..., m:, n:] = schur_root
 
         # Use pseudo-inverse of Z as new inv root
-        # TODO: Replace pseudo-inverse calculation with something more stable than normal equations once
-        # one of torch.svd, torch.qr, or torch.pinverse works in batch mode.
-        cap_mat = new_root.transpose(-2, -1).matmul(new_root)
-        if cap_mat.requires_grad or new_root.requires_grad:
+        # TODO: Consider replacing pseudo-inverse calculation with torch.svd / torch.pinverse once these
+        # methods work in batch mode.
+        Q, R = torch.qr(new_root)
```
Balandat: This is where the interesting stuff is, sorry for mushing this together with the black reformats.

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -196,27 +199,29 @@ def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_
...
+        Rdiag = torch.diagonal(R, dim1=-2, dim2=-1)
+        # if R is almost singular, add jitter (Rdiag is a view, so this works)
+        zeroish = Rdiag.abs() < 1e-6
+        if torch.any(zeroish):
+            Rdiag[zeroish] = 1e-6
+        if R.requires_grad or Q.requires_grad:
```
jacobrgardner: `R` is already triangular, so we should be able to compute `R^{-1}Q^{\T}` with `torch.trtrs`.

jacobrgardner: Sorry, `torch.traingular_solve` as it is now called :-)

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -196,27 +199,29 @@ def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_
...
-        new_root[..., :m, n:] = upper_right
         new_root[..., m:, :n] = lower_left
         new_root[..., m:, n:] = schur_root
 
         # Use pseudo-inverse of Z as new inv root
-        # TODO: Replace pseudo-inverse calculation with something more stable than normal equations once
-        # one of torch.svd, torch.qr, or torch.pinverse works in batch mode.
-        cap_mat = new_root.transpose(-2, -1).matmul(new_root)
-        if cap_mat.requires_grad or new_root.requires_grad:
+        # TODO: Consider replacing pseudo-inverse calculation with torch.svd / torch.pinverse once these
```
jacobrgardner: regarding the todo comment, `torch.qr` should always be preferable to `torch.svd`, being much (probably ~10x) faster and comparably stable. `torch.pinverse` currently uses `torch.svd` internally, but this is a bad thing and it should use `qr`.

gpytorch/models/exact_prediction_strategies.py
Balandat: ha I was looking for that but didn't find it, and I didn't want to manually massage R into an LU factorization... thanks

Balandat:
Changed this to fall back to old implementation when running on stable (not backporting the jitter though).

PR #793, opened 2019-07-17
Update linear_mean.py

techshot25:
Shape mismatch, weights and inputs are both rectangular matrices of the same shape.


jacobrgardner:
It's correct as is - if x is `n x d` then the weights should be `d x 1`

jacobrgardner:
Closing this unless we believe something is actually wrong with the current state.

PR #798, opened 2019-07-19
Fix sigmoid deprecation

dme65:
- Importing sigmoid from ```torch``` instead of ```torch.nn.functional``` as the latter is deprecated
- Fixing a dtype error in ```PolynomialKernelGrad```


jacobrgardner:
Rebased on to master (where the PolynomialKernelGrad stuff was already fixed).

PR #799, opened 2019-07-19
Remove legacy functions

jacobrgardner:
This PR converts all GPyTorch Functions to the non legacy format, and removes `NormalCDF`, which torch now has native support for via `torch.distributions.Normal(0., 1.).cdf(...)`


gpytorch/functions/_log_normal_cdf.py
```diff
@@ -1,12 +1,13 @@
...
 import math
-from . import NormalCDF
 from torch.autograd import Function
 import torch
+from torch.distributions import Normal
 
 
 class LogNormalCDF(Function):
+    @staticmethod
     def forward(self, z):
```
Balandat: `ctx` to follow consistent convention?

gpytorch/functions/_log_normal_cdf.py
```diff
@@ -90,11 +91,12 @@ def forward(self, z):
...
             self.numerator = numerator
 
-        log_phi_z.masked_scatter_(z_is_ordinary, torch.log(NormalCDF().forward(z.masked_select(z_is_ordinary))))
+        log_phi_z.masked_scatter_(z_is_ordinary, torch.log(Normal(0., 1.).cdf(z.masked_select(z_is_ordinary))))
 
         self.save_for_backward(z, log_phi_z)
         return log_phi_z
 
+    @staticmethod
     def backward(self, grad_output):
```
Balandat: `ctx`

jacobrgardner: Good catch, updated.

Balandat:
thanks for the quick fix!

PR #818, opened 2019-07-31
fix tex syntax in a non-raw docstring for LazyTensor

dougalsutherland:
Minor fix; these backslashes should be doubled. As is, it's putting a tab character in mathjax, so not rendering correctly. (Could instead make it a raw string, of course.)


jacobrgardner:
Thanks!

PR #821, opened 2019-08-01
Add a "Newton-Girard additive kernel" of Duvenaud et al 2011

idelbrid:
Paper: https://arxiv.org/abs/1112.4394

This branch implements the "Additive Gaussian Process" kernel from the link above which uses the Newton-Girard formulae to efficiently compute higher-order additive kernels. The implementation is simple, just wrapping a base kernel, similar to AdditiveStructureKernel and in the forward computes the higher-order terms using the Newton-Girard formulae. This branch also implements some fairly basic tests for it.

Since this kernel can achieve state-of-the-art regression performance, it might be nice to include this kernel in GPyTorch.

### DIFF
--- a/gpytorch/kernels/newton_girard_additive_kernel.py
+++ b/gpytorch/kernels/newton_girard_additive_kernel.py
@@ -35,7 +35,6 @@ class NewtonGirardAdditiveKernel(Kernel):
         self.outputscale_constraint = outputscale_constraint
         self.outputscale = [1 / self.max_degree for _ in range(self.max_degree)]
 
-
     @property
     def outputscale(self):
         return self.raw_outputscale_constraint.transform(self.raw_outputscale)
@@ -64,8 +63,7 @@ class NewtonGirardAdditiveKernel(Kernel):
         # e_n = torch.ones(self.max_degree+1, *kern_values.shape[1:], device=kern_values.device)  # includes 0
         # e_n: elementary symmetric polynomial of degree n (e.g. z1 z2 + z1 z3 + z2 z3)
         # e_n is R x n x n, and the array is properly 0 indexed.
-
-        e_n = torch.empty(self.max_degree+1, *kern_values.shape[1:], device=kern_values.device)
+        e_n = torch.empty(self.max_degree + 1, *kern_values.shape[1:], device=kern_values.device)
         e_n[0, ...] = 1.0
 
         # power sums s_k (e.g. sum_i^num_dims z_i^k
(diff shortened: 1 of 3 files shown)


jacobrgardner:
I'll take a look! The doc build failure is unrelated and is now fixed on master. If you rebase, that should resolve the issue.

test/kernels/test_newton_girard_additive_kernel.py
```diff
@@ -0,0 +1,149 @@
...
+from gpytorch.kernels import RBFKernel, ScaleKernel, AdditiveKernel, NewtonGirardAdditiveKernel
+from gpytorch.likelihoods import GaussianLikelihood
+from gpytorch.mlls import ExactMarginalLogLikelihood
+from gpytorch.models import ExactGP
+from gpytorch.means import ConstantMean
+from gpytorch.distributions import MultivariateNormal
+import torch
+
+
+class TestNewtonGirardAdditiveKernel(TestCase):
```
jacobrgardner: There's a `gpytorch.test.BaseKernelTestCase` that you can extend that, if you define the `create_kernel_no_ard` and/or `create_kernel_ard` will also run our standard suite of tests for things like batch mode support and handling active_dims correctly.

test/kernels/test_newton_girard_additive_kernel.py
```diff
@@ -0,0 +1,149 @@
...
+            def forward(self, x):
+                mean_x = self.mean_module(x)
+                covar_x = self.covar_module(x)
+                return MultivariateNormal(mean_x, covar_x)
+
+        model = TestGPModel(data, target, GaussianLikelihood(), ScaleKernel(AddK))
+        optim = torch.optim.Adam(model.parameters(), lr=0.1)
+        mll = ExactMarginalLogLikelihood(model.likelihood, model)
+        model.train()
+        for i in range(10):
```
jacobrgardner: Maybe only 2 steps of optimization for the unit test if we aren't going to assert on a metric like test ll/rmse?

gpytorch/kernels/newton_girard_additive_kernel.py
```diff
@@ -0,0 +1,89 @@
...
+        kern_values = self.base_kernel(x1, x2, diag=diag, last_dim_is_batch=True, **params)
+        # last dim is batch, which gets moved up to pos. 1
+        shape = [-1, 1, 1, 1] if not diag else [-1, 1, 1]
+        kvals = torch.range(1, self.max_degree, device=kern_values.device).reshape(*shape)
+        # kvals = R x 1 x 1 x 1 (these are indexes only)
+
+        # e_n = torch.ones(self.max_degree+1, *kern_values.shape[1:], device=kern_values.device)  # includes 0
+        # e_n: elementary symmetric polynomial of degree n (e.g. z1 z2 + z1 z3 + z2 z3)
+        # e_n is R x n x n, and the array is properly 0 indexed.
+        e_n = torch.empty(self.max_degree + 1, *kern_values.shape[1:], device=kern_values.device)
```
jacobrgardner: Does this work with arbitrary batch dimensions?

In general, we prefer kernels to be written assuming `x1` and `x2` are arbitrarily shaped except for the last two dimensions, e.g. assume `x1` can be `b1 x b2 x ... x n x d`, and assume any sub kernel calls you make return `b1 x b2 x ... x n x n` matrices in non-diag mode or `b1 x b2 x .... x n` vectors in diag mode. This is required for the kernel to work properly with things like deep GPs, some DKL models, etc.

idelbrid: No, it didn't as it was. I just updated the code to be batch agnostic. The update passes the base suite of unit tests.

jacobrgardner:
This looks good to me, thanks for the contribution!

Issue #822, opened 2019-08-01
[Bug] Upstream changes to tensor comparisons breaks things


Balandat:
# 🐛 Bug

After https://github.com/pytorch/pytorch/pull/21113 a bunch of tests are failing b/c of the change in tensor comparison behavior (return type from uint8 to bool). Creating this issue to track the fix.

Balandat:
It appears most of this happens in `batch_symeig` where we do `1 - mask` - will put up a fix shortly.

izdeby:
@Balandat, we have introduced '~' operator for masks to fix this issue. You should be able to just replace '1-mask' with '~mask'

Balandat:
Sure that works. There are a few other places in the code where this causes failures, but they should be easy enough to fix.

Balandat:
Fixed by #823

Balandat:
Ok of course there are now a bunch more warnings emitted now when using uint8 on master, so all the tests that count the warnings fail there ...

Balandat:
It's worse than that, the change actually breaks the jit script code for `_jit_linear_cg_updates`. So the correct thing is to use bool everywhere, and fix the tests that count the warnings...

Balandat:
Won't be able to fix all this properly right now, deferring to the above issue.

jacobrgardner:
@Balandat I'll take a look once the nightly build comes out and I can repro the errors without having to build pytorch from source.

Balandat:
@jacobrgardner sounds good, if you base this on #825 at least you won't have to deal with the warnings.

Balandat:
Fwiw, here's a notebook that shows the failure (commented out the jit decorators locally). 

It only happens when broadcasting. @izdeby, maybe broadcasting semantics for bool tensors are different?

[test_broadcasting_issues.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3459249/test_broadcasting_issues.ipynb.txt)

izdeby:
@Balandat, what exactly failed in broadcasting for you? there shouldn't be any difference in broadcasting semantics for bool tensors

Balandat:
@izdeby looks like this is a different issue related to changing sizes in `addcmul`, see #827

jacobrgardner:
For now do we think we could define like a `gpytorch.bool` type to be torch.bool if that will work (i.e. we are on master), and torch.uint8 otherwise?

jacobrgardner:
[...]

PR #827, opened 2019-08-02
Fix addcmul broadcasting

Balandat:
PyTorch does not allow (any more, for addcmul operator) to resize the input tensor. To allow the output shaped differently from the input, we have to save it in the temporary tensor. We cannot do `res = torch.addcmul(res, -1, a, c)` because functions are supposed to modify res in-place.

Addresses #822

Fixed by @VitalyFedyunin

### DIFF
--- a/gpytorch/lazy/cat_lazy_tensor.py
+++ b/gpytorch/lazy/cat_lazy_tensor.py
@@ -7,10 +7,7 @@ from . import delazify
 from ..utils.broadcasting import _mul_broadcast_shape, _matmul_broadcast_shape
 from ..utils.getitem import _noop_index
 from .. import settings
-
-
-# TODO: Use bool instead of uint8 dtype once pytorch #21113 is in stable release
-bool_compat = (torch.ones(1) > 0).dtype
+from ..utils.deprecation import bool_compat
 
 
 def cat(inputs, dim=0, output_device=None):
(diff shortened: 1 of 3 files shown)


Balandat:
@jacobrgardner, @gpleiss to sign off, not sure if this will have any perf implications.

Balandat:
So the tests actually don't fail, but they emit 37843 warnings, which means the excessive log length makes Travis report failure....

jacobrgardner:
@Balandat in what way is the output ending up differently shaped than the input? The residual and search direction for CG should have a fixed size over the course of the iterations. Is it just initialized to the wrong shape?

Balandat:
Not sure, I'm not very familiar with the cg code. Maybe there was some hidden bug before that just ended up working b/c of broadcasting? Note that this issue only happened for the broadcasting tests, e.g. `test_inv_matmul_matrix_broadcast`, but not `test_inv_matmul_matrix`.

Balandat:
@jacobrgardner If this works for you as a temporary fix can we merge this in? I'm not that comfortable with the linear CG implementation, I'd rather let you take a look whether there are any issues with the shape initialization.

Balandat:
argh great now this breaks the sphinx mocking :(

jacobrgardner:
@Balandat I pushed a commit that maintains usage of the in-place ops by expanding `result` to the correct shape at the beginning of linear CG.

Balandat:
What should we do about the mocking issue for the docs?

Balandat:
nice.

PR #839, opened 2019-08-15
Remove redundant checks in psd_safe_cholesky

vishwakftw:
This has been fixed in MAGMA 2.5.1, whose binaries are now available on anaconda.


Balandat:
Thanks for the PR - I was looking into this, but we internally haven't updated MAGMA yet, so I wanted to hold off for a little bit longer. I'll try to get MAGMA updated so we can merge this.

vishwakftw:
Sure no problem! Thank you.

PR #841, opened 2019-08-17
Add **kwargs to likelihood_i of ModelList

Balandat:
Allows to pass kwargs to the individual likeilihoods of a ModelList. This is useful e.g. for passing noise to models with FixedNoiseGaussianLikelihood.


Balandat:
@jacobrgardner Ideally there would be a way to pass kwargs from the `__call__` method of a `LikelihoodList` to the individual sub-likelihoods (https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/likelihoods/likelihood_list.py#L32-L33), but I'm not sure there is a clean way to do that. lmk if you have any ideas.

PR #842, opened 2019-08-18
Remove custom batch methods, smarter cholesky inverse, no dtype cast

Balandat:
This PR

1. Removes the custom `batch_svd` and `batch_qr` methods, since these are natively supported in pytorch 1.2.0
2. Replaces a `.inverse()` call on a cholesky factor with a triangular solve against the identity (using broadcasting across batches). This results in significant speedups for large matrices.
3. Removes casting to double before performing cholesky decomposition and cholesky solve. Profiling reveals that a non-trivial amount of time is being spent on these conversions. I believe these were added for numerical stability in the past, but with the changes to cholesky (both internal and upstream) this may not be an issue anymore.


gpytorch/lazy/lazy_tensor.py
```diff
@@ -400,7 +398,8 @@ def _cholesky(self):
         if evaluated_mat.requires_grad:
             evaluated_mat.register_hook(_ensure_symmetric_grad)
```
Balandat: @jacobrgardner looks like this has been merged in. Should we remove this?

Balandat: @vishwakftw, have all outstanding issues with the symmetric derivative been resolved?

vishwakftw: @Balandat yes, as far as I am aware, symmetric derivative issues have been resolved.

Balandat: great, I'll remove that hook in this PR as well then.

Balandat: hmm, actually, removing this causes more test failures than I have time to fix :)

PR #857, opened 2019-08-29
Fix in-place operation in get_fantasy_strategy

Balandat:
Previously, if one back-propagated through the fantasy model after jitter was added to the post-QR solve, this would result in autograd failures because of in-place modification of variables. This fixes that.


Balandat:
Failure is b/c travis tests on 1.1, will update

Issue #853, opened 2019-08-26
Non equal-sized grids for GridInterpolationKernel


eytan:
# 🚀 Feature Request

Currently GridInterpolationKernel assumes that the grid size is equal along all dimensions.  I would like to specify the grid size along the x,y or x,y,z axes.

## Motivation

**Is your feature request related to a problem? Please describe.**
In many spatiotemporal applications where GridInterpolationKernel would be very handy, one may have cases where the x and y widths that differ (e.g., for data that is longer than shorter), and may have much longer or shorter z widths (e.g., if time is measured by year or day or hour).
 
## Pitch

**Describe the solution you'd like**
I'd like to specify a tuple for `grid_size`, much like how I can specify the `grid_bounds`.

It looks like it could be easy for me to hack this by modifying `GridInterpolationKernel._create_grid()` but I am not sure if this would mess other things up and how you feel about allowing `grid_size` to optionally be a tuple.

**Describe alternatives you've considered**
I would consider using SGPR if I can't use GridInterpolationKernel.

eytan:
Awesome, thanks @jacobrgardner !

jacobrgardner:
Closed by #888.

PR #861, opened 2019-09-06
Grid size for each dim [WIP]

idelbrid:
* Enables `GridInterpolationKernel` to use different grid sizes for each dimension.

I took a stab at issue #853 by converting the `grid` property of GridInterpolatingKernel to a list of tensors. This change did branch into several modules, but the changes are very superficial. It could make it slightly slower since each dimension is not computed in batch this way, but the grids can't be very high dimensional regardless.

The tests I added to `test_grid_kernel.py` reveal, however, that, while the functionality is retained when the grid sizes are the same, **some bug causes it to break if the grid sizes are different from each other** (even [100, 101]. I will try to fix this bug tomorrow (Friday), but since I heard that a time-sensitive matter depends on this functionality, I wanted to go ahead and get this work in progress in front of maintainers in case someone (@KeAWang, @jacobrgardner ?) is able to finish this before I can. 

(I also added comments to the interpolation code to make it more readable for myself and perhaps others in the future.)

### DIFF
--- a/gpytorch/kernels/grid_interpolation_kernel.py
+++ b/gpytorch/kernels/grid_interpolation_kernel.py
@@ -5,6 +5,7 @@ from .grid_kernel import GridKernel
 from ..lazy import InterpolatedLazyTensor
 from ..models.exact_prediction_strategies import InterpolatedPredictionStrategy
 from ..utils.interpolation import Interpolation
+from collections.abc import Iterable
 
 
 class GridInterpolationKernel(GridKernel):
@@ -47,7 +48,7 @@ class GridInterpolationKernel(GridKernel):
     Args:
         - :attr:`base_kernel` (Kernel):
             The kernel to approximate with KISS-GP
-        - :attr:`grid_size` (int):
+        - :attr:`grid_size` (int, list of ints):
             The size of the grid (in each dimension)
         - :attr:`num_dims` (int):
             The dimension of the input data. Required if `grid_bounds=None`
@@ -83,11 +84,19 @@ class GridInterpolationKernel(GridKernel):
                     "num_dims ({}) disagrees with the number of supplied "
                     "grid_bounds ({})".format(num_dims, len(grid_bounds))
                 )
+        if isinstance(grid_size, int):
+            grid_sizes = [grid_size for _ in range(num_dims)]
+        elif isinstance(grid_size, Iterable):
+            grid_sizes = list(grid_size)
+            if len(grid_sizes) != num_dims:
+                raise RuntimeError("The number of grid sizes provided through grid_size do not match num_dims.")
+        else:
+            raise TypeError("grid_size is neither an integer nor an iterable.")
 
         # Initialize values and the grid
         self.grid_is_dynamic = grid_is_dynamic
         self.num_dims = num_dims
-        self.grid_size = grid_size
+        self.grid_sizes = grid_sizes
         self.grid_bounds = grid_bounds
         grid = self._create_grid()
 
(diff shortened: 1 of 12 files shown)


idelbrid:
...evidently there are also issues revealed by the integration tests.

gpytorch/kernels/grid_interpolation_kernel.py
```diff
@@ -97,18 +107,18 @@ def __init__(self, base_kernel, grid_size, num_dims=None, grid_bounds=None, acti
         self.register_buffer("has_initialized_grid", torch.tensor(has_initialized_grid, dtype=torch.bool))
 
     def _create_grid(self):
-        grid = torch.zeros(self.grid_size, len(self.grid_bounds))
+        grid = [torch.zeros(s) for s in self.grid_sizes]  # ID: now a jagged array of zeros
```
sdaulton: Unnecessary to initialize tensors here. This can just be an empty list and append tensors in the for loop.

gpytorch/kernels/grid_kernel.py
```diff
@@ -34,10 +34,16 @@ def __init__(self, base_kernel, grid, interpolation_mode=False, active_dims=None
...
         self.interpolation_mode = interpolation_mode
         self.base_kernel = base_kernel
-        self.register_buffer("grid", grid)
+        self.register_buffer_list('grid', grid)
         if not self.interpolation_mode:
             self.register_buffer("full_grid", create_data_from_grid(grid))
 
+    def register_buffer_list(self, base_name, tensors):
+        """Helper to register several buffers at once under a single base name"""
+        self.__setattr__(base_name, tensors)
```
sdaulton: This sets `self.grid` to point to `tensors`. But `register_buffer` in L45 makes copies of each tensor, so this pointer no longer points to the right tensors. One work around is remove this `setattr` and instead make a property for called `grid` that returns a list `[self.grid_0,...]`.
e.g.

```
@property
    def grid(self):
        return [getattr(self, "grid" + '_' + str(i)) for i in range(self.num_dims)]
```
Note you'll need to add a `num_dims` attribute

sdaulton:
I have noticed that training and inference is very slow using KissGP with n=1000 data points. An ExactGP is way faster. Is this expected or how can I improve performance? I am running this on a GPU.  cc @jacobrgardner @gpleiss @Balandat

[kiss_gp_perf.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3585485/kiss_gp_perf.ipynb.txt)

jacobrgardner:
@sdaulton Yeah, 1000 points in 3 dimensions is absolutely going to be faster with an exact GP than *any* of the approximate GP methods at this point. In general, I wouldn't recommend the use of approximate GP methods at all for regression at this point unless you have more than a couple hundred thousand data points.

sdaulton:
Cool, I actually have > 400k points, but was just testing this on a small example. However, this is still prohibitive slow especially for active learning, which I would like to use this model for, which involves lots of posterior evaluations and sampling. Even evaluating the posterior mean on 70k points takes 1 minute, using only with 1000 x 3 training points and a grid_size of 9. Any thoughts on how to speed this up?

gpytorch/kernels/grid_kernel.py
idelbrid: Ah, thanks for pointing this out, I didn't realize that `register_buffer` makes copies of tensors. Will fix.

jacobrgardner:
@sdaulton It'll be slower than exact GPs but it shouldn't be THAT slow. Running your notebook locally but changing things to use fp32 instead of fp64 since fp64 is terrible for GPUs gives me 13s to predict on 70k data points with the (1000, 3) training set.

sdaulton:
Maybe I have a lemon GPU, even with FP32 it still takes me 69s for predicting on 70k pts

KeAWang:
@sdaulton I ran your KISSGP notebook on a GPU and a CPU, and found the speeds to be around 20s for predicting on 70k test points with your 1000 `randn` train points when using fp16, much like what Jake found. Have you tried running it on a CPU instead?

jacobrgardner:
#888 Just got merged which adds this functionality.

Issue #725, opened 2019-06-05
[Feature Request] efficient logdet for KroneckerProductLazyTensor


mshvartsman:
# 🚀 Feature Request

It appears that KroneckerProductLazyTensor doesn't compute efficient logdets (as a product of the scaled log-determinants of the factors). This means it costs O(n^2m^2) for a 2-factor kronecker kernel with n x n and m x m factors instead of O(n^2 + m^2) even if we're using efficient inv_quads. 

This is fairly straightforward to add and would speed up multitask models and probably a bunch of other stuff. I'm happy to do it, just a bit confused on (a) what `reduce_inv_quad` does and (b) where the efficient inv_quad is happening right now and whether I can reuse it or also need to implement an efficient kron-solve. 

Simple example: 
```
import gpytorch
import torch
from gpytorch.functions import logdet, inv_matmul
from gpytorch.utils.cholesky import psd_safe_cholesky
LOG2PI = 1.8378770664093453

def lazy_matrixnormal_likelihood(X, R, C):
    """
    Likelihood for matrix-variate normal
    (https://en.wikipedia.org/wiki/Matrix_normal_distribution)
    using lazy logdet and inv_matmul
    """
    row_logdet = logdet(R)
    col_logdet = logdet(C)

    logdet_term = X.shape[0] * col_logdet + X.shape[1] * row_logdet

    trace_term = torch.trace(inv_matmul(R, X).matmul(inv_matmul(C, X.t())))

    return -0.5 * (X.shape[1] * X.shape[0] * LOG2PI + logdet_term + trace_term)


def lazy_kron_mvn(X, K):
    
    """likelihood for multivariate normal using lazy
    """
    logdet_term = logdet(K)

    trace_term = X.t().matmul(inv_matmul(K, X))
    return -0.5 * (X.shape[0] * LOG2PI + logdet_term + trace_term)


def chol_logdet(L):
    """
    log determinant of a matrix with cholesky factor L
    """
    return 2 * torch.sum(torch.log(torch.diag(L)))


def cholesky_solve(x, L):
    """
    Solve positive definite system using the cholesky factor 
    """
    y = torch.triangular_solve(x, L, upper=False).solution
    return torch.triangular_solve(y, L.t(), upper=True).solution

def nonlazy_matrixnormal_likelihood(X, R, C):
    cholr = psd_safe_cholesky(R)
    cholc = psd_safe_cholesky(C)

[...]

PR #872, opened 2019-09-16
Add Functional Kernel Learning

wjmaddox:
Hi,

Opening this PR for our NeurIPS paper, [Function Space Distributions over Kernels](http://roseyu.com/time-series-workshop/submissions/2019/timeseries-ICML19_paper_52.pdf) (sorry it's not on arxiv yet). Basically, we parameterize stationary kernels with (latent) Gaussian processes and then do MCMC over the latent Gaussian process while optimizing the hypers.

Done so far: 

- Integrate kernels ([single dimension](https://github.com/wjmaddox/gpytorch/blob/fkl/gpytorch/kernels/spectral_mixture_kernel.py) and [products](https://github.com/wjmaddox/gpytorch/blob/fkl/gpytorch/kernels/product_spectral_gp_kernel.py)) within gpytorch.kernels.
- Added [quadratic mean class](https://github.com/wjmaddox/gpytorch/blob/fkl/gpytorch/means/quadratic_mean.py) and corresponding LogRBF class.
- Example [notebooks](https://github.com/wjmaddox/gpytorch/tree/fkl/examples/15_Functional_Kernel_Learning) for single dimension, multiple dimensions, and multi-task.
- Example [notebook](https://github.com/wjmaddox/gpytorch/blob/fkl/examples/12_Pyro_LatentFunctionInference/FKL_LatentFunction_Regression.ipynb) demonstrating a SVI implementation using Pyro.

Design choices I'm not sure of:

- Added [gpytorch.samplers](https://github.com/wjmaddox/gpytorch/tree/fkl/gpytorch/samplers) to include support for elliptical slice sampling as well as sampling factories for our models. We could also PR Pyro to add elliptical slice sampling into their library and then use that implementation instead.
- Above LogRBF class - could be written specifically by the user.
- Our specific spectral initialization requires defining models within the kernel class. See [here](https://github.com/wjmaddox/gpytorch/blob/f02eb62231fbf5229ae94d9e809cdec0dd723688/gpytorch/kernels/spectral_gp_kernel.py#L51).

Let me know what changes to make,

Wesley


gpytorch/kernels/grid_kron_wrapper.py
```diff
@@ -0,0 +1,150 @@
...
+        # print("forward being hit")
+        grid = self.grid
+
+        if not self.interpolation_mode:
+            if len(x1.shape[:-2]):
+                full_grid = self.full_grid.expand(*x1.shape[:-2], *self.full_grid.shape[-2:])
+            else:
+                full_grid = self.full_grid
+
+        # print(x1)
```
jacobrgardner: Make sure to clean up commented / debug code here and elsewhere.

gpytorch/kernels/grid_kron_wrapper.py
```diff
@@ -0,0 +1,150 @@
...
+import torch
+import math
+from .. import settings
+from . import GridKernel
+
+from ..lazy import InterpolatedLazyTensor, KroneckerProductLazyLogDet, NonLazyTensor, lazify, delazify, ToeplitzLazyTensor
+from ..models.exact_prediction_strategies import InterpolatedPredictionStrategy
+from ..utils.interpolation import Interpolation
+
+class GridKronWrapper(GridKernel):
```
jacobrgardner: Is this class actually used anywhere? `WrappedSpectralKernel` is commented out above in the `kernels/__init__.py`.

gpytorch/kernels/grid_kron_wrapper.py
```diff
@@ -0,0 +1,150 @@
...
+                if dens is not None:
+                    self.set_latent_params(dens, idx)
+
+        k_uu = self.base_kernel(self.grid, self.grid, last_dim_is_batch=True)
+
+        left_interp_indices, left_interp_values = self._compute_grid(test_x)
+
+        ## this might be the identity, not sure.
+        right_interp_indices, right_interp_values = self._compute_grid(self.grid)
+
```
jacobrgardner: This many empty lines will fail the linter -- maybe run `black -l 120 <file>` on your code to help clean some of this up.

gpytorch/kernels/product_spectral_gp_kernel.py
```diff
@@ -0,0 +1,51 @@
+import torch
+
+from gpytorch.kernels import ProductKernel
+from torch.nn import ModuleList
+
+from .spectral_gp_kernel import SpectralGPKernel
+
+class ProductSpectralGPKernel(ProductKernel):
```
jacobrgardner: If the `SpectralGPKernel` itself correctly supports the `last_dim_is_batch` argument, we should be able to get away with something like a `ProductStructureKernel`. What we want to do is think of `n x d` data as actually `d x n x 1` batch data, compute a `d x n x n` kernel matrix and then product across the `d` "batch" dimension.

Since GP models run in batch mode, getting `SpectralGPKernel` to support batch inputs should be relatively straightforward, and this would make the multi-d version of this kernel basically `d` times faster.

gpytorch/means/fkl_means.py
```diff
@@ -0,0 +1,48 @@
+import torch
+import gpytorch
+from torch.nn.functional import softplus
+from gpytorch.priors import NormalPrior
+
+class LogRBFMean(gpytorch.means.Mean):
```
jacobrgardner: You define two classes called `LogRBFMean`, here and in `means/quadratic_mean.py`

gpytorch/means/fkl_means.py
```diff
@@ -0,0 +1,48 @@
...
+    def forward(self, input):
+        ## CHECK THE OUTPUT SHAPE ##
+        prob_out = torch.zeros_like(input)
+        for mx in range(self.num_mixtures):
+            dist = torch.distributions.Normal(loc=softplus(self.raw_sm_means[mx, 0, 0]),
+                                              scale=softplus(self.raw_sm_scales[mx, 0, 0]))
+            prob_out += dist.log_prob(input).exp() * softplus(self.raw_sm_weights[mx])
+            # print("prob out = ", prob_out)
+
+        return (prob_out + 1e-9).log().squeeze()
```
jacobrgardner:
```suggestion
        return prob_out.clamp_min_(1e-9).log().squeeze()
```

gpytorch/means/fkl_means.py
```diff
@@ -0,0 +1,48 @@
...
+        self.register_prior(name='lengthscale_prior', prior=NormalPrior(torch.zeros(1), 100.*torch.ones(1), transform=torch.nn.functional.softplus),
+            param_or_closure='lengthscale')
+
+    def forward(self, input):
+        # logrbf up to constants is: c - t^2 / 2l
+        out = self.constant - input.pow(2).squeeze(-1) / (2 * (softplus(self.lengthscale.view(-1)) + 1e-7) )
+        # print("output shape = ", out.shape)
+        return out
+
+class SM_Mean(gpytorch.means.Mean):
```
jacobrgardner: `SpectralMixtureMean`. Also, the file name should match the class you are importing (e.g., `spectral_mixture_mean.py`.

Also, do these means make sense to use with GP models outside of with FKL? If so, living here is fine. If not, they might live directly with the kernel.

gpytorch/lazy/lazy_kron_add_diag.py
```diff
@@ -0,0 +1,81 @@
...
+        noise = self._diag_tensor[0,0]
+        ### THIS WORKS FOR 2 DIMENSIONS NOW ###
+        sub_eigs = []
+        for lt in self._lazy_tensor.lazy_tensors:
+            sub_eigs.append(lt.evaluate().eig()[0][:, 0].unsqueeze(-1))
+
+        eigs = sub_eigs[0].matmul(sub_eigs[1].t())
+        return torch.log(eigs + noise).sum()
+
+    def KronSVD(self):
```
jacobrgardner: Functions are `all_lower_case`. Private methods `_start_with_underscore`.

gpytorch/lazy/kron_lazylogdet_lazy_tensor.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+
+import torch
+
+from . import KroneckerProductLazyTensor, DiagLazyTensor, NonLazyTensor
+
+class KroneckerProductLazyLogDet(KroneckerProductLazyTensor):
```
jacobrgardner: Also probably a private class in kronecker_product_plus_diag_lazy_tensor.py.

gpytorch/lazy/lazy_kron_add_diag.py
```diff
@@ -0,0 +1,81 @@
...
+        return inv_quad_term, logdet_term
+
+
+    def logdet(self):
+        res = 0
+        noise = self._diag_tensor[0,0]
+        ### THIS WORKS FOR 2 DIMENSIONS NOW ###
+        sub_eigs = []
+        for lt in self._lazy_tensor.lazy_tensors:
+            sub_eigs.append(lt.evaluate().eig()[0][:, 0].unsqueeze(-1))
```
jacobrgardner: In general, the efficiency of some of the numerics in this class can be improved. In particular, you are using `eig` here and `svd` elsewhere, but presumably the base lazy tensors are all symmetric, so at a minimum `symeig` should be used everywhere. 

In either case, you may also be able to use Cholesky or base `inv_quad_logdet` calls on the base LTs in place of eigendecompositions, see #725.

gpytorch/samplers/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+from .gibbs_alternating_sampler import *
+#from .elliptical_slice_sampling import *
+#from .slice_sampling import *
+from .elliptical_slice import *
+from .sgd import SGD
+from .mean_elliptical_slice import *
+#from .scipy_min import ScipyMinimize
+from .sgld import SGLD
+
+from .sampling_factories import ess_factory, ss_factory, ss_multmodel_factory
```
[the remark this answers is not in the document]
wjmaddox: Good point - from a black box perspective, it should integrate okay with NUTS (just will be slower throughout). Will start looking into it.

gpytorch/means/fkl_means.py
wjmaddox: Thanks! This one was supposed to be removed.

wjmaddox:
Realized I should have done a better job of removing the scalability (Grid/Kronecker structure) from the codebase and will leave that for a later time.

gpytorch/kernels/grid_kron_wrapper.py
wjmaddox: No I removed the `WrappedSpectralKernel` from the PR for now as it's code for scalability and not a base implementation.

gpytorch/kernels/product_spectral_gp_kernel.py
wjmaddox: One question here: we are attempting to include support for both product spectral kernels with shared latent GPs and independent latent GPs. Will `ProductStructureKernel` be able to support both cases?

eytan:
Hi @wjmaddox, I was wondering what your plans were for this PR.  At the very least, we'd love to be able to start using some of your slice sampling routines.  I know that @jacobrgardner has also been doing some work w/ having more examples of using GPyTorch w/ pyro + NUTS, so I wonder if he has any additional thoughts on what to do with the samplers.

wjmaddox:
I'm still working on it - the variational approach is basically done, but I was waiting on the Gpytorch changes to finalize that bit before working on Pyro + NUTS. 

It's still a bit unclear to me what the best way to integrate elliptical slice sampling is - namely, whether I should open a PR in Pyro or just attempt to support the samplers myself?

jacobrgardner:
If someone can demonstrate that ESS outperforms NUTS, then I think the right thing to do would be to write ESS as a pyro sampler and either add it upstream there if they are interested or have it live in the GPyTorch repo if it's too special case to live upstream.

Given that sampling with any pyro sampler is now supported as first class functionality, we'd need to see some strong evidence that building a second redundant code path is necessary before we should consider that.

wjmaddox:
I can start taking a look then. One issue that I remember seeing is related to both versions - how exactly does Pyro deal with the stochasticity in the log determinants when using NUTS?

jacobrgardner:
If we see stochasticity being a problem for sampling (which it seemed to be fine in our testing so far), then #929 might provide a solution.

With that said, fully Bayesian modelling for GPs is generally not super useful past a relatively small amount of data because the hyperparameter posteriors become highly peaked relatively quickly, so you may not even want to be using NUTS or ESS too far past the point where we use Cholesky by default anyways.

wjmaddox:
At least for FKL, it makes sense to use an MCMC approach to sample the latent GPs (as in the experiments of http://proceedings.mlr.press/v9/murray10a/murray10a.pdf) as we may not have a great intuition of the shape of the latent GP given the data. 

In few dimensions, we're also pretty unlikely to ever go beyond Cholesky size for that GP - the critical factor is actually in evaluating the likelihood term somewhat deterministically.

The distinction I'm trying to draw is that the latent GP isn't really a hyper-parameter so to speak.

jacobrgardner:
Hi @wjmaddox -- it looks like this is in mostly good shape. Do you have plans to put the finishing touches on this in the nearish future, or do you consider this closed in the short term?

wjmaddox:
Sorry, I lost track of this in the final set of NeurIPS updates. I was pretty close to finishing up but didn't have a chance to completely update to the new variational inference schemes. I'll attempt to look at it tomorrow and next week.

wjmaddox:
Sorry, I lost track on this before fixing the documentation. Because there's a ton of old wacky attempts to get it working before the last couple of commits, should I just make a new branch and PR with the working kernel, notebook, and unit test?

gpleiss:
@wjmaddox - i'd either squash the commit history or open up a new branch. Whichever is easier.

gpytorch/priors/gp_prior.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+
+from .torch_priors import MultivariateNormalPrior
+
+
+class GaussianProcessPrior(MultivariateNormalPrior):
```
gpleiss, line 6: I don't love having this prior as a public class - since its really specific to FKL and its name might cause confusion. Could you put it as a private class in the SpectralGPKernel file?

PR #880, opened 2019-09-24
Simplify code due to some upstream changes

vishwakftw:
- Use cholesky_solve unconditionally
- Use batched implementations for inverse, logdet

Note that this is partially bc-breaking, and requires at least a nightly build.

cc: @Balandat @jacobrgardner


gpleiss:
@vishwakftw - unfortunately, this code is failing for the stable release of PyTorch because there's no `cholesky_solve` derivative in v1.2 😢 

I'll make a quick fix, and if that passes I'll merge it in.

vishwakftw:
@gpleiss Thank you, and I was aware that it'd fail on the stable build.

gpleiss:
We can also just keep this up until the next stable PyTorch release :)

vishwakftw:
Sure no problem, it should be soon I presume, since the release branch has been cut.

Issue #879, opened 2019-09-24
SumLazyTensor supports adding torch tensors [Feature Request]


wjmaddox:
# 🚀 Feature Request

SumLazy Tensor should be able to handle adding tensors like NonLazyTensors do.

## Motivation

**Is your feature request related to a problem? Please describe.**
I saw this while attempting to have enforced jitter in the covariance in a variational strategy.

## Pitch

**Describe the solution you'd like**
SumLazyTensor allows the following code to work - should be a matter of adding another if statement to the __add__ method.

```
test1 = gpytorch.lazify(torch.randn(30, 30))

test2 = torch.randn(30,30)
res = test1 + test2
res + test2 # currently does not work but should work.
```

**Describe alternatives you've considered**

We could enforce both tensors here to be lazy tensors in the variational strategy instead. See [here](https://github.com/cornellius-gp/gpytorch/blob/e50b9878a090b41eaf38f371cab2b938d5b2ebbc/gpytorch/variational/variational_strategy.py#L216)

[NonLazyTensor](https://github.com/cornellius-gp/gpytorch/blob/e50b9878a090b41eaf38f371cab2b938d5b2ebbc/gpytorch/lazy/non_lazy_tensor.py#L71)  does allow this type of addition.
**Are you willing to open a pull request?**
Yes.

## Additional context

See proposed code above.

jacobrgardner:
Yeah, this is reasonable. In general, I think we should just have `__add__` and `__radd__` in LazyTensor itself support `LT + T` and `T + LT` ops.

The one "gotcha" is we'll need to special case things so that broadcasting still works -- e.g. `torch.randn(30, 1) + SomeArbitraryLazyTensor(30, 30)` is not as simple as `return SumLazyTensor(NonLazyTensor(torch.randn(30, 1), SomeArbitraryLazyTensor(30, 30))`. The first LT needs to be wrapped in some kind of `BroadcastingLazyTensor` whose _matmul does the equivalent of `torch.randn(30, 1).repeat(1, 30).matmul(...)`.

wjmaddox:
Good catch on the broadcasting semantics. It looks like we can use `torch.distributions.utils.broadcast_all` on the shapes to either build the BroadcastingLazyTensor properly or to assign it within the SumLazyTensor itself.

PR #886, opened 2019-09-28
Add ability to add torch.Tensor into SumLazyTensor

wjmaddox:
Provides a partial fix for #879. This PR works provided that the torch.tensor is the tensor that needs to be broadcasted and NOT the lazyTensor itself. 

Resolving lazyTensor broadcasting itself is a bigger issue, given the following code: 
```
res = lazify(torch.randn(30, 1)) + torch.randn(30, 30)
print(res.shape) # 30, 1 not 30, 30
```

Will attempt to look into more detail when I get another chance.


test/lazy/test_broadcasted_lazy_tensor.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
+
+import torch
+import unittest
+from gpytorch import lazify
+from gpytorch.test.lazy_tensor_test_case import LazyTensorTestCase
+
+class TestSumLazyTensorBroadcasting(unittest.TestCase):
```
gpleiss: It would be better if these tests were part of `test_sum_lazy_tensor` and used the LazyTensorTestCase.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1734,6 +1734,7 @@ def __mul__(self, other):
         return self.mul(other)
 
     def __radd__(self, other):
+        #TODO: see if this should be self.add
```
gpleiss: I thought this had been fixed? Definitely should be `self.add`.

gpytorch/lazy/sum_lazy_tensor.py
```diff
@@ -73,6 +71,22 @@ def __add__(self, other):
...
+
+            # update the lazy tensors' shape as well
+            if broadcasted_shape != self.shape:
+                broadcasted_lts = [
+                    lt.expand(*broadcasted_shape, 1).squeeze(-1).transpose(-1, -2) for lt in self.lazy_tensors
+                ]
+            else:
+                broadcasted_lts = list(self.lazy_tensors)
+
+            return SumLazyTensor(*(broadcasted_lts + [broadcasted_other]))
```
gpleiss: Rather than doing the work here, I think it would make sense to add the broadcasting operations in `LazyTensor.__add__` or in the SumLazyTensor constructor. We can also wrap all tensors as LazyTensors as well.

wjmaddox: A couple of comments here - I'm not sure if it's possible to put the broadcasting in `LazyTensor.__add__` and the SumLazyTensor constructor might require a little bit more work to check all of the lazy tensors.

Is there a reason to not put the broadcasting here?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1733,8 +1733,7 @@ def __mul__(self, other):
         return self.mul(other)
 
     def __radd__(self, other):
-        #TODO: see if this should be self.add
-        return self.mul(other)
```
Balandat: wut?

wjmaddox: This was fixed by [0e08be](https://github.com/cornellius-gp/gpytorch/commit/0e08bef757d6ce8055bbaa09fba919b45ff3c851). At the time, it seemed to be in master so I left it.

gpytorch/lazy/sum_lazy_tensor.py
gpleiss: Putting it in `LazyTensor.__add__` or the SLT constructor catches more cases (e.g. if you were to add a ToeplitzLazyTensor and a Tensor), and requires fewer conditional statements which is cleaner/less brittle.

For `LazyTensor.__add__`:

```python
def __add__(self, other):
    other = lazify(other)
    shape = _mul_broadcast_shape(self.shape, other.shape)
    return SumLazyTensor(self.expand(shape), other.expand(shape))
```

For the SLT constructor:

For `LazyTensor.__add__`:

```python
def __init__(self, *lazy_tensors, **kwargs):
    try:
        lazy_tensors = tuple(lazify(lt) for lt in lazy_tensors)
    except TypeError:
        raise TypeError("All arguments of a SumLazyTensor should be LazyTensors or Tensors")
    batch_shape = _mul_broadcast_shape(*[lt.shape for lt in lazy_tensors])
    lazy_tensors = tuple(lt.expand(shape) for lt in lazy_tensors)
    super.__init__(*lazy_tensors, **kwargs)
    self.lazy_tensors = lazy_tensors
```

wjmaddox: Thanks for the help - I ended up using the sum lazy tensor constructor.

test/lazy/test_broadcasted_lazy_tensor.py
wjmaddox: The tests are moved into `test_sum_lazy_tensor` now but given that they aren't dealing with symmetric matrices necessarily I'm not sure if they ought to use LazyTensorTestCase.

gpleiss:
I think I figured out how to make the check work correctly in the constructor as well. Thanks @wjmaddox !

Issue #165, opened 2018-06-16
SKI should support different grid sizes in each dimension


jacobrgardner:
Right now, we support different grid bounds for each dimension, but the grid size is fixed in each dimension. I think this should be simple enough to fix, and will basically only affect `GridInterpolationKernel.__init__`.

gpleiss:
Closed by #888

PR #888, opened 2019-09-30
Independent grid sizes per dimension for GridKernel and GridInterpolationKernel

KeAWang:
This PR finishes #861 and implements independent grid sizes per dimension for GridKernel and GridInterpolationKernel. The changes are mostly the same, except that this PR makes additional changes to pass all the tests and ensure correctness. That PR had a few merges from other branches, which clutters the git log a bit, so I'm opening this new PR.

One thing that will have to be done in the future is to fully deprecate the old method of passing in a `G x D` grid into `GridKernel` where `G` is the number of grid points for every dimension. For now, we use a function that converts this old grid into the new expected format of `List[Tensor]` with each Tensor possibly having different lengths.

In addition, it's important to note that this change causes the computations for each dimension to be done sequentially instead of in parallel through batch mode. It may be useful to special case the scenario when all the grid sizes are the same in the future to ensure maximum parallelism. Still, if we assume that data dimensions for GridKernels and GridInterpolation are of reasonable sizes, say less than 10, then the for-loop slowdowns will not be that significant.

This PR also does the following
- Add previously non-existent docstrings to `ToeplitzLazyTensor`
- Swap to using `torch.meshgrid` for `gpytorch.utils.grid.create_data_from_grid` instead of for-loops
- Added type hints for `GridKernel`, `GridInterpolationKernel` and some of the related code

### DIFF
--- a/gpytorch/kernels/grid_interpolation_kernel.py
+++ b/gpytorch/kernels/grid_interpolation_kernel.py
@@ -1,10 +1,13 @@
 #!/usr/bin/env python3
 
 import torch
+from .kernel import Kernel
 from .grid_kernel import GridKernel
 from ..lazy import InterpolatedLazyTensor
 from ..models.exact_prediction_strategies import InterpolatedPredictionStrategy
 from ..utils.interpolation import Interpolation
+from typing import Union, List, Optional, Tuple
+from gpytorch.utils.grid import create_grid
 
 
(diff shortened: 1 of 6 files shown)


KeAWang:
This would close #165

jacobrgardner:
@KeAWang I'll review this by mid next week.

PR #898, opened 2019-10-12
Roll PyTorch in CI to 1.3.0

vishwakftw:
Since 1.3.0 is out, I thought it would be better to test against the latest stable version.

cc @gpleiss @Balandat


gpleiss:
I was planning on doing this today myself :) Thanks @vishwakftw!

PR #903, opened 2019-10-15
Updates to variational models and objects

### LINKED ISSUE #905: [Bug] add_jitter fails with AttributeError: other must be a LazyTensor at eval time
# 🐛 Bug

Using `add_jitter` with the covariance module fails at eval time.

## To reproduce

1. Start from the example https://github.com/cornellius-gp/gpytorch/blob/master/examples/05_Scalable_GP_Regression_Multidimensional/SVGP_Regression_CUDA.ipynb

2. Optionally, replace `WhitenedVariationalStrategy` with `VariationalStrategy` in `GPModel.__init__`

3. Replace `self.covar_module(x)` with `self.covar_module(x).add_jitter(1e-6)` in `GPModel.forward`

```
AttributeError                            Traceback (most recent call last)
<ipython-input-6-c955f39ee560> in <module>
      4 with torch.no_grad():
      5     for x_batch, y_batch in test_loader:
----> 6         preds = model(x_batch)
      7         means = torch.cat([means, preds.mean.cpu()])
      8 means = means[1:]

~/miniconda3/envs/mbrl/lib/python3.7/site-packages/gpytorch/models/abstract_variational_gp.py in __call__(self, inputs, **kwargs)
     20             inputs = inputs.unsqueeze(-1)
     21 
---> 22         return self.variational_strategy(inputs)

~/miniconda3/envs/mbrl/lib/python3.7/site-packages/gpytorch/variational/variational_strategy.py in __call__(self, x)
    228                 self._memoize_cache = dict()
    229 
--> 230         return super(VariationalStrategy, self).__call__(x)

~/miniconda3/envs/mbrl/lib/python3.7/site-packages/gpytorch/module.py in __call__(self, *inputs, **kwargs)
     20 
     21     def __call__(self, *inputs, **kwargs):
---> 22         outputs = self.forward(*inputs, **kwargs)
     23         if isinstance(outputs, list):
     24             return [_validate_module_outputs(output) for output in outputs]

~/miniconda3/envs/mbrl/lib/python3.7/site-packages/gpytorch/variational/variational_strategy.py in forward(self, x)
    214                     induc_induc_covar.inv_matmul(induc_data_covar)
    215                 )
--> 216                 data_covariance = data_data_covar + neg_induc_data_data_covar
    217             predictive_covar = PsdSumLazyTensor(predictive_covar, data_covariance)
    218 

~/miniconda3/envs/mbrl/lib/python3.7/site-packages/gpytorch/lazy/sum_lazy_tensor.py in __add__(self, other)
     75             return SumLazyTensor(*(list(self.lazy_tensors) + [other]))
     76         else:
---> 77             raise AttributeError("other must be a LazyTensor")
     78 
     79     def diag(self):

AttributeError: other must be a LazyTensor
```

## Expected Behavior

The model to return outputs like always.

## System information

- GPyTorch Version: `0.3.6`
- PyTorch Version: `1.2.0`
- OS Version: `Ubuntu 18.04.3 LTS`

gpleiss:
Features:
- [x] `gpytorch.variational.VariationalStrategy` implements the whitening of Henseman et al (2015)
- [x] Remove WhitenedVariationalStrategy (which was wrong)
- [x] Add `gpytorch.variational.UnwhitenedVariationalStrategy`
- [x] Add `beta` option to `VariationalELBO`
- [x] Add `gpytorch.mlls.PredictiveLogLikelihood` (@jacobrgardner - better name?)
- [x] Redo structure of variational models in `gpytorch.models`
- [x] Update Pyro interface
- [x] Add a `MultitaskVariationalStrategy` that wraps existing variational strategies.
- [x] Low-level Pyro interface
- [x] Legacy mode for variational model loading
- [x] Add mean field and delta variational distributions
- [x] Add Robust SVGP objective

Tests:
- [x] Unit test coverage of variational objects
- [x] Add pyro tests to Travis

Docs:
- [x] Add docstrings to all VariationalStrategies
- [x] Add docstrings to all VariationalDistributions
- [x] Add docstrings to all objective functions

Updates to variational strategies:
- [x] New abstract class: `gpytorch.variational._VariationalStrategy`
- [x] `_VariationalStrategy.variational_distribution` now is the actual MVN q(u), rather than the `gpytorch.variational._VariationalDistribution` object.
- [x] `learn_inducing_points` is True by default
- [x] Get rid of `initialize_variational_distribution` and `kl_divergence` methods (they are no longer necessary since we only do true whitening)

Updates to variational distributions:
- [x] New abstract class: `gpytorch.variational._VariationalDistribution`
- [x] Deprecate `_VariationalDistribution.variational_distribution` (in favor of calling the model instead)
- [x] CholeskyVariationalDistribution adds a bit of noise to the mean initialization

Updates to likelihoods:
- [x] Create a `log_marginal` method for PredictiveLogLikelihood
- [x] Split off `_OneDimensionalLikelihood` from `Likelihood`, which uses 1D Gauss-Hermite quadrature for expected_log_prob and log_marginal (standard likelihood now uses mc-integration)

Small stuff:
- [x] Fix shape method for MVN
- [x] Remove deprecated batch_size parameters
- [x] Deprecate `VariationalMarginalLogLikelihood`

Fixes #905


gpleiss:
@jacobrgardner - this could also be split up into several PRs - but I'm just getting everything out here for now.

gpytorch/mlls/_approximate_mll.py
```diff
@@ -0,0 +1,89 @@
...
+        super().__init__(likelihood, model)
+        self.combine_terms = combine_terms
+        self.num_data = num_data
+        self.beta = beta
+
+    @abstractmethod
+    def _log_likelihood_term(self, approximate_dist_f, target, **kwargs):
+        raise NotImplementedError
+
+    def _kl_divergence(self):
```
jacobrgardner: The predictive loss implemented like this will work fine for `CholeskyVariationalDistribution` and `MeanFieldVariationalDistribution`, but if we wanted to implement the Delta distribution we'll need the regularizer to be `\beta log p(u)`.

Maybe we should stick with mean field as the recommended one and punt on delta? I don't know how important the delta distribution version is -- it seemed like mean field usually did about as well on RMSE but much beter on NLL.

gpleiss: Here's another thought: we could decompose this even further. We could add a `_variational_cross_entropy` term which computes `-\beta E_{q(u)} [ log p(u) ]` and a `_variational` term which computes `-\beta E_{q(u)} [ log q(u) ]`.

With the DeltaVariationalDistribution, this would then collapse to just `\beta log p(u)`. Thoughts?

gpleiss: Actually a better idea:

We create a `gpytorch.distributions.Delta` distribution - similar to [Pyro's](http://docs.pyro.ai/en/stable/distributions.html#delta). We can register a kl_divergence between delta dists and MVNs.

jacobrgardner: @gpleiss Yeah I like that idea

examples/02_Simple_GP_Classification/Simple_GP_Classification.ipynb
```diff
@@ -78,9 +78,9 @@
     "  \n",
     "#### The GP Model\n",
     "  \n",
-    "The `AbstractVariationalGP` model is GPyTorch's simplist approximate inference model. It approximates the true posterior with a distribution specified by a `VariationalDistribution`, which is most commonly some form of MultivariateNormal distribution. The model defines all the variational parameters that are needed, and keeps all of this information under the hood.\n",
+    "The `ApproximateGP` model is GPyTorch's simplist approximate inference model. It approximates the true posterior with a distribution specified by a `VariationalDistribution`, which is most commonly some form of MultivariateNormal distribution. The model defines all the variational parameters that are needed, and keeps all of this information under the hood.\n",
```
martinjankowiak: typo 'simplist'

examples/05_Scalable_GP_Regression_Multidimensional/SVDKL_Regression_GridInterp_CUDA.ipynb
```diff
@@ -137,11 +137,11 @@
    "source": [
     "## Defining the GP Regression Layer\n",
     "\n",
-    "We now define the GP regression module that, intuitvely, will act as the final \"layer\" of our neural network. In this case, because we are doing variational inference and *not* exact inference, we will be using an `AbstractVariationalGP`. To use grid interpolation for variational inference, we'll be using a `GridInterpolationVariationalStrategy`.\n",
+    "We now define the GP regression module that, intuitvely, will act as the final \"layer\" of our neural network. In this case, because we are doing variational inference and *not* exact inference, we will be using an `ApproximateGP`. To use grid interpolation for variational inference, we'll be using a `GridInterpolationVariationalStrategy`.\n",
```
martinjankowiak: typo 'intuitvely'

examples/05_Scalable_GP_Regression_Multidimensional/SVDKL_Regression_SVGP_CUDA.ipynb
```diff
@@ -137,7 +137,7 @@
    "source": [
     "## Defining the GP Regression Layer\n",
     "\n",
-    "We now define the GP regression module that, intuitvely, will act as the final \"layer\" of our neural network. In this case, because we are doing variational inference and *not* exact inference, we will be using an `AbstractVariationalGP`. In this example, because we will be learning the inducing point locations, we'll be using a base `VariationalStrategy` with `learn_inducing_locations=True`.\n",
+    "We now define the GP regression module that, intuitvely, will act as the final \"layer\" of our neural network. In this case, because we are doing variational inference and *not* exact inference, we will be using an `ApproximateGP`. In this example, because we will be learning the inducing point locations, we'll be using a base `VariationalStrategy` with `learn_inducing_locations=True`.\n",
```
martinjankowiak: typo 'intuitvely'

examples/05_Scalable_GP_Regression_Multidimensional/SVGP_Regression_CUDA.ipynb
```diff
@@ -99,7 +99,7 @@
    "source": [
     "## Defining the SVGP Model\n",
     "\n",
-    "We now define the GP regression module that, intuitvely, will act as the final \"layer\" of our neural network. In this case, because we are doing variational inference and *not* exact inference, we will be using an `AbstractVariationalGP`. In this example, because we will be learning the inducing point locations, we'll be using a base `VariationalStrategy` with `learn_inducing_locations=True`.\n",
+    "We now define the GP regression module that, intuitvely, will act as the final \"layer\" of our neural network. In this case, because we are doing variational inference and *not* exact inference, we will be using an `ApproximateGP`. In this example, because we will be learning the inducing point locations, we'll be using a base `VariationalStrategy` with `learn_inducing_locations=True`.\n",
```
martinjankowiak: typo 'intuitvely'

gpytorch/models/pyro/pyro_gp.py
```diff
@@ -0,0 +1,114 @@
...
+        Args:
+            :attr:`input` (`torch.Tensor`):
+                :math:`\mathbf X` The input values values
+            :attr:`target` (`torch.Tensor`):
+                :math:`\mathbf y` The target values
+            :attr:`*args`, :attr:`**kwargs`:
+                Additional arguments passed to the likelihood's `forward` function.
+        """
+        # Hack for getting correct sampling shape
+        pyro.sample("__throwaway__", pyro.distributions.Normal(0, 1)).shape
```
martinjankowiak: what is this hack for?

gpleiss: We're computing the log-likelihood term with MC-integratoin, which requires drawing samples from q(f). However, we don't want to draw samples from q(f) using `pyro.sample` since this would add E(q(f)) as a term in the loss function. Sampling from the throwaway distribution then allows us to get the sampling shape we would've gotten if we had called `pyro.sample` on q(f), and then we can draw non-pyro samples from q(f) by calling `function_dist(sample_shape)`.

martinjankowiak: use pyro.sample but enclose it in pyro.block instead? won't the throwaway log prob leak into your loss?

gpleiss: > won't the throwaway log prob leak into your loss?

It's a sample from a N(0, 1) distribution, so there's no parameters and it's pretty cheap to compute. There's also a `__throwaway__` sample in the guide so the KL component should exactly equal zero.

> use pyro.sample but enclose it in pyro.block instead

I tried `pyro.block` before, but unfortunately it doesn't respect the existing plate structure (i.e. we only end up with a single function_sample as a result).

gpytorch/mlls/predictive_log_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+          \mathbb{E}_{p_\text{data}( y, \mathbf x )} \left[
+            \log p( y \! \mid \! \mathbf x)
+          \right] - \beta \: \text{KL} \left[ q( \mathbf u) \Vert p( \mathbf u) \right]
+          \\
+          &\approx \sum_{i=1}^N \log \mathbb{E}_{q(\mathbf u)} \left[
+            \int p( y_i \! \mid \! f_i) p(f_i \! \mid \! \mathbf u) \: d f_i
+          \right] - \beta \: \text{KL} \left[ q( \mathbf u) \Vert p( \mathbf u) \right]
+       \end{align*}
+
+    where :math:`N` is the amount of data, :math:`q(\mathbf u)` is the variational distribution for
```
martinjankowiak: amount => total number

gpytorch/models/pyro/trace_predictive_log_likelihood.py
```diff
@@ -0,0 +1,142 @@
...
+        loss = self.differentiable_loss(model, guide, *args, **kwargs)
+        return torch_item(loss)
+
+    def _differentiable_loss_particle(self, model_trace, guide_trace):
+        elbo_particle = 0
+        surrogate_elbo_particle = 0
+        log_r = None
+
+        # compute elbo and surrogate elbo
+        for name, site in model_trace.nodes.items():
```
martinjankowiak: i'm not sure if this is what you want. what if you observe other data that has nothing to do with the GP?

gpleiss:
@jacobrgardner - I added the `MultitaskVariationalStrategy` - how do you like that interface? We could also probably use it to re-write SoftmaxLikelihood.

gpytorch/models/pyro/trace_predictive_log_likelihood.py
gpleiss: This is now removed.

docs/source/likelihoods.rst
```diff
@@ -15,9 +15,17 @@ Likelihood
...
-Standard Likelihoods
+One-Dimensional Likelihoods
 -----------------------------
 
+Likelihoods for GPs that are distributions of scalar functions.
+(I.e. for a specific :math:`\mathbf x` we expect that :math:`f(\mathbf x) \in \mathbb{R}`.)
+
+One-dimensional likelihoods should extend :obj:`gpytoch.likelihoods._OneDimensionalLikelihood` to 
+reduce the variance when computing approximate GP objective functions.
+(Variance reduction is accomplished by by using 1D Gauss-Hermite quadrature rather than MC-integration).
```
martinjankowiak: typo 'by by

gpytorch/mlls/predictive_log_likelihood.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
+
+from ._approximate_mll import _ApproximateMarginalLogLikelihood
+
+
+class PredictiveLogLikelihood(_ApproximateMarginalLogLikelihood):
+    r"""
+    An alternative objective function for approximate GPs, proposed in `Jankowiak et al., 2019`_.
+    It typically produces predictive variances than the :obj:`gpytorch.mlls.VariationalELBO` objective.
```
martinjankowiak: missing a word

gpytorch/mlls/predictive_log_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+            \int p( y_i \! \mid \! f_i) p(f_i \! \mid \! \mathbf u) \: d f_i
+          \right] - \beta \: \text{KL} \left[ q( \mathbf u) \Vert p( \mathbf u) \right]
+       \end{align*}
+
+    where :math:`N` is the amount of data, :math:`q(\mathbf u)` is the variational distribution for
+    the inducing function values, and `p(\mathbf u)` is the prior distribution for the inducing function
+    values.
+
+    :math:`\beta` is a scaling constant that reduces the regularization effect of the KL
+    divergence. Setting :math:`\beta=1` (default) results in the true variational ELBO.
```
martinjankowiak: not quite right "Setting :math:`\beta=1` (default) results in the true variational ELBO"

gpytorch/mlls/predictive_log_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+        >>> # model is a gpytorch.models.VariationalGP
+        >>> # likelihood is a gpytorch.likelihoods.Likelihood
+        >>> mll = gpytorch.mlls.PredictiveLogLikelihood(likelihood, model, num_data=100, beta=0.5)
+        >>>
+        >>> output = model(train_x)
+        >>> loss = -mll(output, train_y)
+        >>> loss.backward()
+
+    .. _Jankowiak et al., 2019:
+        http://bit.ly/predictive_gp
```
martinjankowiak: where is this gonna point?

gpytorch/mlls/variational_elbo.py
```diff
@@ -1,60 +1,86 @@
...
+          \mathbb{E}_{p_\text{data}( y, \mathbf x )} \left[
+            \log \mathbb{E}_{q(\mathbf u)} \left[  p( y \! \mid \! \mathbf u) \right]
+          \right] - \beta \: \text{KL} \left[ q( \mathbf u) \Vert p( \mathbf u) \right]
+          \\
+          &\approx \sum_{i=1}^N \mathbb{E}_{q( \mathbf u)} \left[
+            \log \int p( y_i \! \mid \! f_i) p(f_i \! \mid \! \mathbf u) \: d \mathbf f_i
+          \right] - \beta \: \text{KL} \left[ q( \mathbf u) \Vert p( \mathbf u) \right]
+       \end{align*}
+
+    where :math:`N` is the amount of data, :math:`q(\mathbf u)` is the variational distribution for
```
martinjankowiak: amount => number...

gpytorch/mlls/gamma_robust_variational_elbo.py
```diff
@@ -0,0 +1,95 @@
...
+        >>> loss.backward()
+
+    .. _Knoblauch, 2019:
+        https://arxiv.org/pdf/1904.02303.pdf
+    """
+    def __init__(self, likelihood, model, gamma=1.0, *args, **kwargs):
+        if not isinstance(likelihood, _GaussianLikelihoodBase):
+            raise RuntimeError("Likelihood must be Gaussian for exact inference")
+        super().__init__(likelihood, model, *args, **kwargs)
+        self.register_buffer("gamma", torch.tensor(gamma))
```
martinjankowiak: by promote a float to a tensor here?

gpytorch/mlls/gamma_robust_variational_elbo.py
```diff
@@ -0,0 +1,97 @@
...
+        >>> output = model(train_x)
+        >>> loss = -mll(output, train_y)
+        >>> loss.backward()
+
+    .. _Knoblauch, 2019:
+        https://arxiv.org/pdf/1904.02303.pdf
+    .. _Knoblauch, Jewson, Damoulas 2019:
+        https://arxiv.org/pdf/1904.02063.pdf
+    """
+    def __init__(self, likelihood, model, gamma=1.0, *args, **kwargs):
```
gpleiss: Make sure that we're consistent with what value of gamma we're using (either `gamma - 1` or `gamma`). Typical value should be 1.03

gpytorch/mlls/gamma_robust_variational_elbo.py
gpleiss: It was to use `torch.log1p`. But we could just use numpy's log1p instead.

gpytorch/likelihoods/likelihood.py
```diff
@@ -141,12 +140,20 @@ def name_prefix(self):
         def name_prefix(self, val):
             self._name_prefix = val
 
-        def _draw_likelihood_samples(self, function_dist, *args, **kwargs):
-            # Hack to get the current plating structure
-            sample_shape = pyro.sample("__throwaway__", pyro.distributions.Normal(0, 1)).shape
-            if not len(sample_shape):
-                sample_shape = None
-            return super()._draw_likelihood_samples(function_dist, *args, **kwargs, sample_shape=sample_shape)
+        def _draw_likelihood_samples(self, function_dist, *args, sample_shape=None, **kwargs):
```
jacobrgardner: Maybe `_pyro_draw_likelihood_samples`? In both PRs, I think all methods with pyro primitive calls start with `pyro` or `_pyro` so far -- might be a good standard.

gpleiss: I agree about the pyro prefixing in general. However `_draw_likelihood_samples` is also used for non-pyro likelihoods as well.

gpytorch/models/pyro/pyro_gp.py
gpleiss: This is now fixed with the new interface (which is pretty finalized).

gpytorch/likelihoods/likelihood.py
jacobrgardner: If it's used for non pyro likelihoods, should we either remove the pyro statements from the method or make pyro a dependency of the package? as is, this method is uncallable in a fresh environment with `pip install gpytorch`.

gpleiss: The `_draw_likelihood_samples` that you're referring to is for the base likelihood class when Pyro is available. Otherwise it uses this version of the method:

https://github.com/cornellius-gp/gpytorch/blob/f2f442fc95047b3c432d7cbf5e908084ff0d3cb7/gpytorch/likelihoods/likelihood.py#L22

gpytorch/mlls/_approximate_mll.py
gpleiss: Done!

gpytorch/variational/delta_variational_distribution.py
```diff
@@ -0,0 +1,36 @@
...
+
+class DeltaVariationalDistribution(_VariationalDistribution):
+    """
+    This :obj:`~gpytorch.variational._VariationalDistribution` object replaces a variational distribution
+    with a single particle. It is equivalent to doing MAP inference.
+
+    Args:
+        :attr:`num_inducing_points` (int):
+            Size of the variational distribution. This implies that the variational mean
+            should be this size, and the variational covariance matrix should have this many rows and columns.
```
martinjankowiak: there is no cov

gpytorch/likelihoods/gaussian_likelihood.py
```diff
@@ -38,10 +38,37 @@ def _shaped_noise_covar(self, base_shape: torch.Size, *params: Any, **kwargs: An
...
     def forward(self, function_samples: Tensor, *params: Any, **kwargs: Any) -> base_distributions.Normal:
         noise = self._shaped_noise_covar(function_samples.shape, *params, **kwargs).diag()
         return base_distributions.Normal(function_samples, noise.sqrt())
 
+    def log_marginal(
+        self, observations: Tensor, function_dist: MultivariateNormal, *params: Any, **kwargs: Any
+    ) -> Tensor:
+        marginal = self.marginal(function_dist, *params, **kwargs)
+        # We're making everything conditionally independent
+        indep_dist = base_distributions.Normal(marginal.mean, marginal.variance.sqrt())
```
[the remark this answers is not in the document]
gpleiss: Added.

gpytorch/mlls/predictive_log_likelihood.py
gpleiss: That was a placeholder before the paper was on arxiv. Fixed now.

gpytorch/variational/_variational_strategy.py
```diff
@@ -0,0 +1,138 @@
...
+            x = x.expand(*batch_shape, *x.shape[-2:])
+
+        # Get p(u)/q(u)
+        variational_dist_u = self.variational_distribution
+
+        # Get q(f)
+        if isinstance(variational_dist_u, MultivariateNormal):
+            return super().__call__(
+                x, inducing_points,
+                inducing_values=variational_dist_u.mean,
```
jacobrgardner: Not wild about passing the variational mean through “inducing_values”. Think I know why you did it (`Delta`), but could we not just pass `variational_dist_u` along and have the strategy worry about assumptions about what `variational_dist_u` has in it?

gpytorch/variational/cholesky_variational_distribution.py
```diff
@@ -61,3 +44,8 @@ def variational_distribution(self):
         # Now construct the actual matrix
         variational_covar = CholLazyTensor(chol_variational_covar)
         return MultivariateNormal(self.variational_mean, variational_covar)
+
+    def initialize_variational_distribution(self, prior_dist):
+        self.variational_mean.data.copy_(prior_dist.mean)
```
jacobrgardner: Didn’t this once exist on Strategy instead of Distribution, because whitened vs unwhitened initialization was different? I take it we assume whitened now?

gpytorch/variational/mean_field_variational_distribution.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    def __init__(self, num_inducing_points, batch_shape=torch.Size([]), mean_init_std=1e-3, **kwargs):
+        super().__init__(num_inducing_points=num_inducing_points, batch_shape=batch_shape, mean_init_std=mean_init_std)
+        mean_init = torch.zeros(num_inducing_points)
+        covar_init = torch.ones(num_inducing_points)
+        mean_init = mean_init.repeat(*batch_shape, 1)
+        covar_init = covar_init.repeat(*batch_shape, 1)
+
+        self.register_parameter(name="variational_mean", parameter=torch.nn.Parameter(mean_init))
+        self.register_parameter(name="variational_stddev", parameter=torch.nn.Parameter(covar_init))
```
jacobrgardner: Let’s consider adding a `getter` that returns the absolute value of these. I can already see the “variational inference is learning negative standard deviations” github issues now :-).

gpytorch/variational/whitened_variational_strategy.py
```diff
@@ -1,205 +0,0 @@
...
-import torch
-from .. import settings
-from .variational_strategy import VariationalStrategy
-from ..utils.memoize import cached
-from ..lazy import RootLazyTensor, MatmulLazyTensor, CholLazyTensor, \
-    CachedCGLazyTensor, DiagLazyTensor, BatchRepeatLazyTensor, PsdSumLazyTensor
-from ..distributions import MultivariateNormal
-
-
-class WhitenedVariationalStrategy(VariationalStrategy):
```
jacobrgardner: Consider keeping a thin class that raises a RuntimeError explaining what happened (e.g., switch to use VariationalStrategy now, it’s whitened correctly) to deprecate.

jacobrgardner: Ditto for any other deleted classes.

gpleiss:
Alright - @martinjankowiak and @jacobrgardner I'm pretty sure this is no longer a WIP! Ready for a final review.

I'm going to say that re-doing examples and docs should wait for another PR because this one is already so huge.

docs/source/distributions.rst
```diff
@@ -4,12 +4,28 @@
 gpytorch.distributions
 ===================================
 
+GPyTorch distribution objects are essentially the same as torch distribution objects.
+For the most part, GpyTorch relies on torch's distribution library.
+However, we offer two custom distributions.
+
+We implement a custom :obj:`~gpytorch.distributions.MultivariateNormal` that accepts
+:obj:`~gpytorch.lazy.LazyTensor` objects for covariance matrices. This allows us to custom
```
martinjankowiak: typo: 'use custom' or 'customize'

docs/source/distributions.rst
```diff
@@ -4,12 +4,28 @@
...
+linear algebra operations, which makes this more efficient than PyTorch's MVN implementation.
+
+In addition, we implement a :obj:`~gpytorch.distributions.MultitaskMultivariateNormal` which
+can be used with multi-output Gaussian process models.
+
+.. note::
+
+  If Pyro is available, all GPyTorch distribution objects inheret Pyro's distribution methods
```
martinjankowiak: typo inheret

docs/source/marginal_log_likelihoods.rst
```diff
@@ -4,32 +4,61 @@
...
    :members:
 
 
-Variational GP Inference
+Approximate GP Inference
 -----------------------------------
 
-:hidden:`VariationalMarginalLogLikelihood`
+These are MLLs for use with :obj:`~gpytorch.models.ApproximateGP` modules. They are designed for
+when exact inference is intractible (either when the likelihood is non-Gaussian likelihood, or when
```
martinjankowiak: typo => intractable

gpytorch/likelihoods/softmax_likelihood.py
```diff
@@ -18,6 +18,7 @@ def __init__(
             kwargs, "n_classes", "num_classes", num_classes, lambda n: n
         )
         super().__init__()
+        self.max_plate_nesting = 2
```
martinjankowiak: deserved a comment?

gpytorch/mlls/variational_elbo.py
```diff
@@ -1,91 +1,77 @@
...
+
+    :param ~gpytorch.likelihoods.Likelihood likelihood: The likelihood for the model
+    :param ~gpytorch.models.ApproximateGP model: The approximate GP model
+    :param int num_data: The total number of training data points (necessary for SGD)
+    :param float beta: (optional, default=1.) A multiplicative factor for the KL divergence term.
+        Setting it to 1 (default) recovers true variational inference
+        (as derived in `Scalable Variational Gaussian Process Classification`_).
+        Setting it to anything less than 1 reduces the regularization effect of the model
+        (similarly to what was proposed in `the beta-VAE paper`_).
+    :param bool combine_terms: (default=True): Whether or not to sum the
```
martinjankowiak: when beta != 1 and combine_terms=False is the returned KL scaled?

gpytorch/distributions/delta.py
```diff
@@ -0,0 +1,76 @@
...
+            new._validate_args = self._validate_args
+            return new
+
+        def rsample(self, sample_shape=torch.Size()):
+            shape = sample_shape + self.v.shape
+            return self.v.expand(shape)
+
+        def log_prob(self, x):
+            v = self.v.expand(self.batch_shape + self.event_shape)
+            log_prob = (x == v).type(x.dtype).log()
```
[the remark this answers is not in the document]
gpleiss: Yeah... but as far I as know we're never calling `log_prob` for this distribution ever (or we're always calling it with x).

gpytorch/variational/_variational_strategy.py
gpleiss: That's a fair point. The reason I did it this way is to give us some future flexibility if we wanted to use non-Gaussian variational distributions (e.g. normalizing flows) that we'd have to sample from. Then `inducing_values` would correspond to samples from `q(u)`.

Also, this way the casing logic can happen in the super class.

gpytorch/variational/whitened_variational_strategy.py
gpleiss: Yeah... you're right :/
All other classes (except this one) should have deprecation warnings. I just got sloppy here.

gpytorch/variational/mean_field_variational_distribution.py
gpleiss: Yes good point!

gpytorch/variational/_variational_strategy.py
jacobrgardner: Would any of the existing strategies be usable with samples from a NF? If not, then it still might be better to pass the dist along since a NFVariationalStrategy or whatever will need yet another case added here

gpytorch/variational/cholesky_variational_distribution.py
gpleiss: `VariationalStrategy.prior_distribution` returns `N(0, I)`, so this works in both the whitened and unwhitened cases. Also, this way we don't have to re-define `kl_divergence` for VariationalStrategy which is nice.

test/examples/test_pyro_integration.py
```diff
@@ -0,0 +1,324 @@
...
+            return torch.nn.functional.softplus(self.raw_noise)
+
+        def forward(self, x):
+            mean_x = self.mean_module(x)
+            covar_x = self.covar_module(x)
+            res = gpytorch.distributions.MultivariateNormal(mean_x, covar_x)
+            return res
+
+        def guide(self, x, y):
+            name_prefix = "llcmgp"
```
martinjankowiak: why is this defined in 2 places?

gpytorch/variational/mean_field_variational_distribution.py
jacobrgardner: Might make the getter for the variances directly to avoid an unnecessary .abs.pow(2) when just .pow(2) will suffice.

examples/02_Approximate_GPs/Approximate_GP_Objective_Functions.ipynb
```diff
@@ -0,0 +1,282 @@
...
+    "    ax.scatter(train_x, train_y, c='k', marker='.', label=\"Data\")\n",
+    "    ax.legend(loc=\"best\")\n",
+    "    ax.set(xlabel=\"x\", ylabel=\"y\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Objective Funtion 1) The Variational ELBO\n",
```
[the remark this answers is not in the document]
gpleiss: Oooops fixed.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1739,6 +1739,9 @@ def __radd__(self, other):
     def __rmul__(self, other):
         return self.mul(other)
 
+    def __sub__(self, other):
```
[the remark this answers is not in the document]
gpleiss: Added #932

gpytorch/variational/_variational_strategy.py
gpleiss: Atm all of the existing strategies could be used with samples from a NF. Computing q(f | u) (for a sample from u) is the same as computing E[ q(f) ] for delta var dists. is the same as computing E[ q(f) ] for normal var dists.

I agree that it's not the nicest inteface in terms of method names, but it does reduce the amount of repeated code. It's written imagining that the "default" case is getting samples from q(u) (which trivially covers Delta distributions). Normal distributions are written as a special case (where we do marginalization with the covariance), but the code for computing the mean is the same as the "default" case.

Open to suggestions for improving this.

jacobrgardner: I'm kind of just armchair reviewing from my phone, so I'm fine letting it go in as is and opening another issue if it turns out to be problematic down the road.

gpytorch/likelihoods/softmax_likelihood.py
gpleiss: Whoops that's leftover. Removing.

test/examples/test_pyro_integration.py
gpleiss: Shouldn't be.

gpytorch/mlls/variational_elbo.py
gpleiss: With the current implementation, the KL is always scaled (even when combine_terms=False). Open to changing this though.

jacobrgardner: I'm in favor of keeping it as is. If people want to do something more complicated with beta they should set beta=1. and do whatever they want.

jacobrgardner:
lgtm pending missing name_prefix change.

gpytorch/distributions/multivariate_normal.py
```diff
@@ -37,7 +36,8 @@ def __init__(self, mean, covariance_matrix, validate_args=False):
             self._covar = covariance_matrix
             self.__unbroadcasted_scale_tril = None
             self._validate_args = validate_args
-            batch_shape, event_shape = self.loc.shape[:-1], self.loc.shape[-1:]
+            batch_shape = _mul_broadcast_shape(self.loc.shape[:-1], covariance_matrix.shape[:-2])
```
Balandat: @gpleiss this broke our multitask models in botorch. Before I dig deeper into this, what was the rationale to add this here?

Balandat: more precisely, the failure is on batch evaluating a non-batched model

jacobrgardner: @Balandat Do you have an example test that fails?

gpleiss: This is the behavior that `torch.distributions.MultivariateNormal` has - if you pass in a non-batch mean and a batch covariance matrix, then the covariance matrix determines the batch size. This wasn't happening before.

I believe the use case was for training (independent) multi-output variational models. Let's say you want a different lengthscale for each output, but the same mean. This way, you can train a model with non-batch `x` inputs, a non-batch mean function, and a batch kernel function (with different lengthscale). With this fix, the resulting MVN has the expected batch size.

Balandat: Hmm nvm this was actually a bug in our MTGP implementation that would do the wrong thing in batch model. All good.

Balandat: thanks for surfacing the bug :)

PR #909, opened 2019-10-21
Explicit Kronecker Structure Linear Algebra

wjmaddox:
Pulled this out of #872 for now. Consists of solely Kronecker structured linear algebra inverse matmul and log determinants. 

I will need to check #725 for linear algrebra speed checks. 

Current status is that I need to write the unit tests for `KroneckerProductPlusDiagLazyTensor` and `_DiagKroneckerProdLazyTensor` but they are sitting in notebooks somewhere.


gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,139 @@
...
+        eigs = sub_eigs[0].matmul(sub_eigs[1].t())
+        return torch.log(eigs + noise).sum()
+
+    def _kron_svd(self):
+        return [lt.evaluate().svd() for lt in self._lazy_tensor.lazy_tensors]
+
+    def inv_quad(self, rhs):
+        # TODO: check stability of numerics here
+
+        svd_list = self._kron_svd()
```
gpleiss: I can imagine that this could be an issue if this lazy tensor is used with bigger matrices (e.g. if this is used for multi-task models). Can we add a settings flag that peforms this svd-based approach only if the size is sufficiently small?

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,139 @@
...
+            logdet_term = None
+
+        return inv_quad_term, logdet_term
+
+    def logdet(self):
+        noise = self._diag_tensor[0,0]
+        ### THIS WORKS FOR 2 DIMENSIONS NOW ###
+        sub_eigs = []
+        for lt in self._lazy_tensor.lazy_tensors:
+            sub_eigs.append(lt.evaluate().eig()[0][:, 0].unsqueeze(-1))
```
gpleiss: Same thing here. We should have a settings flag for this.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,139 @@
...
+        elif isinstance(lazy_tensors[0], DiagLazyTensor):
+            self._diag_tensor = lazy_tensors[0]
+            self._lazy_tensor = lazy_tensors[1]
+        elif isinstance(lazy_tensors[1], DiagLazyTensor):
+            self._diag_tensor = lazy_tensors[1]
+            self._lazy_tensor = lazy_tensors[0]
+        else:
+            raise RuntimeError("One of the LazyTensors input to AddedDiagLazyTensor must be a DiagLazyTensor!")
+
+    def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True):
```
gpleiss: Is this necessary? Seems like the default implementation.

wjmaddox:
Should be superseded by #1102

wjmaddox:
Actually, it's not superseded and might be useful for certain Kronecker structured applications --- for example multi-output Gaussian processes along the lines of http://proceedings.mlr.press/v89/zhe19a.html .

wjmaddox:
The reason why I'm bringing this back up is because there are situations where the Kronecker structure is more helpful even for reasonably large matrices. 

For example, in the HOGP model (http://proceedings.mlr.press/v89/zhe19a/zhe19a.pdf) i'm trying to implement -- their dataset is 6k data points, and then the size of the latent factors are 60 each respectively so the Kronecker structured covariance matrices are of size 6k x 6k, 60 x 60, and 60 x 60. In dense format in gpytorch CG would try to invert 21 million dimensional matrices which is pretty numerically unstable as comparison to three eigendecompositions of size 6000, 60, and 60.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def logdet(self):
+        noise = self._diag_tensor[0, 0]
+        sub_eigs = [DiagLazyTensor(svd_decomp.S) for svd_decomp in self._kron_svd]
+        sub_eigs_kronecker = KroneckerProductLazyTensor(*sub_eigs).diag()
+        return torch.log(sub_eigs_kronecker + noise).sum()
+
+    @property
+    def _kron_svd(self):
+        return [lt.evaluate().svd() for lt in self._lazy_tensor.lazy_tensors]
```
jacobrgardner, line 82: Why are we running SVD on positive (semi)definite matrices? We should also at a minimum be caching here likely, so we don't compute the same decomposition multiple times.

wjmaddox: The thought is that SVD is more stable than symeig here.

wjmaddox: I agree with the caching bit though, although I thought that the property tag would have caught it?

jacobrgardner: No we'll need to use the cached decorator.

SVD should not be more stable than symeig here, and is likely a fair bit slower.

wjmaddox: ahh, thanks

Balandat: except for small matrices where symeig is really slow :(

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: well the default implementation will use CG for the solves for large tensors, which we'd like to avoid here

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,65 @@
...
+from .non_lazy_tensor import lazify
+
+# TODO: check the speed of the linear algebra - svd vs eig etc.
+
+
+class KroneckerProductAddedDiagLazyTensor(AddedDiagLazyTensor):
+    def __init__(self, *lazy_tensors):
+        super(KroneckerProductAddedDiagLazyTensor, self).__init__(*lazy_tensors)
+        if len(lazy_tensors) > 2:
+            raise RuntimeError("An AddedDiagLazyTensor can only have two components")
```
Balandat, line 20: this seems odd - why not explicitly require the tensors so there is no ambiguity? 

```
def __init__(self, kron: KroneckerLazyTensor, diag DiagLazyTensor):
   ...
```

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if logdet is not None:
+            logdet_term = self.logdet()
+        else:
+            logdet_term = None
+
+        return inv_quad_term, logdet_term
+
+    def logdet(self):
+        noisy_eigenvalues = self._kron_eigenvalues.diag() + self._diag_tensor.diag()
```
Balandat, line 44: naming nit: "noisy" makes it seem that jitttering is the only thing this is useful for (which it's not).

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,65 @@
...
+        return torch.log(noisy_eigenvalues).sum()
```
Balandat, line 45: seems like this won't work with batched matrices, use `sum(dim=-1)` instead?

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
wjmaddox: as it is, i'm not sure if the gradients from `symeig` are trustworthy -- see the most recent commit. Could be a wesley induced bug however.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @property
+    @cached
+    def _eig_cache(self):
+        return [lt.evaluate().symeig(eigenvectors=True) for lt in self._lazy_tensor.lazy_tensors]
```
Balandat, line 50: Doesn't this defeat the purpose of the PR for exploiting structure? We can express the elements of the eigendecomposition of a Kronecker tensor in terms of the eigendecomps of its factors. The `.evaluate()` call prevents us from doing this. 

Instead we should make `symeig` a method on the lazies, with a default implementation as above, but special cases for things like `KroneckerLazyTensor` and `DiagLazyTensor`. Then we can use that with impunity here.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
wjmaddox: ahh yes, i had double checked that it would evaluate likelihoods for batched matrices but hadn't been able to check correctness yet. good catch

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: I might do that on top of my `TriangularLazyTensor` PR, since `TriangularLazyTensor`s have a very nice symeig too...

wjmaddox: I'm not sure I follow.. The point is to compute eigendecompositions of all of the smaller components here so the large case never does anything more than just eigendecompose 1000 x 1000 matrices.

wjmaddox: And as far as I can tell, that's what the `_eig_cache` is doing.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
wjmaddox: one potential issue is that you want this tensor to be called whenever you add `KroneckerLazyTensor` and `DiagLazyTensor` so I'd have to go through the codebase checking this?

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: OK my bad, I misread this. This is the correct thing to do. I think the current setup is fine, but I'd still advocate for adding a `symeig` method to the lazies, as that's more in line with the rest of the package design, and would allow for other generically useful special cases (e.g. for a `TriangularLazyTensor` or `DiagLazyTensor` etc.)

wjmaddox:
Looks like the gradients of the quadratic form are still off by essentially a factor of two, much like what was the issue in #1102. But, Geoff's fix doesn't seem to resolve it here.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: I think we can mostly handle this in `__add__` of both `KroneckerLazyTensor`and `DiagLazyTensor`?

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: Just realized that I started to add `symeig` in #1105. Let me clean that PR up.

Balandat: Ok pushed an update, should have some useful stuff. Should be relatively easy to polish that up and add tests, but whatever goes in first doesn't really matter, we'll just need minor adjustments.

wjmaddox:
Good point, you'll have to do more work to construct an equivalent matrix for the general diagonal case.

test/lazy/test_kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,37 @@
...
+import unittest
+
+import torch
+
+from gpytorch.lazy import DiagLazyTensor, KroneckerProductAddedDiagLazyTensor, KroneckerProductLazyTensor, NonLazyTensor
+from gpytorch.test.lazy_tensor_test_case import LazyTensorTestCase
+
+
+class TestKroneckerProductAddedDiagLazyTensor(unittest.TestCase, LazyTensorTestCase):
+    # this lazy tensor has an explicit inverse so we don't need to run these
```
Balandat, line 12: should we add a small unit tests that checks that this actually uses the right methods / doesn't use CG / Cholesky?

wjmaddox: I think those are mostly dealt with by the `should_call_lanczos` and `should_call_cg` flags. Cholesky might be a good thing to check though.

gpytorch/lazy/kronecker_product_lazy_tensor.py
```diff
@@ -108,7 +122,43 @@ def _expand_batch(self, batch_shape):
     def _size(self):
         left_size = _prod(lazy_tensor.size(-2) for lazy_tensor in self.lazy_tensors)
         right_size = _prod(lazy_tensor.size(-1) for lazy_tensor in self.lazy_tensors)
-        return torch.Size((*self.lazy_tensors[0].batch_shape, left_size, right_size))
+        return Size((*self.lazy_tensors[0].batch_shape, left_size, right_size))
```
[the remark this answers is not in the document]
wjmaddox: Seems to break if I remove it.

test/lazy/test_kronecker_product_added_diag_lazy_tensor.py
wjmaddox: Just added a `inv_matmul` style unit test to double check if the cholesky is called.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
wjmaddox: These are pretty outdated now but could be handled by #1102 and #1105 .

wjmaddox:
I'll see if I can profile out point 1. early next week; there could be a "goldilocks" zone where CG is faster (maybe two large-ish matrices up to a certain point) and/or more stable.

wjmaddox:
It does look like at least on a (TitanX) GPU that the solve for the eigenvalues is definitely better for most sizes of `n` and it's more stable for small noise values.

For a noise level of 0.25:
106945043_606869033284318_8841073214993610326_n

For a noise level of 0.01 where CG starts diverging -- of course you could probably ramp up the preconditioner size. I don't know what happened to symeig at the very last one.
image


Code to reproduce:
```python
num_kernels = 3
num_data = [5, 10, 20, 50, 100, 250, 500, 1000, 3000]

results = []
for data_pt in num_data:
    torch.cuda.empty_cache()
    
    print('Now running ', data_pt)
    #inputs = [torch.randn(data_pt)] * num_kernels
    inputs = [torch.randn(75)] * (num_kernels - 1)
    inputs.append(torch.randn(data_pt))
    kernels = [MaternKernel()] * num_kernels
    
    lts = [lazify(kernel(input)) for kernel, input in zip(kernels, inputs)]
    # may change this line to make the solve harder
    lazy_tensor = KroneckerProductLazyTensor(*lts).add_jitter(0.01).to(device)
    print(lazy_tensor)
    rhs = torch.randn(lazy_tensor.shape[0]).to(device)
    
    start = time.time()
    solve = lazy_tensor.inv_matmul(rhs)
    end = time.time()
    
    elapsed = end - start
    error = (lazy_tensor.matmul(solve) - rhs).norm() / rhs.norm()
    
    results.append(torch.stack([torch.tensor(elapsed), error.cpu().data]))
    
    del solve, lazy_tensor, error, rhs, kernels, lts, inputs
```

gpytorch/lazy/kronecker_product_lazy_tensor.py
```diff
@@ -151,11 +151,26 @@ def add_diag(self, diag):
...
+            # TODO: ensure that the symeig call is also done in this manner
+
+            eval_tensor = lazy_tensor.evaluate()
+            tensor_dtype = eval_tensor.dtype
+
+            evals_, evecs_ = eval_tensor.double().symeig(eigenvectors=eigenvectors)
+
+            # we chop any negative eigenvalues
+            neg_evals = zeros_like(evals_)
+            neg_evals[evals_.data < 0] = -evals_.data[evals_.data < 0]
```
Balandat, line 168: nit: assign `evals_.data < 0` to local var to avoid doing the comparison twice

jacobrgardner: Are these lines just `evals_.clamp_min_(0)`?

wjmaddox: Somehow, I didn't realize that existed.

jacobrgardner:
Dismissed the review requesting changes. This now looks fine to me, although I still feel pretty strongly that we should code up a LogDet function. This would probably avoid the need to default to fp64, too.

I can get around to this this weekend maybe if it isn't done before then.

PR #911, opened 2019-10-24
Fix support for batches in the index kernel

darbour:
There was a bug in the index kernel where things weren't being properly reshaped to support batches. For, example the suggestion for a  Hadamard LMC made on pull request #661 throws an error. This pull request should provide reshaping to avoid these issues.

### DIFF
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -47,9 +47,9 @@ class IndexKernel(Kernel):
         if var_constraint is None:
             var_constraint = Positive()
 
-        self.register_parameter(
-            name="covar_factor", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks, rank))
-        )
+        self.tidcs = torch.tril_indices(num_tasks, num_tasks)
+        self.register_parameter("covar_factor", torch.nn.Parameter(torch.randn(*self.batch_shape, self.tidcs.size(-1))))
+        self.register_buffer("_C", torch.zeros(*self.batch_shape, num_tasks, num_tasks))
         self.register_parameter(name="raw_var", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks)))
         if prior is not None:
             self.register_prior("IndexKernelPrior", prior, self._eval_covar_matrix)
@@ -68,14 +68,8 @@ class IndexKernel(Kernel):
         self.initialize(raw_var=self.raw_var_constraint.inverse_transform(value))
 
     def _eval_covar_matrix(self):
-        var = self.var
-        eye = torch.eye(var.shape[-1], dtype=var.dtype, device=var.device)
-        if len(self.batch_shape) > 0:
-            eye = eye.view((1, var.shape[-1], var.shape[-1])).repeat(*self.batch_shape, 1, 1)
-            D = var.repeat(1, var.shape[-1]).view(*self.batch_shape, var.shape[-1], var.shape[-1]) * eye
-        else:
-            D = var * eye
-        return self.covar_factor.matmul(self.covar_factor.transpose(-1, -2)) + D
+        self._C[..., self.tidcs[0], self.tidcs[1]] = self.covar_factor
+        return self._C @ self._C.transpose(-1, -2) + torch.diag_embed(self.var)
 
     @property
     def covar_matrix(self):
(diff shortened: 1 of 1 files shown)


gpytorch/kernels/index_kernel.py
```diff
@@ -76,7 +69,12 @@ def _set_var(self, value):
...
     def _eval_covar_matrix(self):
         var = self.var
-        D = var * torch.eye(var.shape[-1], dtype=var.dtype, device=var.device)
+        eye = torch.eye(var.shape[-1], dtype=var.dtype, device=var.device)
+        if len(self.batch_shape) > 0:
+            eye = eye.view((1, var.shape[-1], var.shape[-1])).repeat(*self.batch_shape, 1, 1)
+            D = var.repeat(1, var.shape[-1]).view(*self.batch_shape, var.shape[-1], var.shape[-1]) * eye
+        else:
+            D = var * eye
         return self.covar_factor.matmul(self.covar_factor.transpose(-1, -2)) + D
```
Balandat: Here's my suggestion: 

```
def _eval_covar_matrix(self):
    cf =  self.covar_factor
    return cf @ cf.transpose(-1, -2) + torch.diag_embed(self.var)
```

gpytorch/kernels/index_kernel.py
```diff
@@ -87,5 +85,9 @@ def covar_matrix(self):
 
     def forward(self, i1, i2, **params):
         covar_matrix = self._eval_covar_matrix()
-        res = InterpolatedLazyTensor(base_lazy_tensor=covar_matrix, left_interp_indices=i1, right_interp_indices=i2)
+        res = InterpolatedLazyTensor(
+            base_lazy_tensor=covar_matrix,
+            left_interp_indices=i1.repeat(*self.batch_shape, 1, 1),
+            right_interp_indices=i2.repeat(*self.batch_shape, 1, 1),
```
Balandat: Do you need to repeat here? Or can you do expand to avoid allocating memory?
```
ij.expand(self.batch_shape + ij.shape)
```

Balandat:
Also, I wonder if we should parameterize the covariance matrix using a lower triangular matrix, that gets rid of some free parameters:

```
# in the constructor
self.tidcs = torch.tril_indices(num_tasks, num_tasks)
self.register_parameter("covar_factor", torch.randn(*batch_shape, tidcs.size(-1))
self.register_buffer("_C", torch.zeros(*batch_shape, num_tasks, num_tasks))

# in the evaluation 
self._C[..., self.tidcs[0], self.tidcs[1]] = self.covar_factor
return self._C @ self.C.transpose(-1, -2) + torch.diag_embed(self.var)
```

This is another thing that would make a good candidate for a torch native lazy tensor...

gpytorch/kernels/index_kernel.py
darbour: Thanks! This is so much cleaner.

Balandat: let's hope it works :)

Balandat:
I wasn't expecting you to include that parameterization suggestion as part of this PR. This may potentially cause backward-compatibility issues w.r.t.  loading a model see https://github.com/cornellius-gp/gpytorch/pull/912#issuecomment-546200447

darbour:
> I wasn't expecting you to include that parameterization suggestion as part of this PR. This may potentially cause backward-compatibility issues w.r.t. loading a model see [#912 (comment)](https://github.com/cornellius-gp/gpytorch/pull/912#issuecomment-546200447)

darbour:
Clumsy fingers accidentally closed this last night.

gpytorch/kernels/index_kernel.py
```diff
@@ -75,17 +68,22 @@ def _set_var(self, value):
...
-        return self.covar_factor.matmul(self.covar_factor.transpose(-1, -2)) + D
+        cf = self.covar_factor
+        return cf @ cf.transpose(-1, -2) + torch.diag_embed(self.var)
 
     @property
     def covar_matrix(self):
-        var = self.var
-        res = PsdSumLazyTensor(RootLazyTensor(self.covar_factor), DiagLazyTensor(var))
-        return res
+        with torch.autograd.set_detect_anomaly(True):
```
Balandat: delete?

gpytorch/kernels/index_kernel.py
```diff
@@ -75,17 +68,22 @@ def _set_var(self, value):
...
+            var = self.var
+            # res = PsdSumLazyTensor(RootLazyTensor(self.covar_factor), DiagLazyTensor(var))
```
Balandat: delete?

PR #912, opened 2019-10-25
Improved parameterization for correlation matrices

Balandat:
This uses a lower triangular (w/ columns restricted up to the specified rank) parameterization of the to-be-inferred correlation matrix in `MultitaskGaussianLikelihood`.

This parameterization uses
```
rank * (num_tasks - 0.5 * (rank - 1)) - 1
```
instead of
```
num_tasks * (1 + rank)
``` 
parameters for the set of rank `rank` correlation matrices. As the plot below shows the lower the rank the smaller the fraction of parameters (relative to the current parameterization) this uses. 

The black line is the case when `rank=num_tasks`, in which the fraction can be found to be
```0.5 - 1 / (num_tasks * (1 + num_tasks))```. 

Asymptotically (as `num_tasks` -> infinity), the fraction approaches `rank / (1 + rank)` 

param_size

Disclaimer: Haven't tested this in a model yet.

### DIFF
--- a/gpytorch/likelihoods/multitask_gaussian_likelihood.py
+++ b/gpytorch/likelihoods/multitask_gaussian_likelihood.py
@@ -17,22 +17,13 @@ from ..lazy import (
     RootLazyTensor,
 )
 from ..likelihoods import Likelihood, _GaussianLikelihoodBase
-from ..utils.deprecation import _deprecate_kwarg_with_transform
 from .noise_models import MultitaskHomoskedasticNoise
 
 
 class _MultitaskGaussianLikelihoodBase(_GaussianLikelihoodBase):
     """Base class for multi-task Gaussian Likelihoods, supporting general heteroskedastic noise models. """
 
-    def __init__(
-        self,
-        num_tasks,
-        noise_covar,
-        rank=0,
-        task_correlation_prior=None,
-        batch_shape=torch.Size(),
-        **kwargs,
-    ):
+    def __init__(self, num_tasks, noise_covar, rank=0, task_correlation_prior=None, batch_shape=torch.Size()):
         """
         Args:
             num_tasks (int):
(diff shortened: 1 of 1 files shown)


Balandat:
@bletham this may be of interest to you given some of your recent triangular parameterizations.

Balandat:
I guess one potential issue here is that we'd be breaking model loading for old models. One way to fix this would be to reverse engineer the new parameterization from the old one upon loading the state dict. That is, use the old parameterization to compute the correlation matrix, cholesky-decompose it, and index with the rank restricted trilu indices into it. This will result in a parameterization that will result in the same correlation matrix as the initial parameterization did.

The question is whether this is necessary...

gpleiss:
Everything looks good - but I do have the same concern about the breaking change...

I'm not sure how these models are used in practice, and so in general I do think it might be better to have some state_dict hacking, which is definitely annoying. @jacobrgardner your thoughts?

jacobrgardner:
Since this primarily affects relatively small matrices, I think it should be straightforward to maintain backwards compatibility. What I propose is the following:

1. Use `Module._register_load_state_dict_pre_hook` (docs [here](https://github.com/pytorch/pytorch/blob/3f119a5f52210f599ad9f5916c83186b8a01856a/torch/nn/modules/module.py#L702)) on modules we change the parameterization of.
2. Check the value in the supplied `state_dict`. If it is already triangular, do nothing.
3. If it is not triangular, compute the true covariance matrix by doing `RR'` (where `R` is the value in the state_dict), and run Cholesky on the result.
4. Load the Cholesky factor in to the Parameter instead of `R`, and raise a deprecation warning.

Balandat:
I'll work on the deprecation. How do you feel about deprecating the `batch_size` and `param_transform` deprecations? These have been around for a while, I think it's time to kill them.

Balandat:
Tested model, added deprecation, and validated deprecation works:

[test_MTGLH_deprecation.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/3774008/test_MTGLH_deprecation.ipynb.txt)

Balandat:
Hmm I realized in the previous parameterization we'd get results that were not valid correlation matrices. This should fix that issue, too.

PR #916, opened 2019-10-29
Register backward hooks to grad functions for cache clearing.

### LINKED ISSUE #891: [Feature Request] Backward hooks in Exact GPs
# 🚀 Feature Request

In some cases we want to use larger models that are composed of different models, e.g. for modeling noise level with a separate GP in a heteroskedastic likelihood. In this case the noise model needs to be evaluated in eval mode while the other one is in train mode. For this to work properly, we need to make sure the test caches get cleared after backpropagating through them. 

A specific implementation for the noise models was done in https://github.com/cornellius-gp/gpytorch/pull/870, but this would be more generally useful and we should consider making this part of `ExactGP`.

jacobrgardner:
Right now, if you backward through GP predictions to the training data or GP hyperparameters and update either one, the caches don't get cleared but are dirty (the stuff they depend on has changed).

This PR adds hooks to the grad functions of the mean and covariance caches so that if they are backwarded through they get cleared.

Importantly, if you backward through predictions but NOT the caches because `settings.detach_test_caches` is True by default (e.g., you want derivatives w.r.t. test inputs only), the caches will not be cleared because they don't need to be.

In practice, what this does is eliminate the need to manually toggle back and forth between train mode and test mode in models that are trained with a GP component in prediction mode.

Fixes #891

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -19,6 +19,7 @@ from ..lazy import (
 )
 from ..utils.interpolation import left_interp, left_t_interp
 from ..utils.memoize import cached, add_to_cache
+import functools
 
 
 def prediction_strategy(train_inputs, train_prior_dist, train_labels, likelihood):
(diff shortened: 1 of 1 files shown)


KeAWang:
We probably also want to change instances of `module.forward` to `module.__call__` in our internal code so that hooks are registered properly?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -69,10 +70,19 @@ def _exact_predictive_covar_inv_quad_form_cache(self, train_train_covar_inv_root
...
             - A precomputed cache
         """
+        res = train_train_covar_inv_root
         if settings.detach_test_caches.on():
-            return train_train_covar_inv_root.detach()
-        else:
-            return train_train_covar_inv_root
+            res = res.detach()
+
+        def clear_cache_hook(module, *args, **kwargs):
```
Balandat: As this doesn't have any local scope, it might make sense to move this def outside this instance method and into the module. You can re-use that below as well.

jacobrgardner: Done

jacobrgardner:
@KeAWang Most cases like this are intentional. For example, many kernels call `base_covar_module.forward` because we don't want to go through `Kernel.__call__` a second time.

I agree It might be better to do something like `torch.nn.Module.__call__(base_covar_module, ...args)` everywhere we currently call `base_covar_module.forward(...args)` directly.

PR #920, opened 2019-11-04
Add stern warning when calling evaluate() on KeOpsLazyTensor.

jacobrgardner:
This should discourage people from using KeOps + Cholesky together, which is almost always the wrong thing to do (either don't use KeOps, or don't use Cholesky).


jacobrgardner:
This doesn't quite work, since we evaluate at test time and for the preconditioner for legitimate purposes.

Issue #922, opened 2019-11-04
[Discussion] Make has_lengthscale an attr rather than kwarg


jahall:
Currently `has_lengthscale` is passed in as an argument to `kernels.Kernel`. Would this not make more sense as a class attribute which can then be overridden by child classes? I can't think of situation in which it would make sense to allow a user to specify this via a kwarg...

Example proposed implementation:
```
class Kernel(...):
    _has_lengthscale = False

class RBFKernel(Kernel):
    _has_lengthscale = True
```

jacobrgardner:
Yeah, I think this is a good idea. At the moment, a lot of the Kernel kwargs have turned in to a complete mess to be honest. Most of them are absorbed in to `**kwargs` in a lot of the subclasses, and there are a huge number of them, so converting some of them to attributes and cleaning up the `kwargs` sounds like a good thing to do for the next release.

jahall:
Great. Also, I'm not sure it makes sense to have lengthscale support baked in to the base `Kernel`. Maybe in some `StationaryKernel` subclass or something?

gpleiss:
@jahall - we thought about creating a StationaryKernel subclass a while back. Since most of our kernels are stationary we thought it was maybe a bit too much overhead. However, I wouldn't be opposed to adding it into the library.

jahall:
I think it might be a helpful distinction - though I understand your reticence to introduce unnecessary class hierarchy.

KeAWang:
Once we remove it from the kwargs, we can just change `_has_lengthscale` to `_is_stationary` which is a bit cleaner than unnecessarily subclassing 

Also I think the Toeplitz stuff right now doesn't check to see if the kernel is stationary or not.

jahall:
I think `has_lengthscale` and `is_stationary` should be separate attributes because, for example, the product of two stationary kernels would be stationary...but would not have a lengthscale

gpleiss:
Closed by #925

PR #925, opened 2019-11-06
Refactor of has_lengthscale

jahall:
Following the discussion on #922, I would appreciate feedback on this PR. The proposed changes are basically as follows:

1. Make `has_lengthscale` a class attribute
2. Introduce an `is_stationary` property
3. Re-order import in accordance with PEP8 convention of builtins followed by third-party followed by internal

Not sure I have correctly captured exactly which kernels are stationary or not...also would need to add more test coverage, hence PR is in draft.


gpytorch/kernels/grid_kernel.py
```diff
@@ -133,7 +134,7 @@ def forward(self, x1, x2, diag=False, last_dim_is_batch=False, **params):
                 ]  # Each entry i contains a grid_size[i] x grid_size[i] covariance matrix
                 if last_dim_is_batch:
                     # Note that this requires all the dimensions to have the same number of grid points
-                    covar = gpytorch.cat([c.unsqueeze(-3) for c in covars], dim=-3)
+                    covar = gpytorch_cat([c.unsqueeze(-3) for c in covars], dim=-3)
```
jacobrgardner: Any reason for the `gpytorch.cat -> gpytorch_cat` change?

gpytorch/kernels/grid_kernel.py
```diff
@@ -1,13 +1,14 @@
...
-from ..lazy import delazify, ToeplitzLazyTensor, KroneckerProductLazyTensor
 from .. import settings
-import gpytorch
-from gpytorch.utils.grid import create_data_from_grid, convert_legacy_grid
-from typing import List
+from ..utils.grid import create_data_from_grid, convert_legacy_grid
+from ..lazy import delazify, ToeplitzLazyTensor, KroneckerProductLazyTensor, cat as gpytorch_cat
 
 
 class GridKernel(Kernel):
```
jacobrgardner: Now that Kernels define whether they are stationary, we can raise a RuntimeError in the construct of GridKernel if `base_kernel` is not stationary (it is required to be).

gpytorch/kernels/grid_kernel.py
jahall: Couple of reasons:
1. I think it's good to be consistent with relative vs absolute imports and seems like in the vast majority of cases you do relative imports
2. I imported it as `gpytorch_cat` instead of just `cat` to retain the clarity that it is the `gpytorch` version of cat rather than `torch.cat`...though in retrospect, just importing as `cat` would maybe be preferable

gpytorch/kernels/grid_kernel.py
jahall: Happy to add this this check in.

.isort.cfg
```diff
@@ -1,2 +1,2 @@
 [settings]
-known_third_party = matplotlib,numpy,pyro,setuptools,sphinx_rtd_theme,torch
+known_third_party = gpytorch,matplotlib,numpy,pyro,setuptools,sphinx_rtd_theme,torch
```
gpleiss: Not sure about this... @jacobrgardner ?

jahall:
> I think the spectral mixture kernel needs to be marked as `is_stationary` (though it does not have a lengthscale). Other than that LGTM.

Cool - marked it with `is_stationary = True`.

.isort.cfg
jahall: For some reason the `pre-commit` hook is forcing me to add this as per below:
```
Joes-Mac:gpytorch-fork Joe$ git commit -am 'Revert spurious .isort.cfg and mark SpectralMixtureKernel as stationary'
Flake8...................................................................Passed
Check for byte-order marker..............................................Passed
Check for case conflicts.................................................Passed
Check for merge conflicts................................................Passed
Fix End of Files.........................................................Passed
Forbid new submodules....................................................Passed
Mixed line ending........................................................Passed
Trim Trailing Whitespace.................................................Passed
Debug Statements (Python)................................................Passed
black....................................................................Passed
seed isort known_third_party.............................................Failed
hookid: seed-isort-config

Files were modified by this hook.

isort....................................................................Passed
Check file encoding......................................................Passed
Non-executable shell script filename ends in .sh.....(no files to check)Skipped
Forbid binaries......................................(no files to check)Skipped
CRLF end-lines checker...................................................Passed
No-tabs checker..........................................................Passed
```

jahall: I _suspect_ if we consistently used relative imports from within the `gpytorch` package this would go away. I think this would be good to do anyway.

jahall: ^ could related to: https://github.com/pre-commit/pre-commit/issues/842

jacobrgardner: My best guess is it is treating the absolute gpytorch imports from the unit tests as "third party" because they don't live anywhere in the gpytorch module itself.

For now, I think this is mostly harmless relative to the benefits we get from having isort run consistently on the package -- just yesterday isort helped me fix half a dozen bad relative imports that were only working properly because of the order they appeared in.

jacobrgardner:
Thanks for the refactor, this is much cleaner!

.isort.cfg
jahall: I agree that `isort` is great. And this `pre-commit` embedding in git is awesome - I'm going to use it in my projects!

PR #929, opened 2019-11-08
Add deterministic_probes setting to improve L-BFGS convergence

jacobrgardner:
This adds a flag that causes us to not resample probe vectors for our log det estimates each iterations. This will result in some mild bias in our derivatives, but reduced stochasticity in the MLL value. Using this flag should improve the convergence of methods like L-BFGS that don't deal well with even mild amounts of stochasticity in a loss function.

### DIFF
--- a/gpytorch/functions/_inv_quad_log_det.py
+++ b/gpytorch/functions/_inv_quad_log_det.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 
+import warnings
+
 import torch
 from torch.autograd import Function
 
@@ -70,6 +72,10 @@ class InvQuadLogDet(Function):
             num_random_probes = settings.num_trace_samples.value()
             if preconditioner is None:
                 if settings.deterministic_probes.on():
+                    warnings.warn(
+                        "Deterministic probes will currently work only if you aren't training multiple independent"
+                        " models simultaneously."
+                    )
                     if settings.deterministic_probes.probe_vectors is None:
                         probe_vectors = torch.empty(matrix_shape[-1], num_random_probes, dtype=dtype, device=device)
                         probe_vectors.bernoulli_().mul_(2).add_(-1)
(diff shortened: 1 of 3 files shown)


Balandat:
I'm concerned about the scope probe vector caching here - seems like it's global? Will this just break if you compute the mll of two different lazies of different size if this this setting on?

jacobrgardner:
I guess that's fair, yeah. I'm not sure where to store them though because I don't see what we have in the Function that is persistent. Even the lazy tensor is something we reconstruct there.

Balandat:
Hmm yeah it's not an easy thing to do. In botorch we have the `MCSampler` abstraction for this which is a simple wrapper that holds the samples, and sampling from a posterior means doing a forward pass on the sampler on the posterior object: https://github.com/pytorch/botorch/blob/master/botorch/sampling/samplers.py#L24

FWIW, let me try this first to see what the performance of this is, we can worry about making it pretty / less gobal later.

jacobrgardner:
@Balandat Have you had a chance to look in to whether this matters for BFGS training? I'm just wondering if we should bother trying to make this in to 1.0 or just fix it up after, especially since it seems like finding a place to cache the probe vectors might turn in to a bigger project. It's a shame PyTorch moved to static class Functions, this would be really easy if we had a persistent `self` we could keep around...

Balandat:
I have not, and I probably won't get to it for some time. I'm fine with not including this into 1.0 and taking a closer look at this after instead.

gpytorch/settings.py
```diff
@@ -137,6 +137,22 @@ class detach_test_caches(_feature_flag):
     _state = True
 
 
+class deterministic_probes(_feature_flag):
+    """
+    Whether or not to resample probe vectors every iteration of training. If True, we use the same set of probe vectors
+    for computing log determinants each iteration. This introduces small amounts of bias in to the MLL, but allows us
+    to compute a deterministic estimate of it which makes optimizers like L-BFGS more viable choices.
+    """
```
Balandat: Let's add a warning here that this will cause weird behavior when training multiple models with this setting on b/c of the global scope of the caching.

gpytorch/functions/_inv_quad_log_det.py
```diff
@@ -67,15 +67,48 @@ def forward(
         if (probe_vectors is None or probe_vector_norms is None) and logdet:
             num_random_probes = settings.num_trace_samples.value()
             if preconditioner is None:
-                probe_vectors = torch.empty(matrix_shape[-1], num_random_probes, dtype=dtype, device=device)
-                probe_vectors.bernoulli_().mul_(2).add_(-1)
+                if settings.deterministic_probes.on():
+                    if settings.deterministic_probes.probe_vectors is None:
```
Balandat: Let's add a warning here that this will cause weird behavior when training multiple models with this setting on b/c of the global scope of the caching.

gpytorch/settings.py
jacobrgardner: Added a warning comment. If someone finds the setting, they'll see.

PR #934, opened 2019-11-11
Add pre-commit to travis and setup, additional fixes

jacobrgardner:
- pre-commit now installed from setup.py dev
- travis tests all pre-commit hooks to ensure we don't merge PRs that fail some hooks.
- Additional linter + isort fixes to make sure pre-commit hooks all pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,0 @@
-[tool.black]
-line-length = 120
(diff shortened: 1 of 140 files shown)


pyproject.toml
```diff
@@ -1,2 +0,0 @@
-[tool.black]
```
Balandat: Is there any particular reason to remove this file? This way you can just run `black .` instead of `black -l 120` and get the correct option

Issue #943, opened 2019-11-14
[Bug] cholesky singular U


deltaskelta:
# 🐛 Bug

## To reproduce

I followed the directions here for a variational GP (https://gpytorch.readthedocs.io/en/latest/examples/05_Scalable_GP_Regression_Multidimensional/SVGP_Regression_CUDA.html#Overview) with the HEAD build of gpytorch and it trains for a while and then crashes with...

** Stack trace/error message **
```
  File "main.py", line 93, in <module>
    pred = model(x)███████████████████████████████████████████████████████████████████████▉       | 999/1088 [00:42<00:02, 31.22it/s]
  File "/st2/jeff/anaconda3/envs/jeff/lib/python3.7/site-packages/gpytorch-0.3.6-py3.7.egg/gpytorch/models/approximate_gp.py", line 56, in __call__
    return self.variational_strategy(inputs, prior=prior)
  File "/st2/jeff/anaconda3/envs/jeff/lib/python3.7/site-packages/gpytorch-0.3.6-py3.7.egg/gpytorch/variational/whitened_variational_strategy.py", line 240, in __call__
    return Module.__call__(self, x)
  File "/st2/jeff/anaconda3/envs/jeff/lib/python3.7/site-packages/gpytorch-0.3.6-py3.7.egg/gpytorch/module.py", line 24, in __call__
    outputs = self.forward(*inputs, **kwargs)
  File "/st2/jeff/anaconda3/envs/jeff/lib/python3.7/site-packages/gpytorch-0.3.6-py3.7.egg/gpytorch/variational/whitened_variational_strategy.py", line 141, in forward
    induc_induc_covar = CholLazyTensor(induc_induc_covar.cholesky())                                                                
  File "/st2/jeff/anaconda3/envs/jeff/lib/python3.7/site-packages/gpytorch-0.3.6-py3.7.egg/gpytorch/lazy/lazy_tensor.py", line 738, in cholesky
    res = self._cholesky()
  File "/st2/jeff/anaconda3/envs/jeff/lib/python3.7/site-packages/gpytorch-0.3.6-py3.7.egg/gpytorch/utils/memoize.py", line 34, in g
    add_to_cache(self, cache_name, method(self, *args, **kwargs))
  File "/st2/jeff/anaconda3/envs/jeff/lib/python3.7/site-packages/gpytorch-0.3.6-py3.7.egg/gpytorch/lazy/lazy_tensor.py", line 413, in _cholesky
[...]

PR #944, opened 2019-11-14
Cholesky gives a better error when there are NaNs

gpleiss:
Based on a note in #943

### DIFF
--- a/gpytorch/utils/cholesky.py
+++ b/gpytorch/utils/cholesky.py
@@ -4,6 +4,8 @@ import warnings
 
 import torch
 
+from .errors import NanError
+
 
 def psd_safe_cholesky(A, upper=False, out=None, jitter=None):
     """Compute the Cholesky decomposition of A. If A is only p.s.d, add a small jitter to the diagonal.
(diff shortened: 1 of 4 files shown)


jacobrgardner:
@gpleiss Does this actually address #943? Like, definitely agree we want Cholesky to raise a more specific error, but if the real problem is that our SVGP regression notebook is crashing, we probably want to do more than have a nicer error message 😅

gpleiss:
Maybe "addresses" is the wrong word. But I think the nicer error message does make our lives better.

jacobrgardner:
Yeah, no arguments here about merging the PR! I just want to make sure we don't close the issue until the example runs through.

gpytorch/utils/cholesky.py
```diff
@@ -22,6 +24,12 @@ def psd_safe_cholesky(A, upper=False, out=None, jitter=None):
         L = torch.cholesky(A, upper=upper, out=out)
         return L
     except RuntimeError as e:
+        isnan = torch.isnan(A)
```
jacobrgardner: As long as we are doing n^2 work to check this, maybe we should do `torch.isnan(A.sum(-1))` or something to avoid materializing n^2 memory too

jacobrgardner: Approved the PR since it shouldn't be a big deal though -- people likely shouldn't be using Cholesky at the size where GPU memory is a concern. It's just technically possible to check for NaNs without duplicating memory.

gpleiss: Yeah that's a fair point.

Balandat: FWIW I raised this issue with the pytorch folks hopefully we can get them to implement a better error message in the case of NaNs upstream.

PR #946, opened 2019-11-14
max_cholesky_size -> 800

jacobrgardner:
Current benchmarking against our new default tolerances + various code changes suggests that CG overtakes Cholesky around n=800-n=1000. Therefore, we should not use CG until then.

### DIFF
--- a/gpytorch/settings.py
+++ b/gpytorch/settings.py
@@ -360,7 +360,7 @@ class max_cholesky_size(_value_context):
     Default: 128
     """
 
-    _global_value = 128
+    _global_value = 800
 
 
 class max_root_decomposition_size(_value_context):


Balandat:
Is this similar for CPU / GPU? Should there be device-specific default settings?

jacobrgardner:
Yeah possibly. I haven't tested too much with CPU since it's not a typical regime for me -- you have a plot we might pull some guidance from for a CPU threshold I seem to recall?

Balandat:
Hmm the recent plot I had was for computing mvn logprob. The "Cholesky vs CG" is kind of hard to answer generally since it depends on the matrix structure. @gpleiss has the code for generating the logprob plots, that could easily be adapted to evaluate the more proximal question fo solving the linear system.

gpleiss:
I'm okay with this new default - but I'll give @Balandat the final say.
We could also add a `max_cholesky_size_cpu` flag in another PR?

jacobrgardner:
Yeah adding a separate CPU flag somewhere sounds good

PR #957, opened 2019-11-20
Wendland kernel

phumm:
I added the Wendland kernel and hope I made everything in a way that you can easily accept this pull request. Let me know if I can do anything else.
My intention behind this is that the Wendland kernel gives us a sparse kernel gram matrix and I hope to make use of sparse matrix inversion techniques to speed up GP inference. I'll also create an issue to ask you for your opinion and some tips where I can include some code of my own to do the sparse matrix inversion.


jacobrgardner:
@phumm See my comments on the issue you opened (#958). In general, I'm fine merging this if we handle the sparse structure nicely, since it could serve as another example where CG inference is just sort of automatically efficient with a decent matmul. That, and I don't think either implementing `SparseNonLazyTensor` or adapting `NonLazyTensor` to also wrap torch sparse matrices will actually be that much work.

PR #964, opened 2019-11-23
Validation code for MultivariateNormal

rhaps0dy:
I've resolved the TODO item to add validation for the sizes of the mean and variance.

I had also implemented a `__repr__` method that would show you the shape of the MultivariateNormal instead of the values of the tensors. But I think it's best not to upstream that. The `__repr_sizes` method was sharing code with it.

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -77,16 +77,13 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
             variable. The first (second) Tensor is the lower (upper) end of
             the confidence region.
         """
-        std2 = self.stddev.mul(2)
+        std2 = self.stddev.mul_(2)
         mean = self.mean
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Should we validate batch shapes too? One option would be to require they all be identical. Another possibility would be to run something like 

`gpytorch.utils.broadcasting._mul_broadcast_shape(mean.unsqueeze(-1).shape, covar.shape)`. This would ensure things like sampling would work if you had different batch dimensions that you wanted to broadcast, as well as validate the last two dim shapes.

rhaps0dy:
Currently there is already a line doing that by default. It's just below the context show in the diff:

https://github.com/cornellius-gp/gpytorch/pull/964/files#diff-b53061e95ced80fbf7a2fac2150db1b1R42

gpleiss:
Thanks @rhaps0dy !

PR #965, opened 2019-11-23
Indexing MultivariateNormals

rhaps0dy:
if `dist` is a `MultivariateNormal` object, now it is possible to write `dist[index]` and create a new distribution with the appropriate marginalized mean and covariance.

The test provides examples of usage.

This is helpful in my research and it might also be helpful to upstream it?

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -8,7 +8,7 @@ from torch.distributions.kl import register_kl
 from torch.distributions.utils import _standard_normal, lazy_property
 
 from .. import settings
-from ..lazy import LazyTensor, delazify, lazify, DiagLazyTensor
+from ..lazy import DiagLazyTensor, LazyTensor, delazify, lazify
 from ..utils.broadcasting import _mul_broadcast_shape
 from .distribution import Distribution
 
@@ -234,6 +234,7 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
             else:
                 new_cov = self.lazy_covariance_matrix[(*rest_idx, last_idx, slice(None, None, None))][..., last_idx]
         return self.__class__(mean=new_mean, covariance_matrix=new_cov)
+
 
 @register_kl(MultivariateNormal, MultivariateNormal)
 def kl_mvn_mvn(p_dist, q_dist):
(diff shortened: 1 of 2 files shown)


jacobrgardner:
Yeah, we recently added getitem methods to kernels and I've actually found it convenient enough that I've been pondering adding them to GPs and Distributions so that `model[idx]` indexes in to a batch GP. Nice to see it done for distributions :-)!

Test failures look to just be lint and formatting related. if you look at the instructions in the readme for installing precommit hooks on your local copy, everything you need to run to fix these will get run for you when you try to commit. Otherwise just fix lint issues and run black + isort.

rhaps0dy:
Thanks for telling me! I had never used pre-commit hooks so it would have taken me quite a bit of time to find. I managed to run the things on their own, but running `pre-commit install` and then committing shows the following:

```
seed isort known_third_party.............................................Failed
hookid: seed-isort-config
```
while it adds `gpytorch` in the `known_third_party` setting. I don't know why this is the case, maybe you do? Don't worry too much about it though -- I think I managed to get the formatting right anyways.

jacobrgardner:
Yeah our .isort config file seems to be in a state of flux from commit to commit. I've been meaning to look in to it but for now we've been just letting the hooks change it back and forth so that we can get the precommit hooks to be run consistently.

Thanks for the PR!

rhaps0dy:
That was a really prompt reply. Thank you for making `gpytorch`!

rhaps0dy:
I've noticed that when attempting to do: `dist[None]` in the test, to add another dimension, the program fails. I didn't take this case into account originally, but the code fails in [the LazyTensor `__getitem__` implementation](https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/lazy/lazy_tensor.py#L1731), which I suppose also didn't take it into account. Supporting the use of `None`s would require changing `_compute_getitem_size`, which I'm not sure is worth it. Should we add a special error message when `None` items are used? Maybe the error message should be both in `LazyTensor` and `MultivariateNormal`.

I've also noticed a bug in my test, and added a non-implemented `__getitem__` as requested. @Balandat , if it's so easy, would you mind implementing it? I don't really know what the conceptual difference between the MVN and the MultitaskMVN is.

jacobrgardner:
Could we handle None in the top level `__getitem__` by calling unsqueeze and inserting a `:` or a `0` where the None was?

Issue #709, opened 2019-05-29
Remove Woodbury formula from preconditioner computation


jacobrgardner:
Now that PyTorch master has support for batch mode QR, we should finish removing usage of the Woodbury formula and only rely on the QR decomposition, which is significantly more numerically stable.

The following two else branches should disappear:
https://github.com/cornellius-gp/gpytorch/blob/de92cbc6cf5127a9b688bd47ead681d642acf45d/gpytorch/lazy/added_diag_lazy_tensor.py#L97-L104
and:
https://github.com/cornellius-gp/gpytorch/blob/de92cbc6cf5127a9b688bd47ead681d642acf45d/gpytorch/lazy/added_diag_lazy_tensor.py#L117-L124

esass10:
Hi, can I be assigned this issue? I am very interested

Balandat:
Hmm I believe this issue is not current anymore? Note that all the lazy tensor logic has long moved into the `linear_operator` package, it now lives here: https://github.com/cornellius-gp/linear_operator/blob/main/linear_operator/operators/added_diag_linear_operator.py#L96 (but is already using the batch QR decomposition).

PR #968, opened 2019-11-24
Remove use of woodbury

jahall:
Removal of all legacy use of `woodbury.py` as per #709. Also contains a refactoring of the mega `AddedDiagLazyTensor._preconditioner` method to break out the cache initialisation piece. Leaving the original method a little cleaner an easier to understand.

Also added a `.pylintrc` to resolve [this](https://stackoverflow.com/questions/50319943/pytorch-error-message-torch-has-no-member) problem encountered when using VSCode.


jacobrgardner:
@gpleiss @jahall Yes, sorry! This is on my radar but I've just been travelling. I just want to give this one a little more scrutiny because it touches actual math in the package. Hopefully I can get it reviewed before Thursday.

PR #980, opened 2019-12-01
upgrade to the latest version

PingjunChen:
I tried to use `pip install git+https://github.com/cornellius-gp/gpytorch.git` to install the latest version. However, under the circumstance of already installed `gpytorch-0.3.6`, the provided installation command will not upgrade to the latest version. In order to install the latest version, `--upgrade` command should be explicitly added.


gpleiss:
Good catch. Thanks @PingjunChen !

PR #1008, opened 2020-01-01
Get variational batch shape from variational distribution

jacobrgardner:
Alright, so if you have a batch variational GP that are sharing inducing points (so `inducing_points` are `m x d`, but the variational parameters are `b x m` and `b x m x m`) AND you pass a non batch input (e.g., `n x d` instead of explicitly `b x n x d`), the existing expands fail.

This is a bit of an edge case, but it did work before 1.0.0. To fix this on 1.0.0, this PR explicitly grabs the batch shape from the variational distribution object and explicitly expands inducing points and x (which is a no-op if they already match).

We might push out 1.0.1 with this + anything else we find in a week or so. For now, on 1.0.0, this (a) is a pretty rare edge case, and (b) can be worked around by explicitly giving `x` a batch shape (e.g., everything works as intended if you manually do `x.expand(*batch_shape, *x.shape[-2:])`

cc/ @samuelstanton who encountered this initially, and @gpleiss who'll have to review. I also have a quick smoke test I'll add for this to the unit tests, since this is clearly untested currently.

### DIFF
--- a/gpytorch/variational/_variational_strategy.py
+++ b/gpytorch/variational/_variational_strategy.py
@@ -7,7 +7,6 @@ import torch
 from .. import settings
 from ..distributions import Delta, MultivariateNormal
 from ..module import Module
-from ..utils.broadcasting import _mul_broadcast_shape
 from ..utils.memoize import cached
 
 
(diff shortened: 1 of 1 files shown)


samuelstanton:
This fix doesn't exactly replicate the earlier behavior. Previously if both the inducing points and the inputs were shared calling `VariationalStrategy.__call__(model, inputs)` would result in a 2D tensor being passed to the forward method. Now since everything is expanded in the variational strategy, `forward` sees the expanded tensor.

If you're doing DKL that means you have one of two options: 
A) have a flag for shared inputs and just use a slice in the foward method, or 
B) flatten and pass everything through the feature extractor and then reshape, which is wasteful of both memory and operations. 

I certainly want to avoid B) and A) feels very clunky. How was this handled before 1.0.0?

PR #1017, opened 2020-01-13
Handle non-gpytorch submodules in constraint_for_parameter_name

Balandat:
Currently, attaching non-gpytorch torch modules as submodules to gpytorch modules causes issues with the constraint functionality. Specifically, `constraint_for_parameter_name ` assumes that all the submodules are GPyTorch modules and have a `_constraints` dictionary (as initialized here: https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/module.py#L19). 

This catches the `AttributeError` that is raised if that is not the case and returns `None`.

### DIFF
--- a/gpytorch/module.py
+++ b/gpytorch/module.py
@@ -262,8 +262,11 @@ class Module(nn.Module):
             base_module = submodule
             base_name = ".".join(components[1:])
 
-        constraint_name = base_name + "_constraint"
-        return base_module._constraints.get(constraint_name)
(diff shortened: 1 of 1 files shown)


jacobrgardner:
See #1014

Balandat:
Nah not really. I guess it’s more “pythonic” but that’s not a value in itself.

jacobrgardner:
Closes #1014

PR #977, opened 2019-11-29
Custom expand methods for priors

jacobrgardner:
Inheriting the behavior of `expand` from the base torch distributions was causing really strange behavior, I think because our priors also inherit from Module. In addition to the bug in #976, calling `prior.expand(some_torch_size)` was actually modifying `prior` in place in addition to returning a new Module with the correct size.

These issues were easily fixed by just defining our own `expand` methods for the various priors. I also think that this is a rare setting where overriding and defining our own expand behavior will actually make things easier to debug, because the upstream behavior is extremely weird: https://github.com/pytorch/pytorch/blob/dd52f50fc85e6710f020936c1fc5f14673508350/torch/distributions/gamma.py#L50-L57

Fixes #976

PR #1018, opened 2020-01-13
Custom expand method for MultivariateNormalPrior

jacobrgardner:
Fixes an issue where the built in expand method was deleting our prior buffers (see #977).

cc/ @idelbrid

### DIFF
--- a/gpytorch/priors/torch_priors.py
+++ b/gpytorch/priors/torch_priors.py
@@ -118,4 +118,12 @@ class MultivariateNormalPrior(Prior, MultivariateNormal):
         return module
 
     def expand(self, batch_shape):
-        return MultivariateNormal.expand(self, batch_shape, _instance=self)
+        from IPython.core.debugger import set_trace
+
+        set_trace()
+        batch_shape = torch.Size(batch_shape)
+        cov_shape = batch_shape + self.event_shape
+        new_loc = self.loc.expand(batch_shape)
+        new_scale_tril = self.scale_tril.expand(cov_shape)
+
+        return MultivariateNormalPrior(loc=new_loc, scale_tril=new_scale_tril)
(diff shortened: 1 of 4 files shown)


gpytorch/priors/torch_priors.py
```diff
@@ -118,4 +118,12 @@ def cpu(self):
         return module
 
     def expand(self, batch_shape):
-        return MultivariateNormal.expand(self, batch_shape, _instance=self)
+        from IPython.core.debugger import set_trace
+
+        set_trace()
```
gpleiss: Probably should remove this...

gpytorch/priors/torch_priors.py
```diff
@@ -118,4 +118,12 @@ def cpu(self):
...
+        batch_shape = torch.Size(batch_shape)
+        cov_shape = batch_shape + self.event_shape
+        new_loc = self.loc.expand(batch_shape)
```
gpleiss: Should this be `self.loc.expand(batch_shape + torch.Size([self.loc.size(-1)]))`?

jacobrgardner: @gpleiss Ugh. For the code repo as it is it is correct -- I went rogue and made `Prior`'s expand behave like tensors rather than distributions. e.g., to turn `a = torch.randn(3)` in to `1 x 3` you do `a.expand([1, 3])`, not `a.expand([1])` (same for `LazyTensors`).

If we wanted priors to follow the expand semantics of torch Distribution, we'd need to do some hackery here, where I currently call expand on the shape of the underlying Parameter so that the samples always have the same shape as the Parameter: https://github.com/cornellius-gp/gpytorch/blob/707bd988b3b22f0689a81177cea8ee681ebbf7f9/gpytorch/module.py#L395

This might involve adding batch_shapes and event_shapes or something to the parameters themselves, unless you have an obvious clean solution...

gpleiss: Nope - let's keep your rogue solution :)

PR #1025, opened 2020-01-16
batch mode DecoupledVariationalStrategy bug fix

samuelstanton:
adds `_decoupled_inducing_points` flag to `_VariationalStrategy` so that inputs and inducing points are correctly reshaped in `_VariationalStrategy.__call__`.


samuelstanton:
Actually on second thought the flag probably doesn't need to be in the state dict

gpleiss:
Sorry this fell off my radar! Where exactly does the current implementation fail?

Issue #1022, opened 2020-01-16
Saving gpytorch models as ONNX


driesvr:
Hi GPytorch team,

First off, thanks for creating a great piece of software! I was wondering if there's a way to save the models as ONNX files at this point, or if this would be in the plans for the future?

Cheers,

Dries

gpleiss:
Hi @driesvr - I don't think we have ONNX support at the moment. It's definitely something we have on our minds, but not sure exactly when it'd be added.

jacobrgardner:
I believe it may be possible currently by first saving a model as torch script and then saving as ONNX. Let me (1) put up an example notebook saving am exact GP and a variational GP to torch script, and then see if it works.

jacobrgardner:
Closed via #1026 -- just follow the example notebooks for converting to torchscript but instead of calling torch.jit.trace call the ONNX save on the model (which will trace it).

Liyongchao321:
Hi, 

I followed your example "TorchScript_Exact_Models" and tried to export it to ONNX format as below:

with torch.no_grad(), gpytorch.settings.fast_pred_var(), gpytorch.settings.trace_mode():
    model.eval()
    test_x = torch.linspace(0, 1, 51)
    pred = model(test_x)  # Do precomputation
    wrapper = MeanVarModelWrapper(model)
    pred_wrapper = wrapper(test_x)
    torch.onnx.export(wrapper, test_x, 'gpytorch_model.onnx', verbose=True, example_outputs=pred_wrapper) 

However, I got  KeyError: 'diag' .  Could you please help me with it? 
My torch version is 1.4.0.

Thanks,
Yongchao

jacobrgardner:
@Liyongchao321 This one may actually be on PyTorch. Could you open a separate issue for tracking?

supersjgk:
Hi @driesvr @Liyongchao321 were you able to successfully convert gpytorch model to ONNX?

Issue #970, opened 2019-11-25
[Feature Request] Can I use Libtorch to load Gpytorch model?


rangerbottle:
# 🚀 Feature Request

Can I use Libtorch to load a Gpytorch model?

## Motivation

Since I just want to deploy Gpytorch model using C++ frontend.
But after I followed using Libtorch tutorial to get the trained model from gpytorch,
"traced_script_module = torch.jit.trace(model, example)"
I got 
"TracedModules don't support parameter sharing between modules"

So is there any way I can export gpytorch to C++ ?
Thanks a lot

gpleiss:
At the moment, GPyTorch does not support exporting models to C++ (or at least this has never been tested before). We can add this functionality to a future release, but it is not on our radar right now.

jacobrgardner:
Minor note on what Geoff said: actually I have been working quite a lot on exporting GPyTorch models to torch script. At the moment, exact GPs are fully supported if you use fast pred variances and compute the caches, and variational inference should work now as well.

jacobrgardner:
I'll actually just go ahead and add this to my to-do for 1.0. I'll have an example notebook up for both exact and variational GPs either this week or (more likely due to the holiday) early next.

rangerbottle:
Thanks, that will be great!

jacobrgardner:
Closed via #1026

PR #1026, opened 2020-01-16
Example Notebooks for Converting GP Models to TorchScript

jacobrgardner:
These examples demonstrate how GP models can be traced to ScriptModules and therefore used in production. At test time, the traced models are often up to 4x faster as well.

This should mostly address #1022 since I believe ScriptModules can directly be saved to ONNX format.

Additionally, cc/ @rangerbottle regarding #970, since converting your GP model to TorchScript *should* allow you to run it in LibTorch.


jacobrgardner:
Our example notebook CI is failing because the notebooks don't check if cuda is available (it isn't in CI).

I'll put up a fix for that.

gpleiss:
Can you add `if torch.cuda.is_available()` instead? That's what I did for the other notebooks.

jacobrgardner:
Yep that's the plan, just been in meetings straight since putting this up.

examples/08_Advanced_Usage/TorchScript_Exact_Models.ipynb
```diff
@@ -0,0 +1,190 @@
...
+    "    \n",
+    "    def forward(self, x):\n",
+    "        output_dist = self.gp(x)\n",
+    "        return output_dist.mean, output_dist.variance\n",
+    "\n",
+    "with torch.no_grad(), gpytorch.settings.fast_pred_var():\n",
+    "    model.eval()\n",
+    "    test_x = torch.linspace(0, 1, 51)\n",
+    "    pred = model(test_x)  # Do precomputation\n",
+    "    traced_model = torch.jit.trace(MeanVarModelWrapper(model), test_x, check_trace=False)"
```
JasonXu12: Have you guys tried saving the traced model by "traced_model.save('model_state.pt')“ by any chance? Saving will cause error on my side due to "RuntimeError: Could not export Python function call 'RBFCovariance'. Remove calls to Python functions before export." 
We have the need for saving the .pt model and use it for C++ online inferences. Thank you and look forward to your reply!

jacobrgardner: Sorry for the delay finishing up this PR -- things are just getting a little busy as we approach the ICML deadline.

I'll take a look. What we may have to do is somehow avoid using the custom Functions for RBF and Matern when tracing.

jacobrgardner:
Ended up adding a `gpytorch.settings.trace_mode` that we'll need to activate when tracing GPyTorch models for now until https://github.com/pytorch/pytorch/issues/22329 is resolved.

@JasonXu12 calling `traced_model.save('traced_model.pt')` now works.

jacobrgardner:
@gpleiss this is all set. The `trace_mode` setting is I believe unavoidable. We're not allowed to do essentially anything lazy because we can't trace autograd Functions. This adds some minor code complexity to `VariationalStrategy` (and 2 of our kernels) but I think is worth it to have this functionality.

JasonXu12:
Thank you for the reply@jacobrgardner.  As an external developer, my code relies on gpytorch v1.0.1, which doesn't include the trace_mode class yet and will cause error if I try to use trace_mode in ExactGPModel. Any workarounds or plans for the next release? Thank you!

jacobrgardner:
@JasonXu12 Sure, once this gets merged in we can either release 1.0.2 since this is a particularly nice feature to have, or otherwise you can just install from github.

Issue #1023, opened 2020-01-16
Arc Kernel for Hierarchical Spaces


BCJuan:
# 🚀 Feature Request

## Introduction

I am working on Bayesian optimization with gaussian processes for hyperparameter tuning of ML models. I am indirect user of your library (through [Ax](https://ax.dev/) and [Botorch](https://botorch.org/). Many ML models have  conditional configuration spaces. I understand that gaussian processes do not support directly conditional spaces. However, this article [Raiders of the Lost Architecture:Kernels for Bayesian Optimization in Conditional Parameter Spaces](https://arxiv.org/pdf/1409.4011.pdf) defines a kernel which would be useful in such spaces. 

## Motivation

Would be useful to implement this kernel for use in upper libraries. The usage would allow the optimization of conditional configuration spaces by simply implementing this kernel (if I have understood correctly the article).

In the case where conditionality is related to hierarchy, this would solve issues such as [this one](https://github.com/facebook/Ax/issues/140). Also would be a nice try for a flaw consistently assigned to GPs.

## Pitch

In a rough and simplistic way, the kernel consists in a cylindrical embedding in a standard RBF kernel. And then a product of the kernels for each dimension to build the final kernel. I leave the images of the descriptions found in the [Github](https://github.com/duvenaud/hierarchical-kernels/blob/master/latex/talk.pdf) of the project.

![arc](https://user-images.githubusercontent.com/31255521/72534574-57970480-3877-11ea-9714-90da7fe43ef3.png)

![arc1](https://user-images.githubusercontent.com/31255521/72534807-ba889b80-3877-11ea-87e7-a0d417a28ba5.png)

![arc2](https://user-images.githubusercontent.com/31255521/72534841-ce340200-3877-11ea-9eda-e2b470687a18.png)

I have made myself an implementation:

```
class ArcKernel(Kernel):

    def __init__(self, num_parameters, **kwargs):
        super(ArcKernel, self).__init__(has_lengthscale=True, **kwargs)

        self.register_parameter(
            name="raw_angle",
            parameter=torch.nn.Parameter(torch.randn(1, 1, self.ard_num_dims))
            )
        angle_constraint = Interval(0, 1)
        self.register_constraint("raw_angle", angle_constraint)

        self.register_parameter(
            name="raw_radius",
            parameter=torch.nn.Parameter(torch.randn(1, 1, self.ard_num_dims))
            )
        radius_constraint = Positive()
[...]

PR #1027, opened 2020-01-17
Kernel class for arc kernel

BCJuan:
Hi,

I have finally cleaned my implementation. Now I am able to make the PR.

I have reproduced the example of (Exact GPs)[https://gpytorch.readthedocs.io/en/latest/examples/01_Exact_GPs/Simple_GP_Regression.html] with the kernel and seems to work fine. I can upload the notebooks, or something similar; as you wish.

I do only worry about the kernel size definition. Now is simply a vector of the number of dimensions, but maybe would have to be something like

```
        self.register_parameter(
            name="raw_angle",
            parameter=torch.nn.Parameter(torch.zeros(*self.batch_shape, 1, self.ard_num_dims))
            )
```

I have tried to implement it in the tutorial [Botorch with Ax],(https://botorch.org/tutorials/custom_botorch_model_in_ax) but there are numerical  stability problems.

I will post this info also in the issue [#1023](https://github.com/cornellius-gp/gpytorch/issues/1023)

Hope I have done this process in an appropriate manner. 

Thanks!


gpytorch/kernels/arc_kernel.py
```diff
@@ -0,0 +1,181 @@
...
+        :attr:`radius` (Tensor):
+            The radius parameter. Size = `*batch_shape  x 1`.
+        :attr:`angle` (Tensor):
+            The period angle parameter. Size = `*batch_shape  x 1`.
+
+    Example:
+        >>> x = torch.randn(10, 5)
+        >>> # Non-batch: Simple option
+        ... base_kernel = gpytorch.kernels.MaternKernel(nu=2.5)
+        >>> base_kernel.raw_lengthscale.requires_grad_(False)
```
jacobrgardner: It occurs to me just now that we could also turn off the base kernel's lengthscale in the constructor for ArcKernel. Something like: check base_kernel.has_lengthscale, if True set it to 1 and set requires grad to false as here.

What do you think?

gpytorch/kernels/arc_kernel.py
```diff
@@ -0,0 +1,181 @@
...
+                              angle_prior=gpytorch.priors.GammaPrior(0.5,1),
+                              radius_prior=gpytorch.priors.GammaPrior(3,2),
+                              ard_num_dims=x.shape[-1])),
+                num_dims=x.shape[-1])
+        >>> covar = covar_module(x
+        >>> print(covar.shape)
+    """
+
+    def __init__(self,
+                 base_kernel=MaternKernel(nu=2.5),
```
jacobrgardner: I think the doc building won't like having an actual object be the default argument here, so we'll probably need this to be a required argument

gpytorch/kernels/arc_kernel.py
BCJuan: I do agree, let me change the code and the docs

gpytorch/kernels/arc_kernel.py
BCJuan: Ok, I delete it then and leave it as required

gpytorch/kernels/arc_kernel.py
BCJuan: An error occurs:

```
RuntimeError: you can only change requires_grad flags of leaf variables. If you want to use a computed variable in a subgraph that doesn't require differentiation use var_no_grad = var.detach()
```

BCJuan: Added it in the initialization of the kernel

gpytorch/kernels/arc_kernel.py
```diff
@@ -0,0 +1,185 @@
...
+    def embedding(self, x):
+        x_ = x.div(self.lengthscale)
+        x_s = self.radius * torch.sin(pi * self.angle * x_)
+        x_c = self.radius * torch.cos(pi * self.angle * x_)
+        x_ = torch.cat((x_s, x_c), dim=-1).squeeze(0)
+        return x_
+
+    def forward(self, x1, x2, diag=False, **params):
+        x1_, x2_ = self.embedding(x1), self.embedding(x2)
+        return self.base_kernel(x1_, x2_)
```
jacobrgardner: You might need to pass diag onwards in the base kernel. Something like:
```suggestion
        return self.base_kernel(x1_, x2_, diag=diag)
```
You also may need to just verify batch support. If you'd like, you can extend our base kernel test case here:
https://github.com/cornellius-gp/gpytorch/blob/25e9f99f737eb8967580eda6e8c338b2311e13a6/gpytorch/test/base_kernel_test_case.py#L8

If you fill in the two abstract methods, for example like here: https://github.com/cornellius-gp/gpytorch/blob/25e9f99f737eb8967580eda6e8c338b2311e13a6/test/kernels/test_rbf_kernel.py#L12-L17
then the base kernel test case tests generally all the settings your kernel might get called with across all sorts of GPyTorch models. If you pass all of those tests, then you would know the kernel is in pretty good shape.

BCJuan: I am going to try to build the test cases. Asa I can I will upload them. Thank you very much :)

BCJuan: Hi, the kernel now passes the tests. There were problems with adding `self.ard_num_dims` as last dimension for the tensor registerer of radius and angle. I have put a conditional to solve this, but in theory the arc kernel should be used to always with `ard_num_dims`; at least is what makes sense to me. By the way, I have also included the test, but then the checks fail because it cannot find the arc kernel. Should I add also the test or should I add only the kernel?

jacobrgardner: Hi @BCJuan,

You need to add something like `from .arc_kernel import ArcKernel` to `gpytorch/kernels/__init__.py`. This will enable `from gpytorch.kernels import ArcKernel` to work, and will also make your test pass.

BCJuan: Aps, still wondering how could not see that. My apologies and thank you for the clarification. Fixed and added.

BCJuan: I am thinking that to further check the functionality of the arc kernel I could do something of what is found in `test_rbf_kernel.py` in functions such as `test_ard` or `test_ard_batch`, meaning checking an actual computation with the kernel implementation. As in `rest_rbf_kernel` they would go in the test for the arc kernel.

Balandat:
Should we remove this kind of metadata for consistency w/ the rest of the codebase?
```
# -*- coding: utf-8 -*-
"""
Created on Tue Jan  7 11:13:37 2020
@author: blue
"""
```

BCJuan:
Hi,

I do apologize if this is maybe a noob question but I do not know how to proceed. I am trying to merge `master` into arc_kernel and delete the metadata that @Balandat has indicated but the build fails in travis in the `pytorch=MASTER` section. The error is related to the 'test_multitask_gaussian_likelihood.py` and related files. Is there something that I am missing or doing wrong? Thanks and I do apologize for the mess of resets.

Balandat:
@BCJuan This is likely unrelated to your changes and due to a bug in pytorch (can't see the exact failure, but we ran into this too): https://github.com/pytorch/pytorch/issues/33651. This should be fixed relatively soon.

jacobrgardner:
Those tests should pass now as of #1058

BCJuan:
I do apologize, since I might be wrong, but should I do something? Or should this be merged?

BCJuan:
I have added the proper delta functions that select the dimensions and solved the interval problem for the radius parameter. This kernel should be all usable. The default delta selects all dimensions even if they should not appear in the space configuration.

 Meaning if in a neural network, for example, we have four layers but we are modelling the number of neurons in a total of 6 layers. The parameters for the number of neurons in the five and sixth layers should not be modelled. In the default delta function they are; if the user enters a proper delta they are modelled. The delta is a mask for the input.

From the original paper,
arc

gpleiss:
@BCJuan - I just made sure this will work with our docs. Will merge later today :)

gpleiss:
Thanks so much @BCJuan for the PR! Glad we finally got it in :D

BCJuan:
@gpleiss My pleasure. By the way, if there is anything you want to implement but you do not have time or cannot by any other reason, I would more than gladly like to collaborate. I have certain free time which I can devote to this tasks. Of course, depending on the task I might be able or not to do it, or take me more time than expected.

PR #1028, opened 2020-01-20
[Examples] Fix Simple GP function noise

sdsingh:
The function stated in the Simple GP example (https://gpytorch.readthedocs.io/en/latest/examples/01_Exact_GPs/Simple_GP_Regression.html#Introduction)
 has noise from a Gaussian with variance 0.2. 

However, the following code block samples from a Gaussian with _standard deviation_ 0.2. This makes the code snippet/sample consistent with the definition stated in the introduction.

### DIFF
--- a/examples/01_Exact_GPs/Simple_GP_Regression.ipynb
+++ b/examples/01_Exact_GPs/Simple_GP_Regression.ipynb
@@ -12,7 +12,7 @@
     "\n",
     "\\begin{align}\n",
     "y &= \\sin(2\\pi x) + \\epsilon \\\\\n",
-    "  \\epsilon &\\sim \\mathcal{N}(0, 0.2) \n",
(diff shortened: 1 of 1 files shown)


Balandat:
Did you run the full nb though? seems like none of the outputs were updated. Might be easier to just update the text to use standard dev.

jacobrgardner:
Looking at it in the notebook, I think it might be better to just update the text to say standard deviation as Max said. The function just looks pretty noisy with twice-and-a-bit as much noise.

sdsingh:
Sounds good to me!

sdsingh:
I squashed/force-pushed since there's not much of an interesting commit history here, hopefully that's ok! Updated the text + the code sample to make the variance/standard deviation a bit clearer.

PR #1042, opened 2020-02-07
Fix double derivative on x1 eq x2

wecacuee:
The following test case fails without the fixes

```python

def gpytorch_distance_func(x1, x2):
    return Distance().to(dtype=torch.float64)._sq_dist(x1, x2, False, False)

def test_hessian_sq_distance(dist_fun=gpytorch_distance_func):
    """
    d(x₁, x₂)     = |x₁ - x₂|²


    ∂d(x₁, x₂)
    —————————     = 2(x₁ - x₂)
    ∂x₁

    ∂²d(x₁, x₂)
    ——————————–   = -2 I
    ∂x₂ ∂x₁
    """
    #dist_fun = torch.cdist
    with gpytorch.settings.lazily_evaluate_kernels(False):
        x = torch.rand(1, 2, dtype=torch.float64)
        xp = x.clone().detach()
        x.requires_grad_(True)
        xp.requires_grad_(True)

        grad_dist_ne = lambda xq: torch.autograd.grad(
            dist_fun(x, xq), x, create_graph=True)[0][0, 0]
        torch.autograd.gradcheck(grad_dist_ne, xp)
```


gpleiss:
Hi @wecacuee - sorry for the delay (we were swamped with conference deadlines!)

Thanks for the detailed bug report and the PR! It looks like the failure is a bit spurious (a weird test case failure). I just pushed a slightly relaxed bound. Hopefully that fixes things!

PR #1047, opened 2020-02-13
Use non-batched inducing point covariance if possible

Balandat:
This fixes an issue with the caching of the cholesky factor that can cause nasty shape incompatibilty errors when batch-evaluating a model with different batch shapes.

I am not sure whether this may have any adverse side effects..


Balandat:
An alternative solution here would be to check shapes explicitly after the fact:

```
L = self._cholesky_factor(induc_induc_covar)
# If we modify the batch shape when peforming batch evaluation on a non-batched model, the caching can cause
# issues due to shape inconsistencies. We can check for that explicitly and clear the cache.
if not L.shape == induc_induc_covar.shape:
    del self._memoize_cache["cholesky_factor"]
    L = self._cholesky_factor(induc_induc_covar)
```

Balandat:
@jacobrgardner I recall we had some discussion about this somewhere, same issue related to the inducing points covariance cholesky factor caching messing with batch evaluation, but I can't find that right now.

jacobrgardner:
@Balandat This might address https://github.com/pytorch/botorch/issues/347?

gpytorch/variational/variational_strategy.py
```diff
@@ -93,6 +93,11 @@ def forward(self, x, inducing_points, inducing_values, variational_inducing_cova
         induc_data_covar = full_covar[..., :num_induc, num_induc:].evaluate()
         data_data_covar = full_covar[..., num_induc:, num_induc:]
 
+        # If inducing points are not batched, no need to work with batched covariance
+        # Not doing so can cause nasty shape incompatibilies when evaluating with different batch shapes
+        if self.inducing_points.ndim <= 2 and len(induc_induc_covar.size()) > 2:
+            induc_induc_covar = induc_induc_covar[0]
```
jacobrgardner: Better might be something like:
```suggestion
            batch_ind = [0] * (len(induc_induc_covar.size() - 2)
            induc_induc_covar = induc_induc_covar[*batch_ind]
```
for compatability with multibatch

Balandat: Ok this is more painful than that, especially when having to deal with the issue that the inducing point covar is cashed for some batch size, but then is evaluated on another batch size. I went with the pythonic route to catch the RuntimeError and clearing the cache. It's not great, but it works and is much more transparent than any index juggling could be.

Balandat: argh, actually this is also not the solution - say you batch evaluate sth and then you evaluate it in non-batch mode with some other test data, that output will now have batch dimensions since the batch dimensions of the cached cholesky factor broadcast :(

Seems like it is necessary to explicitly check whether the batch shape changes.

Balandat:
@jacobrgardner does the current form of this seem reasonable? It may be doing more work than necessary, but it's much simpler than the alternative of manually keeping track of shapes.

jacobrgardner:
@Balandat I'm generally okay with this as-is. @gpleiss do you see an easier way around this? One point is that in general use cases this will rarely come up I think.

gpleiss:
Yeah I would agree that this is the best solution. Honestly, having to re-cache is the least of our perf issues anyways :p

Issue #1064, opened 2020-02-28
BatchRepeatLazyTensor cannot evaluate when matrix is not square [Bug]


wjmaddox:
# 🐛 Bug

BatchRepeatLazyTensor is trying to expand the `rhs` in the `_matmul` method to look like the expected output when the `rhs` is not the same size as the expected output. 
From my limited tests, using torch batching standards is probably fine and the [batching expansions](https://github.com/cornellius-gp/gpytorch/blob/ae42649da8199b97a4bc2869708b6a97fc99dc70/gpytorch/lazy/batch_repeat_lazy_tensor.py#L98) (line 98-102) are unecessary as LazyTensors ought to be able to handle batched matmuls. 

## To reproduce

** Code snippet to reproduce **
```python
x = torch.randn(256, 128)
x_lazy = gpytorch.lazify(x)
batch_repeated_x = gpytorch.lazy.BatchRepeatLazyTensor(x_lazy, torch.Size((10,)))
batch_repeated_x.evaluate()
```

** Stack trace/error message **
```
RuntimeError: The expanded size of the tensor (256) must match the existing size (128) at non-singleton dimension 1.  Target sizes: [10, 256, 128].  Tensor sizes: [10, 128, 128]
```

## Expected Behavior

It should evaluate. In particular, a 10 x 256 x 128 matrix should be returned.

## System information

**Please complete the following information:**
- gpytorch.1.0.1
- torch 1.3.1

## Additional context
I believe that a potential fix is to have the `_matmul` method be simply `self.base_lazy_tensor._matmul(rhs)`.

gpleiss:
Should be closed by #1068

PR #1068, opened 2020-03-03
Add support to BatchRepeatLazyTensorfor nonsquare Matrices

wjmaddox:
Provides a fix to https://github.com/cornellius-gp/gpytorch/issues/1064 by adding in flags to the `_matmul` and `_quad_form_derivative` methods of `BatchRepeatLazyTensor` when the repeated lazy tensor is not square (as happens when trying to batch expand prediction strategy covariance matrices).

I also added a unit test for the non-square case in the test file for `BatchRepeatLazyTensor`.


gpleiss:
Thanks @wjmaddox !

PR #1072, opened 2020-03-07
Let scale kernel inherit active_dims of base_kernel

tomaskrehlik:
When applying a scale kernel to a kernel with both active dims and
ard, the scale kernel does not properly implement the forward method
unless active_dims is not supplied also in the scale kernel
definition. (See the test case for example.)

The behaviour is changed so that the base_kernel active_dims are
checked before initializing the scale kernel and including the active
dims into the kwargs if present in the base_kernel.


tomaskrehlik:
Let me know if the fix is reasonable and if there is a better way how you would like to test this and I will fix the linting errors.

tomaskrehlik:
Should be fixed now.

PR #1078, opened 2020-03-14
Make _quad_form_derivative for BlockDiagLazy batch-compatible

Balandat:
If the `left_vec` was passed in with an explicit batch dimension but no explicit trailing dimension things ended up breaking in adding batch dimensions (and broke in much less obvious cases if some implicit broadcasting was happening). See the discussion on https://github.com/pytorch/botorch/issues/396 for details.

### DIFF
--- a/gpytorch/lazy/block_lazy_tensor.py
+++ b/gpytorch/lazy/block_lazy_tensor.py
@@ -74,9 +74,12 @@ class BlockLazyTensor(LazyTensor):
         return res
 
     def _quad_form_derivative(self, left_vecs, right_vecs):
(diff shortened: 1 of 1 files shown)


Balandat:
@jacobrgardner, @gpleiss I am not sure if this is the safest way to deal with this. Seems like there were some implicit assumptions being made here that could result in weird undetected behavior...

Also, re the docstring in https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/lazy/lazy_tensor.py#L318-L331, should this read `Given u' (left_vecs) and v (right_vecs)` instead of `Given u (left_vecs) and v (right_vecs)`?

PR #1086, opened 2020-03-27
Add custom cholesky method to KroneckerProductLazyTensor

Balandat:
`chol(A \kron B) = chol(A) \kron chol(B)`

I don't think this is actually used very often (typically there is a `+ sigma^2 I` in there), but if we need it this should be way faster (`O(max(n,m)^3)` vs. `O((nm)^3)`). 

More generally, we can probably speed up other things if we want. e.g. `logdet(A \kron B) = logdet(A) logdet(B)`...

### DIFF
--- a/gpytorch/test/lazy_tensor_test_case.py
+++ b/gpytorch/test/lazy_tensor_test_case.py
@@ -356,6 +356,14 @@ class LazyTensorTestCase(RectangularLazyTensorTestCase):
                 actual[..., i, i] = actual[..., i, i] + other_diag[..., i]
             self.assertAllClose(res, actual, rtol=1e-2, atol=1e-5)
 
+    def test_cholesky(self):
+        lazy_tensor = self.create_lazy_tensor()
+        evaluated = self.evaluate_lazy_tensor(lazy_tensor)
+
+        res = lazy_tensor.cholesky().evaluate()
+        actual = torch.cholesky(evaluated)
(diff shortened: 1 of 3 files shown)


gpytorch/test/lazy_tensor_test_case.py
```diff
@@ -356,6 +356,14 @@ def test_add_diag(self):
                 actual[..., i, i] = actual[..., i, i] + other_diag[..., i]
             self.assertAllClose(res, actual, rtol=1e-2, atol=1e-5)
 
+    def test_cholesky(self):
```
Balandat: added this in b/c it/s never explicitly tested for most lazies (it is indirectly in `inv_quad_log_det` but this makes it explicit).

Balandat:
Hmm, interesting. Should we have something like a `LowerTriangularLazyTensor` then?

jacobrgardner:
Yeah, I think we (1) introduce `LowerTriangularLazyTensor`, and (2) special case Kronecker products of these (because kronecker products of lower triangular matrices are lower triangular) to not use Krylov subspace methods for solves or logdets, both for correctness and honestly efficiency in the small matrix case.

Balandat:
Sounds good. I can probably come up with a basic prototype for this some time later this week.

Balandat:
RFC here:  #1102

Balandat:
Closing this in favor of #1102 (included this there)

PR #1089, opened 2020-03-29
Fix batch evaluation of index kernel

Balandat:
The recent change ##1087 fixed one thing but broke batch evaluation of non-batched models. This fixes it by broadcasting the different batch shapes involved.


Balandat:
not sure why travis isn't running ...

Balandat:
fwiw travis did run successfully: https://travis-ci.org/github/cornellius-gp/gpytorch/builds/668466531?utm_medium=notification&utm_source=github_status

Is there a way to override the check requirement?

gpleiss:
I went nuclear with my admin privileges 😬

PR #1102, opened 2020-04-04
Add TriangularLazyTensor

Balandat:
Adds a new `TriangularLazyTensor` abstraction. This tensor can be upper or lower (default) triangular. This simplifies a bunch of stuff with solves, dets, logprobs etc.

### Some of the changes with larger blast radius in this PR are:
1. `CholLazyTensor` now takes in a `TriangularLazyTensor`
2. The `_cholesky` method is expected to return a `TriangularLazyTensor`
3. The `_cholesky` method now takes an `upper` kwarg (allows to work with both lower and upper variants of `TriangularLazyTensor`)
4. `DiagLazyTensor` is not subclassed from `TriangularLazyTensor`
5. The memoization functionality is updated to allow caching results depending on args/kwargs (required for dealing with the upper/lower kwargs). By setting `ignore_args=False` in the `@cached` decorator, the existing behavior can be replicated.

### Some improvements:
1. `CholLazyTensor` now has a more efficient `inv_matmul` and `inv_quad` methods using the factorization of the matrix.
2. `KroneckerProductLazyTensor` now returns a Cholesky decomposition that itself uses a Kronecker product representation [previously suggested in #1086] 
3. Added a `test_cholesky` test to the `LazyTensorTestCase` (this covers some previously uncovered cases explicitly)
4. There were a number of hard-to-spot issues due to hacky manual cache handling - I replaced all these call sites with the cache helpers from `gpytorch.utils.memoize`, which is the correct way to go about this.

### DIFF
--- a/gpytorch/functions/_inv_matmul.py
+++ b/gpytorch/functions/_inv_matmul.py
@@ -12,7 +12,7 @@ def _solve(lazy_tsr, rhs):
     if isinstance(lazy_tsr, (CholLazyTensor, TriangularLazyTensor)):
         return lazy_tsr.inv_matmul(rhs)
     if settings.fast_computations.solves.off() or lazy_tsr.size(-1) <= settings.max_cholesky_size.value():
-        return lazy_tsr._cholesky()._cholesky_solve(rhs)
+        return lazy_tsr.cholesky()._cholesky_solve(rhs)
     else:
         with torch.no_grad():
             preconditioner = lazy_tsr.detach()._inv_matmul_preconditioner()
--- a/gpytorch/functions/_inv_matmul.py
+++ b/gpytorch/functions/_inv_matmul.py
@@ -7,6 +7,10 @@ from .. import settings
 
 
 def _solve(lazy_tsr, rhs):
+    from ..lazy import CholLazyTensor, TriangularLazyTensor
+
+    if isinstance(lazy_tsr, (CholLazyTensor, TriangularLazyTensor)):
+        return lazy_tsr.inv_matmul(rhs)
     if settings.fast_computations.solves.off() or lazy_tsr.size(-1) <= settings.max_cholesky_size.value():
         return lazy_tsr._cholesky()._cholesky_solve(rhs)
     else:
--- a/gpytorch/lazy/batch_repeat_lazy_tensor.py
+++ b/gpytorch/lazy/batch_repeat_lazy_tensor.py
@@ -35,18 +35,18 @@ class BatchRepeatLazyTensor(LazyTensor):
         self.batch_repeat = batch_repeat
 
     @cached(name="cholesky")
-    def _cholesky(self):
-        res = self.base_lazy_tensor._cholesky()
+    def _cholesky(self, upper=False):
+        res = self.base_lazy_tensor._cholesky(upper=upper)
         res = res.repeat(*self.batch_repeat, 1, 1)
         return res
 
-    def _cholesky_solve(self, rhs):
+    def _cholesky_solve(self, rhs, upper: bool = False):
         output_shape = _matmul_broadcast_shape(self.shape, rhs.shape)
         if rhs.shape != output_shape:
             rhs = rhs.expand(*output_shape)
 
         rhs = self._move_repeat_batches_to_columns(rhs, output_shape)
-        res = self.base_lazy_tensor._cholesky_solve(rhs)
+        res = self.base_lazy_tensor._cholesky_solve(rhs, upper=upper)
         res = self._move_repeat_batches_back(res, output_shape)
         return res
 
--- a/gpytorch/lazy/batch_repeat_lazy_tensor.py
+++ b/gpytorch/lazy/batch_repeat_lazy_tensor.py
@@ -38,7 +38,7 @@ class BatchRepeatLazyTensor(LazyTensor):
     def _cholesky(self, upper=False):
         from .triangular_lazy_tensor import TriangularLazyTensor
 
-        res = self.base_lazy_tensor._cholesky(upper=upper)._tensor
+        res = self.base_lazy_tensor.cholesky(upper=upper)._tensor
         res = res.repeat(*self.batch_repeat, 1, 1)
         return TriangularLazyTensor(res, upper=upper)
 
--- a/gpytorch/lazy/batch_repeat_lazy_tensor.py
+++ b/gpytorch/lazy/batch_repeat_lazy_tensor.py
@@ -38,11 +38,12 @@ class BatchRepeatLazyTensor(LazyTensor):
     def _cholesky(self, upper=False):
         from .triangular_lazy_tensor import TriangularLazyTensor
 
-        res = self.base_lazy_tensor._cholesky(upper=upper)
+        res = self.base_lazy_tensor._cholesky(upper=upper)._tensor
         res = res.repeat(*self.batch_repeat, 1, 1)
         return TriangularLazyTensor(res, upper=upper)
 
     def _cholesky_solve(self, rhs, upper: bool = False):
+        # TODO: Figure out how to deal with this with TriangularLazyTensor if returned by _cholesky
         output_shape = _matmul_broadcast_shape(self.shape, rhs.shape)
         if rhs.shape != output_shape:
             rhs = rhs.expand(*output_shape)
--- a/gpytorch/lazy/batch_repeat_lazy_tensor.py
+++ b/gpytorch/lazy/batch_repeat_lazy_tensor.py
@@ -36,9 +36,11 @@ class BatchRepeatLazyTensor(LazyTensor):
 
     @cached(name="cholesky")
     def _cholesky(self, upper=False):
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
         res = self.base_lazy_tensor._cholesky(upper=upper)
         res = res.repeat(*self.batch_repeat, 1, 1)
-        return res
+        return TriangularLazyTensor(res, upper=upper)
 
     def _cholesky_solve(self, rhs, upper: bool = False):
         output_shape = _matmul_broadcast_shape(self.shape, rhs.shape)
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -37,7 +37,7 @@ class BlockDiagLazyTensor(BlockLazyTensor):
     def _cholesky(self, upper=False):
         from .triangular_lazy_tensor import TriangularLazyTensor
 
-        chol = self.__class__(self.base_lazy_tensor._cholesky(upper=upper))
+        chol = self.__class__(self.base_lazy_tensor.cholesky(upper=upper))
         return TriangularLazyTensor(chol, upper=upper)
 
     def _cholesky_solve(self, rhs, upper: bool = False):
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -35,7 +35,9 @@ class BlockDiagLazyTensor(BlockLazyTensor):
 
     @cached(name="cholesky")
     def _cholesky(self, upper=False):
-        return self.__class__(self.base_lazy_tensor._cholesky(upper=upper))
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
+        return TriangularLazyTensor(self.__class__(self.base_lazy_tensor._cholesky(upper=upper)), upper=upper)
 
     def _cholesky_solve(self, rhs, upper: bool = False):
         rhs = self._add_batch_dim(rhs)
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -34,12 +34,12 @@ class BlockDiagLazyTensor(BlockLazyTensor):
         return other
 
     @cached(name="cholesky")
-    def _cholesky(self):
-        return self.__class__(self.base_lazy_tensor._cholesky())
+    def _cholesky(self, upper=False):
+        return self.__class__(self.base_lazy_tensor._cholesky(upper=upper))
 
-    def _cholesky_solve(self, rhs):
+    def _cholesky_solve(self, rhs, upper: bool = False):
         rhs = self._add_batch_dim(rhs)
-        res = self.base_lazy_tensor._cholesky_solve(rhs)
+        res = self.base_lazy_tensor._cholesky_solve(rhs, upper=upper)
         res = self._remove_batch_dim(res)
         return res
 
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -37,7 +37,8 @@ class BlockDiagLazyTensor(BlockLazyTensor):
     def _cholesky(self, upper=False):
         from .triangular_lazy_tensor import TriangularLazyTensor
 
-        return TriangularLazyTensor(self.__class__(self.base_lazy_tensor._cholesky(upper=upper)), upper=upper)
+        chol = self.__class__(self.base_lazy_tensor._cholesky(upper=upper))
+        return TriangularLazyTensor(chol, upper=upper)
 
     def _cholesky_solve(self, rhs, upper: bool = False):
         rhs = self._add_batch_dim(rhs)
--- a/gpytorch/lazy/cached_cg_lazy_tensor.py
+++ b/gpytorch/lazy/cached_cg_lazy_tensor.py
@@ -135,6 +135,8 @@ class CachedCGLazyTensor(LazyTensor):
         self.base_lazy_tensor.requires_grad = val
 
     def _cholesky(self, upper=False):
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
         res = self.__class__(
             self.base_lazy_tensor._cholesky(upper=upper),
             eager_rhss=self.eager_rhss,
@@ -144,7 +146,7 @@ class CachedCGLazyTensor(LazyTensor):
             probe_vector_solves=self.probe_vector_solves,
             probe_vector_tmats=self.probe_vector_tmats,
         )
-        return res
+        return TriangularLazyTensor(res, upper=upper)
 
     def _cholesky_solve(self, rhs, upper: bool = False):
         # Here we check to see what solves we've already performed
--- a/gpytorch/lazy/cached_cg_lazy_tensor.py
+++ b/gpytorch/lazy/cached_cg_lazy_tensor.py
@@ -8,6 +8,7 @@ from .. import settings
 from ..utils.warnings import ExtraComputationWarning
 from .chol_lazy_tensor import CholLazyTensor
 from .lazy_tensor import LazyTensor
+from .triangular_lazy_tensor import TriangularLazyTensor
 
 
 class CachedCGLazyTensor(LazyTensor):
@@ -109,7 +110,7 @@ class CachedCGLazyTensor(LazyTensor):
         probe_vector_solves=torch.tensor([]),
         probe_vector_tmats=torch.tensor([]),
     ):
-        super(CachedCGLazyTensor, self).__init__(
+        super().__init__(
             base_lazy_tensor,
             eager_rhss=eager_rhss,
             solves=solves,
@@ -160,7 +161,7 @@ class CachedCGLazyTensor(LazyTensor):
                 "LazyTensor should pre-register all vectors to run CG against.".format(rhs.shape),
                 ExtraComputationWarning,
             )
-        return super()._cholesky_solve(rhs, upper=upper)
+        return torch.cholesky_solve(rhs, self.evaluate(), upper=upper)
 
     def _expand_batch(self, batch_shape):
         return self.base_lazy_tensor._expand_batch(batch_shape)
@@ -192,7 +193,7 @@ class CachedCGLazyTensor(LazyTensor):
                         "CachedCGLazyTensor did not recognize the supplied probe vectors for tridiagonalization.",
                         ExtraComputationWarning,
                     )
-                return super(CachedCGLazyTensor, self)._solve(rhs, preconditioner, num_tridiag=num_tridiag)
+                return super()._solve(rhs, preconditioner, num_tridiag=num_tridiag)
 
         # Here we check to see what solves we've already performed
         truncated_rhs = rhs[..., (num_tridiag or 0) :]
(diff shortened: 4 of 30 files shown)


gpytorch/lazy/triangular_lazy_tensor.py
```diff
@@ -0,0 +1,101 @@
...
+    def exp(self):
+        return self.__class__(self._tensor.exp(), upper=self._upper)
+
+    def inverse(self):
+        eye = torch.eye(self._tensor.shape[-2:], device=self._tensor.device, dtype=self._tensor.dtype)
+        inv = self.inv_matmul(eye)
+        return self.__class__(inv, upper=self._upper)
+
+    def inv_matmul(self, right_tensor, left_tensor=None):
+        tsr = self.evaluate()
```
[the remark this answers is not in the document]
gpleiss: Yeah - i think at some point i added a `kron` method to LazyTensors. We can have that method return something different for triangular lazy tensors.

gpytorch/lazy/triangular_lazy_tensor.py
```diff
@@ -0,0 +1,101 @@
...
+        self._tensor = tensor
+
+    def __add__(self, other):
+        if isinstance(other, TriangularLazyTensor) and not self._upper ^ other.upper:
+            return self.__class__(self._tensor + other._tensor, upper=self._upper)
+        return self._tensor + other
+
+    @cached(name="cholesky")
+    def _cholesky(self):
+        return self.__class__(self._tensor._cholesky(), upper=self._upper)
```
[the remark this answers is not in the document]
gpleiss: In general - we could maybe add a subclass for tensors that we do not expect to be PSD square matrices.

gpleiss:
I like this idea! Can the PR also make CholeskyLazyTensor do the subclassing?

gpytorch/lazy/triangular_lazy_tensor.py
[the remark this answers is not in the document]
Balandat: I guess what I really should do is have `DiagLazyTensor` subclass `TriangularLazyTensor`, then the cholesky issue is handled correctly.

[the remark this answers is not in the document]
Balandat: re the non-PSD square matrices: This would make sense but also create a bunch of complications. Like how to handle things like `SumLazyTensor` - would we mark it PSD or non-PSD depending on whether all constituent tensors are psd (sufficient)?

gpytorch/lazy/triangular_lazy_tensor.py
```diff
@@ -0,0 +1,101 @@
...
+        tsr = self.evaluate()
+        res = torch.triangular_solve(right_tensor, tsr, upper=self._upper).solution
+        if left_tensor is not None:
+            res = left_tensor @ res
+        return res
+
+    def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True):
+        # triangular, inv_matmul is cheap
+        inv_quad_term = inv_quad_rhs.transpose(-1, -2) @ self.inv_matmul(inv_quad_rhs)
+        logdet_term = self.diag().prod(-1) if logdet else None
```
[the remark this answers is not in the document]
Balandat: > More serious addendum: if we want TLT to represent arbitrary triangular matrices and not just Cholesky roots, we'll also need to compute the sign of the determinant ahead of time, then do self.diag().abs().log().sum(-1) since we can't feed negative numbers in to the log.

So I guess the standing assumption throughout the lazies is psd. So if the det is negative we can return `NaN`, but that will likely mess with a lot of our assumptions. Maybe we should for now assume that things are psd and revisit the bigger picture of non-PSD lazies separately from this?

Balandat:
> Can the PR also make CholeskyLazyTensor do the subclassing? 

So there is some interpretation ambiguity here: `CholLazyTensor` is a `RootLazyTensor`, and thus actually doesn't represent the Choleksy decomposition of a lazy, it represents the lazy itself by means of its Cholesky decomposition. What we can do is have `CholLazyTensor` require a `TriangularLazyTensor` as its factor. But we can't really subclass `CholLazyTensor` from `TriangularLazyTensor` since the semantics are different.  Does that make sense?

gpleiss:
@Balandat yeah that makes sense, my bad! I like the idea of having CholLazyTensor requiring a TriangularLazyTensor. Also the `_cholesky` method of LazyTensors can also have a TriangularLazyTensor return type.

Balandat:
OK made a bunch of changes. A lot of tests fail, some of these are actually b/c some of the changes result in the cholesky gradient not being symmetric anymore, will have to fix that. 

Before I try to fix all these test failures, how does this look more generally?

gpytorch/lazy/batch_repeat_lazy_tensor.py
```diff
@@ -35,18 +35,20 @@ def __init__(self, base_lazy_tensor, batch_repeat=torch.Size((1,))):
...
     @cached(name="cholesky")
-    def _cholesky(self):
-        res = self.base_lazy_tensor._cholesky()
+    def _cholesky(self, upper=False):
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
+        res = self.base_lazy_tensor._cholesky(upper=upper)
         res = res.repeat(*self.batch_repeat, 1, 1)
-        return res
+        return TriangularLazyTensor(res, upper=upper)
```
jacobrgardner, line 43: Do we want this? This is creating probably a TriangularLazyTensor(BatchRepeatLT(TriangularLT))

gpytorch/lazy/block_diag_lazy_tensor.py
```diff
@@ -34,12 +34,14 @@ def _add_batch_dim(self, other):
         return other
 
     @cached(name="cholesky")
-    def _cholesky(self):
-        return self.__class__(self.base_lazy_tensor._cholesky())
+    def _cholesky(self, upper=False):
+        from .triangular_lazy_tensor import TriangularLazyTensor
 
-    def _cholesky_solve(self, rhs):
+        return TriangularLazyTensor(self.__class__(self.base_lazy_tensor._cholesky(upper=upper)), upper=upper)
```
jacobrgardner, line 40: Same issue here

gpytorch/lazy/cached_cg_lazy_tensor.py
```diff
@@ -134,19 +134,21 @@ def requires_grad(self):
...
+            self.base_lazy_tensor._cholesky(upper=upper),
             eager_rhss=self.eager_rhss,
             solves=self.solves,
             probe_vectors=self.probe_vectors,
             probe_vector_norms=self.probe_vector_norms,
             probe_vector_solves=self.probe_vector_solves,
             probe_vector_tmats=self.probe_vector_tmats,
         )
-        return res
+        return TriangularLazyTensor(res, upper=upper)
```
jacobrgardner, line 149: And here

gpytorch/lazy/block_diag_lazy_tensor.py
gpleiss: I think that's okay - in some sense that's what we want (an block-triangular matrix of triangular matrices)

gpytorch/utils/errors.py
```diff
@@ -5,4 +5,8 @@ class NanError(RuntimeError):
     pass
 
 
-__all__ = ["NanError"]
+class NotPSDError(RuntimeError):
```
gpleiss, line 8: Nit: can we also use this error for when `psd_safe_cholesky` fails?

gpytorch/lazy/batch_repeat_lazy_tensor.py
gpleiss:
```suggestion
        return TriangularLazyTensor(res._tensor, upper=upper)
```

gpytorch/lazy/cached_cg_lazy_tensor.py
```diff
@@ -134,19 +134,21 @@ def requires_grad(self):
     def requires_grad(self, val):
         self.base_lazy_tensor.requires_grad = val
 
-    def _cholesky(self):
+    def _cholesky(self, upper=False):
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
         res = self.__class__(
-            self.base_lazy_tensor._cholesky(),
+            self.base_lazy_tensor._cholesky(upper=upper),
```
gpleiss, line 141:
```suggestion
            self.base_lazy_tensor._cholesky(upper=upper)._tensor,
```

gpytorch/lazy/cached_cg_lazy_tensor.py
gpleiss: I think here we want `TriangularLazyTensor` wrapping the CachedCGLazyTensor, but maybe could remove the innr TriangularLazyTensor.

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -8,6 +8,7 @@
 from .non_lazy_tensor import NonLazyTensor
 
 
+# TODO: Subclass DiagLazyTensor from TriangularLazyTensor
```
gpleiss, line 11: I was going to suggest this :) Any reason it can't be sub-classed now?

gpytorch/lazy/kronecker_product_lazy_tensor.py
```diff
@@ -59,6 +60,11 @@ def __init__(self, *lazy_tensors):
         super(KroneckerProductLazyTensor, self).__init__(*lazy_tensors)
         self.lazy_tensors = lazy_tensors
 
+    @cached(name="cholesky")
+    def _cholesky(self, upper=False):
+        chol = KroneckerProductLazyTensor(*[lt._cholesky(upper=upper) for lt in self.lazy_tensors])
+        return TriangularLazyTensor(chol, upper=upper)
```
gpleiss, line 66: 👍

gpytorch/lazy/lazy_tensor.py
```diff
@@ -424,7 +424,7 @@ def _cholesky_solve(self, rhs):
         Returns:
             (LazyTensor) Cholesky factor
         """
-        return torch.cholesky_solve(rhs, self.evaluate())
+        return torch.cholesky_solve(rhs, self.evaluate(), upper=upper)
```
gpleiss, line 427: Now that we have `TriangularLazyTensor`, this method could move to `TriangularLazyTensor` and throw an error for the base LazyTensor.

gpytorch/lazy/diag_lazy_tensor.py
Balandat: time and more test failures :) I'll try to get it done though so it's logically more consistent

gpytorch/lazy/batch_repeat_lazy_tensor.py
Balandat: I guess I could just call the repeat on the `_tensor` directly. Amounts to the same thing, but might be less convoluted.

gpytorch/lazy/lazy_tensor.py
Balandat: makes sense. What I will do is raise a `NotImplementedError` here. Then in `TriangularLazyTensor._cholesky_solve` I will try calling the `_cholesky_solve` method of `_tensor` and catch the `NotImplementedError` and use the `TriangularLazyTensor` method. That way we can still use the custom implementations of all the other lazy tensors.

gpytorch/lazy/cached_cg_lazy_tensor.py
Balandat: That has been causing issues if this is recursively called with `TriangularLazyTensor`s. What I've done instead is the constructor of `TriangularLazyTensor` strip out the underlying tensor if instantiated with `TriangularLazyTensor`.

Balandat: there may be a more elegant way to do this.

Balandat:
OK so this should address most of the comments. The main question now is how to deal with non-symmetric gradients. The mock-based test failures probably just require some changing of the tests themselves b/c of the changes in computation.

gpytorch/lazy/kronecker_product_lazy_tensor.py
```diff
@@ -112,3 +117,45 @@ def _size(self):
...
+
+    @cached
+    def inverse(self):
+        # here we use that (A \kron B)^-1 = A^-1 \kron B^-1
+        inverses = [lt.inverse() for lt in self.lazy_tensors]
+        return self.__class__(*inverses, upper=self.upper)
+
+    def inv_matmul(self, right_tensor, left_tensor=None):
+        # Since the individual tensors are triangular, their inverses can be computed cheaply
+        res = self.inverse() @ right_tensor
```
mshvartsman, line 158: Could you do something without making the explicit inverse here? From https://github.com/cornellius-gp/gpytorch/issues/725: 
```
        tensor_shapes = [q.shape[0] for q in self.lazy_tensors]
        multipliers = torch.Tensor([torch.prod(torch.Tensor([ts for i, ts in enumerate(tensor_shapes) if i!=j]))
                                for j in range(len(tensor_shapes))])

        y = inv_quad_rhs.clone()
        for i, q in enumerate(self.lazy_tensors):
            n = q.shape[0]
            y = q.inv_matmul(y.contiguous().view(n, -1))
            y = y.view(n, nrows//n, -1).permute(1, 0, 2)
```

gpleiss: Yeah this seems analogous to what we have for `KPLT._matmul`.

Balandat: Added that in, got slightly more complicated to allow arbitrary batch shapes, but it works. I implemented it generically for `KroneckerProductLazyTensor`, but made sure it's the default only for `KroneckerProductTriangularLazyTensor` for now (in which case it's just a loop over batched triangular solves, so really cheap and should probably always be done).

gpytorch/functions/_inv_matmul.py
```diff
@@ -7,6 +7,10 @@
 
 
 def _solve(lazy_tsr, rhs):
+    from ..lazy import CholLazyTensor, TriangularLazyTensor
+
+    if isinstance(lazy_tsr, (CholLazyTensor, TriangularLazyTensor)):
```
Balandat, line 12: For the `TriangularLazyTensor` this is necessary (otherwise downstream we try to compute a Cholesky decomposition of a triangular matrix). For the `CholLazyTensor` this is something that should be done anyway (it in fact is, but that's hidden inside the overwritten `_solve` below). I guess this highlights a pretty fundamental change, namely that with `TriangularLazyTensor` we now have a lazy tensor that is not symmetric.

gpytorch/lazy/triangular_lazy_tensor.py
```diff
@@ -0,0 +1,159 @@
...
+                If True, the tensor is considered to be upper-triangular, otherwise lower-triangular.
+        """
+        if isinstance(tensor, TriangularLazyTensor):
+            # this is a null-op, we can just use underlying tensor directly.
+            tensor = tensor._tensor
+        elif isinstance(tensor, BatchRepeatLazyTensor):
+            # things get kind of messy when interleaving repeats and triangualrisms
+            if not isinstance(tensor.base_lazy_tensor, TriangularLazyTensor):
+                tensor = tensor.__class__(
+                    TriangularLazyTensor(tensor.base_lazy_tensor, upper=upper), batch_repeat=tensor.batch_repeat,
```
Balandat, line 37: This works but it's not really pretty. If we want to build this out more generally we probably will need to rethink the design and work with Mixin classes to make this less cumbersome.

Balandat:
Alright I think I fixed most of the obvious issues. There are two outstanding things that I need help with:
1. The `quad_form_derivative` in `CholLazyTensor` and `MulLazyTensor` test cases is off  by a factor of 2. I don't think that I dropped some `0.5` factor anywhere, so I think that has to do with the changes to the `CholLazyTensor` somehow, but I'm not entirely sure what's going on (could also be deeper in the `RootLazyTensor` since that is used inside `CholLazyTensor` as well as explicitly in the `MulLazyTensor` test case).
2. The variational tests are all failing b/c of the Cholesky mock of `torch.cholesky` defined [here](https://github.com/cornellius-gp/gpytorch/blob/5ea8483114813655c792321b334a1f8e15788d74/test/variational/test_whitened_variational_strategy.py#L141) being called twice, instead of the single call expected in the tests. I'm not familiar enough with the variational part of the codebase to understand if that's an issue or whether that might even be desirable (e.g. performing two small Cholesky decomps + double triangular substitution vs. a single large Cholesky decomp).

gpytorch/lazy/chol_lazy_tensor.py
```diff
@@ -1,51 +1,71 @@
...
+        return res.evaluate()
+
+    @cached
+    def inverse(self):
+        Linv = self.root.inverse()  # this could be slow in some cases w/ structured lazies
+        return CholLazyTensor(Linv, upper=not self.upper)
 
     def inv_matmul(self, right_tensor, left_tensor=None):
-        with settings.fast_computations(solves=False):
-            return super().inv_matmul(right_tensor, left_tensor=left_tensor)
```
Balandat, line 48: just realizing that this may be doing a bunch of unnecessary work currently if we compute `inv_quad` on a `CholLazyTensor` (at least if it's large enough to use linear CG by default for the solve)

Balandat:
bump :)

(lint errors are due to flake8 update)

gpleiss:
Ooops sorry to leave you hanging @Balandat ! I'll give a review later today (though @jacobrgardner lmk if you can beat me to it)

gpytorch/lazy/triangular_lazy_tensor.py
```diff
@@ -0,0 +1,159 @@
...
+            :attr:`tensor` (Tensor or LazyTensor):
+                A `b1 x ... x bk x n x n` Tensor, representing a `b1 x ... x bk`-sized batch
+                of `n x n` triangular matrices.
+            :attr:`upper` (bool):
+                If True, the tensor is considered to be upper-triangular, otherwise lower-triangular.
+        """
+        if isinstance(tensor, TriangularLazyTensor):
+            # this is a null-op, we can just use underlying tensor directly.
+            tensor = tensor._tensor
+        elif isinstance(tensor, BatchRepeatLazyTensor):
```
jacobrgardner, line 33: This is basically my main concern. If BatchRepeatLT.cholesky returns a TriangularLT but we didn't have this, everything would become num_repeats more expensive than necessary. We need to make sure that the new LT doesn't destroy structure when we return it from Cholesky calls. We had this bug with MulLazyTensor once and the entire package was like 10x slower

Balandat:
So I still could need some help with the two points I raised here: https://github.com/cornellius-gp/gpytorch/pull/1102#issuecomment-614369747, not quite sure what to do with those.

gpytorch/lazy/chol_lazy_tensor.py
gpleiss: I think the `with settings.fast_computations(solves=False)` flag makes it so that it uses cholesky solve rather than CG.

gpleiss:
@Balandat sorry for the delay getting to this.

> The quad_form_derivative in CholLazyTensor and MulLazyTensor test cases is off by a factor of 2.

I think this is some issue with the default `_quad_form_derivative` that `TriangularLazyTensor` was using. I copied over the `_quad_form_derivative` from `NonLazyTensor` and that seems to fix those tests.

> The variational tests are all failing b/c of the Cholesky mock of torch.cholesky defined here being called twice, instead of the single call expected in the tests.

I think the changes you made to `utils.memoize` is preventing the caching from actually occurring. We should be caching the Cholesky factorization used in variational GPs (https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/variational/variational_strategy.py#L69), which is why the test expects Cholesky to only be called once. It looks like the Cholesky function is called both times, which means the `cached` decorator is failing.

Balandat:
> I think the changes you made to utils.memoize is preventing the caching from actually occurring. We should be caching the Cholesky factorization used in variational GPs (https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/variational/variational_strategy.py#L69), which is why the test expects Cholesky to only be called once. It looks like the Cholesky function is called both times, which means the cached decorator is failing.

So turns out the caching works fine, the issue seems to be that the changes I made to cholesky generate additional `LazyTensor` objects so the cache on the previously existing one doesn't get reused. I will need to dig deeper here.

Balandat:
Actually, the fact that caching is working properly is what causes the issues here. What I mean by that is that on master the `@cache` implementation means that the following https://github.com/cornellius-gp/gpytorch/blob/8ab41ccd4131883555c229ff9a44f5a5db786623/gpytorch/variational/variational_strategy.py#L70-L71
will technically not do proper memoization - that is, if `_cholesky_factor` will be called with different instances of `induc_induc_covar`, the `@cached` decorator will ignore any subsequent calls and always return the same cached result. This ends up working in this particular case where the content of `induc_induc_covar` is the same even though the instances are different. But such caching will generally break if you have other arguments (e.g. `upper` for `cholesky`, hence my change to memoize). Let me see if I can make this cleaner here to avoid relying on a "broken" memoization.

jacobrgardner:
If I understand your changes to memoize correctly, it looks like it now tracks args and kwargs as part of the key in addition to merely the name? I agree that seems like a reasonable addition to avoid erroneous caching.

Balandat:
> If I understand your changes to memoize correctly, it looks like it now tracks args and kwargs as part of the key in addition to merely the name? I agree that seems like a reasonable addition to avoid erroneous caching.

Yes, that's what it's doing. I just pushed some updates that adds a `ignore_args` kwarg to the `cached` decorator so that we can easily achieve the current behavior if we want it, and select it explicitly. 

I'm still running into some variational strategy test failures (some were solved by the proper caching above) - presumably it has to do with the generation of new LazyTensor objects so that the cache that was carried along earlier doesn't get carried over. Looking for the root cause now.

Balandat:
Hmm a lot of stuff has changed on the variational side - I guess I should use the `TriangularLazyTensor` abstractions instead of the manual triangular solves.

Balandat:
> If I understand your changes to memoize correctly, it looks like it now tracks args and kwargs as part of the key in addition to merely the name? I agree that seems like a reasonable addition to avoid erroneous caching.

Yeah I actually had to manually hackfix this in the past where this messed up batch evaluation: https://github.com/cornellius-gp/gpytorch/blob/8ab41ccd4131883555c229ff9a44f5a5db786623/gpytorch/variational/variational_strategy.py#L101-L103

Balandat:
OK so i stared at this for a long time now, but I'm kind of stuck. I did run a cprofile of `TestUnwhitenedVariationalGP.test_training_iteration()` to count the calls, and on this PR `LazyTensor._cholesky` gets called twice (once on master), and `CachedCGLazyTensor._cholesky` gets called four times (six times on master). 

Maybe someone who knows the `CachedCGLazyTensor` setup better can take a look to see what's going on...

[tlt_debug_master.cprofile.txt](https://github.com/cornellius-gp/gpytorch/files/4841445/tlt_debug_master.cprofile.txt)
[tlt_debug.cprofile.txt](https://github.com/cornellius-gp/gpytorch/files/4841446/tlt_debug.cprofile.txt)

Balandat:
This should be ready now.

gpytorch/lazy/triangular_lazy_tensor.py
Balandat: I made sure to also add an `_expand_batch` method to TLT so that we don't unnecessarily wrap TLT in BRLTs.

PR #1103, opened 2020-04-05
Remove tuple() null-ops

Balandat:
This is not necessary.


Balandat:
Presumably this is the first time anything is run on sphinx 3.0.0 (which was released yesterday and messed with our docs too). You can either pin sphinx to < 3.0.0 or do the right thing and fix the docs :)

PR #1104, opened 2020-04-05
Remove manual symmetrization of Chloesky gradients

Balandat:
This has long been changed upstream to use a symmetric gradient instead.

```
import torch

mat = torch.randn(4, 4, dtype=torch.float64)
mat = (mat @ mat.transpose(-1, -2)).div_(5).add_(torch.eye(4, dtype=torch.float64))
mat = mat.detach().clone().requires_grad_(True)
mat_clone = mat.detach().clone().requires_grad_(True)

inv_mat1 = mat_clone.inverse()  # A^{-1} = A^{-1}

chol_mat = mat.cholesky()
chol_inv_mat = chol_mat.inverse().transpose(-2, -1)
inv_mat2 = chol_inv_mat @ chol_inv_mat.transpose(-2, -1)  # A^{-1} = L^{-T}L^{-1}

# This was always True
print('Are these both A^{-1}?', bool(torch.norm(inv_mat1 - inv_mat2) < 1e-8))

inv_mat1.trace().backward()
inv_mat2.trace().backward()

# This now True as well 
print('Are the gradients the same?', bool(torch.norm(mat_clone.grad - mat.grad) < 1e-8))
```


jacobrgardner:
Ah good call! This'll probably buy a small bit of performance back

gpytorch/lazy/lazy_tensor.py
```diff
@@ -18,7 +18,6 @@
 from ..utils.cholesky import psd_safe_cholesky
 from ..utils.deprecation import _deprecate_renamed_methods
 from ..utils.getitem import _compute_getitem_size, _convert_indices_to_tensors, _noop_index
-from ..utils.gradients import _ensure_symmetric_grad
```
jacobrgardner, line 21: Can probably also rm the file / remove from __init__ if applicable

Balandat: ah yeah I thought this was used somewhere else, but apparently not

gpleiss:
Okay the tests failing looks like it might be a semi-related issue. Our `inv_matmul` function (where all of the failing tests are coming from) computes `d l^T K^{-1} r / dK` as K^{-1} l r^T K^{-1}`, which makes things non-symmetric. We should probably be doing `1/2 K^{-1} l r^T K^{-1} + 1/2 K^{-1} r l^T K^{-1}`. That should still be more efficient than the manual symmetry from `ensure_symmetric_gradient` though.

@Balandat i can add this fix to the PR tomorrow.

PR #1105, opened 2020-04-06
symeig and svd methods on LazyTensors, allow to choose for root_decomposition

Balandat:
This adds high-level `LazyTensor.symeig` and `LazyTensor.svd` functions. These are implemented via `_symeig` and `_svd` on the respective LT subclasses. 

These are useful when dealing with small-ish singular (or near-singular) covariances (typically posterior covariance at repeated points or points that are very close together. They are also useful for performing explicit Kronecker algebra, as implemented in #909 - one contribution of this PR is therefore to ensure that using these operations in the Kronecker algebra does not break the structure of the LT, and that computations can be performed efficiently of structured tensors are themselves used in structured algebra.

This PAR also adds a `method` argument to `LazyTensor`'s `root_decomposition` that allows to use either `svd`, `symeig`, or `pivoted_cholesky` for performing the root decomposition in addition to the existing methods (`cholesky`, `lanczos`).

### DIFF
--- a/gpytorch/lazy/added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/added_diag_lazy_tensor.py
@@ -10,7 +10,7 @@ from .. import settings
 from ..utils import broadcasting, pivoted_cholesky
 from ..utils.memoize import cached
 from ..utils.warnings import NumericalWarning
-from .diag_lazy_tensor import DiagLazyTensor
+from .diag_lazy_tensor import ConstantDiagLazyTensor, DiagLazyTensor
 from .lazy_tensor import LazyTensor
 from .psd_sum_lazy_tensor import PsdSumLazyTensor
 from .root_lazy_tensor import RootLazyTensor
@@ -132,18 +132,16 @@ class AddedDiagLazyTensor(SumLazyTensor):
 
     @cached(name="svd")
     def _svd(self) -> Tuple["LazyTensor", Tensor, "LazyTensor"]:
-        diag = self._diag_tensor.diag()
-        if torch.equal(diag, diag[..., :1].expand(diag.shape)):
+        if isinstance(self._diag_tensor, ConstantDiagLazyTensor):
             U, S_, V = self._lazy_tensor.svd()
-            S = S_ + diag  # this assumes all diagonal entries are positive
+            S = S_ + self._diag_tensor.diag()  # this assumes all diagonal entries are positive
             return U, S, V
         return super()._svd()
 
     @cached(name="symeig")
     def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
-        diag = self._diag_tensor.diag()
-        if torch.equal(diag, diag[..., :1].expand(diag.shape)) and torch.all(diag[..., :1] >= 0):
+        if isinstance(self._diag_tensor, ConstantDiagLazyTensor):
             evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
-            evals = evals_ + diag  # this assumes all diagonal entries are positive
+            evals = evals_ + self._diag_tensor.diag()  # this assumes all diagonal entries are positive
             return evals, evecs
         return super()._symeig(eigenvectors=eigenvectors)
--- a/gpytorch/lazy/added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/added_diag_lazy_tensor.py
@@ -6,6 +6,7 @@ import torch
 
 from .. import settings
 from ..utils import broadcasting, pivoted_cholesky
+from ..utils.memoize import cached
 from ..utils.warnings import NumericalWarning
 from .diag_lazy_tensor import DiagLazyTensor
 from .psd_sum_lazy_tensor import PsdSumLazyTensor
@@ -125,3 +126,21 @@ class AddedDiagLazyTensor(SumLazyTensor):
         logdet = self._r_cache.diagonal(dim1=-1, dim2=-2).abs().log().sum(-1).mul(2)
         logdet -= (1.0 / self._noise).log().sum([-1, -2])
         self._precond_logdet_cache = logdet.view(*batch_shape) if len(batch_shape) else logdet.squeeze()
+
+    @cached(name="svd")
+    def _svd(self):
+        diag = self._diag_tensor.diag()
+        if torch.equal(diag, diag[..., :1].expand(diag.shape)):
+            U, S_, V = self._lazy_tensor.svd()
+            S = S_ + diag  # this assumes all diagonal entries are positive
+            return U, S, V
+        return super()._svd()
+
+    @cached(name="symeig")
+    def _symeig(self, eigenvectors=False):
+        diag = self._diag_tensor.diag()
+        if torch.equal(diag, diag[..., :1].expand(diag.shape)):
+            evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
+            evals = evals_ + diag  # this assumes all diagonal entries are positive
+            return evals, evecs
+        return super()._symeig(eigenvectors=eigenvectors)
--- a/gpytorch/lazy/added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/added_diag_lazy_tensor.py
@@ -8,7 +8,8 @@ from torch import Tensor
 
 from .. import settings
 from ..utils import broadcasting, pivoted_cholesky
-from ..utils.memoize import cached
+from ..utils.errors import CachingError
+from ..utils.memoize import cached, pop_from_cache
 from ..utils.warnings import NumericalWarning
 from .diag_lazy_tensor import ConstantDiagLazyTensor, DiagLazyTensor
 from .lazy_tensor import LazyTensor
@@ -140,6 +141,10 @@ class AddedDiagLazyTensor(SumLazyTensor):
 
     @cached(name="symeig")
     def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
+        try:
+            return pop_from_cache(self, "symeig", eigenvectors=True)
+        except CachingError:
+            pass
         if isinstance(self._diag_tensor, ConstantDiagLazyTensor):
             evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
             evals = evals_ + self._diag_tensor.diag()  # this assumes all diagonal entries are positive
--- a/gpytorch/lazy/added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/added_diag_lazy_tensor.py
@@ -142,7 +142,7 @@ class AddedDiagLazyTensor(SumLazyTensor):
     @cached(name="symeig")
     def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
         diag = self._diag_tensor.diag()
-        if torch.equal(diag, diag[..., :1].expand(diag.shape)):
+        if torch.equal(diag, diag[..., :1].expand(diag.shape)) and torch.all(diag[..., :1] >= 0):
             evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
             evals = evals_ + diag  # this assumes all diagonal entries are positive
             return evals, evecs
--- a/gpytorch/lazy/added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/added_diag_lazy_tensor.py
@@ -1,14 +1,17 @@
 #!/usr/bin/env python3
 
 import warnings
+from typing import Optional, Tuple
 
 import torch
+from torch import Tensor
 
 from .. import settings
 from ..utils import broadcasting, pivoted_cholesky
 from ..utils.memoize import cached
 from ..utils.warnings import NumericalWarning
 from .diag_lazy_tensor import DiagLazyTensor
+from .lazy_tensor import LazyTensor
 from .psd_sum_lazy_tensor import PsdSumLazyTensor
 from .root_lazy_tensor import RootLazyTensor
 from .sum_lazy_tensor import SumLazyTensor
@@ -128,7 +131,7 @@ class AddedDiagLazyTensor(SumLazyTensor):
         self._precond_logdet_cache = logdet.view(*batch_shape) if len(batch_shape) else logdet.squeeze()
 
     @cached(name="svd")
-    def _svd(self):
+    def _svd(self) -> Tuple["LazyTensor", Tensor, "LazyTensor"]:
         diag = self._diag_tensor.diag()
         if torch.equal(diag, diag[..., :1].expand(diag.shape)):
             U, S_, V = self._lazy_tensor.svd()
@@ -137,7 +140,7 @@ class AddedDiagLazyTensor(SumLazyTensor):
         return super()._svd()
 
     @cached(name="symeig")
-    def _symeig(self, eigenvectors=False):
+    def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
         diag = self._diag_tensor.diag()
         if torch.equal(diag, diag[..., :1].expand(diag.shape)):
             evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
--- a/gpytorch/lazy/added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/added_diag_lazy_tensor.py
@@ -8,8 +8,7 @@ from torch import Tensor
 
 from .. import settings
 from ..utils import broadcasting, pivoted_cholesky
-from ..utils.errors import CachingError
-from ..utils.memoize import cached, pop_from_cache
+from ..utils.memoize import cached
 from ..utils.warnings import NumericalWarning
 from .diag_lazy_tensor import ConstantDiagLazyTensor, DiagLazyTensor
 from .lazy_tensor import LazyTensor
@@ -139,12 +138,7 @@ class AddedDiagLazyTensor(SumLazyTensor):
             return U, S, V
         return super()._svd()
 
-    @cached(name="symeig")
     def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
-        try:
-            return pop_from_cache(self, "symeig", eigenvectors=True)
-        except CachingError:
-            pass
         if isinstance(self._diag_tensor, ConstantDiagLazyTensor):
             evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
             evals = evals_ + self._diag_tensor.diag()  # this assumes all diagonal entries are positive
--- a/gpytorch/lazy/added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/added_diag_lazy_tensor.py
@@ -134,13 +134,13 @@ class AddedDiagLazyTensor(SumLazyTensor):
     def _svd(self) -> Tuple["LazyTensor", Tensor, "LazyTensor"]:
         if isinstance(self._diag_tensor, ConstantDiagLazyTensor):
             U, S_, V = self._lazy_tensor.svd()
-            S = S_ + self._diag_tensor.diag()  # this assumes all diagonal entries are positive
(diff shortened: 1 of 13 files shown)


Balandat:
Added drafts for custom methods for `symeig` and `svd` that exploit matrix structure for some lazies. A lot more can be done, e.g. for Kronecker matrices we (since everything is diagonalizable b/c psd) should be able to express the eigendecomposition  in terms of Kronecker products of the eigenvectors and products of the eigenvalues. We'll also want to make sure that the return type of these methods are `LazyTensor` for the `evec` and `U` and `V` matrices.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1321,21 +1323,42 @@ def root_decomposition(self):
...
-                warnings.warn(
-                    "Runtime Error when computing Cholesky decomposition: {}. Using RootDecomposition.".format(e)
-                )
+                warnings.warn(f"Runtime Error when computing Cholesky decomposition: {e}. Using eigendecomposition.")
+                method = "symeig"
+
+        if method == "pivoted_cholesky":
+            return RootLazyTensor(pivoted_cholesky(self.evaluate(), max_iter=100))
+
+        if method == "symeig":
```
gpleiss, line 1346: We might want a feature flag that chooses to do symeig for small matrices

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1531,6 +1556,36 @@ def sum(self, dim=None):
...
+
+        Returns:
+            :obj:`torch.Tensor`:
+                The left singular vectors (`U`).
+            :obj:`torch.Tensor`:
+                The singular values (`S`).
+            :obj:`torch.Tensor`:
+                The right singular vectors (`V`).
+        """
+        return torch.svd(self.evaluate())
```
Balandat, line 1572: We'll want to return `U` and `V` here as `LazyTensor` objects, as for Kronecker tensors these can be themselves expressed as `KroneckerProductLazyTensor`s. Same for the eigenvectors in `symeig`.

gpytorch/test/lazy_tensor_test_case.py
```diff
@@ -578,3 +578,35 @@ def test_sqrt_inv_matmul_no_lhs(self):
...
+        if idxr.ndim == 1:
+            evecs = evecs[:, idxr]
+        # TODO: Fix rearranging of evecs in the batch setting
+        elif idxr.ndim == 2:
+            batch_idxr = torch.arange(idxr.shape[-2]).unsqueeze(-1).expand(idxr.shape)
+            evecs = evecs[batch_idxr, idxr]
+        elif idxr.ndim == 3:
+            batch_idxr = torch.arange(idxr.shape[-1]).unsqueeze(-2).expand(idxr.shape)
+            multi_batch_idxr = torch.arange(idxr.shape[-2]).unsqueeze(-2).expand(idxr.shape)
+            evecs = evecs[multi_batch_idxr, batch_idxr, idxr]
```
Balandat, line 602: there is a better (and correct) way of doing this, I'm sure

gpytorch/test/lazy_tensor_test_case.py
```diff
@@ -578,3 +578,29 @@ def test_sqrt_inv_matmul_no_lhs(self):
...
+        lazy_tensor_copy = lazy_tensor.clone().detach_().requires_grad_(True)
+        evaluated = self.evaluate_lazy_tensor(lazy_tensor_copy)
+
+        # Perform forward pass
+        evals, evecs = lazy_tensor.symeig(eigenvectors=True)
+        evecs = evecs.evaluate()
+
+        # since LazyTensor.symeig does not sort evals, we do this here for the check
+        evals_sorted, idxr = torch.sort(evals, dim=-1, descending=False)
+        evecs_sorted = torch.gather(evecs, dim=-1, index=idxr.unsqueeze(-2).expand(evecs.shape))
```
Balandat, line 593: this is kind of dense, torch should really have a more convenient batched indexing functionality....

Balandat:
@wjmaddox, @jacobrgardner, I believe this is in pretty good shape at this point.

Testing is a little harder b/c unlike torch our `symeig` and `svd` methods do not sort the evals/singvals (the reason being that this would be messy and could destroy structure, e.g. for KroneckerLTs), and so we need to do some sorting. I'm also skipping explicit comparisons of evecs/singvecs for repeated evals/singvals, since the sorting here would get pretty messy. I do verify though that we get a correct eigendecomposition / SVD for all tensors.

There are a few tests that are still failing, I haven not have had time to dig very deep into those. It seems that these might just be due to numerical issues (I might well be sth else though). Apart from these test failures, any comments on this PR?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1335,22 +1338,45 @@ def root_decomposition(self):
...
+        if method == "pivoted_cholesky":
+            return RootLazyTensor(pivoted_cholesky(self.evaluate(), max_iter=100))
+
+        if method == "symeig":
+            evals, evecs = self.symeig(eigenvectors=True)
+            # TODO: only use non-zero evals (req. dealing w/ batches...)
+            F = evecs * evals.clamp(0.0).sqrt().unsqueeze(-2)
+            return RootLazyTensor(F)
+
+        if method == "svd":
```
jacobrgardner, line 1370: When would we ever use SVD for this task? The only cases where it'd be applicable are cases where symeig would apply.

Balandat: Small singular matrices on the GPU, since symeig I think is still terribly slow there. But yes, it's true that if that wasn't the case `symeig` would be preferable. Can't hurt to provide the option I guess?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1335,22 +1338,45 @@ def root_decomposition(self):
...
             except RuntimeError as e:
                 warnings.warn(
-                    "Runtime Error when computing Cholesky decomposition: {}. Using RootDecomposition.".format(e),
+                    f"Runtime Error when computing Cholesky decomposition: {e}. Using RootDecomposition.".format(e),
                     NumericalWarning,
                 )
+                method = "symeig"
+
+        if method == "pivoted_cholesky":
+            return RootLazyTensor(pivoted_cholesky(self.evaluate(), max_iter=100))
```
jacobrgardner, line 1362: We might want to use the max_root_decomposition_size here for max_iter

gpytorch/lazy/kronecker_product_lazy_tensor.py
```diff
@@ -149,6 +150,33 @@ def _size(self):
...
+            S.append(S_)
+            V.append(V_)
+        S = KroneckerProductLazyTensor(*[DiagLazyTensor(S_) for S_ in S]).diag()
+        U = KroneckerProductLazyTensor(*U)
+        V = KroneckerProductLazyTensor(*V)
+        return U, S, V
+
+    @cached(name="symeig")
+    def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
+        evals, evecs = [], []
```
jacobrgardner, line 168: If we wanted to sort here, we could sort evals and then use like a `PermutationLazyTensor` to represent shuffling the eigenvectors.

gpytorch/lazy/added_diag_lazy_tensor.py
```diff
@@ -125,3 +129,21 @@ def _init_cache_for_non_constant_diag(self, eye, batch_shape, n):
...
+        if torch.equal(diag, diag[..., :1].expand(diag.shape)):
+            U, S_, V = self._lazy_tensor.svd()
+            S = S_ + diag  # this assumes all diagonal entries are positive
+            return U, S, V
+        return super()._svd()
+
+    @cached(name="symeig")
+    def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
+        diag = self._diag_tensor.diag()
+        if torch.equal(diag, diag[..., :1].expand(diag.shape)):
```
jacobrgardner, line 145:
```suggestion
        if torch.equal(diag, diag[..., :1].expand(diag.shape)) and torch.all(diag[..., :1] >= 0):
```
Let's just explicitly check that all the diagonal elements are positive insteaad of assuming it.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -46,8 +46,8 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
         return inv_quad_term, logdet_term
 
     def _logdet(self):
-        eig_matrix, _ = self.lazy_tensor._symeig()
-        evals_plus_diag = eig_matrix.diag() + self.diag_tensor.diag()
+        evals, _ = self.lazy_tensor.symeig(eigenvectors=False)
+        evals_plus_diag = evals + self.diag_tensor.diag()
```
jacobrgardner, line 50: Might as well check that the diagonal tensor is (1) constant and (2) positive here as well -- I don't think `add` checks for that, so you could technically end up with this LT not meeting those conditions right?

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -46,8 +46,8 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
         return inv_quad_term, logdet_term
 
     def _logdet(self):
-        eig_matrix, _ = self.lazy_tensor._symeig()
-        evals_plus_diag = eig_matrix.diag() + self.diag_tensor.diag()
+        evals, _ = self.lazy_tensor.symeig(eigenvectors=False)
```
jacobrgardner, line 49: From what I understand of your `@cached` changes, will having the call with `eigenvectors=False` here cause us to do another symeig? We should avoid this, since it's a strict subset of the information we would be computing had we already done a `_solve` (e.g., we don't want to `symeig` twice for every `mll` computation.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1794,6 +1853,29 @@ def __getitem__(self, index):
...
+        U, S, V = torch.svd(self.evaluate())
+        return NonLazyTensor(U), S, NonLazyTensor(V)
+
+    @cached(name="symeig")
+    def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional["LazyTensor"]]:
+        """Method that allows implementing special-cased symeig computation. Should not be called directly"""
+        from gpytorch.lazy.non_lazy_tensor import NonLazyTensor
+
+        dtype = self.dtype  # perform decomposition in double precision for numerical stability
+        evals, evecs = torch.symeig(self.evaluate().to(dtype=torch.double), eigenvectors=eigenvectors)
```
jacobrgardner, line 1870: With this and some of the recent VI PRs, we've started becoming a little prescriptive about what precision ops happen in.  It might be nice for us to find a better solution that hard require people to use fp64 here, since for a lot of cases fp32 will be just fine.

Maybe we could have an internal of registry of methods we run in fp64 that can be user modified? We could accomplish this without directly patching e.g. `torch.symeig` / `torch.cholesky` by defining `gpytorch.symeig` / `gpytorch.cholesky` that get used internally.

Not necessarily important for this PR, but it seems like we could avoid hard coding fp64 casts everywhere.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
wjmaddox: This one is really on me; as it is, this is really only KP + ConstantDiag throughout. I need to make a PR for the general Kronecker structured case.

gpytorch/lazy/kronecker_product_lazy_tensor.py
Balandat: yeah - not sure if we do want to sort though. Consistency w/ pytorch methods would be nice but if we never use that assumption then why do the extra work?

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: good call - we can check the cache for the `eigenvectors =True` case and re-use those results if present

Balandat: I guess this actually does not include the `@cached` changes, these are part of the TLT PR (hint, hint).

gpytorch/lazy/lazy_tensor.py
Balandat: yeah good idea, created an issue for that: #1213

I'll leave it out of this PR for now.

gpytorch/lazy/added_diag_lazy_tensor.py
```diff
@@ -125,3 +129,21 @@ def _init_cache_for_non_constant_diag(self, eye, batch_shape, n):
...
         logdet -= (1.0 / self._noise).log().sum([-1, -2])
         self._precond_logdet_cache = logdet.view(*batch_shape) if len(batch_shape) else logdet.squeeze()
+
+    @cached(name="svd")
+    def _svd(self) -> Tuple["LazyTensor", Tensor, "LazyTensor"]:
+        diag = self._diag_tensor.diag()
+        if torch.equal(diag, diag[..., :1].expand(diag.shape)):
+            U, S_, V = self._lazy_tensor.svd()
+            S = S_ + diag  # this assumes all diagonal entries are positive
+            return U, S, V
```
wjmaddox, line 139: We probably don't want to include these because the backwards call of the SVD of the matrix plus the diagonal doesn't necessarily match the result of this operation, due to gradient un-identifiability. This should be what's causing the test failures.

In this case, it's also blindly assuming that the matrix is PSD, which may or may not be true; shifting doesn't necessarily hold for SVD of non PD matrices (thanks Jake for pointing that one out).

```suggestion

```

gpytorch/lazy/added_diag_lazy_tensor.py
```diff
@@ -125,3 +129,21 @@ def _init_cache_for_non_constant_diag(self, eye, batch_shape, n):
...
+        return super()._svd()
+
+    @cached(name="symeig")
+    def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
+        diag = self._diag_tensor.diag()
+        if torch.equal(diag, diag[..., :1].expand(diag.shape)) and torch.all(diag[..., :1] >= 0):
+            evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
+            evals = evals_ + diag  # this assumes all diagonal entries are positive
+            return evals, evecs
```
wjmaddox, line 148: As in the other comment: We probably don't want to include these because the backwards call of the symeig of the matrix plus the diagonal doesn't necessarily match the result of this operation, due to gradient un-identifiability. This should be what's causing the test failures.
```suggestion
```

wjmaddox: The better fix is to create an `AddedConstDiag` lazy tensor and update the unit tests to better match that.

gpytorch/lazy/added_diag_lazy_tensor.py
Balandat: I mean isn't the standing assumption for lazies throughout that they are psd? At least prior to my `TriangularLazyTensor` PR...

jacobrgardner: My main issue is that then the most efficient implementation of SVD is to just call symeig.

wjmaddox: I think so, but maybe @jacobrgardner would know of some cases where they are not necessarily?

Balandat: Fair enough - but aren't there still some serious issues with `symeig` speed on smaller matrices, especially on the GPU?

Balandat:
Hmm I'm not seeing any test failures locally, wonder what's going on there...

wjmaddox:
```
RuntimeError: isDifferentiableType(variable.scalar_type()) INTERNAL ASSERT FAILED at "/pytorch/torch/csrc/autograd/functions/utils.h":59, please report a bug to PyTorch. 
```

That's a weird bug? It seems to be happening in all of the current `svd`, `symeig` unit tests.

Balandat:
> RuntimeError: isDifferentiableType(variable.scalar_type()) INTERNAL ASSERT FAILED at "/pytorch/torch/csrc/autograd/functions/utils.h":59, please report a bug to PyTorch. 

This seems to be on pytorch master only. Let me see if i can repro that on fb infra.

Balandat:
Explanation here: https://github.com/pytorch/pytorch/issues/41389

wjmaddox:
Do we need `unique` to be differentiable in the tests? If not, there's probably no harm in preventing that by detaching.

The other test failures are due to an svd issue in `BlockInterleavedLT` but only once, I can check for that tomorrow.

Balandat:
Fixed up the failures on master. interestingly, I cannot repro the `BlockInterleavedLT` failure locally.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1800,6 +1859,30 @@ def __getitem__(self, index):
         # We're done!
         return res
 
+    @cached(name="svd")
+    def _svd(self) -> Tuple["LazyTensor", Tensor, "LazyTensor"]:
+        """Method that allows implementing special-cased SVD computation. Should not be called directly"""
+        from gpytorch.lazy.non_lazy_tensor import NonLazyTensor
+
+        U, S, V = torch.svd(self.evaluate())
+        return NonLazyTensor(U), S, NonLazyTensor(V)
```
wjmaddox, line 1868: Following some testing, using symeig throughout is preferable:

```suggestion
        U, S = self.symeig(eigenvectors=True)
        V = U.transpose(-1, -2)

        return U, S, V
```

Balandat:
Made `_svd` use `_symeig` by default (will need to overwrite this for non-psd LTs such as TLT). This had the added benefit of getting rid of the numerical issues causing test failures. 

This should be ready to go in.

gpytorch/lazy/added_diag_lazy_tensor.py
```diff
@@ -125,3 +130,23 @@ def _init_cache_for_non_constant_diag(self, eye, batch_shape, n):
...
+
+    @cached(name="symeig")
+    def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
+        try:
+            return pop_from_cache(self, "symeig", eigenvectors=True)
+        except CachingError:
+            pass
+        if isinstance(self._diag_tensor, ConstantDiagLazyTensor):
+            evals_, evecs = self._lazy_tensor.symeig(eigenvectors=eigenvectors)
+            evals = evals_ + self._diag_tensor.diag()  # this assumes all diagonal entries are positive
```
jacobrgardner, line 150: nit: not sure why this assumes all diagonal entries are positive. If v is an eigenvector of A, then `(A - s I)v = Av - s*Iv = \lambda*v - s*v = (\lambda - s)*v`.

gpytorch/lazy/batch_repeat_lazy_tensor.py
```diff
@@ -283,3 +286,22 @@ def repeat(self, *sizes):
...
+        U_, S_, V_ = self.base_lazy_tensor.svd()
+        U = U_.repeat(*self.batch_repeat, 1, 1)
+        S = S_.repeat(*self.batch_repeat, 1)
+        V = V_.repeat(*self.batch_repeat, 1, 1)
+        return U, S, V
+
+    @cached(name="symeig")
+    def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional[LazyTensor]]:
+        try:
+            return pop_from_cache(self, "symeig", eigenvectors=True)
```
jacobrgardner, line 301: I guess *technically* for these "wrapper" LTs we could check the base LT for its cache too?

jacobrgardner: Oh I see, calling `symeig` down below on the base class should handle that.

gpytorch/lazy/added_diag_lazy_tensor.py
Balandat: that's a typo - this was meant to be "the same" rather than "positive". But I guess it's kid of moot now that we have an explicit `ConstantDiagLazyTensor` for that.

PR #1111, opened 2020-04-07
Address all warnings in the test suite

### LINKED ISSUE #1034: Fix deprecated add / addmm / etc torch calls
On Torch master, some of our `addmm` / `add` etc calls are now deprecated, and doing pretty much anything raises deprecation warnings. We should fix this before the next torch release.

gpleiss:
This ensure we're using the most up-to-date calls for `addmm`, `add`, `addcmul`, `true_divide`, etc.

Also suppresses warnings that we expect to see in the test set.

[Fixes #1034]


gpleiss:
I think I got all of the warnings.

**Controversial decision** - travis now errors out on warnings for the PyTorch stable versions. @Balandat @jacobrgardner does this seem reasonable?

gpleiss:
Okay never mind - travis is erroring out on all of our non-raw docstrings. Seems like something we should fix, but probably as part of a different PR (because it'll add an overwhelming number of changes to this PR)

Balandat:
Re raising errors on warnings: Is that pytorch specific or all test warnings? This might be a little aggressive if warnings are raised in other places, say in numpy, and could create a bunch of thrash. Maybe for now we can go with the guideline to avoid warnings but not error out?

gpleiss:
I set it to run `python -W error -m unittest` for pytorch stable (but not for pytorch master). At the moment neither is raising any warnings. I figured it was okay because the pytorch/numpy versions are locked down on these runs, and so all new warnings on these builds are probably our fault.

More than happy to revert that commit though.

Balandat:
That's reasonable, though this is still failing on a sphinx warning, so that's another potential source of external noise.

gpleiss:
Alright - i removed the "fail on error," and the sphinx issues should be fixed.

jacobrgardner:
I suppose one option would be to run another build in the matrix against pytorch stable that errors on warnings but filters warnings that aren't thrown from torch, which I think is possible based on the warnings docs.

We could set up another PR for something like that though.

gpleiss:
That sounds like a good approach. This PR is ready to go then!

PR #1120, opened 2020-04-16
Adding RQ kernel to documentation

ndalchau:
Very simple PR that adds the RQ kernel to the docs.
Also fixing the position of the body for AdditiveStructureKernel, which went for walk elsewhere.


ndalchau:
N.B. I noticed the Newton-Girard Additive Kernel is also not in the docs. I'd add it, but don't know which category it should be part of.

PR #1126, opened 2020-04-22
New PyTorch version for travis

gpleiss:
Updating to 1.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
       pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html;
       python setup.py build develop;
     else
-      pip install torch==1.3.1+cpu -f https://download.pytorch.org/whl/torch_stable.html;
+      pip install torch==1.5+cpu -f https://download.pytorch.org/whl/torch_stable.html;
     fi
   - if [[ $WITH_PYRO == true ]]; then
       pip install git+https://github.com/pyro-ppl/pyro@dev;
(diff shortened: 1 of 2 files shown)


gpleiss:
@Balandat @jacobrgardner I was making sure we were compatible with 1.5 before our new release. There's a really strange bug with the pyro integration that only happens with `MeanFieldVariationalStrategy` - and I threw in a very hacky fix for it.

Not sure if this is an issue with the new version of PyTorch (I know there's some changes to PyTorch optimizers related to inplace changes https://github.com/pytorch/pytorch/pull/33640 and https://github.com/pytorch/pytorch/pull/34211). It could also be an issue on Pyro's end - and I'm not really sure how to proceed debugging.

Since this seems to be an edge case I'm okay throwing in a temporary hacky fix that will hopefully get ironed out soon. Is this okay with you two?

Balandat:
Curious what causes this. What exactly is the bug? Another possibility would be recent changes to type promotion, but it seems like the `ones_like` wouldn't affect that?

gpleiss:
I was getting "one of the variables needed for gradient computation has been modified by an inplace operation." This was only happening with `MeanFieldVariationalDistribution`, not `Cholesky` or `Delta`.

The `MeanFieldVariationalDistribution` has the following line:

```python
# self._variational_stddev is a parameter
variance = self._variational_stddev.pow(2)  # This line is causing the issue
covar = DiagLazyTensor(variance)
```

If I removed the `.pow(2)` - everything worked. If I did `covar = CholLazyTensor(DiagLazyTensor(self._variational_stddev))` - everything worked. The current solution (multiplying `self._variational_stddev` by an all ones mask) works. Really not sure why this is failing.

Balandat:
huh, weird

PR #1136, opened 2020-05-06
Add cholesky_jitter to settings.

Balandat:
Currently it's not easily possible to customize the jitter value in the cholesky computations inside the lazy tensors. This adds a setting so that we can now do
```
with settings.cholesky_jitter(1e-5):
    lt.cholesky()
```

Naturally, this will also apply this to the solves (when performing posterior inference). This also adds this setting to other uses of `psd_safe_cholesky` in the code.

Note that this PR does not change current behavior as the default setting value is `None`, which is the default argval of `psd_safe_cholesky`.

### DIFF
--- a/gpytorch/kernels/inducing_point_kernel.py
+++ b/gpytorch/kernels/inducing_point_kernel.py
@@ -5,6 +5,7 @@ import math
 
 import torch
 
+from .. import settings
 from ..distributions import MultivariateNormal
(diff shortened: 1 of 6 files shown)


jacobrgardner:
Failures look like a misplaced parenthesis causing jitter to be passed to delazify as an arg

Balandat:
Hah, what would we do without tests?

Balandat:
Fixed bad parentheses.

Issue #1185, opened 2020-06-18
Add Kernels for Distributional Input [Feature Request]


wjmaddox:
# 🚀 Feature Request

Rather than requiring a variational method to incorporate uncertainty over input data points, a more natural method is to compute the kernel directly on the distribution of input space. 
One of the original methods in this space was an [exponentiated KL kernel](https://papers.nips.cc/paper/2351-a-kullback-leibler-divergence-based-kernel-for-svm-classification-in-multimedia-applications.pdf), while probability product kernels became popular for a while. Now, kernel mean embeddings have become very popular and useful and supporting them would expand the usefulness of the package.

## Motivation

A better way of dealing with input uncertainty than simply variational methods. This is highly relevant for some Bayesian optimization applications such as embedding a VAE model (which technically outputs a distribution) in a gpytorch model and [here](https://arxiv.org/abs/1902.07908).

## Pitch

Basically this forwards for the kernel --- assuming that x1, x2, represent the mean and (log) variances of the Gaussian distribution:
```python
def forward(x1, x2):
      covar_dist = -_symmetrized_kl(x1, x2)
      return covar_dist.div(self.lengthscale).exp()
```

where symmetrized kl will be: $$KL(p(\mu_1, D_1) || p(\mu_2, D_2)) + KL(p(\mu_2, D_2) || p(\mu_1, D_1))$$

Using the mean in a standard kernel will produce something sensible as well.

PR coming in a minute.

PR #1186, opened 2020-06-18
Include Exponentiated Symmetrized KL Divergence Kernel.

wjmaddox:
See #1185 for more details. But, this is an implementation of the exponentiated symmetrized KL divergence kernel between two Gaussians that are diagonal. To do so, I assume that the first `num_dims` observations are the mean, and the second half are the log variance.

This boils down to $\sum_{i=1}^D d1_i / d2_i + d2_i  / d1_i - 2D + (mu1 - mu2)'(1/d1 + 1/d2)(mu1 - mu2)$ where $p(x) = N(mu1, d1)$ and $p(x') = N(mu2, d2).$ So it should be PSD, and we can compute by the standard torch kl_divergence implementation.

Not quite finished writing(/blocking?) the unit tests.


gpytorch/kernels/distributional_input_kernel.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+
+from torch.distributions import Normal, kl_divergence
+
+from .kernel import Kernel
+
+
+def _symmetrized_kl(dist1, dist2, num_dims=2, eps=1e-8):
```
Balandat, line 8: maybe add some docstring here to make clear that this assumes that the inputs are assumed MVNs with diagonal covars

gpytorch/kernels/distributional_input_kernel.py
```diff
@@ -0,0 +1,66 @@
...
+    dist1_mean = dist1[..., :num_dims].unsqueeze(-3)
```
Balandat, line 9: Would we ever have the case where `num_dims != dist1.shape[-1] // 2`? If not, we can get rid of the arg and just infer it this way to simplify the interface.

gpytorch/kernels/distributional_input_kernel.py
```diff
@@ -0,0 +1,66 @@
...
+    dist1_var = eps + dist1_logvar.exp()
+
+    dist2_mean = dist2[..., :num_dims].unsqueeze(-3)
+    dist2_logvar = dist2[..., num_dims:].unsqueeze(-2)
+    dist2_var = eps + dist2_logvar.exp()
+
+    dist1 = Normal(dist1_mean, dist1_var ** 0.5)
+    dist2 = Normal(dist2_mean, dist2_var ** 0.5)
+
+    res = (kl_divergence(dist1, dist2) + kl_divergence(dist2, dist1)).sum(-1)
```
Balandat, line 20: Since we know / assume these are Gaussian, we could just use `_kl_normal_normal` instead to shortcut all the registry business. But I guess not using the "private" API is also virtuous. Alternatively we could just compute this directly without having to construct any distributions:

```
var_ratio12 = dist1_var / dist2_var
log_var_ratio12 = var_ratio12.log()
loc_sqdiffs = (dist1_mean - dist2_mean).pow(2)
kl1 = 0.5 * (var_ratio12 + loc_sqdiffs / dist2_var - 1 - log_var_ratio12)
kl2 = 0.5 * (var_ratio12.reciprocal() + loc_sqdiffs / dist1_var - 1 + log_var_ratio12)
kl = kl1 + kl2
```

gpytorch/kernels/distributional_input_kernel.py
```diff
@@ -0,0 +1,66 @@
...
+    (https://papers.nips.cc/paper/2351-a-kullback-leibler-divergence-based-kernel-for-svm-\
+    classification-in-multimedia-applications.pdf) for the symmetrized KL. For
+    general distributions, these kernels may not produce PSD kernels; however, they do for
+    Gaussian distributions which are implemented here.
+
+    Args:
+        :attr:`distance` (string, optional, default=`symmetrized_kl`):
+            `symmetrized_kl`. Determines whether to use the
+            exponentiated symmetrized KL divergence.
+        :attr:`num_dims` (int): The first `num_dims` dimensions of the last dimension are
```
Balandat, line 45: seems unnecessary to specify this - see comment above

gpytorch/kernels/distributional_input_kernel.py
```diff
@@ -0,0 +1,66 @@
...
+            assumed to be the mean. The second num_dims are assumed to be the log variance.
+    """
+    has_lengthscale = True
+
+    def __init__(self, distance="symmetrized_kl", num_dims=4, **kwargs):
```
Balandat, line 50: instead of using the `distance` arg, you could make `DistributionalInputKernel` abstract, and give it an abstract `distance_function` method. Then you can implement `SymmetrizedKLKernel(DistributionalInputKernel)` by implementing the `distance_function`. That seems maybe a little more modular than this approach?

gpytorch/kernels/distributional_input_kernel.py
wjmaddox: My original plan for the PR was to support more than just the symmetrized_KL but I couldn't immediately verify that the Wasserstein implementation was behaving properly.

I'll remove for now.

gpytorch/kernels/distributional_input_kernel.py
wjmaddox: Not entirely sure. For other distributions, very possibly?

gpytorch/kernels/distributional_input_kernel.py
wjmaddox: Switched to your implementation and resolving.

gpytorch/kernels/gaussian_symmetrized_kl_kernel.py
```diff
@@ -0,0 +1,54 @@
...
+    dist2_mean = dist2[..., :num_dims].unsqueeze(-2)
+    dist2_logvar = dist2[..., num_dims:].unsqueeze(-2)
+    dist2_var = eps + dist2_logvar.exp()
+
+    var_ratio12 = dist1_var / dist2_var
+    log_var_ratio12 = var_ratio12.log()
+    loc_sqdiffs = (dist1_mean - dist2_mean).pow(2)
+    kl1 = 0.5 * (var_ratio12 + loc_sqdiffs / dist2_var - 1 - log_var_ratio12)
+    kl2 = 0.5 * (var_ratio12.reciprocal() + loc_sqdiffs / dist1_var - 1 + log_var_ratio12)
+    symmetrized_kl = kl1 + kl2
```
Balandat, line 35: `log_var_ratio12` completely cancels here. Might as well get rid of it (can leave a comment here though it's clear)

wjmaddox: Yeah, i should have read your code more closely.

PR #1191, opened 2020-06-20
[WIP] Natural gradient descent

### LINKED ISSUE #894: Natural Gradient Optimizer
Hello, 

I have seen in the [literature](https://arxiv.org/abs/1803.09151) for GP algorithms that use inducing points and variational inference to perform inference can sometimes converge slowly when only using optimizers such as stochastic gradient descent. This is apparently due to the difficulty of optimizing the variational parameters as well as the likelihood and kernel parameters. The authors in that paper suggest that using the **natural gradient optimizer** has shown to have big improvements especially in large-scale GP methods like the **SVGP** and **DeepGP**.

I've seen the NatGrad optimizer has already been implemented in this [branch](https://github.com/cornellius-gp/gpytorch/tree/natural) within the GPyTorch library. I wanted to know what was everyone's experience using it and why is it not in the main branch? Perhaps the nature of using matrix-vector-multiplication methods don't suffer from the same issue of convergence or joint optimization? Or it's just a matter of code coverage?

gpleiss:
- [x] Add `NaturalVariationalDistribution`
- [ ] Add unit/example tests
- [ ] Add tutorial notebook

Closes #894

### DIFF
--- a/gpytorch/module.py
+++ b/gpytorch/module.py
@@ -138,9 +138,14 @@ class Module(nn.Module):
         return _extract_named_added_loss_terms(module=self, memo=None, prefix="")
 
     def named_hyperparameters(self):
+        from .variational._variational_distribution import _VariationalDistribution
+
         for name, param in self.named_parameters():
-            if "variational_" not in name:
-                yield name, param
+            gen = self._named_members(
+                lambda module: module._parameters.items() if not isinstance(module, _VariationalDistribution) else [],
+            )
+        for elem in gen:
+            yield elem
 
     def named_priors(self, memo=None, prefix=""):
         """Returns an iterator over the module's priors, yielding the name of the prior,
@@ -159,9 +164,14 @@ class Module(nn.Module):
         return _extract_named_constraints(module=self, memo=None, prefix="")
 
     def named_variational_parameters(self):
+        from .variational._variational_distribution import _VariationalDistribution
+
         for name, param in self.named_parameters():
-            if "variational_" in name:
-                yield name, param
+            gen = self._named_members(
+                lambda module: module._parameters.items() if isinstance(module, _VariationalDistribution) else [],
+            )
+        for elem in gen:
+            yield elem
 
     def register_added_loss_term(self, name):
         self._added_loss_terms[name] = None
--- a/docs/source/variational.rst
+++ b/docs/source/variational.rst
@@ -118,3 +118,10 @@ distributions.
 
 .. autoclass:: MeanFieldVariationalDistribution
    :members:
+
+
+:hidden:`NaturalVariationalDistribution`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. autoclass:: NaturalVariationalDistribution
+   :members:
--- a/examples/04_Variational_and_Approximate_GPs/Natural_Gradient_Descent.ipynb
+++ b/examples/04_Variational_and_Approximate_GPs/Natural_Gradient_Descent.ipynb
@@ -1,0 +1,35 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Natural Gradient Descent with SVGP Regression\n",
+    "\n",
+    "TODO"
+   ]
+  }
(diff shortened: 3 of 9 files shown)


rhaps0dy:
Hi Geoff. I have an implementation of `NaturalVariationalDistribution` which is quite similar to yours, but it also has tests. I'd like to contribute it to GPytorch. We should maybe merge the two? At least you can use my tests.

It's pretty cool that we've converged to the same solution to implement natural gradients :)

https://gist.github.com/rhaps0dy/6174458e99b1a6aa76bf9da6c434d97a

jacobrgardner:
@rhaps0dy Awesome! I'll look at your implementation more closely soon, but if you have a working solution with unit tests, then maybe it might make sense to fork the `ngd` branch in favor of a new PR that merges the two? Looking at the diff for this PR, it includes some other nice to have changes (like `model.variational_parameters` no longer depending on string matching *cough cough*...)

gpytorch/module.py
```diff
@@ -138,9 +138,14 @@ def named_added_loss_terms(self):
         return _extract_named_added_loss_terms(module=self, memo=None, prefix="")
 
     def named_hyperparameters(self):
+        from .variational._variational_distribution import _VariationalDistribution
+
         for name, param in self.named_parameters():
-            if "variational_" not in name:
-                yield name, param
+            gen = self._named_members(
```
rhaps0dy, line 144: What if there are several `_VariationalDistribution` modules? This will give the wrong answer silently. At least, the `for elem in gen:` line and the one below should be indented 4 spaces to the right. I'd even go further, and remove all non-public `nn.Module` methods:

```python
    for module_prefix, module in self.named_modules():
        if not isinstance(module, _VariationalDistribution):
            for elem in module.named_parameters(prefix=module_prefix, recurse=False):
                yield elem
```

rhaps0dy:
@jacobrgardner, I'm happy to do that. My implementation also includes `TrilNaturalDistribution`, which represents the natural matrix with a lower-triangular matrix, and thus avoids giving Cholesky errors after a while of running SGD.

I've also put more care in the `forward` method, I think, than Geoff's implementation here. I avoid calling an `inverse()` every time.

I really like Geoff's way of initialising the variational distribution, though, and names for the parameters.

rhaps0dy:
I had beautiful string docs that used Unicode to display math nicely, but these had to go because documents are forced to be encoded in ascii by `black`.

Why is it like this?

The nice and readable
```
        """Calculate dout/d(η1, η2), which are:
        η1 = μ
        η2 = μμᵀ + LLᵀ = μμᵀ + Σ

        Thus:
        dout/dη1 = dout/dμ + dout/dL dL/dη1
        dout/dη2 = dout/dL dL/dη1

        For L = chol(η2 - η1⋅η1ᵀ).
        dout/dΣ = _cholesky_backward(dout/dL, L)
        dout/dη2 = dout/dΣ
        dΣ/dη1 = -2* (dout/dΣ) μ
        """
```

turned into

```
        """Calculate dout/d(eta1, eta2), which are:
        eta1 = mu
        eta2 = mu*mu^T + LL^T = mu*mu^T + Sigma

        Thus:
        dout/deta1 = dout/dmu + dout/dL dL/deta1
        dout/deta2 = dout/dL dL/deta1

        For L = chol(eta2 - eta1*eta1^T).
        dout/dSigma = _cholesky_backward(dout/dL, L)
        dout/deta2 = dout/dSigma
        dSigma/deta1 = -2* (dout/dSigma) mu
        """
```

which is OK I suppose, but not amazing.

jacobrgardner:
@rhaps0dy I think we'd prefer math docstrings typeset with LaTeX anyways since RTD renders LaTeX ([e.g.](https://docs.gpytorch.ai/en/v1.1.1/variational.html#variational-strategies))

rhaps0dy:
It's a docstring for an internal function, I'm not sure it appears there?

Either way, turning it into LaTeX is too much work, and it would hamper readability of the raw docstring, which is what's important (since it's what you see on the text editor)

I've added the natural gradient implementation and tests to the other pull request #1207. Only the tutorial notebook remains.

gpleiss:
closed in favor of #1207

PR #1193, opened 2020-06-23
[WIP] Add DSPPs

jacobrgardner:
- [x] Various internal changes (e.g., add `to_data_independent_dist` to MVN / MMVN).
- [x] Cleaned up `predictive_deep_gp.py`
- [x] Add DSPP example notebook
- [x] Rename Predictive* -> DSPP
- [x] Fix batch dimensions so that the quadrature sites correspond to the outermost batch dim (dim 0).
- [x] Move quadrature weights to DSPP class, remove DeepPredictive likelihoods in favor of `DeepPredictiveLogLikelihood`


gpytorch/models/deep_gps/predictive_deep_gp.py
```diff
@@ -0,0 +1,119 @@
...
+        else:
+            self.quad_grid = torch.nn.Parameter(torch.randn(num_sample_sites, input_dims))
+
+    def __call__(self, inputs, are_samples=False, expand_for_quadgrid=True, **kwargs):
+        if isinstance(inputs, MultitaskMultivariateNormal):
+            # inputs is definitely in the second layer, and mean is n x t
+            mus, sigmas = inputs.mean, inputs.variance.sqrt()
+
+            if expand_for_quadgrid:
+                xi_mus = mus.unsqueeze(-3)  # 1 x n x t
```
gpleiss, line 30: I think we should make this `unsqueeze(0)` for two reasons:

1) It will be more similar to DeepGPs this way (the sample dimension for deep gps is always the left-most batch dimension)
2) It will allow us to do batch DSPPs easier.

gpytorch/models/deep_gps/predictive_deep_gp.py
```diff
@@ -0,0 +1,119 @@
...
+                xi_sigmas = sigmas.unsqueeze(-3)  # 1 x n x t
+            else:
+                xi_mus = mus
+                xi_sigmas = sigmas
+
+            # unsqueeze sigmas to 1 x n x t, locations from [q] to Q^T x 1 x T.
+            # Broadcasted result will be Q^T x N x T
+            qg = self.quad_grid.unsqueeze(-2)
```
gpleiss, line 38: See comment above. This could be something like `self.quad_grid.view([self.num_sample_sites] + [1] * (xi_mus.dim() - 2) + [self.input_dims])`.

gpytorch/models/deep_gps/predictive_deep_gp.py
```diff
@@ -0,0 +1,119 @@
...
+                mean = output.loc.transpose(-1, -2)
+                covar = BlockDiagLazyTensor(output.lazy_covariance_matrix, block_dim=-3)
+                output = MultitaskMultivariateNormal(mean, covar, interleaved=False)
+        else:
+            output = output.loc.transpose(-1, -2)  # this layer provides noiseless kernel interpolation
+
+        return output
+
+
+class DeepPredictiveGaussianLikelihood(GaussianLikelihood):
```
gpleiss, line 74: Given that `DeepLikelihood` is deprecated (in favor of `DeepApproximateMLL`), I would get rid of this class.

gpytorch/models/deep_gps/predictive_deep_gp.py
```diff
@@ -0,0 +1,119 @@
...
+
+        deep_log_prob = deep_log_marginal.logsumexp(dim=-2)
+
+        return deep_log_prob
+
+    def forward(self, *args, **kwargs):
+        pass
+
+
+class MultitaskDeepPredictiveGaussianLikelihood(MultitaskGaussianLikelihood):
```
gpleiss, line 98: Same here.

gpytorch/models/deep_gps/predictive_deep_gp.py
```diff
@@ -0,0 +1,119 @@
...
+from gpytorch import settings
+from gpytorch.distributions import MultitaskMultivariateNormal
+from gpytorch.lazy import BlockDiagLazyTensor
+from gpytorch.likelihoods import GaussianLikelihood, MultitaskGaussianLikelihood
+
+from ..approximate_gp import ApproximateGP
+from .deep_gp import DeepGPLayer as AbstractDeepGPLayer
+
+
+class AbstractPredictiveDeepGPLayer(AbstractDeepGPLayer):
```
gpleiss, line 12: Would it make sense to change the name of this class to `AbstractDSPPLayer`? cc/ @martinjankowiak

gpytorch/models/deep_gps/predictive_deep_gp.py
jacobrgardner: I think for DSPPs the likelihood makes more sense than it did in the standard deep GP case because it currently handles summing over the quadrature weights (and is no longer just a matter of dealing with the additional samples dimension).

What do you think?

jacobrgardner: I guess we could make like a `DeepPredictiveLogLikelihood` or something similar

gpytorch/models/deep_gps/predictive_deep_gp.py
jacobrgardner: Yeah this makes sense :+1:

gpytorch/models/deep_gps/predictive_deep_gp.py
gpleiss: I'd be more in favor of `DeepPredictiveLogLikelihood` - otherwise we'll have a DSPPGaussianLikelihood and a DSPPMultitaskGaussianLikelihood and maybe a DSPPBernoulliLikelihood and DSPPSoftmaxLikelihood if we end up doing classification.

gpytorch/models/deep_gps/predictive_deep_gp.py
jacobrgardner: Probably makes sense. I'll almost certainly also change the file name as well

gpytorch/models/deep_gps/predictive_deep_gp.py
jacobrgardner: @gpleiss I looked at this a bit more to refresh my memory, and we need the quadrature weights to use the predictions of the DSPP. E.g., the `log_marginal` code here gives the marginal predictive log likelihood for the input test point.

We need the quadrature grid to live somewhere, and it shouldn't be on the loss object since that doesn't make much sense for users to save and load. Moreover, I'm not sure it really makes sense for them to live on any particular DSPP layer. 

What I think I'm going to do is make a `DSPP(GP):` class analogous to the `DeepGP` container in `deep_gp.py` and put them there.

gpytorch/mlls/deep_predictive_log_likelihood.py
```diff
@@ -0,0 +1,36 @@
+from ..models.deep_gps.dspp import DSPP
+from ._approximate_mll import _ApproximateMarginalLogLikelihood
+
+
+class DeepPredictiveLogLikelihood(_ApproximateMarginalLogLikelihood):
+    """
+    A wrapper to make a GPyTorch approximate marginal log likelihoods compatible with Deep GPs.
```
gpleiss, line 7: nit: DSPPs not Deep GPs.

gpleiss: Also very nit: could you do "compatible with :obj:`~gpytorch.models.deep_gps.DSPP`?

gpytorch/models/deep_gps/dspp.py
```diff
@@ -0,0 +1,114 @@
...
+                xi_mus = mus.unsqueeze(0)  # 1 x n x t
+                xi_sigmas = sigmas.unsqueeze(0)  # 1 x n x t
+            else:
+                xi_mus = mus
+                xi_sigmas = sigmas
+
+            # unsqueeze sigmas to 1 x n x t, locations from [q] to Q^T x 1 x T.
+            # Broadcasted result will be Q^T x N x T
+            qg = self.quad_sites.view([self.num_quad_sites] + [1] * (xi_mus.dim() - 2) + [self.input_dims])
+            # qg = qg + torch.randn_like(qg) * 1e-2
```
martinjankowiak, line 61: remove?

PR #1194, opened 2020-06-23
Add Contour integral quadrature for exact GP sampling

jacobrgardner:
Implements contour integral quadrature as described **[here](https://arxiv.org/abs/2006.11267)** for sampling from exact GP posteriors. Any time `rsample` is called on a `MultivariateNormal` with a `LazyTensor` covariance matrix (or more specifically, when `LazyTensor.zero_mean_mvn_samples` is called), if it is done in a `with gpytorch.settings.ciq_samples():` context, the samples will be drawn using contour integral quadrature instead of Cholesky.

Support for whitened variational inference using CIQ will come in a follow up PR.

Setting is off by default because, like CG, it is not faster than Cholesky until N ~ 1000. Once the feature has been on master for a while, we can have it automatically switch to using CIQ in certain conditions. This will be slightly trickier, because Cholesky is faster than CIQ when drawing a very large number (e.g., O(N)) of samples.


gpytorch/lazy/lazy_tensor.py
```diff
@@ -1585,15 +1621,33 @@ def zero_mean_mvn_samples(self, num_samples):
             :obj:`torch.tensor`:
                 Samples from MVN (num_samples x batch_size x num_dim) or (num_samples x num_dim)
         """
-        if self.size()[-2:] == torch.Size([1, 1]):
-            covar_root = self.evaluate().sqrt()
+        if settings.ciq_samples.on():
+            from ..utils.contour_integral_quad import contour_integral_quad
```
martinjankowiak, line 1625: why conditional importing?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1585,15 +1621,33 @@ def zero_mean_mvn_samples(self, num_samples):
...
+
+            base_samples = torch.randn(
+                *self.batch_shape, self.size(-1), num_samples, dtype=self.dtype, device=self.device
+            )
+            base_samples = base_samples.permute(-1, *range(self.dim() - 1)).contiguous()
+            base_samples = base_samples.unsqueeze(-1)
+            solves, weights, _, _ = contour_integral_quad(
+                self.evaluate_kernel(),
```
martinjankowiak, line 1633: what's the plan for custom mvm support?

gpytorch/utils/cholesky.py
```diff
@@ -8,7 +8,7 @@
 from .warnings import NumericalWarning
 
 
-def psd_safe_cholesky(A, upper=False, out=None, jitter=None):
+def psd_safe_cholesky(A, upper=False, out=None, jitter=None, max_tries=3):
```
martinjankowiak, line 11: add `max_tries` to docstring

gpytorch/utils/contour_integral_quad.py
```diff
@@ -0,0 +1,137 @@
...
+    :param torch.Tensor rhs: Right hand side tensor :math:`\mathbf b`
+    :param bool inverse: (default False) whether to compute :math:`\mathbf K^{1/2} \mathbf b` (if False)
+        or `\mathbf K^{-1/2} \mathbf b` (if True)
+    :param int max_lanczos_iter: (default 10) Number of Lanczos iterations to run (to estimate eigenvalues)
+    :param int num_contour_quadrature: (default 15) How many quadrature samples to use for approximation
+    :rtype: torch.Tensor
+    :return: Approximation to :math:`\mathbf K^{1/2} \mathbf b` or :math:`\mathbf K^{-1/2} \mathbf b`.
+    """
+    import numpy as np
+    from scipy.special import ellipj, ellipk
```
martinjankowiak, line 39: i take it this is already a gpytorch requirement?

gpytorch/utils/contour_integral_quad.py
```diff
@@ -0,0 +1,137 @@
...
+        # Compute an approximate condition number
+        # We'll do this with Lanczos
+        try:
+            approx_eigs = lanczos_mat.symeig()[0]
+            if approx_eigs.min() <= 0:
+                raise RuntimeError
+        except RuntimeError:
+            approx_eigs = lazy_tensor.diag()
+
+        max_eig = approx_eigs.max(dim=-1)[0]
```
martinjankowiak, line 87: should we make these more conserative?

gpytorch/utils/contour_integral_quad.py
```diff
@@ -0,0 +1,137 @@
...
+    max_lanczos_iter=20,
+    num_contour_quadrature=7,
+    shift_offset=0,
+):
+    r"""
+    Performs :math:`\mathbf K^{1/2} \mathbf b` or `\mathbf K^{-1/2} \mathbf b`
+    using contour integral quadrature.
+
+    .. note::
+        Right now this only works for non-batch matrices
```
martinjankowiak, line 27: will this fail gracefully?

gpytorch/utils/minres.py
```diff
@@ -0,0 +1,257 @@
...
+        uses the constant stored in :obj:`gpytorch.settings.max_cg_iterations`.
+    :rtype: torch.Tensor
+    :return: The solves :math:`\mathbf x`. The shape will correspond to the size of `rhs` and `shifts`.
+    """
+    # Default values
+    if torch.is_tensor(matmul_closure):
+        matmul_closure = matmul_closure.matmul
+    mm_ = matmul_closure
+    if preconditioner is None:
+        preconditioner = lambda x: x.clone()
```
martinjankowiak, line 30: why clone?

gpytorch/utils/minres.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # Epsilon (to prevent nans)
+    eps = torch.tensor(eps, dtype=rhs.dtype, device=rhs.device)
+
+    # Create space for matmul product, solution
+    prod = mm_(rhs)
+    if value is not None:
+        prod.mul_(value)
+
+    # Reisze shifts
```
martinjankowiak, line 59: typo

gpytorch/utils/minres.py
```diff
@@ -0,0 +1,257 @@
...
+    shifts = _pad_with_singletons(shifts, 0, prod.dim() - shifts.dim() + 1)
+    solution = torch.zeros(shifts.shape[:1] + prod.shape, dtype=rhs.dtype, device=rhs.device)
+
+    # Variasbles for Lanczos terms
```
martinjankowiak, line 63: typo

gpytorch/lazy/lazy_tensor.py
jacobrgardner: This just evaluates any outstanding kernel calls, it doesn't necessarily explicitly materialize an `n x n` matrix (i.e., `evaluate_kernel` vs `evaluate`).

gpytorch/utils/contour_integral_quad.py
jacobrgardner: Actually I looked and we explicitly don't require you to have numpy / scipy installed anywhere and have by convention imported them directly in functions/features where they are used (the same way we handle pyro). I ended up moving the imports down here to follow this convention

gpytorch/utils/contour_integral_quad.py
jacobrgardner: I don't know how we would do this in a principled way. For many matrices, our condition number estimate will be accurate to machine precision.

gpytorch/utils/minres.py
jacobrgardner: We later modify the output of the preconditioner in place:
https://github.com/cornellius-gp/gpytorch/blob/fac03c2622e72288e2edf5d5b3edce7cbc7b3846/gpytorch/utils/minres.py#L75

gpytorch/utils/contour_integral_quad.py
jacobrgardner: Oops, that note actually isn't even true anymore

gpytorch/lazy/lazy_tensor.py
martinjankowiak: ah gotcha thanks

gpytorch/utils/contour_integral_quad.py
martinjankowiak: why not just do a factor of 2 fudge or something?

gpytorch/settings.py
```diff
@@ -484,6 +511,14 @@ class prior_mode(_feature_flag):
...
 
 
+class single_precision_cholesky(_feature_flag):
+    """
+    Use single precision when computing the Cholesky decomposition (and solves)
+    """
+
+    _state = False
+
+
```
jacobrgardner, line 521:
```suggestion
```
This setting is currently unused on any branch as far as I can tell

gpytorch/utils/contour_integral_quad.py
jacobrgardner: Just multiplying the max eigenvalue by 2 almost surely gets us further from the true max eigenvalue than the estimate here, and I'm not clear that intentionally overestimating the condition number by a factor of sqrt(2) buys us anything in terms of CIQ convergence? 

These estimates are going to be significantly more accurate than, e.g., 20 iterations of power iteration. They should be pretty darn near exact.

gpytorch/utils/contour_integral_quad.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def contour_integral_quad(
+    lazy_tensor,
+    rhs,
+    inverse=False,
+    weights=None,
+    shifts=None,
+    max_lanczos_iter=20,
+    num_contour_quadrature=7,
```
dme65, line 19: Default should be 15?

PR #1208, opened 2020-07-10
Use trace estimation for logdet backwards in KPAddedDiagLT

wjmaddox:
This PR should fix any possible bugs with the logdet backwards in KPAddedDiagLazyTensor --- that I would have introduced.

Basically, the gradients aren't well defined for the log det if there are repeated eigenvalues and backpropagating through `symeig` is known to be unstable. Instead, we can use stochastic trace estimation to estimate the gradient, which is easy in this case because solves are cheap and can be done in batch with the original solve.

The trick I used here is to skip the logdet forward Lanczos call in the `inv_quad_logdet` function while detaching the `_logdet()` so that we add zero to the correctly computed forwards pass in the logdet while also propagating the gradients.

### DIFF
--- a/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
@@ -30,7 +30,7 @@ class KroneckerProductAddedDiagLazyTensor(AddedDiagLazyTensor):
         # solve but we only want to cache the probe vectors for the backwards
         with skip_logdet_forward(True):
             inv_quad_term, func_logdet_term = super().inv_quad_logdet(
-                inv_quad_rhs=inv_quad_rhs, logdet=logdet is not None, reduce_inv_quad=reduce_inv_quad
+                inv_quad_rhs=inv_quad_rhs, logdet=logdet, reduce_inv_quad=reduce_inv_quad
             )
 
         if logdet is not None:
--- a/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
@@ -26,13 +26,6 @@ class KroneckerProductAddedDiagLazyTensor(AddedDiagLazyTensor):
             raise RuntimeError("One of the LazyTensors input to AddedDiagLazyTensor must be a DiagLazyTensor!")
 
     def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True):
-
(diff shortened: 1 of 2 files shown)


gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -25,19 +26,23 @@ def __init__(self, *lazy_tensors, preconditioner_override=None):
             raise RuntimeError("One of the LazyTensors input to AddedDiagLazyTensor must be a DiagLazyTensor!")
 
     def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True):
+        # we want to call the standard InvQuadLogDet to easily get the probe vectors and do the
+        # solve but we only want to cache the probe vectors for the backwards
+        with skip_logdet_forward(True):
+            inv_quad_term, func_logdet_term = super().inv_quad_logdet(
+                inv_quad_rhs=inv_quad_rhs, logdet=logdet is not None, reduce_inv_quad=reduce_inv_quad
```
jacobrgardner, line 33:
```suggestion
                inv_quad_rhs=inv_quad_rhs, logdet=logdet, reduce_inv_quad=reduce_inv_quad
```
As is I think this will always compute the logdet in the forward, since `False is not None`

wjmaddox: Oops, totally thought that it could be None instead of False.

gpytorch/lazy/kronecker_product_lazy_tensor.py
```diff
@@ -188,7 +188,7 @@ def _symeig(self, eigenvectors=True):
             evals_, evecs_ = eval_tensor.double().symeig(eigenvectors=eigenvectors)
 
             # we chop any negative eigenvalues
-            evals_ = evals_.clamp_min(0.0)
+            evals_.clamp_min_(0.0)
```
Balandat, line 191: I think we should still avoid doing in-place tensor updates, as there could be other more general use cases in which we'd want to backprop through `symeig`. The additional mem allocation is dwarfed by the actual computations anyway, so there is no efficiency issue.

jacobrgardner: At the moment as coded it will never backprop through symeig. If a PR changes this behavior then this should probably not be in place, yeah. Can change this now to avoid a gotcha, I suppose

Balandat: I'll put up a quick PR

Issue #1157, opened 2020-05-22
[Bug] Fantasizing is very slow on GPU due to torch.qr


danielrjiang:
# 🐛 Bug

torch.qr is causing things to be very slow on the GPU, when using fantasies. This is likely caused by https://github.com/pytorch/pytorch/issues/22573.

## To reproduce

See the following notebook example.
[KG_GPU_investigate.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/4665346/KG_GPU_investigate.ipynb.txt)

** Stack trace/error message **
See the following profile results.
<img width="1211" alt="CPU" src="https://user-images.githubusercontent.com/18407088/82617537-388de380-9b85-11ea-91c7-c5d86fe103b7.png">
<img width="1107" alt="GPU" src="https://user-images.githubusercontent.com/18407088/82617539-3b88d400-9b85-11ea-9ecc-c5171d8380b7.png">
[kg_cpu.cprofile.txt](https://github.com/cornellius-gp/gpytorch/files/4665343/kg_cpu.cprofile.txt)
[kg_gpu.cprofile.txt](https://github.com/cornellius-gp/gpytorch/files/4665344/kg_gpu.cprofile.txt)

## Expected Behavior

Running on CUDA should show speedups compared to CPU.

## System information

**Please complete the following information:**
- latest master
- 1.6.0a0
- Mac OS X Catalina 10.15.4

jacobrgardner:
Maybe we should put in some logic to use the Woodbury formula for small enough matrices? QR is better stability but likely slower for small matrices. @Balandat @gpleiss

Balandat:
Yeah that makes sense. Also, looking through the code above that section, we know that the `new_root` here is diagonal if we use cholesky for the root decomposition (i.e. if `L` is lower triangualr). So we don't have to do any work, we can just do a triangular solve against eye to the get the inverse. So we should probably have two code paths, one where `L` is diagonal and one where it's not. For the general case of tall non-triangular `L` we can probably do woodbury.

jacobrgardner:
Fixed for now by #1224

PR #1224, opened 2020-07-23
Perform QR composition in exact prediction strategy on the CPU

Balandat:
Fixes (temporarily) #1157, until pytorch/pytorch#22573 is addressed


Balandat:
This should use a size check, will update

PR #1231, opened 2020-07-30
Implement Spectral Kernel Learning with Deltas

jacobrgardner:
This implements a simple kernel that does spectral kernel learning with a simple mixture of point masses. It supports linear time and space for training and inference, and is pretty robust in practice. The method is very similar e.g. to the SSGP of Lazaro-Gredilla et al., 2010, but without the goal of recovering an existing kernel.


andrewgordonwilson:
The issue with this approach is that there are often many hyperparameters to learn for only marginal flexibility and risk of over-fitting. A la Carte builds on this approach to greatly reduce the parametrization, and improve the inductive biases, without losing flexibility: http://proceedings.mlr.press/v38/yang15b.pdf

jacobrgardner:
The approach does introduce a lot of parameters which can definitely require a large amount of training, but in the method's defense it's pretty scalable 5-10 years later now that GPUs let us scale to thousands of point masses / bases, since it's linear time/space in N and doesn't rely on somewhat harder to parallelize operations like the FWHT. 

It actually works surprisingly well when scaled up to huge numbers of point masses in the multitask setting where the deltas can be shared across tasks (see e.g. the example notebook). You get spectral densities that look more like "samples from a smooth distribution" than "a handful of delta distributions."

jacobrgardner:
Looks like someone a while back implemented a (possibly outdated) FWHT in PyTorch: https://discuss.pytorch.org/t/fast-walsh-hadamard-transform/19341

Including both a Fast food kernel and an implementation of a-la carte could definitely be interesting.

andrewgordonwilson:
Interesting! It could be quite practical to reduce the SSGP parametrization by assuming the point masses are drawn from a latent mixture of Gaussians, and then just estimating the parameters of the mixture. I wonder if similar results could be obtained with far fewer parameters in this case, especially if the point masses look like they are forming smooth densities.
This would be like A la Carte without Fastfood.

jacobrgardner:
Yeah, that's an interesting idea. In the most general case, we could have it switch over to fastfood at some very large setting when doing FWHTs might win out due to the asymptotics.

One interesting thing with large enough numbers of point masses is that optimization tends to not fall in to local optima with trivial extrapolations. We were thinking this is an artifact of being able to move individual point masses around instead of entire mixtures, and moving a single point mass (out of thousands) affects the spectral density little enough that you're less likely to be unable to make progress and get stuck.

I wonder if there's a happy medium of fewer parameters or some regularization encouraging smoother fits, but changing individual parameters still doesn't lead to as drastic a change in the spectral density.

PR #1234, opened 2020-07-31
add smoke tests to two notebooks where it was missing

martinjankowiak:
the only GPUs i currently have access to are in the cloud so someone will need to re-run the CIQ notebook and push to this branch in order to get reasonable performance timings...


jacobrgardner:
@martinjankowiak Haha -- I saw the timing results in the file changes before I saw your comment and it gave me a mini heart attack :-).

I'll rerun the notebook on a GPU and reupload it to this branch :+1:.

gpleiss:
huh my gpu timings seem weird. Cholesky is exceptionally slow... @jacobrgardner maybe take a look?

gpleiss:
Ah okay - it looks like the ill-conditioning makes Cholesky not work, so the samples have to be drawn with an eigendecomposition. Will merge after checks pass.

Balandat:
The reason it takes so long is, sadly, pytorch error handling: https://github.com/pytorch/pytorch/issues/34272

PR #1237, opened 2020-08-04
Make sure MVN variances are positive (raise warning if not)

gpleiss:
Ideally we'd be able to check than any MVN covariance is positive definite, but that's too expensive. In practice, negative posterior/variational variances are the more common issue. This ensures that all variances are at least 1e-6 (value determined by a context manager). It fixes any negative variance values, but raises a warning.

@Balandat @jacobrgardner thoughts?

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import math
+import warnings
 
 import torch
 from torch.distributions import MultivariateNormal as TMultivariateNormal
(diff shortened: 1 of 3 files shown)


gpytorch/distributions/multivariate_normal.py
```diff
@@ -201,15 +202,32 @@ def to_data_independent_dist(self):
...
-            return diag.expand(self._batch_shape + self._event_shape)
+            variance = diag.expand(self._batch_shape + self._event_shape)
         else:
-            return super().variance
+            variance = super().variance
+
+        # Check to make sure that variance isn't lower than minimum allowed value (default 1e-6).
+        # This ensures that all variances are positive
+        min_variance = settings.min_variance.value()
+        if variance.lt(min_variance):
```
Balandat, line 223: should be `torch.any(variance.lt(min_variance))`, no?

gpleiss: Oh yeah... good call.

gpleiss:
@Balandat agree about the dtype. I'm thinking about introducing this new setting context manager:

```python
class _dtype_value_context(object):
    _global_float_value = None
    _global_double_value = None
    _global_half_value = None

    @classmethod
    def value(cls):
        return cls._global_value

    @classmethod
    def _set_value(cls, float_value, double_value, half_value):
        if float_value is not None:
            cls._global_float_value = float_value
        if double_value is not None:
            cls._global_double_value = double_value
        if half_value is not None:
            cls._global_half_value = half_value

    def __init__(self, float=None, double=None, half=None):
        self._orig_float_value = self.__class__.value()
        self._instance_float_value = float
        self._orig_double_value = self.__class__.value()
        self._instance_double_value = double
        self._orig_half_value = self.__class__.value()
        self._instance_half_value = half

    def __enter__(self,):
        self.__class__._set_value(self._instance_float_value, self._instance_double_value, self._instance_half_value)

    def __exit__(self, *args):
        self.__class__._set_value(self._orig_float_value, self._orig_double_value, self._orig_half_value)
        return False
```

Thoughts?

gpleiss:
K this is finally ready for review.

jacobrgardner:
Yeah, this seems reasonable to me. I like the dtype context manager, and generally agree with the values you chose.

Issue #1247, opened 2020-08-11
[Bug] Incorrect gradient sizes when back-propagating through TLT - tensor product


Balandat:
# 🐛 Bug

An error in gradient shape incompatibility gets triggered when back-propagating through a TLT - tensor product. This was discovered in the context of sampling from a MVN posterior using base samples in https://github.com/pytorch/botorch/issues/513

This issue is caused by the new TriangularLazyTensor from #1102: The following code works fine pre-merge on 3e87f849c56c5ad018ac535052080b6649b244da, but fails on the merge commit 4e6f2d0b0988409f312cdfe97eff0754bc356d48, 

I haven't been able to dig much deeper, but the plot thickens.

## To reproduce

** Code snippet to reproduce **
```python
from gpytorch.lazy import BlockDiagLazyTensor, NonLazyTensor 

E = torch.stack([
    torch.eye(2, requires_grad=True),
    torch.eye(2),
])
bdlt = BlockDiagLazyTensor(NonLazyTensor(E))
clt = bdlt.root_decomposition()
root = clt.root

base_samples = torch.randn(4, 1)
out = root.matmul(base_samples)
out.sum().backward()
```

** Stack trace/error message **
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-74-0ac2634cc9e4> in <module>
     11 base_samples = torch.randn(4, 1)
     12 out = root.matmul(base_samples)
---> 13 out.sum().backward()

~/miniconda3/envs/botorch/lib/python3.7/site-packages/torch/tensor.py in backward(self, gradient, retain_graph, create_graph)
    183                 products. Defaults to ``False``.
    184         """
--> 185         torch.autograd.backward(self, gradient, retain_graph, create_graph)
    186 
    187     def register_hook(self, hook):

~/miniconda3/envs/botorch/lib/python3.7/site-packages/torch/autograd/__init__.py in backward(tensors, grad_tensors, retain_graph, create_graph, grad_variables)
    125     Variable._execution_engine.run_backward(
    126         tensors, grad_tensors, retain_graph, create_graph,
--> 127         allow_unreachable=True)  # allow_unreachable flag
    128 
    129 

RuntimeError: Function MatmulBackward returned an invalid gradient at index 1 - got [4, 4] but expected shape compatible with [2, 2, 2]
[...]

PR #1250, opened 2020-08-12
[WIP] Remove _quad_form_deriv methods from TriangularLT and RootLt

wjmaddox:
I'm not entirely sure what's going on here. Motivated by upstream Ax test failures caused by TriangularLT, https://github.com/pytorch/botorch/issues/513, and #1247.

Removing the explicit `_quad_form_derivative` method from `TriangularLazyTensor` seems to resolve those issues, but creates some test failures in `CholLazyTensor` that are resolved by removing the `_quad_form_derivative` method from `RootLazyTensor`.

### DIFF
--- a/gpytorch/lazy/root_lazy_tensor.py
+++ b/gpytorch/lazy/root_lazy_tensor.py
@@ -59,18 +59,6 @@ class RootLazyTensor(LazyTensor):
         # Matrix is symmetric
         return self._matmul(rhs)
 
-    def _quad_form_derivative(self, left_vecs, right_vecs):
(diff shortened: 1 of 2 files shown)


Balandat:
FWIW, this is the same issue that I ran in when working on that PR: https://github.com/cornellius-gp/gpytorch/pull/1102#issuecomment-614369747

@gpleiss "fixed" this in https://github.com/cornellius-gp/gpytorch/pull/1102#issuecomment-646944330, but there seems to be something weird going on in the representation tree.

PR #1274, opened 2020-09-02
Fix bad cache invalidation in variational strategies

Balandat:
This got overlooked in #1102. This fix is not the most elegant one, we really should use some kind of hook (as the code states). This only occurred when batch-evaluating with different batch sizes. That logic generally could be improved..


Balandat:
@mshvartsman

mshvartsman:
Thanks for the quick turnaround! Indeed fixes the `CachingError` I was seeing.

Issue #1292, opened 2020-09-22
[Bug] Bad type of elements of ZeroLazyTensor._args


PFMassiani:
# 🐛 Bug

The elements of the `_args` property of a `ZeroLazyTensor` have the wrong type: the current type is `int`, whereas it should be `torch.Tensor`. Currently, `_args` is the shape of the tensor, stored as a `tuple`.
The consequence is that the `representation` method of `ZeroLazyTensor` fails. In particular, it is not possible to evaluate a `CatLazyTensor` involving a `ZeroLazyTensor`.

## To reproduce

```python
import traceback
from gpytorch.lazy import ZeroLazyTensor, CatLazyTensor

lazy_tensor = ZeroLazyTensor(2, 2)
print(f'Type of lazy_tensor._args: {type(lazy_tensor._args[0])}')
print('---- `lazy_tensor.representation()` fail:')
try:
	representation = lazy_tensor.representation()
except RuntimeError:
	traceback.print_exc()
print('---- ZeroLazyTensor concatenation fail:')
try:
	CatLazyTensor(lazy_tensor, lazy_tensor).evaluate()
except RuntimeError:
	traceback.print_exc()
```

Stack trace:
```
Type of lazy_tensor._args: <class 'int'>
---- `lazy_tensor.representation()` fail:
Traceback (most recent call last):
  File "bug.py", line 9, in <module>
    representation = lazy_tensor.representation()
  File "<module>/gpytorch/gpytorch/lazy/lazy_tensor.py", line 1286, in representation
    raise RuntimeError("Representation of a LazyTensor should consist only of Tensors")
RuntimeError: Representation of a LazyTensor should consist only of Tensors
---- ZeroLazyTensor concatenation fail:
Traceback (most recent call last):
  File "bug.py", line 14, in <module>
    CatLazyTensor(lazy_tensor, lazy_tensor).evaluate()
  File "<module>/gpytorch/gpytorch/utils/memoize.py", line 59, in g
    return _add_to_cache(self, cache_name, method(self, *args, **kwargs), *args, kwargs_pkl=kwargs_pkl)
  File "<module>/gpytorch/gpytorch/lazy/lazy_tensor.py", line 906, in evaluate
    res = self.matmul(eye)
  File "<module>/gpytorch/gpytorch/lazy/lazy_tensor.py", line 1117, in matmul
    return func.apply(self.representation_tree(), other, *self.representation())
[...]

PR #1293, opened 2020-09-22
Fix type of elements of ZeroLazyTensor._args

PFMassiani:
Fixes https://github.com/cornellius-gp/gpytorch/issues/1292.
Adds a unittest to check that the representation of a `ZeroLazyTensor` can be computed.

The constructor of `ZeroLazyTensor` now accepts three types of parameters: some integers (same as before), a `torch.Tensor` object, or a `torch.Size` object.


jacobrgardner:
Looks like the main failure is just a minor issue in the new unit test checking a return type. `representation` returns a tuple of tensors, so should be a quick fix.

The giant pile of warnings in travis' run look to basically just be torch deciding to move some stuff under a new `torch.linalg`.

Issue #1294, opened 2020-09-22
[Bug] Matrix multiplication of rectangular ZeroLazyTensor


PFMassiani:
# 🐛 Bug

The matrix multiplication of `ZeroLazyTensor` does not allow rectangular tensors, while it is allowed at construction. More specifically, the methods `ZeroLazyTensor.matmul`, `ZeroLazyTensor._matmul` and `ZeroLazyTensor._t_matmul` are the problematic ones.
There are two problems (batch sizes are omitted):
- when multiplying a `ZeroLazyTensor` of shape `(m, n)` to a LazyTensor or PyTorch tensor of shape `(n, ?)`, the output is _systematically_ of shape `(n, ?)`, whereas it should be of shape `(m, ?)`
- when doing the transpose multiplication (method `_t_matmul`) of a `ZeroLazyTensor` of shape `(m, n)` to a PyTorch tensor of shape `(m, ?)`, the operation _fails_ unless `m == n`: the `ZeroLazyTensor` is not transposed.
This issue also prevents the concatenation of `ZeroLazyTensor` as described in https://github.com/cornellius-gp/gpytorch/issues/1292.

## To reproduce

```python
import traceback
import torch
from gpytorch.lazy import ZeroLazyTensor

def check_equality(product, actual):
        if product.shape != actual.shape:
                print(f'Size mismatch: {product.shape} != {actual.shape}')
        elif torch.max((product - actual).abs()) > 1e-4:
                print(f'Value mismatch: {product}\n!=\n{actual}')
        else:
                print('Tensors are equal.')

print('------- TEST 1')
zero = ZeroLazyTensor(5, 4, 3)
lazy_square = ZeroLazyTensor(5, 3, 3)
actual = torch.zeros(5, 4, 3)
product = zero.matmul(lazy_square)
check_equality(product.evaluate(), actual)

print('------- TEST 2')
tensor_square = torch.eye(3).repeat(5, 1, 1)
product = zero._matmul(tensor_square)
check_equality(product, actual)

print('------- TEST 3')
tensor_square = torch.eye(4).repeat(5, 1, 1)
actual = torch.zeros(5, 3, 4)
try:
        product = zero._t_matmul(tensor_square)
        check_equality(product, actual)
except RuntimeError:
        print(f'Fatal RuntimeError! The ZeroLazyTensor is not transposed before multiplication!')
        traceback.print_exc()
```

Output:
```
------- TEST 1
[...]

PR #1295, opened 2020-09-22
Fix matrix multiplication of rectangular ZeroLazyTensor

PFMassiani:
Fixes https://github.com/cornellius-gp/gpytorch/issues/1294.
The code given in the bug report to reproduce the issue is also added as a unit test.

I bring your attention to the returns of `ZeroLazyTensor._matmul` and 'ZeroLazyTensor._t_matmul': a new `torch.tensor` is created, and therefore all the gradient information that may be present in the parameter `rhs` is lost. I do not think this is a problem (backpropagation throught a `ZeroLazyTensor` does not make a lot of sense), but this should definitely be double-checked.


jacobrgardner:
I'll merge this one in after the previous PR is merged and we rebase this on top of that. I think the backpropagation is a non-issue.

**Technically**, multiplying a tensor that requires grad by 0 results in a tensor with a `grad_fn`:

```python
In [1]: import torch

In [2]: a = torch.randn(5).requires_grad_(True)

In [3]: b = 0 * a

In [4]: b
Out[4]: tensor([-0., 0., -0., 0., 0.], grad_fn=<MulBackward0>)
```
This could result in insidious bugs where we rely on simply checking whether a tensor requires grad or not to clear a cache or something?

gpytorch/lazy/zero_lazy_tensor.py
```diff
@@ -43,7 +43,14 @@ def _matmul(self, rhs):
...
             raise RuntimeError("Size mismatch, self: {}, rhs: {}".format(self.size(), rhs.size()))
-        return rhs * 0
+        new_m = self.size(-2)
+        if rhs_size_ind == -1:
+            *batch_shape, m = rhs.shape
+            output_shape = (*batch_shape, new_m)
+        else:
+            *batch_shape, m, n = rhs.shape
+            output_shape = (*batch_shape, new_m, n)
+        return torch.zeros(*output_shape)
```
jacobrgardner, line 53:
```suggestion
        return torch.zeros(*output_shape, device=rhs.device, dtype=rhs.dtype)
```
At a minimum, these should inherit

gpytorch/lazy/zero_lazy_tensor.py
```diff
@@ -72,9 +79,16 @@ def _sum_batch(self, dim):
...
             raise RuntimeError("Size mismatch, self: {}, rhs: {}".format(self.size(), rhs.size()))
-        return rhs * 0
+        new_m = self.size(-1)
+        if rhs_size_ind == -1:
+            *batch_shape, m = rhs.shape
+            output_shape = (*batch_shape, new_m)
+        else:
+            *batch_shape, m, n = rhs.shape
+            output_shape = (*batch_shape, new_m, n)
+        return torch.zeros(*output_shape)
```
jacobrgardner, line 91:
```suggestion
        return torch.zeros(*output_shape, dtype=rhs.dtype, device=rhs.device)
```
ditto here

PFMassiani:
True: the device and dtype should be preserved now. I think you forgot to point it in the standard `matmul` method: this is also fixed.

About the gradient: the `requires_grad` property of ZeroLazyTensors seems to be always False. Therefore, if this variable is correctly checked before using gradient information, it should not be a problem?
```python
from gpytorch.lazy import ZeroLazyTensor
lz = ZeroLazyTensor(5, 4, 3)
print(f'Requires grad: {lz.requires_grad}')
lz.requires_grad_(True)
print(f'Requires grad: {lz.requires_grad}')
```
Output:
```
Requires grad: False
Requires grad: False
```

jacobrgardner:
Not sure why the stable build failed the first time -- the error message made no sense. Hopefully it will pass this time.

gpleiss (approved):

PR #1312, opened 2020-10-15
Fix shape issue for MMVN with broadcasted means

gpleiss:
If I have a MMVN with `mean.shape = [1 x d]` and `cover.shape = `[nd x nd]`, I would expect the singleton dimension in the mean to broadcast. Similarly for `mean.shape = [n x 1]` and `cover.shape = `[nd x nd]`.

This PR fixes a small bug and allows for this broadcasting.

### DIFF
--- a/gpytorch/distributions/multitask_multivariate_normal.py
+++ b/gpytorch/distributions/multitask_multivariate_normal.py
@@ -3,6 +3,7 @@
 import torch
 
 from ..lazy import BlockDiagLazyTensor, BlockInterleavedLazyTensor, CatLazyTensor, LazyTensor, lazify
+from ..utils.broadcasting import _mul_broadcast_shape
 from .multivariate_normal import MultivariateNormal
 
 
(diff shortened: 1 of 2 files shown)


gpytorch/distributions/multitask_multivariate_normal.py
```diff
@@ -33,6 +34,19 @@ def __init__(self, mean, covariance_matrix, validate_args=False, interleaved=Tru
         if mean.dim() < 2:
             raise RuntimeError("mean should be a matrix or a batch matrix (batch mode)")
 
+        batch_shape = _mul_broadcast_shape(mean.shape[:-2], covariance_matrix.shape[:-2])
+        if mean.shape[-2:].numel() != covariance_matrix.size(-1):
+            if mean.size(-2) == 1:
+                mean = mean.expand(*batch_shape, covariance_matrix.size(-1) // mean.size(-1), mean.size(-1))
```
Balandat, line 40: Should we for good measure add a check that `covariance_matrix.size(-1)` is a multiple of `mean.size(-1)`? Otherwise this could result in some hard-to-debug errors...

gpleiss: Yeah that makes sense. I'll update it

gpleiss: Updated

Balandat (approved):
should we add a unit test to verify this broadcasting behavior?

PR #1314, opened 2020-10-16
Leave one out cross-validation 

dme65:
This is an implementation of the leave one out cross-validation (pseudo likelihood), from RW 5.4.2.  The objective is to maximize the sum of the leave-one-out log probabilities, which we can do in `O(n^3)` via a bordered system formulation. I explicitly use Cholesky as I think this is mostly interesting in the small-data regime, where incorrect model assumptions may influence the resulting model a lot. 

The two computationally expensive pieces are:
1) Computing `diag(inv(K))`, which we can do `O(n^3)` via Cholesky.  We can potentially do diagonal estimation, see, e.g., https://www-users.cs.umn.edu/~saad/PDF/umsi-2005-082.pdf, but it requires working with inv(K) rather than K, so we probably need a pretty large number of points to beat Cholesky as we have a CG step in the inner loop.
2) Solve `K x = target - mean`, which is trivial with either CG or Cholesky.

I've verified that the los function works in both the `Simple_GP_regression.ipynb` and `Simple_Batch_Mode_GP_Regression.ipynb`, so hopefully I got the batching right. There are no specific tests as this is an isolated piece, but I can add some if you want.

I probably need some pointers and thoughts on what to do with the `pyro_factor`, `named_priors`, `added_loss_terms`, that are being added to the loss in `exact_marginal_log_likelihood.py`.

### DIFF
--- a/gpytorch/mlls/exact_marginal_log_likelihood.py
+++ b/gpytorch/mlls/exact_marginal_log_likelihood.py
@@ -60,7 +60,7 @@ class ExactMarginalLogLikelihood(MarginalLogLikelihood):
         # Get the log prob of the marginal distribution
         output = self.likelihood(function_dist, *params)
         res = output.log_prob(target)
-        self._add_other_terms(res, params)
+        res = self._add_other_terms(res, params)
 
         # Scale by the amount of data we have
         num_data = target.size(-1)
--- a/gpytorch/mlls/exact_marginal_log_likelihood.py
+++ b/gpytorch/mlls/exact_marginal_log_likelihood.py
@@ -33,7 +33,7 @@ class ExactMarginalLogLikelihood(MarginalLogLikelihood):
             raise RuntimeError("Likelihood must be Gaussian for exact inference")
         super(ExactMarginalLogLikelihood, self).__init__(likelihood, model)
 
-    def _add_other_terms(self, res):
+    def _add_other_terms(self, res, params):
         # Add additional terms (SGPR / learned inducing points, heteroskedastic likelihood models)
         for added_loss_term in self.model.added_loss_terms():
             res = res.add(added_loss_term.loss(*params))
@@ -60,7 +60,7 @@ class ExactMarginalLogLikelihood(MarginalLogLikelihood):
         # Get the log prob of the marginal distribution
         output = self.likelihood(function_dist, *params)
         res = output.log_prob(target)
-        self._add_other_terms(res)
+        self._add_other_terms(res, params)
 
         # Scale by the amount of data we have
         num_data = target.size(-1)
--- a/gpytorch/mlls/exact_marginal_log_likelihood.py
+++ b/gpytorch/mlls/exact_marginal_log_likelihood.py
@@ -33,6 +33,17 @@ class ExactMarginalLogLikelihood(MarginalLogLikelihood):
             raise RuntimeError("Likelihood must be Gaussian for exact inference")
         super(ExactMarginalLogLikelihood, self).__init__(likelihood, model)
 
+    def _add_other_terms(self, res):
+        # Add additional terms (SGPR / learned inducing points, heteroskedastic likelihood models)
+        for added_loss_term in self.model.added_loss_terms():
+            res = res.add(added_loss_term.loss(*params))
+
+        # Add log probs of priors on the (functions of) parameters
+        for _, prior, closure, _ in self.named_priors():
+            res.add_(prior.log_prob(closure()).sum())
+
+        return res
+
     def forward(self, function_dist, target, *params):
         r"""
         Computes the MLL given :math:`p(\mathbf f)` and :math:`\mathbf y`.
@@ -49,14 +60,7 @@ class ExactMarginalLogLikelihood(MarginalLogLikelihood):
         # Get the log prob of the marginal distribution
         output = self.likelihood(function_dist, *params)
         res = output.log_prob(target)
-
-        # Add additional terms (SGPR / learned inducing points, heteroskedastic likelihood models)
-        for added_loss_term in self.model.added_loss_terms():
-            res = res.add(added_loss_term.loss(*params))
-
-        # Add log probs of priors on the (functions of) parameters
-        for _, prior, closure, _ in self.named_priors():
-            res.add_(prior.log_prob(closure()).sum())
+        self._add_other_terms(res)
 
         # Scale by the amount of data we have
         num_data = target.size(-1)
--- a/gpytorch/mlls/leave_one_out_likelihood.py
+++ b/gpytorch/mlls/leave_one_out_likelihood.py
@@ -1,0 +1,66 @@
+import math
+
+import torch
+from torch import Tensor
+
+from ..distributions import MultivariateNormal
+from .exact_marginal_log_likelihood import ExactMarginalLogLikelihood
+
+
+class LeaveOneOutLikelihood(ExactMarginalLogLikelihood):
+    """
+    The leave one out cross-validation (LOO-CV) likelihood from RW 5.4.2 for an exact Gaussian process with a
+    Gaussian likelihood. This offers an alternative to the exact marginal log likelihood where we
+    instead maximize the sum of the leave one out log probabilities :math:`\log p(y_i | X, y_{-i}, \theta)`.
+
+    Naively, this will be O(n^4) with Cholesky as we need to compute `n` Cholesky factorizations. Fortunately,
+    given the Cholesky factorization of the full kernel matrix (without any points removed), we can compute
+    both the mean and variance of each removed point via a bordered system formulation making the total
+    complexity O(n^3).
+
+    The LOO-CV approach can be more robust against model mis-specification as it gives an estimate for the
+    (log) predictive probability, whether or not the assumptions of the model is fulfilled.
+
+    .. note::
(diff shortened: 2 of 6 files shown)


gpytorch/mlls/leave_one_out_likelihood.py
```diff
@@ -0,0 +1,66 @@
+import math
```
Balandat, line 1:
```suggestion
#!/usr/bin/env python3
import math
```

gpytorch/mlls/leave_one_out_likelihood.py
```diff
@@ -0,0 +1,66 @@
...
+
+import torch
+from torch import Tensor
+
+from ..distributions import MultivariateNormal
+from .exact_marginal_log_likelihood import ExactMarginalLogLikelihood
+
+
+class LeaveOneOutLikelihood(ExactMarginalLogLikelihood):
```
Balandat, line 10: could call this `LeaveOneOuPseudoLikelihood` if we wanted to be precise

gpytorch/mlls/leave_one_out_likelihood.py
```diff
@@ -0,0 +1,66 @@
...
+        >>> loocv = gpytorch.mlls.LeaveOneOutLikelihood(likelihood, model)
+        >>>
+        >>> output = model(train_x)
+        >>> loss = -loocv(output, train_y)
+        >>> loss.backward()
+    """
+
+    def __init__(self, likelihood, model):
+        super().__init__(likelihood=likelihood, model=model)
+        self.likelihood = likelihood
```
Balandat, line 44: should probably raise an error here if a non-gaussian LH is passed in

gpytorch/mlls/leave_one_out_likelihood.py
```diff
@@ -0,0 +1,66 @@
...
+        m, K = output.mean, output.covariance_matrix
+        m = m.reshape(*target.shape)
+        L = torch.cholesky(K, upper=False)
+        I = torch.eye(*K.shape[-2:], dtype=K.dtype, device=K.device)
+        sigma2 = 1.0 / torch.cholesky_solve(I, L, upper=False).diagonal(dim1=-1, dim2=-2)  # 1 / diag(inv(K))
+        mu = target - torch.cholesky_solve((target - m).unsqueeze(-1), L, upper=False).squeeze(-1) * sigma2
+        term1 = -0.5 * torch.log(sigma2)
+        term2 = -0.5 * (target - mu).pow(2.0) / sigma2
+        log_loocv = term1 + term2 - 0.5 * math.log(2 * math.pi)
+        return log_loocv.sum(dim=-1)
```
Balandat, line 66: can you add support for added loss terms and priors (https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/mlls/exact_marginal_log_likelihood.py#L53-L59) as well here?

dme65: Sure! Just adding them to log_loocv makes sense, right?

gpytorch/mlls/leave_one_out_likelihood.py
dme65: That should happen in the `super` call since we inherit from `ExactMarginalLogLikelihood`.

gpytorch/mlls/leave_one_out_likelihood.py
dme65: Good suggestion, will do!

gpytorch/mlls/leave_one_out_likelihood.py
Balandat: yeah, just as in `ExactMarginalLogLikelihood`.

Probably a good idea to move that logic in to a separate `_add_other_loss_terms` instance method that we can then call from both `ExactMarginalLogLikelihood` and `LeaveOneOutPseudoLikelihood` without having to copy & paste that code into the both forward methods.

dme65: Added!

gpytorch/mlls/leave_one_out_pseudo_likelihood.py
```diff
@@ -0,0 +1,73 @@
...
+        r"""
+        Computes the leave one out likelihood given :math:`p(\mathbf f)` and `\mathbf y`
+
+        :param ~gpytorch.distributions.MultivariateNormal output: the outputs of the latent function
+            (the :obj:`~gpytorch.models.GP`)
+        :param torch.Tensor target: :math:`\mathbf y` The target values
+        :param dict kwargs: Additional arguments to pass to the likelihood's :attr:`forward` function.
+        """
+        output = self.likelihood(function_dist, *params)
+        m, K = output.mean, output.covariance_matrix
```
Balandat, line 58: maybe add a comment here why using cholesky on the dense matrix is fine

Balandat (approved):
I was going to say let's add a test like for `ExactMarginaLogLikelihood`, but it looks like we don't have one for that ... Should probably add for both. 

Other than that this lgtm.

gpytorch/mlls/leave_one_out_pseudo_likelihood.py
jacobrgardner: Might still be better here to rely on `output.lazy_covariance_matrix.cholesky()`. It'll do the same thing in most cases, but in some edge cases may be more efficient than explicitly evaluating still.

dme65: I changed to `output.lazy_covariance_matrix.cholesky()`. Is `_cholesky_solve` what I want to use to do a solve with the Cholesky of K?

Balandat: yes

dme65: @Balandat: Added three tests: (1) + (2) checks that the code runs with and without batch mode and (3) compares the O(n^3) LOOCV with the naive O(n^4) approach as a sanity check.

gpytorch/mlls/exact_marginal_log_likelihood.py
```diff
@@ -49,14 +60,7 @@ def forward(self, function_dist, target, *params):
...
         res = output.log_prob(target)
-
-        # Add additional terms (SGPR / learned inducing points, heteroskedastic likelihood models)
-        for added_loss_term in self.model.added_loss_terms():
-            res = res.add(added_loss_term.loss(*params))
-
-        # Add log probs of priors on the (functions of) parameters
-        for _, prior, closure, _ in self.named_priors():
-            res.add_(prior.log_prob(closure()).sum())
+        self._add_other_terms(res, params)
```
Balandat, line 63: I don't think this actually does anything (it returns the value but that return value is never assigned, this doesn't modify things in-place). 

Can awe instead make a `_compute_other_loss_terms` function that only takes in `params`, returns the additional value, and then add that explicitly in the forward method? That way things are less confusing.

Balandat (changes requested):

gpytorch/mlls/exact_marginal_log_likelihood.py
dme65: I meant to do the same as in `leave_one_out_pseudo_likelihood` where the returned value is assigned to `res`. Thanks for catching that. I have fixed that part.

Sure, we can do that, but what is the cleanest way for me to get the `dtype`, `device`, and shapes in that case since I'm not passing in res anymore? I can add three additional arguments to `_compute_other_loss_terms`, but maybe there is a better way of doing it.

Balandat: Hmm do you need that? I don't see these being used anywhere. Is there some implicit casting going on in `_add_other_terms` that you couldn't rely on otherwise? You could also have `_compute_other_loss_terms` be a generator that yields the individual terms, and then do `res = res + sum(self._compute_other_loss_terms(*params))` in the `forward).

dme65: Is the way these losses are added up in the original code correct? If you have a batch-GP, then shouldn't the losses be added on independently for each model? The original code does stuff like this: `res.add_(prior.log_prob(closure()).sum())`, which would sum up the losses from all priors and then add them to each batch dimension, counting the losses multiple times when you eventually call `sum()` on `res` to do a backward pass?

Before changing this I need to understand what the shapes of these additional losses are. They will all have to be identical for us to call sum on them. Right now, they are all 0-dimensional tensors and it will be pretty easy to just sum them up, but it depends on if things are currently being handled correctly.

Balandat: Hmm yeah good point. I haven't been using batched priors with different values for different batch elements (though the code should allow for it). Same thing for the added loss term. Could this be generalized in a straightforward way if we just do `sum(dim=-1)` instead?

dme65: That causes a lot of other tests to fail. As an example, the `test_train_on_batch_test_on_batch` test has four priors with sizes: `[2, 2]`, `[2]`, `[]`, [2, 1, 1]`, which obviously don't broadcast.

It's probably out of scope to fix this in this PR and it probably makes more sense to leave things as is and open a separate issue to track down how these additional losses are added up. What do you think?

Balandat (approved):

PR #1319, opened 2020-10-21
Bypass keops for small/rectangular kernels

gpleiss:
For CIQ BayesOpt sampling, it makes sense to use Cholesky to compute the GP marginal log likelihood and the posterior. Unfortunately, this causes multiple KeOps compilation steps - each time Cholesky is called on a new matrix with a different size, or when the `k_Xx` term changes size.

This PR bypasses KeOps when a KeOps kernel is called with a small or rectangular kernel. Here, a small/rectangular kernel is one where `n` or `m` is smaller than `gpytorch.settings.max_cholesky_size`.

For the BayesOpt experiment in [this repo](https://github.com/gpleiss/ciq_experiments/tree/main/bayesopt/Hartmann6.ipynb) - KeOps is now only used to perform matmuls with (large) prior covariance matrix when performing Thompson sampling.

(We could also include a different context manager to control the KeOps bypass, but I'm trying to avoid adding new context managers to keep complexity down).

cc/ @dme65 @KeAWang

### DIFF
--- a/gpytorch/kernels/keops/matern_kernel.py
+++ b/gpytorch/kernels/keops/matern_kernel.py
@@ -27,21 +27,33 @@ try:
             super(MaternKernel, self).__init__(**kwargs)
             self.nu = nu
 
+        def _nonkeops_covar_func(self, x1, x2, diag=False):
+            distance = self.covar_dist(x1, x2, diag=diag)
+            exp_component = torch.exp(-math.sqrt(self.nu * 2) * distance)
+
+            if self.nu == 0.5:
+                constant_component = 1
+            elif self.nu == 1.5:
+                constant_component = (math.sqrt(3) * distance).add(1)
+            elif self.nu == 2.5:
+                constant_component = (math.sqrt(5) * distance).add(1).add(5.0 / 3.0 * distance ** 2)
+            return constant_component * exp_component
+
         def covar_func(self, x1, x2, diag=False):
+            # We only should use KeOps on big kernel matrices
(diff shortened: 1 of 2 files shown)


gpytorch/kernels/keops/matern_kernel.py
```diff
@@ -43,8 +44,17 @@ def covar_func(self, x1, x2, diag=False):
...
                 with torch.autograd.enable_grad():
-                    x1_ = KEOLazyTensor(x1[..., :, None, :])
-                    x2_ = KEOLazyTensor(x2[..., None, :, :])
+                    # We only should use KeOps on big kernel matrices
+                    # If we would otherwise be performing Cholesky inference, then don't apply KeOps
+                    if (
+                        x1.size(-2) < settings.max_cholesky_size.value()
+                        or x2.size(-2) < settings.max_cholesky_size.value()
+                    ):
+                        x1_ = x1[..., :, None, :]
```
jacobrgardner, line 53: I know that N is gonna be inherently smallish here, but I think we should avoid materializing an n x n x d tensor here when we get to the distance calculation below just because it's easy enough to avoid. Let's move the distance calculation inside and use covar_dist / the RBF Function as we do in RBFKernel.

gpleiss: Cool. Updated the files, and checked that they both work in the CIQ sampling example.

jacobrgardner (approved):

Issue #1331, opened 2020-11-04
Don't Use Stochastic Trace Estimation for KroneckerProductAddedDiagLT logdets


wjmaddox:
Quit using stochastic trace estimation to estimate backwards passes through `KroneckerProductAddedDiagLazyTensor` log determinants. This change will remove the logdet backwards setup call in https://github.com/cornellius-gp/gpytorch/blob/8f9b44fc57dbb0a13b568946f07a37e9332f92c4/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py#L32 in order to improve memory efficiency and instead use pytorch's native backwards pass through symeig instead.

```
import torch

from gpytorch.lazy import KroneckerProductLazyTensor, KroneckerProductAddedDiagLazyTensor
from gpytorch.kernels import MaternKernel

xlist = [torch.randn(x, 1).cuda() for x in [100, 200, 210, 5]]
kernel_list = [MaternKernel().cuda()] * 4

eval_kernel = KroneckerProductLazyTensor(*[k(x) for k, x in zip(kernel_list, xlist)])

eval_kernel_with_noise = eval_kernel.add_jitter(0.1)

# is extremely memory intensive and OOMs on many GPUs.
solve, ld = eval_kernel_with_noise.inv_quad_logdet(torch.randn(100 * 200 * 210 * 5).cuda(), logdet=True)
```

This OOMs because internally we use 10 samples for the trace estimation part of the log determinant rather than backpropagating through the output of symeig (which would mean multiplication with slightly smaller matrices).

Stack tracing in the `matmul` here gives this output
```
> /..../gpytorch/gpytorch/lazy/kronecker_product_lazy_tensor.py(192)_matmul()
    190             rhs = rhs.unsqueeze(-1)
    191 
--> 192         res = _matmul(self.lazy_tensors, self.shape, rhs.contiguous())
    193 
    194         if is_vec:

ipdb> rhs.shape
torch.Size([21000000, 11])
```

I know that stability with backpropagation through the eigenvalues is going to be an issue for some kernels (especially RBF), but we could maybe implement something to fall back onto if the backwards pass nans out or something to resolve that issue?
Looking at the math further, I could alternatively implement something non-stochastic for the backwards pass given that the derivative is pretty clear --- this approach could also allow for varying types of diagonalization in the solves (e.g. one component is large enough that we would use Lanczos on while we use symeig on the others).

gpleiss:
[...]

PR #1332, opened 2020-11-04
Change KroneckerProductAddedDiagLazyTensor log det to pytorch logdet

wjmaddox:
See #1331 for more details but this changes `KroneckerProductAddedDiagLazyTensor` to use pytorch's symeig backwards through the logdeterminant rather than gpytorch's stochastic trace estimation.

I don't think that there's a better structure exploiting version of this right now than what's in here, but am open to discussions about it.

### DIFF
--- a/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
@@ -31,7 +31,7 @@ class KroneckerProductAddedDiagLazyTensor(AddedDiagLazyTensor):
             inv_quad_rhs=inv_quad_rhs, logdet=False, reduce_inv_quad=reduce_inv_quad
         )
 
-        if logdet is not None:
+        if logdet is not False:
             logdet_term = self._logdet()
         else:
             logdet_term = None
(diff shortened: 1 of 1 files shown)


gpleiss:
@wjmaddox can you get some rough profiling numbers for how much this affects wall-clock time? I'm for merging this PR as is (once the issue on master is sorted out), but it might also be worth considering using the eigendecomposition to also compute the `inv_quad` term rather than relying on CG. (This would reduce the overall computation.)

wjmaddox:
image

In the setting I'm looking at here, it seems slightly faster actually, although this might be the best case scenario for symeig. Here's the code I'm using to roughly test:
```
def construct_kp_and_compute_logdet(x1_size):
    xlist = [torch.randn(x, 1).to(device) for x in [x1_size, 200, 210, 5]]
    kernel_list = [MaternKernel().to(device)] * 4
    
    eval_kernel = KroneckerProductLazyTensor(*[k(x) for k, x in zip(kernel_list, xlist)])

    eval_kernel_with_noise = eval_kernel.add_jitter(0.1)

    # is extremely memory intensive and OOMs on many GPUs.
    solve, ld = eval_kernel_with_noise.inv_quad_logdet(torch.randn(x1_size * 200 * 210 * 5).to(device), logdet=True)
    ld.backward()

nn_list = [5, 10, 20, 30, 50, 70, 100, 200, 400, 500]
timing_list = []
for nn in nn_list:
    print("now trying", nn)
    try:
        start = time.time()
        construct_kp_and_compute_logdet(nn)
        end = time.time()
        timing_list.append(end - start)
    except:
        print(nn, "failed")
        timing_list.append(float('nan'))
```

I'm using torch 1.6.0 and a Titan RTX gpu for these comparisons.

wjmaddox:
Looks like there was something flaky on pytorch master when I committed originally because the tests are passing now and I was unable to reproduce the issues via building a new conda environment.

gpleiss (approved):

Issue #907, opened 2019-10-21
Saving kernels (pickle)


Galto2000:
Howdy folks, 

I am aware that one can save the state_dict using torch.save(). But since I am experimenting with different kernel structures I thought it would be nice to keep track of them by putting them in a simple data base like `shelve` and then be able to recall them when I decided I wanted to train (new) data on it.

So I was thinking of something simple like this:

```
import shelve

class KernelRepo:
    def __init__(self, shelve_db_filename="default_kernel_shelve.db"):
        self._db = shelve_db_filename


    def store_kernel(self, key, kernel):
        assert(isinstance(key,str))
        s = shelve.open(self._db)
        try:
            s[key] = kernel
        finally:
            s.close()


    def get_kernel(self, key):
        assert(isinstance(key,str))
        kernel = None
        s = shelve.open(self._db, flag='r')
        try:
            kernel = s[key]
        finally:
            s.close()

        return kernel
```

And then when I have a kernel that I would like to store, I would do something like this:
```

from prototype.kernel_repository import KernelRepo
import gpytorch

default_kernel = gpytorch.kernels.RBFKernel()*gpytorch.kernels.PolynomialKernel(power=2) + \
                            gpytorch.kernels.ScaleKernel(gpytorch.kernels.LinearKernel())

kernel_repo = KernelRepo()
kernel_repo.store_kernel("default kernel", default_kernel)
```

Unfortunately I am running in the following issue. 

`_pickle.PicklingError: Can't pickle <built-in function softplus>: import of module 'torch._C._nn' failed`

I am relatively new to Python so I am still trying to work out what the root cause of this issue is, but perhaps someone has already run into something similar, or has a better solution for "managing kernels".

My ultimate goal would be to automatically create key-tags so that I can keep track of different kernels, and automate the kernel selection process by using something like an ABC-SMC method.

But for now I would be really happy with being able to store data kernels so I can recall them programmatically using something like `shelve` or something else.

Any ideas, suggestions or solutions?

Thanks in advance

Galto

david-waterworth:
Maybe try the library 'dill' (or maybe cloudpickle) in place of pickle? I'm not sure the exact problem, but pickle cannot pickle lambda (anonymous) functions which may be the problem you're seeing.

Galto2000:
Thanks for the suggestion , I'll take a look at dill and report back soon

albertcthomas:
[...]

PR #1336, opened 2020-11-09
Make kernels pickleable

PhilippThoelke:
This solves issue #907 by using an instance of the module `torch.nn.Softplus` instead of `torch.nn.functional.softplus` because the functional version can't be pickled. Also added a test case for this.

### DIFF
--- a/gpytorch/constraints/constraints.py
+++ b/gpytorch/constraints/constraints.py
@@ -9,7 +9,6 @@ from torch.nn import Module
 from .. import settings
 from ..utils.transforms import _get_inv_param_transform, inv_sigmoid, inv_softplus
 
-
 # define softplus here instead of using torch.nn.functional.softplus because the functional version can't be pickled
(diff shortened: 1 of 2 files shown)


PhilippThoelke:
Checks are not passing because isort wants
```python
 from ..utils.transforms import _get_inv_param_transform, inv_sigmoid, inv_softplus
```
to be
```python
from ..utils.transforms import (_get_inv_param_transform, inv_sigmoid,
                                inv_softplus)
```
I'm unsure if I should fix this because it's not in my code.

jacobrgardner (approved):
This is great, thanks!

Definitely feel free to fix the import--no idea why that sometimes passes on master and sometimes doesn't.

PhilippThoelke:
I am also quite confused by isort now, locally it still wants to change this one import line but the pre-commit call in the CI passes for some reason. I hope it works now...

PR #1338, opened 2020-11-09
add lazy evals option to kplt symeig + root

wjmaddox:
The current "symeig" root decomposition for `KroneckerProductLazyTensor` uses a Hadamard product of a KPLT and a dense vector https://github.com/cornellius-gp/gpytorch/blob/8f9b44fc57dbb0a13b568946f07a37e9332f92c4/gpytorch/lazy/lazy_tensor.py#L1366, which is quite memory intensive for large kronecker products. This PR adds in a lazy option for returning the eigenvalues (maybe this should be done throughout the LT framework instead of returning solely the vector?), which allows writing the KPLT root decomposition as a Kronecker product of the product of the eigenvectors and eigenvalues of the components of the KP.

Example usage that should now be possible:

```python

kernels = [MaternKernel()] * 4
xlist = [torch.randn(x, 1) for x in [10, 30, 50, 100]]

kplt = KroneckerProductLazyTensor(*[kernel(x) for kernel, x in zip(kernels, xlist)])

kplt.root_decomposition(method = "symeig")
```

### DIFF
--- a/gpytorch/lazy/kronecker_product_lazy_tensor.py
+++ b/gpytorch/lazy/kronecker_product_lazy_tensor.py
@@ -195,6 +195,22 @@ class KroneckerProductLazyTensor(LazyTensor):
             res = res.squeeze(-1)
         return res
 
+    @cached(name="root_decomposition")
+    def root_decomposition(self, method: Optional[str] = None):
+        from gpytorch.lazy import RootLazyTensor
+
+        if method == "symeig":
+            evals, evecs = self._symeig(eigenvectors=True, return_evals_as_lazy=True)
+            # TODO: only use non-zero evals (req. dealing w/ batches...)
+            f_list = [
+                evec * eval.diag().clamp(0.0).sqrt().unsqueeze(-2)
+                for eval, evec in zip(evals.lazy_tensors, evecs.lazy_tensors)
+            ]
+            F = KroneckerProductLazyTensor(*f_list)
+            return RootLazyTensor(F)
+        else:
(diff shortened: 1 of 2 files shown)


Balandat (approved):
Thanks, this is quite useful. I am usually not a big fan of boolean flags that change what a functions returns, but I guess this is not user-facing and so should be ok. Maybe we can add some more comment in the code describing what this option is for, when it should / shouldn't be used. 

I think doing this via the `LazyTensors` would be the "proper" approach, but that would really be a `LazyVector` and it seems that would probably violate a lot of gpytorch's implicit assumptions about shapes of lazies...

wjmaddox:
Sure, I'll add a comment to that extent. Well, it's not so much a `LazyVector` so much as a `DiagLazyTensor` (or in this case a `KroneckerProductLazyTensor(DiagLazyTensors(....))`. However, `torch.symeig` doesn't return a diagonal matrix for eigenvalues, so I'm not entirely sure if we want our `symeig` to return different things than `torch.symeig` does.

Balandat:
Yeah I mean it makes sense to return the evals as a vector rather than a diagonal matrix, and we should be consistent with torch on that. So if we had a `LazyVector` (that internally could just wrap the `DiagLazyTensor`) then we would achieve that. But it seems a little premature to introduce this concept unless there are more use cases of that kind.

wjmaddox:
Yeah, none of the default options are changed here currently. It's just an option that's added.

Issue #1346, opened 2020-11-10
[Feature Request] Know whether feature flags have been set or are the default


Balandat:
# 🚀 Feature Request

Right now, the feature flags (such as `fast_pred_var`) are either `True` or `False`, but when checking the setting there is no way of knowing whether the setting is from the gpytorch default value or was set manually. It would be nice to have this ability.

## Motivation

Knowing whether a setting is the default or explicitly selected by the user would allow more elegant usage of these settings in code using gpytorch. For instance, in BoTorch we set some feature flags to default values that are different from those in gpytorch.

But we also want to be able to retain the ability to overwrite these flags for particular non-standard models. Since currently there is no way of knowing whether a gpytoch setting has been set manually or comes from the default, we essentially have to wrap every single gpytorch flag in our own container flag.

If we knew that the setting came from the default value, we could chose to override it depending on that, which would allow using gpytorch context managers with BoTorch models in a transparent way. 

## Pitch

One simple way of doing this would be to have the `_state` class property set to `None` by default, and define a simple `_default` property for each flag that induces the desired default behavior (this would also make it clear in code that the state is the default setting in code):

```
class _feature_flag(object):
    _default 
    _state = None

    @property
    def state(self):
        if self.is_default:
            return self._default
        return self._state

    @property
    def is_default(self):
        return self._state is None
```

A concrete flag would then look like
```
class _fast_solves(_feature_flag):
    _default = True
```

PR #1347, opened 2020-11-11
Add is_default method to feature flags.

Balandat:
Implements #1346


Balandat:
We could do a similar thing for the value contexts, but it would have to work a bit differently (since `None` is a valid value to be set).

gpleiss (approved):

gpleiss:
Looks good to me!

PR #1355, opened 2020-11-18
Use the base lazy tensor's root decomposition in ConstMulLazyTensor

### LINKED ISSUE #1354: [Bug] ConstantMulLazyTensor recomputes its own root
# 🐛 Bug

`ConstantMulLazyTensor` recomputes its own root decomposition even when the component lazy tensor already has its own root. This is quite slow because it means more root decompositions are called than are strictly necessary.

## To reproduce

```python
import torch
from gpytorch.lazy import lazify, RootLazyTensor

a = torch.randn(50, 10)
lazy_square_a = RootLazyTensor(lazify(a))
(4. * lazy_square_a).root_decomposition().root.evaluate() # this is not 2 * a.
```

## Expected Behavior

At least for positive constants, this should return 2 * a.

## System information

gpytorch master
pytorch 1.7.0
mac os

## Additional context
I'll make a PR in a minute.

wjmaddox:
This will fix #1354; however, one outstanding issue is that while the fix itself works, it might break something in botorch's sampling mechanisms. 

```
import torch
from gpytorch.lazy import lazify, RootLazyTensor
from gpytorch.distributions import MultivariateNormal

a = torch.randn(50, 100)
lazy_square_a = RootLazyTensor(lazify(a))

mm_a = 4 * lazy_square_a

torch.norm((mm_a).root_decomposition().root.evaluate() - (2 * a)) # should be zero now
# note that this is 100 x 50
mm_a.zero_mean_mvn_samples(100).shape

# sampling also works
dist = MultivariateNormal(torch.zeros(50), mm_a)
dist.rsample(torch.Size((100,)))
```

The sampling bit might break botorch's sampling in this case.


Balandat:
Hmm how would it break the BoTorch sampling mechanisms? not quite sure I see why this would be the case, it doesn't change any shapes, right?

wjmaddox:
It does actually change sampling shapes. Sorry for not putting it in the original comment but continuing on from that example:

```python
from botorch.sampling import IIDNormalSampler
from botorch.posteriors import GPyTorchPosterior

posterior = GPyTorchPosterior(dist)
sampler = IIDNormalSampler(num_samples=256)

sampler(posterior)
# RuntimeError: Incompatible dimension of `base_samples`
```

The error is here:
```
gpytorch/gpytorch/distributions/multivariate_normal.py(177)rsample()
    175                 base_samples = base_samples[..., : covar_root.shape[-1], :]
    176             elif covar_root.shape[-1] > base_samples.shape[-2]:
--> 177                 raise RuntimeError("Incompatible dimension of `base_samples`")
    178             res = covar_root.matmul(base_samples) + self.loc.unsqueeze(-1)
 ```
Debugging, the covariance matrix is `50 x 50`, the `covar_root` is `50 x 100` but the base_samples are `50 x 256`

**Comment by wjmaddox** [comment]
On digging a bit deeper, the proper thing to do is to modify `gpytorch.distributions.MultivariateNormal` rather than botorch's event shape because even a naive fix like enforcing the event shape to be correct here produces errors.

**Review by gpleiss — APPROVED** [review]

**Comment by wjmaddox** [comment]
This commit fixes these issues but looks a little bit funky to do it this way on initialization.

File: gpytorch/lazy/constant_mul_lazy_tensor.py [hunk 1]
```diff
@@ -153,3 +155,11 @@ def diag(self):
     def evaluate(self):
         res = self.base_lazy_tensor.evaluate()
         return res * self.expanded_constant
+
+    @cached(name="root_decomposition")
+    def root_decomposition(self, method: Optional[str] = None):
+        if (self._constant > 0).sum() == self._constant.numel():
```
Balandat, line 161: Not sure how much `>=` messes with p.d. assumptions anywhere, but should be ok 

```suggestion
        if torch.all(self._constant >= 0):
```

gpytorch/distributions/multivariate_normal.py
```diff
@@ -154,8 +164,13 @@ def rsample(self, sample_shape=torch.Size(), base_samples=None):
             res = res.view(sample_shape + self.loc.shape)
 
         else:
+            covar_root = covar.root_decomposition().root
+
             # Make sure that the base samples agree with the distribution
-            if self.loc.shape != base_samples.shape[-self.loc.dim() :]:
+            if (
+                self.loc.shape != base_samples.shape[-self.loc.dim() :]
+                and covar_root.shape[-1] < base_samples.shape[-1]
```
Balandat, line 172: To make sure I fully understand: The higher level concern is that the covar root is not full rank, right? Is the idea that we initially get base samples for the full rank and leave those fixed during SAA optimization so if we use low-rank root decomps we need need to tolerate the fact that the root loses a rank between evaluations at different test points?

wjmaddox: I don't think that's a new concern; it was already there. This check is just ensuring that we still check for the base samples having the correct shape (e.g. we don't lose too many ranks during the SAA approximation), but that it's okay for the base samples to not match in the case where the covar root is large.

Balandat: Hmm so my understanding would be that we'd start off with base samples compatible with a full rank covar root, and then just use however many we need given the rank of the covar root. This seems to be the opposite of what you say - did you mean to say "it's okay for the base samples to not match in the case where the covar root is *large*." (in terms of rank)?

wjmaddox: Yes, this allows root decompositions of size `n x m` where `m > n` to be used.

Issue #1349, opened 2020-11-14
[Question] Sparse GPs for Batch Independent MultiOutputs slow prediction?


monabf:
# Question

Thanks for this great package!

Not really a bug, rather a question. I need to train a GP with 20,000x7 training inputs, 20,000x5 training outputs, and 40,000x7 test inputs. My first call was to look into the implementation of sparse GPs, because that's what I usually use with GPy. I implemented a sparse + batch independent multioutput GP model class as in issue #1043 and tested it with the UCI elevators dataset from the [SGPR tutorial](https://docs.gpytorch.ai/en/v1.1.1/examples/02_Scalable_Exact_GPs/SGPR_Regression_CUDA.html). 

However, I noticed that while training runs as expected, prediction is unreasonably slow. Attached is a small example comparing training and prediction time with GPyTorch and GPy. The difference gets much larger as the number of output dimensions, training samples or inducing points grows. 

So here are my questions:
1) How come prediction takes so long with the sparse + multioutput GP model from #1043? Is there a bug somewhere, or is this actually expected?
2) Is it just a bad call to try using SGPR with multioutputs? I don't care which method I am using, I just need to be able to train large, multioutput GPs efficiently. This will mostly be done on GPU, but I would still like it to run reasonably on CPU so I can do some tests locally. With sparse GPs from GPy my use case ran in around 2h, which I would consider reasonable on CPU for tests, but with my current GPyTorch implementation prediction alone takes about 20h... I am open to suggestions!

Thanks a lot for your help.

## To reproduce

```python
import time
import urllib.request
from math import floor

import GPy
import gpytorch
import torch
from gpytorch.distributions import MultivariateNormal
from gpytorch.kernels import ScaleKernel, RBFKernel, InducingPointKernel
from gpytorch.means import ConstantMean
from scipy.io import loadmat

if __name__ == '__main__':
    # Run GPyTorch SGPR + independent multioutputs example
    # https://docs.gpytorch.ai/en/v1.2.1/examples/02_Scalable_Exact_GPs/SGPR_Regression_CUDA.html
    # https://github.com/cornellius-gp/gpytorch/issues/1043
    print('Downloading \'elevators\' UCI dataset...')
    urllib.request.urlretrieve(
        'https://drive.google.com/uc?export=download&id=1jhWL3YUHvXIaftia4qeAyDwVxo6j1alk',
        '../elevators.mat')
    output_size = 2
[...]

PR #1356, opened 2020-11-19
Add LowRankRootLazyTensor / LowRankRootAddedDiagLazyTensor, refactor SGPR

jacobrgardner:
Adds two new LTs that allow for the use of Woodbury on matrices of the form `UV + diag(a)` for which `U` and `V` are low rank enough that we believe this will be faster. 

SGPR is an obvious use case here, where `U = V^{T} = K_{XU}K_{UU}^{-1/2}`.

The motivation for this PR is to solve #1349, and it does so pretty well. The example code that @monabf included in that issue now trains 5 times faster on the GPU and now has prediction times on the CPU that are in line with GPy.

@gpleiss -- this PR also fixes one other GPyTorch bug in a slightly hacky way (see the new `AddedDiagLazyTensor.evaluate_kernel` where I describe the problem), but in a way that will go away when we switch over to linear operators that don't use representation trees. If you're not happy with this fix as a temporary one, we'll need to do some thinking about how to redesign the representation tree. Probably it will like need the option to optionally pass in `_cls` or something to `LazyTensorRepresentationTree`. I'm not 100% sure.

### DIFF
--- a/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
@@ -6,7 +6,7 @@ from ..utils.cholesky import psd_safe_cholesky
 from ..utils.memoize import cached
 from . import delazify
 from .added_diag_lazy_tensor import AddedDiagLazyTensor
-from .diag_lazy_tensor import DiagLazyTensor, ConstantDiagLazyTensor
+from .diag_lazy_tensor import ConstantDiagLazyTensor, DiagLazyTensor
 from .low_rank_root_lazy_tensor import LowRankRootLazyTensor
 
 
@@ -28,7 +28,7 @@ class LowRankRootAddedDiagLazyTensor(AddedDiagLazyTensor):
         A_inv = self._diag_tensor.inverse()  # This is fine since it's a DiagLazyTensor
         U = self._lazy_tensor.root
         V = self._lazy_tensor.root.transpose(-2, -1)
-        C = ConstantDiagLazyTensor(torch.ones(*V.batch_shape, device=V.device, dtype=V.dtype), V.shape[-1])
+        C = ConstantDiagLazyTensor(torch.ones(V.batch_shape, device=V.device, dtype=V.dtype), V.shape[-2])
 
         cap_mat = delazify(C + V.matmul(A_inv.matmul(U)))
         chol_cap_mat = psd_safe_cholesky(cap_mat)
@@ -56,7 +56,6 @@ class LowRankRootAddedDiagLazyTensor(AddedDiagLazyTensor):
         logdet_term = logdet_cap_mat + logdet_A
 
         return logdet_term
-
 
     def __add__(self, other):
         from .diag_lazy_tensor import DiagLazyTensor
--- a/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
@@ -2,12 +2,12 @@
 
 import torch
 
+from ..utils.cholesky import psd_safe_cholesky
+from ..utils.memoize import cached
 from . import delazify
 from .added_diag_lazy_tensor import AddedDiagLazyTensor
 from .diag_lazy_tensor import DiagLazyTensor
 from .low_rank_root_lazy_tensor import LowRankRootLazyTensor
-from ..utils.cholesky import psd_safe_cholesky
-from ..utils.memoize import cached
 
 
 class LowRankRootAddedDiagLazyTensor(AddedDiagLazyTensor):
--- a/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
+++ b/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
@@ -22,12 +22,13 @@ class LowRankRootAddedDiagLazyTensor(AddedDiagLazyTensor):
 
         super().__init__(*lazy_tensors, preconditioner_override=preconditioner_override)
 
+    @property
     @cached(name="chol_cap_mat")
     def chol_cap_mat(self):
         A_inv = self._diag_tensor.inverse()  # This is fine since it's a DiagLazyTensor
         U = self._lazy_tensor.root
         V = self._lazy_tensor.root.transpose(-2, -1)
-        C = DiagLazyTensor(torch.ones(*V.batch_shape, V.shape[-2], V.shape[-2], device=V.device, dtype=V.dtype))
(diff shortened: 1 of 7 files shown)


jacobrgardner:
Two other points:

1. It's trivial to use the new LTs for the linear kernel, but we'll just need to decide on a threshold of `d` to switch over. Obviously, as `d` approaches `n`, CG is probably the better choice over woodbury. I think in the SGPR case it should probably always be worth it to just use woodbury, though.

2. @wjmaddox the bugfix for `AddedDiagLazyTensor` is possibly relevant to you, since before this fix it would not have been possible for a `KroneckerProductAddedDiagLazyTensor` to be created by a GPyTorch model -- you'd always end up with a vanilla `AddedDiagLazyTensor`.

wjmaddox:
Shouldn't this also speed up SKI + KISS-GP?

jacobrgardner:
@wjmaddox I think folding `K_{UU}^{-1/2}` in to `W` might still be slower since you lose the sparsity of `W` doing that. Could make a generalization to handle full `UWU' + diag(a)` matrices.

gpytorch/lazy/low_rank_root_lazy_tensor.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
+from .root_lazy_tensor import RootLazyTensor
+
+
+class LowRankRootLazyTensor(RootLazyTensor):
+    """
+    Very thin wrapper around MatmulLazyTensor that denotes that the matmul specifically represents a low rank
```
Balandat, line 8: ?
```suggestion
    Very thin wrapper around RootLazyTensor that denotes that the tensor specifically represents a low rank
```

gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,107 @@
...
+
+        super().__init__(*lazy_tensors, preconditioner_override=preconditioner_override)
+
+    @property
+    @cached(name="chol_cap_mat")
+    def chol_cap_mat(self):
+        A_inv = self._diag_tensor.inverse()  # This is fine since it's a DiagLazyTensor
+        U = self._lazy_tensor.root
+        V = self._lazy_tensor.root.transpose(-2, -1)
+        C = DiagLazyTensor(torch.ones(*V.batch_shape, V.shape[-2], device=V.device, dtype=V.dtype))
```
Balandat, line 31:
```suggestion
        C = ConstantDiagLazyTensor(torch.ones(*V.batch_shape, device=V.device, dtype=V.dtype), V.shape[-1])
```

gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,107 @@
...
+    def _logdet(self):
+        chol_cap_mat = self.chol_cap_mat
+        logdet_cap_mat = 2 * torch.diagonal(chol_cap_mat, offset=0, dim1=-2, dim2=-1).log().sum(-1)
+        logdet_A = self._diag_tensor.logdet()
+        logdet_term = logdet_cap_mat + logdet_A
+
+        return logdet_term
+
+    def add_diag(self, added_diag):
+        return self.__class__(self._lazy_tensor, self._diag_tensor.add_diag(added_diag))
```
Balandat, line 61: this is already properly inherited

```suggestion
```

monabf:
Are there any news on this PR? I'm starting to need this fix for #1349. Thanks a lot!

gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
+
+import torch
+
+from ..utils.cholesky import psd_safe_cholesky
+from ..utils.memoize import cached
+from . import delazify
+from .added_diag_lazy_tensor import AddedDiagLazyTensor
+from .diag_lazy_tensor import DiagLazyTensor
```
jacobrgardner, line 9:
```suggestion
from .diag_lazy_tensor import DiagLazyTensor, ConstantDiagLazyTensor
```

Balandat (approved):

monabf:
After updating to the latest version of GPyTorch, running the exact same code from my issue #1349 raises an error in the forward function. @gpleiss @jacobrgardner could something have gone wrong with this fix?

Thanks for your help!

Output of the example code after updating GPyTorch:
```
  File "issue.py", line 87, in <module>
    loss = -mll(output, train_y).sum()
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/module.py", line 28, in __call__
    outputs = self.forward(*inputs, **kwargs)
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/mlls/exact_marginal_log_likelihood.py", line 62, in forward
    res = output.log_prob(target)
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/distributions/multivariate_normal.py", line 141, in log_prob
    inv_quad, logdet = covar.inv_quad_logdet(inv_quad_rhs=diff.unsqueeze(-1), logdet=True)
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py", line 98, in inv_quad_logdet
    self_inv_rhs = self._solve(inv_quad_rhs)
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py", line 42, in _solve
    chol_cap_mat = self.chol_cap_mat
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/utils/memoize.py", line 59, in g
    return _add_to_cache(self, cache_name, method(self, *args, **kwargs), *args, kwargs_pkl=kwargs_pkl)
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/lazy/low_rank_root_added_diag_lazy_tensor.py", line 31, in chol_cap_mat
    C = ConstantDiagLazyTensor(torch.ones(V.batch_shape, device=V.device, dtype=V.dtype), V.shape[-2])
  File "/Users/mona/PhD_code/observer_add_GPyTorch/venv/lib/python3.8/site-packages/gpytorch/lazy/diag_lazy_tensor.py", line 221, in __init__
    self._diag = diag_values.expand(*diag_values.shape[:-1], diag_shape)
RuntimeError: The expanded size of the tensor (500) must match the existing size (2) at non-singleton dimension 0.  Target sizes: [500].  Tensor sizes: [2]
```

PR #1361, opened 2020-11-24
Pass module to prior closures

### LINKED ISSUE #1360: [Bug] priors are lost after deepcopy (and ExactGP.get_fantasy_model)
# 🐛 Bug

GPyTorch modules register priors using a (name, prior, closure, setting_closure) tuple. The closure and setting_closure are callables that reference the instance (self) as a global variable. When a module is deepcopied. Those closure methods are copied over, but still reference the original instance (not the new copy). Therefore the closure returns the parameters attached to the old module, not the new copy. This makes the loss term from the priors constant in the MLL evaluation.

## To reproduce
[test_deepcopy.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/5585605/test_deepcopy.ipynb.txt)

## Expected Behavior

Ideally, the prior distributions would work as expected on fantasy models.

Balandat:
This fixes #1360

Test Plan:
[test_deepcopy_fixed.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/5586510/test_deepcopy_fixed.ipynb.txt)

### DIFF
--- a/examples/00_Basic_Usage/Hyperparameters.ipynb
+++ b/examples/00_Basic_Usage/Hyperparameters.ipynb
@@ -17,6 +17,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "# smoke_test (this makes sure this example notebook gets tested)\n",
+    "\n",
     "import math\n",
     "import torch\n",
     "import gpytorch\n",
@@ -394,7 +396,7 @@
     "likelihood.noise_covar.register_prior(\n",
     "    \"noise_std_prior\",\n",
     "    gpytorch.priors.NormalPrior(0, 1),\n",
-    "    lambda: likelihood.noise.sqrt()\n",
+    "    lambda module: module.noise.sqrt()\n",
     ")\n",
     "\n",
     "# Create a GaussianLikelihood with a normal prior for the noise\n",
@@ -505,7 +507,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.3"
+   "version": "3.7.0"
   }
  },
  "nbformat": 4,
--- a/gpytorch/kernels/arc_kernel.py
+++ b/gpytorch/kernels/arc_kernel.py
@@ -123,7 +123,7 @@ class ArcKernel(Kernel):
         )
         if angle_prior is not None:
             self.register_prior(
-                "angle_prior", angle_prior, lambda: self.angle, lambda v: self._set_angle(v),
+                "angle_prior", angle_prior, lambda m: m.angle, lambda m, v: m._set_angle(v),
             )
 
         self.register_constraint("raw_angle", angle_constraint)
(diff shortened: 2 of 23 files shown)


Balandat:
@jacobrgardner @gpleiss This makes some not-insignificant changes to the internals of priors.

I should probably get rid of the term "closure" since that was the core issue with this setup in the first place. Maybe `evaluator` and `setter`?

sdaulton:
This looks good to me. The only concern is backward compatibility for scenarios where a downstream user call's register_prior on the gpytorch module (e.g. https://github.com/pytorch/botorch/blob/master/botorch/models/transforms/input.py#L639-L644). What are your thoughts on maintaining backwards compatibility?

Balandat:
Yeah that's a valid concern. I think compatibility here will be kind of pain to ensure due to the fact that these are generic callables, but maybe we can add an actionable deprecation error inside `register_prior` to make sure that the signature of the callable is as expected.

jacobrgardner:
This must surely break out basic usage Hyperparameters.ipynb right? I've no idea why the examples smoke test run wouldn't have caught that

Balandat:
Yeah it would break the likelihood setup in `Hyperparameters.ipynb`, but it looks like nothing is actually being done with the likelihood. This will only break once you actually evaluate the prior, e.g. compute the mll, which is not done in that tutorial.

Balandat:
@jacobrgardner Apart from updating the tutorial, are there any concerns with making these changes here? As I mentioned, since we're passing in generic callables, it's not going to be easy to make this work in a backward-compatible fashion - unless we wrap a try around the evaluation of the provided closure, and then wrap that closure into a trivial container of the form `lambda m, v: closure(v)`.

jacobrgardner:
@Balandat What do you think? My feeling would be that trying to maintain backward compatibility might be pretty important here, since people are probably pretty likely to be directly registering priors (e.g., it's how I do it).

Maybe at a minimum we can detect old style closures and raise an error explaining how to fix things?

Balandat:
Yeah that sounds reasonable to me - let me give it a try.

Balandat:
OK this should do it. Even added a test...

Balandat:
@jacobrgardner, @gpleiss Any idea why the doc test is failing?

Balandat:
bump

gpleiss:
So I can confirm that this does break `hyperparameters.ipynb` (it wasn't part of our smoke tests). I added it to our example smoke tests and I fixed the registered prior.

If @jacobrgardner is okay with the breaking change, and throwing and error, then I'm okay with it as well. It might be useful to make the backward-compatibility error message a little more verbose. Maybe:

```
As of version 1.4, `param_or_closure` must operate on a module instance. For example:

likelihood.noise_covar.register_prior(
    "noise_std_prior",
    gpytorch.priors.NormalPrior(0, 1),
    lambda module: module.noise.sqrt()
)
```

Thoughts?

Balandat:
sure I will make this change

gpleiss (approved):

PR #1374, opened 2020-11-30
Bump Pytorch up to 1.7

gpleiss:
- Use new `torch.fft` module
- Remove old `gpytorch.utils.fft` stuff


jacobrgardner (approved):

Balandat:
> this 1.7 version that seems to address the problem

What do you mean by that? Don't the newer pytorch versions (>1.7) also address the problem?

samfairman:
Actually, conda automatically downgraded pytorch while I was testing. I'm not sure why, but my comment is resolved. Thanks for responding so quickly.

PR #1410, opened 2020-12-31
Add multitask DGP example.

### LINKED ISSUE #1240: [Docs] DGP/DSPP Multi-task Example
# 📚 Documentation/Examples

** Is there a feature that needs some example code? **

Currently multi-task examples are given for exact and variational models, but it's not clear how
multi-task models should be built using the DGP/DSPP infrastructure.

- a

gpleiss:
[Closes #1240]


gpleiss:
View the multitask DGP example here: https://gpytorch--1410.org.readthedocs.build/en/1410/examples/05_Deep_Gaussian_Processes/DGP_Multitask_Regression.html

jacobrgardner:
Only comment, do you think it would be worth it to include LMCVariationalStrategy as a possibility in the example?

gpleiss:
@jacobrgardner I originally tried that - and it was a bit challenging to get it to work with DeepGPs. We probably want to create a LMCDeepGP layer if we want LMC functionality.

jacobrgardner:
Interesting -- a student and I have been using LMC with DSPPs with no issues really. If it's just not working with DGPs then we can merge without for now.

Issue #1387, opened 2020-12-10
[Question] MultiOutput Kernels that are not of the MultiTask-form


arthus701:
Hi,
I am really amazed by the BBMM approach to GP inference and would love to use GPyTorch. However, in my use case (Geosciences) we deal with MultiInput-MultiOutput GPs. Think of a vector field defined on the Earth's surface. Additionally, we have good physical motivation to implement a certain MultiOutput-Kernel, which is not of the MultiTask form described in the docs. We already know the correlation of the different output dimensions.

Is it possible to implement such a model (with fixed and known input and output dimensions) in GPyTorch?

If so, I think the documentation would really benefit from a section describing how to translate an already existing, custom model to GPyTorch. Once I accomplished implementing my use case, I could start working on this and open a PR later...

Thank you so much,
Arthus

Balandat:
That should certainly be possible. Do you have some more specifics on the Kernel you're trying to implement?

arthus701:
Sure! I'm trying to implement the gradient of the Legendre kernel. I'm not sure which information you need...

For two points y, y' on the sphere, set x = |y| |y'| / R^2 and t = y^T y / R^2, with some reference radius R. The kernel is defined via the [generating function of the Legendre polynials](https://en.wikipedia.org/wiki/Legendre_polynomials#Definition_via_generating_function). The gradient wrt. y,y' then defines a Multivariate/MultiOutput kernel on the sphere.

Let me know what else you need to know. I guess my question could also be abstracted from this specific kernel to: 

"I have a model with n input and m output dimensions and a kernel K: R^n x R^n -> R^{mxm}. 
How can I implement this in GPyTorch? 
Should I start off of MultiTaskKernel or Kernel?
What about the vector-likelihood?"

Thank you for your time!

gpleiss:
If `x1` is `n1 x d` and `x2` is `n2 x d`, then the kernel's forward method will need to return a `(n1 x m) x (n2 x m)` tensor. It should look something like:

```python
class LegendreKernel(gpytorch.kernels.Kernel):
    def __init__(self, m, ...):
        self.m = m
        # ...

    def forward(self, x1, x2, **params):
        # ...
        return tensor # size (n1 x m) x (n2 x m)
        
    def num_outputs_per_input(self, x1, x2):
        return x1.size(-2) * x2.size(-2) * self.m
```

[...]

PR #1421, opened 2021-01-06
Add custom kernel example.

arthus701:
After the brief discussion over at #1387, I started to sketch a notebook on how to implement a custom kernel.

I am a bit unhappy with the repeated training and evaluation, although I think it emphasizes how simple it is to implement a kernel and how the kernel can be improved by adding hyperparameters, which was the idea behind it.

Instead I could also add a section on the MultiOutput stuff we discussed over at #1387. I do not yet have an idea for a good general use case though. And I'm not sure whether my implementation of the Legendre Kernel is actually "the right way" (although it works...). 

I also did not know where to put the notebook, so I started by placing it under "Basic Usage".

One more thing: From my implementation I figured the results should be similar to just setting `has_lengthscale = True`, but as can be seen from the end of the notebook, it appears the results are different. Do you see why that is?

I'm looking forward to your opinion!

Cheers,
Arthus


examples/00_Basic_Usage/index.rst
```diff
@@ -26,3 +26,6 @@ Before checking these out, you may want to check out our `simple GP regression t
 
 .. _Saving and Loading Models:
   Saving_and_Loading_Models.ipynb
+
+.. _Implementing a custom Kernel
```
gpleiss:
```suggestion
.. _Implementing a custom Kernel:
```

examples/00_Basic_Usage/Implementing_a_custom_Kernel.ipynb
```diff
@@ -0,0 +1,428 @@
...
+    "        ax.legend(['Observed Data', 'Mean', 'Confidence'])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### A first kernel\n",
+    "\n",
+    "To implement a custom kernel, we derive one from GPyTorch's [`kernel` class](https://docs.gpytorch.ai/en/latest/kernels.html) and implement the `forward()` method. The base class provides many useful routines. For example, `__call__()` is implemented, so that the kernel may be called directly, without resorting to the `forward()` routine. Among other things, the `Kernel` class provides a method `covar_dist()`, which may be used to calculate the Euclidian distance between point pairs conveniently.\n",
```
gpleiss, line 95:
```suggestion
    "To implement a custom kernel, we derive one from GPyTorch's [kernel class](https://docs.gpytorch.ai/en/latest/kernels.html) and implement the `forward()` method. The base class provides many useful routines. For example, `__call__()` is implemented, so that the kernel may be called directly, without resorting to the `forward()` routine. Among other things, the `Kernel` class provides a method `covar_dist()`, which may be used to calculate the Euclidian distance between point pairs conveniently.\n",
```

gpleiss (approved):

gpleiss:
Sorry for the delay @arthus701  - this notebook looks excellent! I'm just going to add a few changes that fix the build issues.

PR #1422, opened 2021-01-06
Add arg_constraints attribute to Delta distribution

neerajprad:
Some of the tests in `test/examples` are failing after https://github.com/pytorch/pytorch/pull/48743 was merged turned on distributions validation. This adds the `arg_constraints` attribute that was missing in `distributions.Delta` which is needed for generic parameter support validation. Note that this failure only happens if Pyro isn't installed (since Pyro's implementation has this attribute). 

cc. @Balandat

### DIFF
--- a/gpytorch/distributions/delta.py
+++ b/gpytorch/distributions/delta.py
@@ -28,7 +28,8 @@ except ImportError:
             under differentiable transformation.
         :param int event_dim: Optional event dimension, defaults to zero.
         """
-        arg_constraints = {'v': constraints.real, 'log_density': constraints.real}
+
+        arg_constraints = {"v": constraints.real, "log_density": constraints.real}
         has_rsample = True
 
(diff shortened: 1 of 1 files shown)


gpytorch/distributions/delta.py
```diff
@@ -28,18 +29,29 @@ class Delta(Distribution):
         :param int event_dim: Optional event dimension, defaults to zero.
         """
 
+        arg_constraints = {"v": constraints.real, "log_density": constraints.real}
         has_rsample = True
 
         def __init__(self, v, log_density=0.0, event_dim=0, validate_args=None):
             if event_dim > v.dim():
-                raise ValueError("Expected event_dim <= v.dim(), actual {} vs {}".format(event_dim, v.dim()))
+                raise ValueError(
```
neerajprad, line 37: All changes from here below are due to `black` reformatting.

Balandat (approved):
thanks!

PR #1430, opened 2021-01-15
Improvements to Kronecker LazyTensors

Balandat:
This is a rather large PR that changes the structured LinAlg internals of Kronecker lazy tensors. The main changes are

- New `KroneckerProductDiagLazyTensor`
- New `KroneckerSumLazyTensor` for efficient solves and log dets
- Changes to `KroneckerProductAddedDiagLazyTensor`:
  - Custom implementations for solves/logdets/roots with non-constant diagonal tensor (Fixes a bug where we implicitly assumed the diagonal to be constant, and returned incorrect results if that was not the case)
  - Previously, `_root_decomposition` and `_inv_root_decomposition` were returning the (inv) root from the eigendecomposition as a dense tensor, which can be inefficient. This now returns the root as `MatmulLazyTensor` instead. E.g. a matrix vector product of the root with some vector `v` is now implicitly computed as `q_matrix @ (evals \dot v)` rather than `(q_matrix @ diag(evals)) @ v`, which can make a big difference since `q_matrix` is a Kronecker product. This can help runtime, but more importantly it significantly reduces memory footprint, since we don't need to instantiate the (inv) root, but only the constituent components.
- Refactor of `MultitaskGaussianLikelihood` to use the new lazy tensor goodies, alias  `MultitaskGaussianLikelihoodKronecker` to `MultitaskGaussianLikelihood` (to be removed in a later release, not quite sure how you want to handle the deprecation). Adds the following two options:
  - `has_global_noise`: whether to include a `\sigma^2 I_{nt}` term in the noise model.
  - `has_task_noise`: whether to include task-specific noise terms, which add `I_n \kron D_T` into the noise model.


Balandat:
@wjmaddox

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -71,11 +65,11 @@ def _solve(self, rhs, preconditioner=None, num_tridiag=0):
     def _root_decomposition(self):
         evals, q_matrix = self.lazy_tensor.symeig(eigenvectors=True)
         updated_evals = DiagLazyTensor((evals + self.diag_tensor.diag()).pow(0.5))
-        matrix_root = q_matrix.matmul(updated_evals)
+        matrix_root = MatmulLazyTensor(q_matrix, updated_evals)
         return matrix_root
 
     def _root_inv_decomposition(self, initial_vectors=None):
         evals, q_matrix = self.lazy_tensor.symeig(eigenvectors=True)
         inv_sqrt_evals = DiagLazyTensor((evals + self.diag_tensor.diag()).pow(-0.5))
```
wjmaddox, line 73: May want to clamp these at 0 for numerical stability

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -71,11 +65,11 @@ def _solve(self, rhs, preconditioner=None, num_tridiag=0):
     def _root_decomposition(self):
         evals, q_matrix = self.lazy_tensor.symeig(eigenvectors=True)
         updated_evals = DiagLazyTensor((evals + self.diag_tensor.diag()).pow(0.5))
```
wjmaddox, line 67:
```suggestion
        updated_evals = DiagLazyTensor((evals + self.diag_tensor.diag()).clamp(min=0.).pow(0.5))
```

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
wjmaddox, line 73:
```suggestion
        inv_sqrt_evals = DiagLazyTensor((evals + self.diag_tensor.diag()).clamp(min=0.).pow(-0.5))
```

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: I guess `diag_tensor` is assumed to be non-negative (should maybe check?) and `evals` should already be clamped in https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/lazy/lazy_tensor.py#L1881.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: this happens here if I'm not mistaken: https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/lazy/lazy_tensor.py#L1881

Balandat:
I am not sure why in the case of using a `ConstantDiagLazyTensor` gradients aren't being returned properly (hence the test failure).

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -25,57 +22,63 @@ def __init__(self, *lazy_tensors, preconditioner_override=None):
...
-        return inv_quad_term, logdet_term
+        if isinstance(self.diag_tensor, ConstantDiagLazyTensor):
+            # we want to call the standard InvQuadLogDet to easily get the probe vectors and do the
+            # solve but we only want to cache the probe vectors for the backwards
+            inv_quad_term, _ = super().inv_quad_logdet(
+                inv_quad_rhs=inv_quad_rhs, logdet=False, reduce_inv_quad=reduce_inv_quad
+            )
+            logdet_term = self._logdet() if logdet else None
+            return inv_quad_term, logdet_term
+        return super().inv_quad_logdet(inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
```
wjmaddox, line 33:
```suggestion
        return super().inv_quad_logdet(inv_quad_rhs=inv_quad_rhs, logdet=logdet, reduce_inv_quad=reduce_inv_quad)
```

This is causing most of the test failures. Fixing this locally resolves all but one of the failures for me, where it seems like `gpytorch...LazyTensor._quad_form_derivative(...)` is missing the gradient wrt the diagonal.

Balandat: Any idea why? I don't really see the difference compared to the previous `ConstantDiagLazyTensor`.

wjmaddox: Not sure yet, but it doesn't appear if it's a `DiagLazyTensor`. Will next take a look through the changes to `ConstantDiagLazyTensor`.

wjmaddox:
A partial fix can be found by storing diag_values instead of _diag and using _diag as a property, which passes the failing unit tests.

But, this breaks a bunch of gradient based unit tests in `MulLazyTensor` that I haven't been able to make any progress in debugging.
```python
    @property
    def _diag(self):
        return self._diag_values.expand(*self._diag_values.shape[:-1], self.diag_shape)
    
    def _quad_form_derivative(self, left_vecs, right_vecs):
        res = super()._quad_form_derivative(left_vecs, right_vecs)
        if res[0] is not None:
            res = res[0].sum(-1,keepdim=True)
            return (res,)
        return res
```

Balandat:
Hmm yeah having the same issue. My last commit adds a bunch of special cases to `ConstantDiagLazyTensor` so that it will return `ConstantDiagLazyTensor` e.g. for all the element-wise operations, but that's orthogonal to this issue.

Balandat:
Added a `KroneckerProductDiagLazyTensor` in order to simplify implementation of the Woodbury matrix identity for `KroneckerProductAddedDiagLazyTensor` in case the diagonal has the same kronecker structure as the full matrix. 

I am running into some failures regarding construction of these that I don't really understand. 
```
from gpytorch.lazy.kronecker_product_lazy_tensor import KroneckerProductDiagLazyTensor
from gpytorch.lazy.diag_lazy_tensor import DiagLazyTensor

dlts = [DiagLazyTensor(torch.ones(2)), DiagLazyTensor(torch.ones(3))]

# this works fine: 
kplt = KroneckerProductLazyTensor(*dlts)

# this fails:
kpdlt = KroneckerProductDiagLazyTensor(*dlts)
```

Am I missing something obvious? The error is 
```
TypeError                                 Traceback (most recent call last)
<ipython-input-16-cd50cf88eed5> in <module>
----> 1 kpdlt = KroneckerProductDiagLazyTensor(*dlts)

~/Code/gpytorch/gpytorch/lazy/kronecker_product_lazy_tensor.py in __init__(self, *lazy_tensors)
    316         if not all(isinstance(lt, DiagLazyTensor) for lt in lazy_tensors):
    317             raise RuntimeError("Components of KroneckerProductDiagLazyTensor must be DiagLazyTensor.")
--> 318         super(KroneckerProductTriangularLazyTensor, self).__init__(*lazy_tensors)
    319 
    320     @cached(name="cholesky")

~/Code/gpytorch/gpytorch/lazy/kronecker_product_lazy_tensor.py in __init__(self, *lazy_tensors)
     79                     "same batch shapes. Got {}.".format([lv.batch_shape for lv in lazy_tensors])
     80                 )
---> 81         super().__init__(*lazy_tensors)
     82         self.lazy_tensors = lazy_tensors
     83 

TypeError: __init__() takes 2 positional arguments but 3 were given
```

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -46,33 +47,54 @@ def _preconditioner(self):
...
+                # TODO: Use fp64 registry once #1213 is addressed
+                rhs = rhs.double()
+                lt = self.lazy_tensor.to(torch.double)
+                dlt = self.diag_tensor.to(torch.double)
+
+                KDinv = KroneckerProductLazyTensor(
+                    *[tfull.matmul(tdiag.inverse()) for tfull, tdiag in zip(lt.lazy_tensors, dlt.lazy_tensors)]
+                )
+                # TODO: Figure out how to cache the decompositon for use in later solves
+                Lambda, S = KDinv.symeig(eigenvectors=True)
```
wjmaddox, line 91: This is technically incorrect in my understanding because KDinv is not a symmetric matrix, but we can't use `torch.eig` because the eigenvectors aren't differentiable.

There's also a SVD solution without squaring anything that's similarly numerically stable.

wjmaddox, line 91:
```suggestion
                kron_times_diag_list = [tfull.matmul(tdiag.inverse()) for tfull, tdiag in zip(lt.lazy_tensors, dlt.lazy_tensors)]
                kron_times_diag_symm = KroneckerProductLazyTensor(*[k.matmul(k.transpose(-1, -2)) for k in kron_times_diag_list])
                Lambda_square, S = kron_times_diag_symm.symeig(eigenvectors=True)
                LambdaI = DiagLazyTensor(Lambda_square.sqrt() + 1)
```

wjmaddox:
https://github.com/cornellius-gp/gpytorch/blob/fca285821740bcc013bfe27b5dd11b7fdb9b9812/gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py#L96

To ensure that the kronecker solve rather than linear_cg is getting called, this line should be 
```python
res = lt._inv_matmul(rhs - tmp_term)
```

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -23,59 +21,92 @@ def __init__(self, *lazy_tensors, preconditioner_override=None):
...
-        # symeig requires computing the eigenvectors so that it's differentiable
-        evals, _ = self.lazy_tensor.symeig(eigenvectors=True)
-        evals_plus_diag = evals + self.diag_tensor.diag()
-        return torch.log(evals_plus_diag).sum(dim=-1)
+        if self._diag_is_constant:
+            # symeig requires computing the eigenvectors so that it's differentiable
+            evals, _ = self.lazy_tensor.symeig(eigenvectors=True)
+            evals_plus_diag = evals + self.diag_tensor.diag()
+            return torch.log(evals_plus_diag).sum(dim=-1)
+        return super()._logdet()
```
wjmaddox, line 43:
```suggestion
        if len(self.lazy_tensor.lazy_tensors) == len(self.diag_tensor.lazy_tensors) and all(
                isinstance(dt, ConstantDiagLazyTensor) for dt in self.diag_tensor.lazy_tensors
                ):
            # here the log determinant identity is |D + K| = | D| |I + D^{-1} K|
            # as D is assumed to have constant components, we can look solely at the diag_values
            evals, _ = self.lazy_tensor.symeig(eigenvectors = True)
            const_times_evals = KroneckerProductLazyTensor(*[ee * d.diag_values for ee, d in zip(evals.lazy_tensors, self.diag_tensor.lazy_tensors])
            first_term = (const_times_evals.diag() + 1.).log().sum(dim=-1)
            
            diag_term = self.diag_tensor.diag().clamp(min=1e-7).log().sum(dim=-1)
            
            return first_term + diag_term
            
        return super()._logdet()
```

wjmaddox (commented):
I think this is the correct matrix determinant lemma to exploit the structure in the log determinant.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
Balandat: I made sure to use this only if the kronecker factors are indeed constant diagonal tensors

wjmaddox (commented):
Here's the diag non-constant log det.

gpytorch/lazy/diag_lazy_tensor.py
```diff
@@ -63,10 +64,7 @@ def _mul_constant(self, constant):
         return self.__class__(self._diag * constant.unsqueeze(-1))
 
     def _mul_matrix(self, other):
-        if isinstance(other, DiagLazyTensor):
-            return self.__class__(self._diag * other._diag)
-        else:
-            return self.__class__(self._diag * other.diag())
```
gpleiss, line 69: 🤦 for the way the code was originally...

gpytorch/lazy/kronecker_sum_lazy_tensor.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
+
+from .. import settings
+from .kronecker_product_lazy_tensor import KroneckerProductLazyTensor
+from .sum_lazy_tensor import SumLazyTensor
+
+
+class KroneckerSumLazyTensor(SumLazyTensor):
```
gpleiss, line 8: We may want to consider renaming this? The Kronecker sum has an alternative definition: https://mathworld.wolfram.com/KroneckerSum.html

gpytorch/lazy/kronecker_sum_lazy_tensor.py
```diff
@@ -0,0 +1,75 @@
...
+        (C^{-1/2}AC^{-1/2} \kron D^{-1/2} B D^{-1/2} + I_|C| \kron I_|D|)(C^{1/2} \kron D^{1/2})^{T}
+    where .^{1/2} = Q_.\Lambda_.^{1/2} (e.g. an eigen-decomposition.)
+
+    This formulation admits efficient solves and log determinants.
+
+    The original reference is [https://papers.nips.cc/paper/2013/file/59c33016884a62116be975a9bb8257e3-Paper.pdf].
+
+    Args:
+        :`lazy_tensors`: List of lazy tensors
+    """
```
gpleiss, line 22: It looks like this LT only supports summing  2 LTs, not a variable number? Should we add an explicit constructor for it?

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
```diff
@@ -36,53 +36,76 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
...
+                # as D is assumed to have constant components, we can look solely at the diag_values
+                diag_term = self.diag_tensor.diag().clamp(min=1e-7).log().sum(dim=-1)
+                # symeig requires computing the eigenvectors for it to be differentiable
+                evals, _ = self.lazy_tensor.symeig(eigenvectors=True)
+                const_times_evals = KroneckerProductLazyTensor(
+                    *[ee * d.diag_values for ee, d in zip(evals.lazy_tensors, self.diag_tensor.lazy_tensors)]
+                )
+                first_term = (const_times_evals.diag() + 1).log().sum(dim=-1)
+                return diag_term + first_term
+
```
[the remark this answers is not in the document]
gpleiss: Is this suggested change still relevant?

gpleiss (approved):
Everything looks good to me. I didn't dive into too much detail on the KP math methods, but the test coverage is very comprehensive so I trust the changes :)

I have two minor comments about `KroneckerSumLazyTensor`, but that's it. @Balandat want to take a look and then merge?

gpleiss:
Regarding the deprecation, I think it's okay to just remove it - since it was already marked as deprecated. We'll make the next release a 1.4 release then.

gpytorch/lazy/kronecker_product_added_diag_lazy_tensor.py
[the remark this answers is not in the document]
wjmaddox: These are not relevant any more. Actually, the second one was slightly incorrect.

PR #1476, opened 2021-02-11
psd_safe_cholesky should obey cholesky_jitter settings

### LINKED ISSUE #1457: [Feature Request] should psd_safe_cholesky always respect cholesky jitter settings?
# 🚀 Feature Request

`psd_safe_cholesky` takes max jitter as an argument. Looking across the codebase, sometimes `gpytorch.settings.cholesky_jitter.value()` is passed in and sometimes it's not. Is there a pattern I'm missing? If not, should it default to the global setting if nothing is passed rather than defaulting to a hardcoded 1e-6 or 1e-8 depending on the input dtype? 

## Motivation

In our active learning setting, we occasionally run into crashes where the default jitter isn't big enough. I've set the global jitter value to be higher to avoid them, so I was surprised to see that it's still crashing with the smaller jitter. 

## Pitch

Defaulting to the global setting value is the behavior I would expect. An alternative is for users to change the `psd_safe_cholesky` calls directly but they're almost always buried pretty deep elsewhere in gpytorch (or in my case, botorch) rather than my client code so there's no easy path there. 

Willing to PR this. This would be a breaking behavior change though, and I'm not sure of other downstream effects on other client code.

mshvartsman:
This potentially resolves #1457, but is a breaking-change way of doing it. 

Old behavior: the default value in `settings` is `None`, which when passed into `psd_safe_cholesky` gives dtype-dependent behavior. But `settings.cholesky_jitter` is a dtype-independent `_value_context`. 

One possible new behavior (in this PR) Turn the setting into a `_dtype_value_context` subclass, which would probably break a bunch of downstream things that expect to call `settings.cholesky_jitter.value()`. I've removed these calls in gpytorch, but at least one place in botorch will need to be changed as well. 

Other possible new behavior: just have a dtype-independent jitter (this won't throw `TypeError`s but will slightly change behavior). 

Thoughts?


Balandat:
What about

```
class cholesky_jitter(_dtype_value_context):
    """
    The jitter value passed to `psd_safe_cholesky` when using cholesky solves.
    Default: None
    """

    _global_value = None

    @classmethod
    def value(cls, dtype=None):
        if dtype is None:
            warnings.warn(
                "cholesky_jitter is now a _dtype_value_context and should be called with a dtype argument",
                DeprecationWarning,
            )
            return None
        return super().value(dtype=dtype)
```

mshvartsman:
I like it, though if we return None I think it breaks the new `psd_safe_cholesky` if you pass `None` into it? Let me know what you think of just returning the float or double default there (or 0?).

Balandat:
Well on the first call it returns `None` to the `psd_safe_cholesky`, but then internally `psd_safe_cholesky` calls it again with an actual data type so all should be good? 

Looks you have some isort failure, o/w this looks good.

gpleiss (approved):

gpleiss:
This looks good to me - and I think the deprecation we have in place will work for BoTorch. Thanks @mshvartsman ! I'll fix the linting issues.

Issue #1415, opened 2021-01-04
[Bug] Possible Memory Leak in UnwhitenedVariationalStrategy


wjmaddox:
# 🐛 Bug

This is a gpytorch-only issue reproducing https://github.com/pytorch/botorch/issues/641 . When using an `UnwhitenedVariationalStrategy`, it looks like something is building up in the (GPU) memory usage over multiple posterior calls that is causing out of memory issues.

## To reproduce

** Code snippet to reproduce **
```python
import torch
from gpytorch.kernels import MaternKernel, ScaleKernel
from gpytorch.means import ZeroMean

from gpytorch.models import ApproximateGP
from gpytorch.variational import CholeskyVariationalDistribution
from gpytorch.variational import UnwhitenedVariationalStrategy, VariationalStrategy

### define variational model
class VariationalGPModel(ApproximateGP):
    def __init__(self, inducing_points, likelihood, *args, **kwargs):
        variational_distribution = CholeskyVariationalDistribution(inducing_points.size(0))
        # this works
#         variational_strategy = VariationalStrategy(
#             self, inducing_points, variational_distribution, learn_inducing_locations=False
#         )       
        # this OOMs
        variational_strategy = UnwhitenedVariationalStrategy(
            self, inducing_points, variational_distribution, learn_inducing_locations=False
        )
        super(VariationalGPModel, self).__init__(variational_strategy)
        self.mean_module = ZeroMean()
        self.covar_module = ScaleKernel(
            MaternKernel(
                ard_num_dims=2,
                nu=0.5,
            ),
        )
        self.likelihood = likelihood

    def forward(self, x):
        mean_x = self.mean_module(x)
        covar_x = self.covar_module(x)
        return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)

#### run simulated BO loop
  if torch.cuda.is_available():
      device = torch.device("cuda:3")
  else:
      device = torch.device("cpu")
  
  current_x = torch.randn(10, 2, device=device)
  current_y = torch.randn(10, device=device)
  
  test_x = torch.randn(1000, 2, device=device)

  induc_points = torch.rand(900, 2)
  
  model = VariationalGPModel(
      inducing_points = induc_points,
      likelihood=None,
  ).to(device)

  for step in range(500):
      if step > 0 and step % 25 == 0:
          print("Beginning step ", step)

      start = time.time()
      model.train()
      model.zero_grad()
          
      model.eval()
      pred = model(test_x)
[...]

PR #1480, opened 2021-02-12
Clone Cached Solve in CachedCGLazyTensor

wjmaddox:
Provides a resolution to #1415. I think the slicing in https://github.com/cornellius-gp/gpytorch/blob/11dce4f1172bdfbf55d090b313b18ff554eb7587/gpytorch/lazy/cached_cg_lazy_tensor.py#L199 is in place so we need to perform an operation to ensure that the solve is not being selected in place (which is likely the cause of the gradient buildup).

### DIFF
--- a/gpytorch/lazy/cached_cg_lazy_tensor.py
+++ b/gpytorch/lazy/cached_cg_lazy_tensor.py
@@ -202,7 +202,7 @@ class CachedCGLazyTensor(LazyTensor):
                 if num_tridiag:
                     return torch.cat([probe_vector_solves, solve], -1), tmats
                 else:
-                    return solve
+                    return solve.clone()
 
         if settings.debug.on():
             warnings.warn(


Balandat (commented):
I don't really understand why this causes the issue since `solve` according to https://github.com/cornellius-gp/gpytorch/blob/11dce4f1172bdfbf55d090b313b18ff554eb7587/gpytorch/lazy/cached_cg_lazy_tensor.py#L124 should already be detached?

Maybe @gpleiss or @jacobrgardner who have a better understanding of this code have thoughts.

jacobrgardner:
I know this is somewhat orthogonal to this specific PR, but I feel like in the long run `UnwhitenedVariationalStrategy` should just be refactored to use Cholesky at this point. It would be a massive simplification (`CachedCGLazyTensor` is super gross), and CIQ+Whitened is a much better iterative method for variational inference than CG+Unwhitened.

Balandat:
> `CachedCGLazyTensor` is super gross

Couldn't agree more...

gpleiss:
Agreed. @wjmaddox since I'm doing some other spring cleaning, I'll put up a PR to get rid of CachedCGLazyTensor.

Issue #1481, opened 2021-02-15
[Bug] Unexpected behavior -- initial value for constraint sets raw value, not transformed value


syncrostone:
# 🐛 Bug

Setting the initial value of a parameter constraint sets the raw parameter to that value, not the transformed parameter. I expect the transformed parameter to be set as that is what I and I expect other users would want to set as the initial value. 

If setting the raw parameter is the expected / desired behavior, then that should be clarified in the docs.

## To reproduce

** Code snippet to reproduce **
```python
import torch
import gpytorch
initial_value = torch.tensor(1.)
kernel = gpytorch.kernels.RBFKernel(lengthscale_constraint=gpytorch.constraints.Positive(initial_value=initial_value))
print("kernel lengthscale: %f \nkernel raw_lengthscale: %f"% (kernel.lengthscale, kernel.raw_lengthscale))
```

** Stack trace/error message **
```
kernel lengthscale: 1.313262 
kernel raw_lengthscale: 1.000000
```

## Expected Behavior

I expect the following output
```
kernel lengthscale: 1.00000
kernel raw_lengthscale: 0.541325
```

because I expect the above code snippet to be equivalent to 
```python
import torch
import gpytorch
initial_value = torch.tensor(1.)
kernel = gpytorch.kernels.RBFKernel()
kernel.lengthscale= initial_value
print("kernel lengthscale: %f \nkernel raw_lengthscale: %f"% (kernel.lengthscale, kernel.raw_lengthscale))
```

## System information

**Please complete the following information:**
- GPyTorch Version 1.3.1
- PyTorch Version 1.7.1+cu101

gpleiss:
Will be closed by #1482 . Thanks @syncrostone !

PR #1482, opened 2021-02-15
inverse transform the initial value in constraints.py

syncrostone:
Addresses #1481 if you agree that #1481 is indeed unexpected behavior


gpytorch/constraints/constraints.py
```diff
@@ -41,6 +41,11 @@ def __init__(self, lower_bound, upper_bound, transform=sigmoid, inv_transform=in
         if transform is not None and inv_transform is None:
             self._inv_transform = _get_inv_param_transform(transform)
 
+        if initial_value is not None:
+            if not isinstance(initial_value, torch.Tensor):
+                initial_value = torch.tensor(initial_value)
+            self._initial_value = self.inverse_transform(initial_value)
```
gpleiss, line 47:
```suggestion
            self._initial_value = self.inverse_transform(initial_value)
        else:
            self._initial_value = None
```

gpleiss (approved):
Looks good!  I removed the earlier `self._initial_value = initial_value` just to clean up the code a bit.

PR #1484, opened 2021-02-15
Add a Dirichlet Classification Likelihood

wjmaddox:
Provides an implementation of https://papers.nips.cc/paper/2018/file/b6617980ce90f637e68c3ebe8b9be745-Paper.pdf which performs classification as fixed noise heteroscedastic regression. I've included an example notebook as well as unit tests for the non-batch case.


gpleiss:
Looks good! A few nits that I will fix:
- `DirichletClassificationLikelihood` needs to be added to the likelihood docs
- The headers for the example notebook should be second-level headings: otherwise it screws up the TOC

Screen Shot 2021-02-17 at 3 26 28 PM

gpleiss (approved):

PR #1494, opened 2021-02-18
Add a Diagonalization cache to LazyTensor

### LINKED ISSUE #1395: [Feature Request] Allow a diagonalize operation analogous to symeig + root decomposition
# 🚀 Feature Request

Create a `lazy_tensor.diagonalize()` operation analogous to `torch.symeig` that caches the "eigenvectors" and "eigenvalues" similar to how `root_decomposition()` caches the root of the matrix. 

Diagonalize is slightly more general than `root_decomposition()` and `root_inv_decomposition()` as root decompositions can be straightforwardly created from diagonalized lazy tensors.

This will also enable setting different methods to perform root_inv_decompositions, rather than just Lanczos.

## Motivation

**Is your feature request related to a problem? Please describe.**

a) It's currently tricky to use approximate eigen-decompositions for Kronecker solves of the form `(K + \sigma^2 I)` while also ensuring that `K` has the correct root decomposition, e.g. for Lanczos. This pops up a bit in sampling routines for larger Kronecker structured models, e.g. [botorch PR #631](https://github.com/pytorch/botorch/pull/631).

b) The current Kronecker plus diagonal solve is only correct for `K + c I` and I'd like to implement `K + D` for general diagonal, but Kronecker structured diagonals.

d) Implementing Jordan normal form decompositions for general lazy tensors.

## Pitch

**Describe the solution you'd like**
```python
cached(name="diagonalization")
def diagonalization(self, method: Optional[str] = None): 
    if method == "lanczos":
          qmat, tmat = lanczos(self.matmul, .....)
          evals, eigenvectors = lanczos_tridiag_to_diag(tmat)
          return evals, qmat @ eigenvectors
    if method == "symeig":
         evals, eigenvectors = torch.symeig(self.evaluate(), eigenvectors = True)
         return evals, eigenvectors

def root_decomposition(self, method=None):
    if hasattr("diagonalization"):
        evals, eigenvectors = self.diagonalization(method=method)
        return eigenvectors.matmul(evals.clamp(0.0).sqrt())
```
    

**Describe alternatives you've considered**
It's possible to do all of these manually but the book keeping is pretty intense.

**Are you willing to open a pull request?** (We LOVE contributions!!!)

Yes, I can take a look at it and setup a base PR.

## Additional context

Will be quite useful for LinearOperator.

wjmaddox:
Resolves #1395 and adds a diagonalization cache to lazy tensor. If the matrix is under the max cholesky size, we call symeig, otherwise, we directly call lanczos.

The usecases here should enable expanded usage of large scale Kronecker lazy tensors (e.g. large multitask models) as well as computing Jordan normal forms in linearoperator (eventually).

ToDos (Edit):
- [x] Diagonalization function for autograd.
- [x] Switch Kronecker symeig solves for diagonalization.

### DIFF
--- a/gpytorch/functions/_diagonalization.py
+++ b/gpytorch/functions/_diagonalization.py
@@ -44,7 +44,7 @@ class Diagonalization(Function):
             q_mat = q_mat.unsqueeze(0)
             t_mat = t_mat.unsqueeze(0)
 
-        mins = torch.diagonal(t_mat, dim1=-1, dim2=-2).min(dim=-1, keepdim=True).solution
+        mins = torch.diagonal(t_mat, dim1=-1, dim2=-2).min(dim=-1, keepdim=True)[0]
         jitter_val = settings.tridiagonal_jitter.value()
         jitter_mat = torch.diag_embed(jitter_val * mins).expand_as(t_mat)
         eigenvalues, eigenvectors = lanczos.lanczos_tridiag_to_diag(t_mat + jitter_mat)
--- a/gpytorch/functions/_diagonalization.py
+++ b/gpytorch/functions/_diagonalization.py
@@ -45,10 +45,9 @@ class Diagonalization(Function):
             q_mat = q_mat.unsqueeze(0)
             t_mat = t_mat.unsqueeze(0)
 
-        mins = lazify(t_mat).diag().min(dim=-1, keepdim=True)[0].unsqueeze(-1)
-        jitter_mat = (settings.tridiagonal_jitter.value() * mins) * torch.eye(
-            t_mat.size(-1), device=t_mat.device, dtype=t_mat.dtype
-        ).expand_as(t_mat)
+        mins = torch.diagonal(t_mat, dim1=-1, dim2=-2).min(dim=-1, keepdim=True).solution
+        jitter_val = settings.tridiagonal_jitter.value()
+        jitter_mat = torch.diag_embed(jitter_val * mins).expand_as(t_mat)
         eigenvalues, eigenvectors = lanczos.lanczos_tridiag_to_diag(t_mat + jitter_mat)
 
         # Get orthogonal matrix and eigenvalues
@@ -79,7 +78,7 @@ class Diagonalization(Function):
         # (\tilde K)_{ij} = 1_{i\neq j} (\sigma_i - \sigma_j)^{-1}
         # add a small amount of jitter to ensure that no zeros are produced
         kmat = (eigenvalues.unsqueeze(-1) - eigenvalues.unsqueeze(-2) + 1e-10).reciprocal()
-        kmat[..., torch.arange(kmat.shape[-1]), torch.arange(kmat.shape[-1])] = 0.0
+        torch.diagonal(kmat, dim1=-1, dim2=-2).zero_()
 
         # dU = U(\tilde K^T \hadamard (U^T dL/dU)U^T
         inner_term = kmat.transpose(-1, -2) * q_mat.transpose(-1, -2).matmul(evecs_grad_output)
--- a/gpytorch/functions/_diagonalization.py
+++ b/gpytorch/functions/_diagonalization.py
@@ -68,32 +68,28 @@ class Diagonalization(Function):
 
     @staticmethod
     def backward(ctx, evals_grad_output, evecs_grad_output):
-        # backwards pass uses Power Iteration style gradients from
-        # Backpropagation Friendly Eigen-Decomposition, Wang, et al, NeurIPS, 2019
-        # and Matrix Backpropagation for Deep Networks with Structured Lazyers,
-        # Ionescu, CVPR, 2015.
+        # backwards pass uses explicit gradients from
+        # Matrix Backpropagation for Deep Networks with Structured Lazyers,
+        # Ionescu, et al CVPR, 2015. https://arxiv.org/pdf/1509.07838.pdf
+        # TODO: check matrix friendly backpropagation
 
-        # TODO: ensure that this is actually Wang et al's Eq. 10
-
-        # matrix_args = ctx.saved_tensors[:-2]
         q_mat = ctx.saved_tensors[-2]
         eigenvalues = ctx.saved_tensors[-1]
 
-        v_k = q_mat[..., -1].unsqueeze(-1)
-        # num_iter = q_mat.shape[-1]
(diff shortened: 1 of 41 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -1329,6 +1329,40 @@ def requires_grad_(self, val):
...
+            )
+
+        if method is None:
+            if self.size(-1) <= settings.max_cholesky_size.value():
+                method = "symeig"
+            else:
+                method = "lanczos"
+
+        if method == "lanczos":
+            # from ..utils.lanczos import lanczos_tridiag, lanczos_tridiag_to_diag
```
Balandat, line 1347:
```suggestion
```

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1381,6 +1415,19 @@ def root_decomposition(self, method: Optional[str] = None):
...
+            diagonalization = None
+            try:
+                diagonalization = get_from_cache(self, "diagonalization")
+            except CachingError:
+                pass
+
+            if diagonalization is not None:
+                evals, evecs = diagonalization
+                F = evecs * evals.clamp(1e-7).sqrt().unsqueeze(-2)
+                return RootLazyTensor(F)
```
Balandat, line 1428:
```suggestion
            try:
                evals, evecs = get_from_cache(self, "diagonalization")
                F = evecs * evals.clamp(1e-7).sqrt().unsqueeze(-2)
                return RootLazyTensor(F)
            except CachingError:
                pass                        
```

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1422,6 +1469,19 @@ def root_inv_decomposition(self, initial_vectors=None, test_vectors=None):
...
+        diagonalization = None
+        try:
+            diagonalization = get_from_cache(self, "diagonalization")
+        except CachingError:
+            pass
+
+        if diagonalization is not None:
+            evals, evecs = diagonalization
+            F = evecs * evals.clamp(1e-7).sqrt().reciprocal().unsqueeze(-2)
+            return RootLazyTensor(F)
```
Balandat, line 1482:
```suggestion
        # check to see if we have already run lanczos for a diagonalization
        try:
            evals, evecs = get_from_cache(self, "diagonalization")
            F = evecs * evals.clamp(1e-7).sqrt().reciprocal().unsqueeze(-2)
            return RootLazyTensor(F)
        except CachingError:
            pass
```

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1329,6 +1329,40 @@ def requires_grad_(self, val):
         self.requires_grad = val
         return self
 
+    @cached(name="diagonalization")
+    def diagonalization(self, method: Optional[str] = None):
```
Balandat, line 1333: Add a docstring here? In particular, list the available methods.

Balandat:
I'm running into a similar test failure with BoTorch:

> a view of a leaf Variable that requires grad is being used in an in-place operation.

I believe this is due to some upstream changes in PyTorch. Looking into this now.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1329,6 +1329,39 @@ def requires_grad_(self, val):
...
+        if not self.is_square:
+            raise RuntimeError(
+                "diagonalization only operates on (batches of) square (symmetric) LazyTensors. "
```
Balandat, line 1336: I we wanted to be fancy we could also allow diagonalization of non-symmetric matrices if we don't restrict things to unitary matrices. E.g. for a `TriangularLazyTensor` we can easily read off the evals and construct the evecs by backward substitution. I guess that's something for down the road maybe...

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1329,6 +1330,45 @@ def requires_grad_(self, val):
         self.requires_grad = val
         return self
 
+    @cached(name="diagonalization")
+    def diagonalization(self, method: Optional[str] = None):
+        """
+        Returns a (usually partial) diagonalization of a symmetric PSD matrix.
+        Options are either "lanczos" or "symeig". "lanczos" runs Lanczos while
+        "symeig" runs torch.symeig.
```
Balandat, line 1338: Maybe clarify that this doesn't necessarily mean calling symeig on the full dense matrix.
```suggestion
        "symeig" runs LazyTensor.symeig.
```

Balandat (commented):
some specific comments, will have to take a closer look at the reference

gpytorch/functions/_diagonalization.py
```diff
@@ -0,0 +1,95 @@
...
+            q_mat = q_mat.unsqueeze(-3)
+            t_mat = t_mat.unsqueeze(-3)
+        if t_mat.ndimension() == 3:  # If we only used one probe vector
+            q_mat = q_mat.unsqueeze(0)
+            t_mat = t_mat.unsqueeze(0)
+
+        mins = lazify(t_mat).diag().min(dim=-1, keepdim=True)[0].unsqueeze(-1)
+        jitter_mat = (settings.tridiagonal_jitter.value() * mins) * torch.eye(
+            t_mat.size(-1), device=t_mat.device, dtype=t_mat.dtype
+        ).expand_as(t_mat)
```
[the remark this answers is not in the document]
wjmaddox: It was copied over from root decomposition but sure.

gpytorch/functions/_diagonalization.py
```diff
@@ -0,0 +1,95 @@
...
+import torch
+from torch.autograd import Function
+
+from .. import settings
+from ..utils import lanczos
+
+
+class Diagonalization(Function):
+    @staticmethod
+    def forward(ctx, representation_tree, device, dtype, matrix_shape, max_iter, batch_shape, *matrix_args):
```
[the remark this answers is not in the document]
wjmaddox: Passing in the lazy tensor instead of the representation tree causes errors in lazy evaluated kernel tensor (which doesn't want to use _matmul without checkpointing).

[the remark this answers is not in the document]
wjmaddox: So it's probably best to leave it as it is and just modify when i put it into LinearOperator (after this PR I guess).

Balandat (approved):

PR #1495, opened 2021-02-18
Avoid Lanczos decomposition with non-deterministic logdet probes

### LINKED ISSUE #1491: [Bug] `gpytorch.settings.deterministic_probes` computes an extra Lanczos factorization
# 🐛 Bug

In preconditioned CG mode, we make the following lining calls for each training iteration:

```
LinAlg (Verbose) - DEBUG - Running Pivoted Cholesky on a torch.Size([5000, 5000]) RHS for 15 iterations.

# these two calls shouldn't exist
LinAlg (Verbose) - DEBUG - Running Lanczos on a torch.Size([5000, 5000]) matrix with a torch.Size([5000, 1]) RHS for 100 iterations.
LinAlg (Verbose) - DEBUG - Running symeig on a matrix of size torch.Size([9, 9]).

LinAlg (Verbose) - DEBUG - Running CG on a torch.Size([5000, 11]) RHS for 1000 iterations (tol=1). Output: torch.Size([5000, 11]).
LinAlg (Verbose) - DEBUG - Running symeig on a matrix of size torch.Size([10, 4, 4]).

LinAlg (Verbose) - INFO - Iter 1/50 - Loss: 0.808   lengthscale: 0.693   noise: 0.693
```

This Lanczos/symeig calls come from this line in [inv_quad_logdet](https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/functions/_inv_quad_log_det.py#L98)

```python
            else:  # When preconditioning, probe vectors must be drawn from N(0, P)
                if precond_lt.size()[-2:] == torch.Size([1, 1]):
                    covar_root = precond_lt.evaluate().sqrt()
                else:
                    covar_root = precond_lt.root_decomposition().root
```

Before the `gpytorch.settings.deterministic_probes` option was added (#929), the code looked like this:

```python
            else:  # When preconditioning, probe vectors must be drawn from N(0, P)
                probe_vectors = precond_lt.zero_mean_mvn_samples(num_random_probes)
                # ...
```

This is much more efficient, as it does not perform a root decomposition.

@jacobrgardner there is an obvious easy fix: we don't call `root_decomposition` unless we're using deterministic probes. However, I think this makes the `deterministic_probes` setting really undesirable - the slight variance reductions gains we get from it come at the cost of an extra linalg routine.

Thoughts?

gpleiss:
[Fixes #1491]

### DIFF
--- a/gpytorch/functions/_inv_quad_log_det.py
+++ b/gpytorch/functions/_inv_quad_log_det.py
@@ -92,12 +92,15 @@ class InvQuadLogDet(Function):
                     probe_vectors = probe_vectors.expand(*batch_shape, matrix_shape[-1], num_random_probes)
                     probe_vector_norms = probe_vector_norms.expand(*batch_shape, 1, num_random_probes)
             else:  # When preconditioning, probe vectors must be drawn from N(0, P)
-                if precond_lt.size()[-2:] == torch.Size([1, 1]):
-                    covar_root = precond_lt.evaluate().sqrt()
-                else:
-                    covar_root = precond_lt.root_decomposition().root
-
                 if settings.deterministic_probes.on():
+                    # NOTE: calling precond_lt.root_decomposition() is expensive
+                    # because it requires Lanczos
+                    # We don't have any other choice for when we want to use deterministic probes, however
+                    if precond_lt.size()[-2:] == torch.Size([1, 1]):
+                        covar_root = precond_lt.evaluate().sqrt()
+                    else:
+                        covar_root = precond_lt.root_decomposition().root
+
                     warnings.warn(
                         "Deterministic probes will currently work only if you aren't training multiple independent"
                         " models simultaneously.",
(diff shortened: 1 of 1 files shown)


gpytorch/functions/_inv_quad_log_det.py
```diff
@@ -116,13 +119,6 @@ def forward(
 
                     probe_vectors = covar_root.matmul(base_samples).permute(-1, *range(precond_lt.dim() - 1))
                 else:
-                    base_samples = torch.randn(
-                        *precond_lt.batch_shape,
-                        covar_root.size(-1),
-                        num_random_probes,
-                        dtype=precond_lt.dtype,
-                        device=precond_lt.device,
-                    )
```
Balandat, line 125: gotta love 'em nullops

Balandat (approved):

PR #1499, opened 2021-02-19
Root Updates for Low Rank Additions and Row/Column Appending

wjmaddox:
This is a PR to add support for #1497. 

- Enables a more general usage of the row/column appending root update strategy in [exact_prediction_strategy](https://github.com/cornellius-gp/gpytorch/blob/c74b8ed8d590fcb1d79808d9ee7cde168588ef99/gpytorch/models/exact_prediction_strategies.py#L183).
- Implements a strategy for general root updating of low rank matrices, e.g. `B = A + dd'`. 

The current implementation version is to just update the lazy tensor but then to overwrite the root / root inverse decomposition in the memoize cache to use these updates.

One gotcha is that the low rank matrix update will currently return a SumLazyTensor but if we do many SumLazyTensor updates, this can get quite memory intensive.

ToDos:
- [x] Unit test for row/column appending
- [x] Implement inverse root updates in the added rows setting.
- [x] Add a flag into `root_inv_decomposition` to ensure that there's control over which type of root inverse is used (lanczos or cholesky).
- [x] Change over the exact prediction strategy to use this version.
- [x] Unit test for low rank updates.

Example usage:
```python
import torch
import gpytorch
from gpytorch.lazy import lazify
from gpytorch.kernels import MaternKernel

# low rank addition
a = torch.randn(500, 50)
lt = lazify(a.t() @ a)
res = lt.add_low_rank(torch.randn(50, 5))

res.root_decomposition() #is accomplished without calling lanczos/love
res.root_inv_decomposition() # same here

# adding rows and columns
kernel = MaternKernel()
xtrain = torch.randn(30)
xtest = torch.randn(2)

lt = kernel(xtrain)
fant_train = kernel(xtest, xtrain)
fant_fant = kernel(xtest)
new_lt = lt.cat_rows(fant_train, fant_fant)
new_lt.root_decomposition() # is a root lt rather than a cholesky
```

Why:

- Faster fantasy strategies for gpytorch exact gps.
- efficient Matheron's rule posterior sampling.
- fast low rank updates to PSD matrices generally.

### DIFF
--- a/gpytorch/lazy/chol_lazy_tensor.py
+++ b/gpytorch/lazy/chol_lazy_tensor.py
@@ -106,6 +106,6 @@ class CholLazyTensor(RootLazyTensor):
 
         return inv_quad_term, logdet_term
 
-    def root_inv_decomposition(self, initial_vectors=None, test_vectors=None):
+    def root_inv_decomposition(self, method=None, initial_vectors=None, test_vectors=None):
         inv_root = self.root.inverse()
         return RootLazyTensor(inv_root._transpose_nonbatch())
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -823,7 +823,7 @@ class LazyTensor(ABC):
             # we don't need the batch shape here, thanks to broadcasting
             Eye = torch.eye(new_root.shape[-2], device=new_root.device, dtype=new_root.dtype)
             new_inv_root = torch.triangular_solve(Eye, new_root, upper=False)[0]
-            new_inv_root = lazify(new_inv_root.transpose(-1, -2))
+            new_inv_root = new_inv_root.transpose(-1, -2)
         else:
             Q, R = stable_qr(new_root)
 
@@ -833,8 +833,8 @@ class LazyTensor(ABC):
                 # solve with least squares
                 new_inv_root = torch.lstsq(Q.transpose(-2, -1), R).solution.transpose(-2, -1)
 
-        add_to_cache(new_lazy_tensor, "root_decomposition", RootLazyTensor(new_root))
-        add_to_cache(new_lazy_tensor, "root_inv_decomposition", RootLazyTensor(new_inv_root))
+        add_to_cache(new_lazy_tensor, "root_decomposition", RootLazyTensor(TriangularLazyTensor(new_root)))
+        add_to_cache(new_lazy_tensor, "root_inv_decomposition", RootLazyTensor(TriangularLazyTensor(new_inv_root)))
 
         return new_lazy_tensor
 
@@ -869,6 +869,7 @@ class LazyTensor(ABC):
         from . import lazify
         from .root_lazy_tensor import RootLazyTensor
         from .sum_lazy_tensor import SumLazyTensor
+        from .triangular_lazy_tensor import TriangularLazyTensor
 
         if not isinstance(self, SumLazyTensor):
             new_lazy_tensor = self + lazify(low_rank_mat.matmul(low_rank_mat.transpose(-1, -2)))
@@ -894,6 +895,8 @@ class LazyTensor(ABC):
 
         # first get LL^T = A
         current_root = self.root_decomposition(method=root_decomp_method).root
+        return_triangular = isinstance(current_root, TriangularLazyTensor)
+
         # and MM^T = A^{-1}
         current_inv_root = self.root_inv_decomposition(method=root_inv_decomp_method).root.transpose(-1, -2)
 
@@ -932,6 +935,10 @@ class LazyTensor(ABC):
         inner_inv_root = U.matmul(torch.diag_embed(stacked_inv_root_S))
         # finally \tilde{L}^{-1} = L^{-1} U \tilde{S}^{-1}
         updated_inv_root = current_inv_root.transpose(-1, -2).matmul(inner_inv_root)
+
+        if return_triangular:
+            updated_root = TriangularLazyTensor(updated_root)
+            updated_inv_root = TriangularLazyTensor(updated_inv_root)
 
         add_to_cache(new_lazy_tensor, "root_decomposition", RootLazyTensor(updated_root))
         add_to_cache(new_lazy_tensor, "root_inv_decomposition", RootLazyTensor(updated_inv_root))
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -877,9 +877,8 @@ class LazyTensor(ABC):
             low_rank_mat (:obj:`torch.tensor`): the matrix `B` that we are adding to `A`.
             root_decomp_method (:obj:`str`): how to compute the root decomposition of `A`.
             root_inv_decomp_method (:obj:`str`): how to compute the root inverse decomposition of `A`.
-            generate_roots (:obj:`bool`): whether to generate the root decomposition of :math:`A` even if it has
-                not been created yet.
-
+            generate_roots (:obj:`bool`): whether to generate the root decomposition of :math:`A` even if it
+            has not been created yet.
 
         Returns:
             :obj:`SumLazyTensor`: addition of A and BB^T.
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -800,11 +800,18 @@ class LazyTensor(ABC):
             # haven't found a more elegant way to add a jitter diagonal yet...
             jitter_diag = 1e-6 * torch.sign(Rdiag) * zeroish.to(Rdiag)
             R = R + torch.diag_embed(jitter_diag)
+
         new_inv_root = torch.triangular_solve(Q.transpose(-2, -1), R)[0].transpose(-2, -1)
 
         cross_mat = lazify(cross_mat)
         new_mat = lazify(new_mat)
-        new_lazy_tensor_1 = CatLazyTensor(self, cross_mat, dim=-2)
+        if self.ndimension() < cross_mat.ndimension():
+            from .batch_repeat_lazy_tensor import BatchRepeatLazyTensor
+
+            old_tsr = BatchRepeatLazyTensor(self, batch_shape)
+        else:
+            old_tsr = self
+        new_lazy_tensor_1 = CatLazyTensor(old_tsr, cross_mat, dim=-2)
         new_lazy_tensor_2 = CatLazyTensor(cross_mat.transpose(-1, -2), new_mat, dim=-2)
         new_lazy_tensor = CatLazyTensor(new_lazy_tensor_1, new_lazy_tensor_2, dim=-1)
 
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -801,7 +801,11 @@ class LazyTensor(ABC):
             jitter_diag = 1e-6 * torch.sign(Rdiag) * zeroish.to(Rdiag)
             R = R + torch.diag_embed(jitter_diag)
 
-        new_inv_root = torch.triangular_solve(Q.transpose(-2, -1), R)[0].transpose(-2, -1)
+        if R.shape[-2] == R.shape[-1]:
+            new_inv_root = torch.triangular_solve(Q.transpose(-2, -1), R)[0].transpose(-2, -1)
+        else:
+            # solve with least squares
+            new_inv_root = torch.lstsq(Q.transpose(-2, -1), R)[0].transpose(-2, -1)
 
         cross_mat = lazify(cross_mat)
         new_mat = lazify(new_mat)
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -739,6 +739,7 @@ class LazyTensor(ABC):
         TODO: add reference.
         """
         from .cat_lazy_tensor import CatLazyTensor
+        from .root_lazy_tensor import RootLazyTensor
 
         batch_shape = fant_train.shape[:-2]
         L = self.root_decomposition().root.evaluate()
@@ -761,7 +762,8 @@ class LazyTensor(ABC):
         new_lazy_tensor_1 = CatLazyTensor(self, fant_train, dim=-2)
         new_lazy_tensor_2 = CatLazyTensor(fant_train.transpose(-1, -2), fant_fant, dim=-2)
         new_lazy_tensor = CatLazyTensor(new_lazy_tensor_1, new_lazy_tensor_2, dim=-1)
-        add_to_cache(new_lazy_tensor, "root_decomposition", new_root)
+
+        add_to_cache(new_lazy_tensor, "root_decomposition", RootLazyTensor(new_root))
 
         # TODO: update the inverse root too.
         return new_lazy_tensor
@@ -774,6 +776,7 @@ class LazyTensor(ABC):
         TODO: add reference.
         """
         from . import lazify
+        from .root_lazy_tensor import RootLazyTensor
 
         # TODO: add a check if this should NOT be a lazy addition
         new_tensor = self + lazify(low_rank_mat.matmul(low_rank_mat.transpose(-1, -2)))
@@ -823,8 +826,8 @@ class LazyTensor(ABC):
         # finally \tilde{L}^{-1} = L^{-1} U \tilde{S}^{-1}
         updated_inv_root = current_inv_root.transpose(-1, -2).matmul(inner_inv_root)
 
-        add_to_cache(new_tensor, "root_decomposition", updated_root)
-        add_to_cache(new_tensor, "root_inv_decomposition", updated_inv_root)
+        add_to_cache(new_tensor, "root_decomposition", RootLazyTensor(updated_root))
+        add_to_cache(new_tensor, "root_inv_decomposition", RootLazyTensor(updated_inv_root))
 
         return new_tensor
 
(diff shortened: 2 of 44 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -731,6 +731,106 @@ def add_jitter(self, jitter_val=1e-3):
         diag = torch.tensor(jitter_val, dtype=self.dtype, device=self.device)
         return self.add_diag(diag)
 
+    def cat_rows(self, fant_train, fant_fant):
+        """
+        Concatenates new rows and columns to the matrix that this LazyTensor represents, e.g.
+        C = ((A; B); B', D).
```
Balandat, line 737: ... where `A` is the existing lazy tensor and `B` and `D` are the new components...

gpytorch/lazy/lazy_tensor.py
```diff
@@ -731,6 +731,106 @@ def add_jitter(self, jitter_val=1e-3):
...
+
+        TODO: add reference.
+        """
+        from .cat_lazy_tensor import CatLazyTensor
+        from .root_lazy_tensor import RootLazyTensor
+
+        batch_shape = fant_train.shape[:-2]
+        L = self.root_decomposition().root.evaluate()
+
+        L_inverse = self.root_inv_decomposition().root.evaluate()
```
Balandat, line 747: Why do things with dense tensors?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -731,6 +731,106 @@ def add_jitter(self, jitter_val=1e-3):
...
+        # Form new root Z = [L 0; lower_left schur_root]
+        num_fant = schur_root.size(-2)
+        new_root = torch.zeros(*batch_shape, m + num_fant, n + num_fant, device=L.device, dtype=L.dtype)
+        new_root[..., :m, :n] = L
+        new_root[..., m:, : lower_left.shape[-1]] = lower_left
+        new_root[..., m:, n : (n + schur_root.shape[-1])] = schur_root
+
+        new_lazy_tensor_1 = CatLazyTensor(self, fant_train, dim=-2)
+        new_lazy_tensor_2 = CatLazyTensor(fant_train.transpose(-1, -2), fant_fant, dim=-2)
+        new_lazy_tensor = CatLazyTensor(new_lazy_tensor_1, new_lazy_tensor_2, dim=-1)
```
Balandat, line 764: I wonder if it makes sense to have some kind of `LazyTensorWithNewRowsAndCols` object that then internally just uses the base lazy tensor and the `fant_train` and `fant_fant` components.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -731,6 +731,106 @@ def add_jitter(self, jitter_val=1e-3):
         diag = torch.tensor(jitter_val, dtype=self.dtype, device=self.device)
         return self.add_diag(diag)
 
+    def cat_rows(self, fant_train, fant_fant):
```
Balandat, line 734: probably want to use different variables here to be more generic than fantasizing

Balandat (commented):
some initial comments, will have to get back to `add_low_rank`

gpytorch/lazy/lazy_tensor.py
wjmaddox: These need to be dense roots so that the matrix multiplications in L751 below is okay. In general, fant_train will probably be dense.

wjmaddox:
This PR will slightly change the signature of root inv decomposition so that
```python
lazy_tensor.root_inv_decomposition("lanczos", probe_vectors, test_vectors)
```
is needed to enforce a lanczos call for small matrices.

gpytorch/lazy/root_lazy_tensor.py
```diff
@@ -66,6 +66,9 @@ def _t_matmul(self, rhs):
         # Matrix is symmetric
         return self._matmul(rhs)
 
+    def add_low_rank(self, low_rank_mat, root_decomp_method=None, root_inv_decomp_method="pinverse"):
+        return super().add_low_rank(low_rank_mat, root_inv_decomp_method=root_inv_decomp_method)
```
Balandat, line 70: This shouldn't be necessary, `method` of `root_decomposition` will already be ignored in `RootLazyTensor`

wjmaddox: This is done more to pass in the "pinverse" method for the root inverse decomposition, and to allow that to change if possible.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -731,6 +731,168 @@ def add_jitter(self, jitter_val=1e-3):
...
+    def add_low_rank(
+        self, low_rank_mat, root_decomp_method: Optional[str] = None, root_inv_decomp_method: Optional[str] = None
+    ):
+        """
+        Adds a low rank matrix to the matrix that this LazyTensor represents, e.g.
+        computes A + BB'. We then update both the tensor and its root decomposition.
+
+        We have access to, L and M where A \approx LL^T and A^{-1} \approx MM^T.
+        We compute \tilde{A} = A + VV^T = L(I + M V V^T M^T)L' and then decompose
+        (I + M VV^T M^T) \approx RR^T, using LR as our new root decomposition.
```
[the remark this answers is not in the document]
wjmaddox: I was originally worried that adding these checks would clobber the usecase of `cat_rows` in `get_fantasy_strategy` but it looks like a root inverse decomposition is called there beforehand so it shouldn't actually clobber it.

gpytorch/lazy/chol_lazy_tensor.py
```diff
@@ -106,6 +106,6 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
 
         return inv_quad_term, logdet_term
 
-    def root_inv_decomposition(self, initial_vectors=None, test_vectors=None):
+    def root_inv_decomposition(self, method=None, initial_vectors=None, test_vectors=None):
```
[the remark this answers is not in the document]
Balandat: bump

gpytorch/lazy/lazy_tensor.py
```diff
@@ -731,6 +731,221 @@ def add_jitter(self, jitter_val=1e-3):
...
+        from .root_lazy_tensor import RootLazyTensor
+        from .triangular_lazy_tensor import TriangularLazyTensor
+        from . import lazify
+
+        cross_mat = lazify(cross_mat)
+        batch_shape = cross_mat.shape[:-2]
+        new_mat = lazify(new_mat)
+        old_tsr = self
+        if old_tsr.ndimension() < cross_mat.ndimension():
+            from .batch_repeat_lazy_tensor import BatchRepeatLazyTensor
```
Balandat, line 777: may as well do the import above when importing all the other lazies

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1422,57 +1643,73 @@ def root_inv_decomposition(self, initial_vectors=None, test_vectors=None):
...
-            _, best_solve_index = residuals.min(0)
-            inv_root = inv_roots[best_solve_index].squeeze(0)
+                # Choose solve that best fits
+                _, best_solve_index = residuals.min(0)
+                inv_root = inv_roots[best_solve_index].squeeze(0)
 
-        else:
-            inv_root = inv_roots
+            else:
+                inv_root = inv_roots
```
Balandat, line 1697: this is a pretty big if block. For better readability / avoiding huge indents, how about moving this into a separate function and just call that here?

gpytorch/lazy/chol_lazy_tensor.py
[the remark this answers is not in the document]
wjmaddox: resolved.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -731,6 +731,168 @@ def add_jitter(self, jitter_val=1e-3):
...
+        # Use pseudo-inverse of Z as new inv root
+
+        if new_root.shape[-1] <= 2048:
+            # Dispatch to CPU so long as pytorch/pytorch#22573 is not fixed
+            device = new_root.device
+            Q, R = torch.qr(new_root.cpu())
+            Q = Q.to(device)
+            R = R.to(device)
+        else:
+            Q, R = torch.qr(new_root)
```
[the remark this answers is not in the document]
wjmaddox: Where should this one go? `gpytorch.utils.qr`?

gpytorch/lazy/lazy_tensor.py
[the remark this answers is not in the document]
wjmaddox: I've implemented it to not compute a root decomposition unless the root already exists.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -732,6 +762,230 @@ def add_jitter(self, jitter_val=1e-3):
...
+
+        # if the old lazy tensor does not have either a root decomposition or a root inverse decomposition
+        # don't create one
+        does_not_have_roots = _is_in_cache_ignore_args(self, "root_decomposition") or _is_in_cache_ignore_args(
+            self, "root_inv_decomposition"
+        )
+        if not generate_roots and not does_not_have_roots:
+            return new_lazy_tensor
+
+        L = self.root_decomposition().root
```
Balandat, line 823: You're calling `L = delazify(L)` below anyway, might as well `evaluate` once here...

```suggestion
        L = self.root_decomposition().root.evaluate()
```

wjmaddox: This is in there because I need to pluck out if L was a cholesky when lazy. let me move some stuff around.

Balandat (approved):
lgtm. This touches a lot of the internals though, so maybe @jacobrgardner and @gpleiss can take a look and merge it in if this is to their liking.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -732,6 +734,217 @@ def add_jitter(self, jitter_val=1e-3):
...
+            :obj:`LazyTensor`: concatenated lazy tensor with the new rows and columns.
+        """
+        from . import lazify
+        from .batch_repeat_lazy_tensor import BatchRepeatLazyTensor
+        from .cat_lazy_tensor import CatLazyTensor
+        from .root_lazy_tensor import RootLazyTensor
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
+        cross_mat = lazify(cross_mat)
+        batch_shape = cross_mat.shape[:-2]
```
gpleiss, line 777: Is it worth it to make sure that self, cross_mat, and new_mat all have the same batch shape (using the broadcasting helper)?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -732,6 +734,217 @@ def add_jitter(self, jitter_val=1e-3):
         diag = torch.tensor(jitter_val, dtype=self.dtype, device=self.device)
         return self.add_diag(diag)
 
+    def cat_rows(self, cross_mat, new_mat, generate_roots=True):
```
gpleiss, line 737: Maybe this method should take in a **root_decomp_args that get passed into `root_decomposition` - if `generate_roots` is true?

gpytorch/utils/qr.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+
+import torch
+
+
+def stable_qr(mat):
```
gpleiss, line 6: 👍👍👍👍 glad to have this helper method :)

gpytorch/test/lazy_tensor_test_case.py
```diff
@@ -405,6 +405,58 @@ def test_add_diag(self):
...
+        summed_rhs = summed_lt.matmul(rhs)
+        root_rhs = new_lt.root_decomposition().matmul(rhs)
+        self.assertAllClose(root_rhs, summed_rhs, **self.tolerances["root_decomposition"])
+
+        # check that the inverse root decomposition is close
+        summed_solve = torch.solve(rhs.unsqueeze(-1), summed_lt)[0].squeeze(-1)
+        root_inv_solve = new_lt.root_inv_decomposition().matmul(rhs)
+        self.assertAllClose(root_inv_solve, summed_solve, **self.tolerances["root_inv_decomposition"])
+
+    def test_cat_rows(self):
```
gpleiss, line 431: I want to be sure that this test case is capturing all the branching logic. Just to be clear, one of the sub LT classes will go into the `if l_is_triang:` blocks, right?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -732,6 +734,217 @@ def add_jitter(self, jitter_val=1e-3):
...
+
+        return new_lazy_tensor
+
+    def add_low_rank(
+        self,
+        low_rank_mat,
+        root_decomp_method: Optional[str] = None,
+        root_inv_decomp_method: Optional[str] = None,
+        generate_roots: Optional[bool] = True,
+    ):
```
gpleiss, line 847: Same thing here - might be good to pass **root_decomp_kwargs?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -732,6 +734,217 @@ def add_jitter(self, jitter_val=1e-3):
...
+
+        Returns:
+            :obj:`SumLazyTensor`: addition of A and BB^T.
+        """
+        from . import lazify
+        from .root_lazy_tensor import RootLazyTensor
+        from .sum_lazy_tensor import SumLazyTensor
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
+        if not isinstance(self, SumLazyTensor):
```
gpleiss, line 874: I don't think this if block is necessary - this casing happens as part of the SLT constructor.

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -183,68 +183,10 @@ def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_
...
-        # if R is almost singular, add jitter
-        zeroish = Rdiag.abs() < 1e-6
-        if torch.any(zeroish):
-            # can't use in-place operation here b/c it would mess up backward pass
-            # haven't found a more elegant way to add a jitter diagonal yet...
-            jitter_diag = 1e-6 * torch.sign(Rdiag) * zeroish.to(Rdiag)
-            R = R + torch.diag_embed(jitter_diag)
-        new_covar_cache = torch.triangular_solve(Q.transpose(-2, -1), R)[0].transpose(-2, -1)
+        # now update the root and root inverse
+        new_lt = self.lik_train_train_covar.cat_rows(fant_train_covar, fant_fant_covar)
```
gpleiss, line 187: This is awesome! I love how simple this method becomes.

gpytorch/lazy/lazy_tensor.py
wjmaddox: I don't think so. In general, we want to be able to have self be non-batched but cross_mat and new_mat to be batched. That is something like this currently works
```python
x = torch.randn(10, 5)
a = lazify(x.t() @ x) # 5 x 5 full rank

newx = torch.randn(32, 7, 5)
newa = lazify(newx.transpose(-1, -2) @ newx)

crossa = 0.01 * torch.randn(32, 5, 5)
result = a.cat_rows(crossa, newa) # is of shape 32 x 5 x 5
```

gpytorch/test/lazy_tensor_test_case.py
wjmaddox: Yes, most of the lazy tests are going to hit `l_is_triang` but enforcing an error in the false case hits at least constant_mul_lazy_tensor (and probably others that have specialized roots).

gpytorch/lazy/lazy_tensor.py
wjmaddox: I don't see it in [here](https://github.com/cornellius-gp/gpytorch/blob/d433003402e09195a76054b0fb7c3cf4aa9189dd/gpytorch/lazy/sum_lazy_tensor.py#L14). This is really designed to prevent caching memory issues where we continually are adding low rank matrices and then the recursion just builds up a lot (have experienced it in previous iterations of this code). E.g. the situation like `A_T = A_0 + \sum_{i=1}^T a_i a_i' = A_{T-1} + a_T a_T`

gpytorch/lazy/lazy_tensor.py
wjmaddox: And a non-batched cross matrix fails currently, e.g. 
```python
cross = 0.01 * torch.randn(5, 5)
result = a.cat_rows(crossa, newa) # failed
```
which is also the desired behavior.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -732,6 +741,218 @@ def add_jitter(self, jitter_val=1e-3):
...
+        from .cat_lazy_tensor import CatLazyTensor
+        from .root_lazy_tensor import RootLazyTensor
+        from .triangular_lazy_tensor import TriangularLazyTensor
+
+        cross_mat = lazify(cross_mat)
+        batch_shape = cross_mat.shape[:-2]
+        new_mat = lazify(new_mat)
+        old_tsr = self
+        if old_tsr.ndimension() < cross_mat.ndimension():
+            old_tsr = BatchRepeatLazyTensor(old_tsr, batch_shape)
```
Balandat, line 788: @wjmaddox I believe this breaks the broadcasting that we do when computing fantasy models. The issue is that `old_tsr` can itself be batched, so that the `BatchRepeatLazyTensor` messes up the dimensions here. 

Specifically, say `old_tsr` is `3 x 4 x 4` and `cross_mat` is of shape `2 x 3 x 2 x 4` then this `BatchRepeatLazyTensor` will be of shape `2 x 9 x 4 x 4`. Let me try to fix this consistently

gpytorch/lazy/lazy_tensor.py
Balandat: Just saw this resolved one. Didn't age all that well given my comment here: https://github.com/cornellius-gp/gpytorch/pull/1499/files#r594720889

let me see if I can put together a fix for that.

PR #1502, opened 2021-02-22
Add a base sample shape property to MultivariateNormal

wjmaddox:
Adds a base sample shape property to MultivariateNormal. I should have just added this in #1355 instead of the fix I did. I checked and it resolves the failing unit test in botorch (botorch/models/test_higher_order_gp.py), but botorch will need another PR to properly use the base_sample_shape property in its samplers.

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -128,8 +128,7 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
         """
         base_sample_shape = self.event_shape
         if isinstance(self.lazy_covariance_matrix, RootLazyTensor):
-            if self.lazy_covariance_matrix.root.shape[-1] > self.lazy_covariance_matrix.root.shape[-2]:
-                base_sample_shape = self.lazy_covariance_matrix.root.shape[-1:]
+            base_sample_shape = self.lazy_covariance_matrix.root.shape[-1:]
 
         return base_sample_shape
 
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -42,16 +42,8 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
             self.__unbroadcasted_scale_tril = None
             self._validate_args = validate_args
             batch_shape = _mul_broadcast_shape(self.loc.shape[:-1], covariance_matrix.shape[:-2])
-            if not isinstance(self._covar, RootLazyTensor):
-                should_have_separate_event = False
-            else:
-                should_have_separate_event = True
-            if should_have_separate_event:
-                should_have_separate_event = self._covar.root.shape[-1] > self._covar.root.shape[-2]
-            if not should_have_separate_event:
-                event_shape = self.loc.shape[-1:]
-            else:
-                event_shape = self._covar.root.shape[-1:]
+
+            event_shape = self.loc.shape[-1:]
 
             # TODO: Integrate argument validation for LazyTensors into torch.distribution validation logic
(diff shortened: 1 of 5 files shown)


gpytorch/distributions/multivariate_normal.py
```diff
@@ -114,6 +120,19 @@ def get_base_samples(self, sample_shape=torch.Size()):
...
+    def base_sample_shape(self):
+        """
+        Returns the shape of a base sample (without batching) that is used to
+        generate a single sample.
+        """
+        base_sample_shape = self.event_shape
+        if isinstance(self.lazy_covariance_matrix, RootLazyTensor):
+            if self.lazy_covariance_matrix.root.shape[-1] > self.lazy_covariance_matrix.root.shape[-2]:
+                base_sample_shape = self.lazy_covariance_matrix.root.shape[-1:]
+
```
Balandat, line 133: maybe I'm missing something here, but what about the case where the root is a. tall tensor? Shouldn't `self.lazy_covariance_matrix.root.shape[-1:]` always be the `base_sample_shape`? If the reason is to allow easy dealing with varying posterior shapes for different test points (<= matrix size), this is probably something that we should deal with outside of the mvn to separate concerns better.

test/distributions/test_multivariate_normal.py
```diff
@@ -296,6 +296,18 @@ def test_getitem(self):
         assert torch.equal(d.mean, dist.mean[1, 2, 2, :])
         self.assertAllClose(d.covariance_matrix, dist_cov[1, 2, 2, :, :])
 
+    def test_base_sample_shape(self):
```
Balandat, line 299: shoudl probably add explicit checks for the property for all the cases of `lazy_covariance_matrix`.

gpytorch/distributions/multivariate_normal.py
wjmaddox: It's already dealt with gracefully but the updated commit allows for it as well.

Balandat (approved):
Thanks for putting this up, this lgmt on my end. I'll wait for @gpleiss to merge this in so he can double check this doesn't cause any other issues.

gpytorch/distributions/multivariate_normal.py
```diff
@@ -114,6 +120,18 @@ def get_base_samples(self, sample_shape=torch.Size()):
...
 
+    @property
+    def base_sample_shape(self):
+        """
+        Returns the shape of a base sample (without batching) that is used to
+        generate a single sample.
+        """
+        base_sample_shape = self.event_shape
+        if isinstance(self.lazy_covariance_matrix, RootLazyTensor):
+            base_sample_shape = self.lazy_covariance_matrix.root.shape[-1:]
```
gpleiss, line 131: @wjmaddox `event_shape` for multitask MVNs will be a length-2 tuple. It might be worth overwriting this property on the MMVN class to explicitly return `self.event_shape`, just to be safe.

gpleiss:
Thanks for the quick work @wjmaddox ! I just left a single comment - otherwise lgtm.

gpleiss (approved):

gpytorch/distributions/multivariate_normal.py
Balandat: hmm so what if the root decomp of the underlying full covariance of the MTMVN is not square? Shouldn't the handling here apply there as well? I guess it's not clear how a smaller/larger number of samples can be squeezed into the expected two-tuple form.

wjmaddox: I supoose it would only work in some sort of non-interleaved setting? But, ideally the interleaving? will cause the non-full rank LT to just not be a RootLazyTensor.

Issue #1509, opened 2021-02-26
[Bug] num_gauss_hermite quadrature value setter isn't working


wjmaddox:
# 🐛 Bug

Trying to set `num_gauss_hermite_quadrature(int)` isn't affecting the number of locations in the quadrature.

I think it's with how the default is set in the initialization because using `num_likelhood_samples(int)` works fine so it's not just the setter method.

## To reproduce

** Code snippet to reproduce **
```python
from gpytorch.utils.quadrature import GaussHermiteQuadrature1D
from gpytorch.settings import num_gauss_hermite_locs

with num_gauss_hermite_locs(25):
    quadrature = GaussHermiteQuadrature1D()
    print(num_gauss_hermite_locs._global_value) # 25
    print(num_gauss_hermite_locs.value()) # 25
    print(quadrature.num_locs) # 20
```

** Stack trace/error message **
see above

## Expected Behavior

`quadrature.num_locs` should be 20

## System information

**Please complete the following information:**
- gpytorch master
- torch 1.7.0

gpleiss:
Closed by #1513

PR #1513, opened 2021-02-26
Use Gauss Hermite Locs value in init body

wjmaddox:
Easy resolution to #1509, must have been something with the global state in the initialization.

```python
from gpytorch.utils.quadrature import GaussHermiteQuadrature1D
from gpytorch.settings import num_gauss_hermite_locs

with num_gauss_hermite_locs(25):
    quadrature = GaussHermiteQuadrature1D()
    print(num_gauss_hermite_locs._global_value) # 25
    print(num_gauss_hermite_locs.value()) # 25
    print(quadrature.num_locs) # 25
```

It's still sort of weird behavior to have to call the setting on the likelihood init but at least you can change it now.
```python
with num_gauss_hermite_locs(25):
    likelihood = Likelihood()
```

### DIFF
--- a/gpytorch/utils/quadrature.py
+++ b/gpytorch/utils/quadrature.py
@@ -20,8 +20,10 @@ class GaussHermiteQuadrature1D(Module):
     should initialize one time, but that should obey parent calls to .cuda(), .double() etc.
     """
 
-    def __init__(self, num_locs=settings.num_gauss_hermite_locs.value()):
+    def __init__(self, num_locs=None):
         super().__init__()
+        if num_locs is None:
+            num_locs = settings.num_gauss_hermite_locs.value()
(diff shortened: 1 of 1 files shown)


Balandat (approved):
> Easy resolution to #1509, must have been something with the global state in the initialization.

Python’s default arguments are evaluated once when the function is defined. So previously this created the default value upon import of this function, using the default value. Calling this within a context manager will change the value of the setting only after, so this doesn't have any effect. That's why you should never call a function to construct a default value for function args (many linters will complain about this)

> It's still sort of weird behavior to have to call the setting on the likelihood init but at least you can change it now.

Yeah agreed that this is kind of weird, but this PR is a certainly a Pareto improvement.

PR #1518, opened 2021-03-01
Ensure LazyEvaluatedKernelTensor requires grad.

### LINKED ISSUE #1511: [Bug] LazyEvaluatedKernelTensor does not require gradient if inputs do not require gradient
# 🐛 Bug

Not entirely sure if this is intended behavior, but it does produce some downstream problems. 
If the input to `kernel(x)` doesn't require a gradient `LazyEvaluatedKernelTensor` does not portray as requiring a gradient, even though it does. Crucially `kernel(x).evaluate_kernel()` does. 

As a result, it causes some weirdness downstream in objects that can manipulate lazy kernel tensors themselves:

## To reproduce

```python
import torch
from gpytorch.kernels import MaternKernel
from gpytorch.lazy import KroneckerProductLazyTensor

torch.random.manual_seed(3)

kernels = [MaternKernel()] * 3
x = [torch.randn(5)] * 3

kernels[0](x[0]).requires_grad # False

# breaking behavior
kplt = KroneckerProductLazyTensor(*[k(a) for k, a in zip(kernels, x)])

rhs = torch.randn(kplt.shape[-1])
print(kplt.matmul(rhs).requires_grad) # True, as expected

kdiag = kplt.diag()
print(kdiag.requires_grad) # False ??

# the fix is that this works fine
kplt = KroneckerProductLazyTensor(*[k(a).evaluate_kernel() for k, a in zip(kernels, x)])
kdiag = kplt.diag()
print(kdiag.requires_grad) # True
```

## Expected Behavior

LazyKernelEvaluatedTensor should probably require a gradient to prevent this sort of behavior from happening.

## System information

gpytorch master

## Additional context

I'm trying to implement structured variational MTGP models with non-gaussian likelihoods and am struggling to get reasonable gradients, noticed while debugging.

gpleiss:
Currently, LazyEvaluatedKernelTensor only checks the kernel inputs during a  call. This is not desired behavior, because the kernel object itself likely has parameters that require gradients.
By passing the kernel parameters into the LazyTensor super constructor, we ensure that these are included as part of the  call.

[Fixes #1511]

### DIFF
--- a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
@@ -37,6 +37,16 @@ class LazyEvaluatedKernelTensor(LazyTensor):
     def device(self):
         return self.x1.device
 
+    @property
+    def requires_grad(self):
+        return super().requires_grad or any(param.requires_grad for param in self.kernel.parameters())
+
+    def _set_requires_grad(self, val):
+        super()._set_requires_grad(val)
+        # The behavior that differs from the base LazyTensor setter
+        for param in self.kernel.parameters():
+            param.requires_grad_(val)
+
     def _expand_batch(self, batch_shape):
         return self.evaluate_kernel()._expand_batch(batch_shape)
 
@@ -91,8 +101,8 @@ class LazyEvaluatedKernelTensor(LazyTensor):
         except IndexError:
             if any(not isinstance(bi, slice) for bi in batch_indices):
                 raise RuntimeError(
-                    f"Attempting to tensor index a non-batch matrix's batch dimensions. "
-                    "Got batch index {batch_indices} but my shape was {self.shape}"
+                    "Attempting to tensor index a non-batch matrix's batch dimensions. "
+                    f"Got batch index {batch_indices} but my shape was {self.shape}"
                 )
             x1 = x1.expand(*([1] * (len(batch_indices) - self.x1.dim() + 2)), *self.x1.shape)
             x1 = x1[(*batch_indices, row_index, dim_index)]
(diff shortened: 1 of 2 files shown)


gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -13,15 +13,16 @@
...
 
-    def _check_args(self, x1, x2, kernel, last_dim_is_batch=False, **params):
+    def _check_args(self, x1, x2, *kernel_param, kernel, last_dim_is_batch=False, **params):
         if not torch.is_tensor(x1):
             return "x1 must be a tensor. Got {}".format(x1.__class__.__name__)
         if not torch.is_tensor(x2):
             return "x1 must be a tensor. Got {}".format(x1.__class__.__name__)
 
     def __init__(self, x1, x2, kernel, last_dim_is_batch=False, **params):
+        kernel_params = list(kernel.parameters())
```
Balandat, line 23: shouldn't need this
```suggestion
        kernel_params =kernel.parameters()
```

Balandat (commented):
This should work. Is there maybe a cleaner way to do this? Or is passing the parameters explicitly as args of the constructor necessary for the backward pass to work properly?

gpleiss:
@Balandat 

> This should work. Is there maybe a cleaner way to do this?

I changed the way this is implemented. Rater than changing the `_args` method of LazyEvaluatedKernelTensor, I instead changed the `requires_grad` property and setter to look at the kernel parameters. See what you think.

Balandat:
I like it. Maybe we can move the meat of the `requires_grad` setter method into a `_set_requires_grad` method on `LazyTensor` and then just call that in both the setter on `LazyTensor` and `LazyEvaluatedKernelTensor`? That would at least avoid copy pasting the code.

gpleiss:
Added a `_set_requires_grad` method.

Balandat (approved):
gtmtl

Issue #1318, opened 2020-10-20
[Bug] Prior losses are incorrectly added to the mll in batch-mode


dme65:
# 🐛 Bug

Prior losses are currently being added up incorrectly in `ExactMarginalLogLikelihood`. The line: 
```
res.add_(prior.log_prob(closure()).sum())
```` 
will sum up all of the losses and then add them to the mll. If you are using a batch model this sum gets added to all of the batch dimensions which will count the losses multiple times when eventually calling `loss.sum().backward()`. It looks like the priors may not support batch mode which leads to a large variety of different shapes, but the `.sum()` call masks this issue since it just sums everything up anyway.

## To reproduce

**Code snippet (taken from `test_train_on_batch_test_on_batch`):**

```
import math

import torch

import gpytorch
from gpytorch.distributions import MultivariateNormal
from gpytorch.kernels import RBFKernel, ScaleKernel
from gpytorch.likelihoods import GaussianLikelihood
from gpytorch.means import ConstantMean

train_x1 = torch.linspace(0, 2, 11).unsqueeze(-1)
train_y1 = torch.sin(train_x1 * (2 * math.pi)).squeeze()
train_x2 = torch.linspace(0, 1, 11).unsqueeze(-1)
train_y2 = torch.sin(train_x2 * (2 * math.pi)).squeeze()
train_x12 = torch.cat((train_x1.unsqueeze(0), train_x2.unsqueeze(0)), dim=0).contiguous()
train_y12 = torch.cat((train_y1.unsqueeze(0), train_y2.unsqueeze(0)), dim=0).contiguous()


class ExactGPModel(gpytorch.models.ExactGP):
    def __init__(self, train_inputs, train_targets, likelihood, batch_shape=torch.Size()):
        super(ExactGPModel, self).__init__(train_inputs, train_targets, likelihood)
        self.mean_module = ConstantMean(batch_shape=batch_shape, prior=gpytorch.priors.SmoothedBoxPrior(-1, 1))
        self.covar_module = ScaleKernel(
            RBFKernel(
                batch_shape=batch_shape,
                lengthscale_prior=gpytorch.priors.NormalPrior(
                    loc=torch.zeros(*batch_shape, 1, 1), scale=torch.ones(*batch_shape, 1, 1)
                ),
            ),
            batch_shape=batch_shape,
            outputscale_prior=gpytorch.priors.SmoothedBoxPrior(-2, 2),
        )

    def forward(self, x):
[...]

Issue #1317, opened 2020-10-20
[Bug] Sampling from priors doesn't match shape of hyperparameters


mshvartsman:
# 🐛 Bug
I found some unexpected interactions between `ard_num_dims` and the shapes of priors for kernels -- a few settings where if I sample from a hyperparameter prior I don't get a tensor the same shape as the hyperparameter. I'm not sure if all of these are intended or not, but looks like a bug to me. 

## To reproduce

```python
import torch
from gpytorch.priors import GammaPrior, NormalPrior
from gpytorch.kernels import RBFKernel

# make a kernel
scales = torch.Tensor([1,1])
kernel = RBFKernel(
    ard_num_dims=2,
    lengthscale_prior=GammaPrior(3.0, 6.0 / scales),
)
new_lengthscale = kernel.lengthscale_prior.sample(kernel.lengthscale.shape)
print(kernel.lengthscale.shape) # size 1,2
print(new_lengthscale.shape) # size 1,2,2, if I try to assign it back I get an error

# same with another prior
kernel2 = RBFKernel(
    ard_num_dims=2,
    lengthscale_prior=NormalPrior(loc=10, scale=scales)
)

new_lengthscale = kernel2.lengthscale_prior.sample(kernel2.lengthscale.shape)
print(kernel2.lengthscale.shape) # size 1,2
print(new_lengthscale.shape) # size 1, 2, 2

# ard_num_dims is only 1 but we have a higher-dim prior. Is this behavior defined?
kernel3 = RBFKernel(
    ard_num_dims=1,
    lengthscale_prior=NormalPrior(loc=10, scale=scales)
)

new_lengthscale = kernel3.lengthscale_prior.sample(kernel3.lengthscale.shape)
print(kernel3.lengthscale.shape) # size 1, 1 -- but shouldn't we expect 1,2? 
print(new_lengthscale.shape) # size 1, 1, 2

# ok, ard_num_dims is 2 but my prior is 1d, now it works correctly
kernel4 = RBFKernel(
    ard_num_dims=2,
    lengthscale_prior=NormalPrior(loc=10, scale=1)
)

new_lengthscale = kernel4.lengthscale_prior.sample(kernel4.lengthscale.shape)
print(kernel4.lengthscale.shape) # size 1,2
print(new_lengthscale.shape) # size 1, 2
```

## Expected Behavior

It would be nice if we got a warning/error earlier for undefined/unsupported behavior, and otherwise shapes matched correctly. 

[...]

PR #1520, opened 2021-03-03
More explicit prior shape broadcasting

mshvartsman:
This is an attempt to address #1317 and #1318 -- incomplete but hopefully helps concretize discussion. I think these bugs basically have to do with unexpected implicit broadcasting behavior of priors (where you think your prior should broadcast over events but it silently broadcasts over batches, or vice versa). So this PR makes prior broadcasting universal and explicit, and throws if it can't expand shapes. Without something like `to_event` (which I think we'd need pyro for) or a restriction to not to broadcast priors over event dimensions, we (still) have a mismatch between gpytorch batch/event shapes and the underlying torch distributions shapes, but I think that the errors are correct (i.e. they are thrown if "event" and "batch" dimensions don't match even if they're not encoded as such attributes). 

I'm not sure it's the correct thing. For example, you need to `prior.sample()` without args to get a full gpytorch-batch of priors, and if you `prior.sample(torch.Size([3])` you get 3 batches of batched priors. 

Below are some examples of what happens with these changes, based on @gpleiss' taxonomy of parameters that take priors. If they make sense, they can become tests in this PR. 

``` 
# setup
import torch
from gpytorch.priors import NormalPrior, GammaPrior
from gpytorch.kernels import RBFKernel, ScaleKernel
from gpytorch.likelihoods.noise_models import HomoskedasticNoise
from gpytorch.means import ConstantMean
```

# Case 1:  Noise prior
`[*batch_shape, 1]`

```
# no broadcast
noise = HomoskedasticNoise(noise_prior=GammaPrior(concentration=2., rate=2.))
new_noise = noise.noise_prior.sample()
assert noise.noise.shape == new_noise.shape

# batched broadcast
noise = HomoskedasticNoise(noise_prior=GammaPrior(concentration=2., rate=2.), batch_shape=torch.Size([3,5]))
new_noise = noise.noise_prior.sample()
assert noise.noise.shape == new_noise.shape

# fails (can't broadcast)
noise = HomoskedasticNoise(noise_prior=GammaPrior(concentration=2., rate=torch.ones(2)))

# also fails, because mixing event and batch dims
noise = HomoskedasticNoise(noise_prior=GammaPrior(concentration=2., rate=torch.ones(2)), batch_shape=torch.Size([2]))

# succeeds with batched priors
noise = HomoskedasticNoise(noise_prior=GammaPrior(concentration=2., rate=torch.ones(2, 1)), batch_shape=torch.Size([2]))
```
And here's @dme65's example: 
```
# throws, noise is [*batch_shape, 1] and we give a [*batch_shape] prior which cannot broadcast
likelihood = GaussianLikelihood(
    noise_prior=gpytorch.priors.NormalPrior(loc=torch.zeros(2), scale=torch.ones(2)),
    batch_shape=torch.Size([2]),
)

# after this PR, the correct way of doing this would be:
likelihood = GaussianLikelihood(
    noise_prior=gpytorch.priors.NormalPrior(loc=torch.zeros(2, 1), scale=torch.ones(2, 1)),
    batch_shape=torch.Size([2]),
)
```

# Case 2: means
`[*batch_shape, 1]`, this is basically the same as noise. 
```
# unbatched, works
mean = ConstantMean(prior=NormalPrior(loc=10, scale=3))
new_mean = mean.mean_prior.sample()
assert mean.constant.shape == new_mean.shape

# fails (can't broadcast)
mean = ConstantMean(prior=NormalPrior(loc=10, scale=torch.ones(2)))

# also fails, batch shape isn't event shape
mean = ConstantMean(prior=NormalPrior(loc=10, scale=torch.ones(2)), batch_shape = torch.Size([2]))

# succeeds 
mean = ConstantMean(prior=NormalPrior(loc=10, scale=torch.ones(2, 1)), batch_shape = torch.Size([2]))
```

# Case 3: outputscales
`[*batch_shape]`
```
# unbatched, no broadcast
kernel = ScaleKernel(RBFKernel(), outputscale_prior=NormalPrior(loc=10, scale=1.3))
new_outputscale = kernel.outputscale_prior.sample()
assert kernel.outputscale.shape == new_outputscale.shape

# batched
kernel = ScaleKernel(RBFKernel(), outputscale_prior=NormalPrior(loc=10, scale=1.3), batch_shape=torch.Size([3,4]))
new_outputscale = kernel.outputscale_prior.sample()
assert kernel.outputscale.shape == new_outputscale.shape

# ScaleKernel has no ARD, error
kernel = ScaleKernel(RBFKernel(), outputscale_prior=NormalPrior(loc=10, scale=torch.ones(2)))

# this works. I'm not sure if it should work -- since there's no event dim, the last prior dim is actually batch dim here
kernel = ScaleKernel(RBFKernel(), outputscale_prior=NormalPrior(loc=10, scale=torch.ones(2)), batch_shape=torch.Size([2]))
``` 

# Lengthscales
[*batch_shape, 1, d]

```
# no broadcasting
kernel = RBFKernel(ard_num_dims=2,lengthscale_prior=GammaPrior(3.0, torch.ones(2)))
new_lengthscale = kernel.lengthscale_prior.sample()
assert kernel.lengthscale.shape == new_lengthscale.shape

# broadcast event
kernel = RBFKernel(ard_num_dims=2, lengthscale_prior=NormalPrior(loc=10, scale=1))
new_lengthscale = kernel.lengthscale_prior.sample()
assert kernel.lengthscale.shape == new_lengthscale.shape

# broadcast batch
kernel = RBFKernel(ard_num_dims=1, lengthscale_prior=NormalPrior(loc=10, scale=1.3), batch_shape=torch.Size([3,5]))
new_lengthscale = kernel.lengthscale_prior.sample()
assert kernel.lengthscale.shape == new_lengthscale.shape

# broadcast batch + event -- this is possibly the most semantically strange one
kernel = RBFKernel(ard_num_dims=2, lengthscale_prior=NormalPrior(loc=10, scale=1.3), batch_shape=torch.Size([3,5]))
new_lengthscale = kernel.lengthscale_prior.sample()
assert kernel.lengthscale.shape == new_lengthscale.shape

# ARD, no way to broadcast, error
kernel = RBFKernel(ard_num_dims=1, lengthscale_prior=NormalPrior(loc=10, scale=torch.ones(2)))

# batch is not event, errors
kernel = RBFKernel(ard_num_dims=1,lengthscale_prior=NormalPrior(loc=10, scale=torch.ones(2)), batch_shape=torch.Size([2]),)
kernel = RBFKernel(ard_num_dims=1,lengthscale_prior=NormalPrior(loc=10, scale=torch.ones(2, 1)), batch_shape=torch.Size([2]),)

# succeeds, batched prior
kernel = RBFKernel(ard_num_dims=2,lengthscale_prior=NormalPrior(loc=10, scale=torch.ones(2, 1, 1)), batch_shape=torch.Size([2]))
new_lengthscale = kernel.lengthscale_prior.sample()
assert kernel.lengthscale.shape == new_lengthscale.shape
```

### DIFF
--- a/gpytorch/module.py
+++ b/gpytorch/module.py
@@ -235,6 +235,20 @@ class Module(nn.Module):
 
         else:
             closure = param_or_closure
+        
+        hyperparameter_shape = closure().shape
+        prior_shape = prior.shape()
+
+        if prior_shape != hyperparameter_shape:
+            try:
+                prior = prior.expand(hyperparameter_shape)
+            except RuntimeError: 
+                raise RuntimeError(
+                    "Attempting to broadcast a prior that is not broadcastable! "
+                    + f"Got parameter shape {hyperparameter_shape} "
+                    + f"but prior shape {prior_shape}!"
+                )
+
         self.add_module(name, prior)
         self._priors[name] = (prior, closure, setting_closure)
 
--- a/gpytorch/kernels/arc_kernel.py
+++ b/gpytorch/kernels/arc_kernel.py
@@ -121,24 +121,24 @@ class ArcKernel(Kernel):
         self.register_parameter(
             name="raw_angle", parameter=torch.nn.Parameter(torch.zeros(*self.batch_shape, 1, self.last_dim)),
         )
+        self.register_constraint("raw_angle", angle_constraint)
+
         if angle_prior is not None:
             self.register_prior(
                 "angle_prior", angle_prior, lambda: self.angle, lambda v: self._set_angle(v),
             )
 
-        self.register_constraint("raw_angle", angle_constraint)
-
         self.register_parameter(
             name="raw_radius", parameter=torch.nn.Parameter(torch.zeros(*self.batch_shape, 1, self.last_dim)),
         )
+        radius_constraint = Positive()
+        self.register_constraint("raw_radius", radius_constraint)
 
         if radius_prior is not None:
             self.register_prior(
                 "radius_prior", radius_prior, lambda: self.radius, lambda v: self._set_radius(v),
             )
 
-        radius_constraint = Positive()
-        self.register_constraint("raw_radius", radius_constraint)
 
         self.base_kernel = base_kernel
         if self.base_kernel.has_lengthscale:
--- a/gpytorch/kernels/cosine_kernel.py
+++ b/gpytorch/kernels/cosine_kernel.py
@@ -65,6 +65,7 @@ class CosineKernel(Kernel):
 
         if period_length_constraint is None:
             period_length_constraint = Positive()
+        self.register_constraint("raw_period_length", period_length_constraint)
 
         if period_length_prior is not None:
             self.register_prior(
@@ -74,7 +75,6 @@ class CosineKernel(Kernel):
                 lambda v: self._set_period_length(v),
             )
 
-        self.register_constraint("raw_period_length", period_length_constraint)
 
     @property
     def period_length(self):
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -53,10 +53,11 @@ class IndexKernel(Kernel):
             name="covar_factor", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks, rank))
         )
         self.register_parameter(name="raw_var", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks)))
+        self.register_constraint("raw_var", var_constraint)
+
         if prior is not None:
             self.register_prior("IndexKernelPrior", prior, self._eval_covar_matrix)
 
-        self.register_constraint("raw_var", var_constraint)
 
     @property
     def var(self):
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -166,12 +166,13 @@ class Kernel(Module):
                 name="raw_lengthscale",
                 parameter=torch.nn.Parameter(torch.zeros(*self.batch_shape, 1, lengthscale_num_dims)),
             )
+            self.register_constraint("raw_lengthscale", lengthscale_constraint)
+
             if lengthscale_prior is not None:
                 self.register_prior(
                     "lengthscale_prior", lengthscale_prior, lambda: self.lengthscale, lambda v: self._set_lengthscale(v)
                 )
 
-            self.register_constraint("raw_lengthscale", lengthscale_constraint)
 
         self.distance_module = None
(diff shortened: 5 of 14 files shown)


gpytorch/module.py
```diff
@@ -261,6 +261,19 @@ def setting_closure(module, val):
                     """
                 )
             closure = param_or_closure
+        
+        hyperparameter_shape = closure().shape
```
Balandat, line 265:
```suggestion
        hyperparameter_shape = closure(self).shape
```

(I hope this works)

mshvartsman:
Still failing: `SmoothedBoxPrior` has `event_shape` of 1 whereas all pytorch-imported priors have no `event_shape` , so if we are being pedantic about broadcasting you shouldn’t be able to use a `SmoothedBoxPrior` as a prior for `outputscale` (which also has no `event_shape`). Some options: 
1. Modify `SmoothedBoxPrior` to have the same semantics as pytorch scalar priors (no `event_shape`). Haven't checked yet whether we'll have the same issue for other gpytorch priors. It's weird because it's basically converting to incorrect behavior. 
2. Modify all the pytorch-imported priors to have pyro-style `event_shape`s, redo the broadcasting code to handle batch vs event correctly. 
3. Modify outputscales to have an event dimension. 

I'm slightly leaning towards 3, but all of these seem like relatively big changes. 

Separately: I added a guard in case `prior` is getting passed as `None`, which sometimes happens. Is this intentional? 

I'm also wondering if we should do a more minor change that just warns if you're doing something that may lead to unexpected prior broadcasting behavior, because this PR feels like it's turning into a larger set of prior changes.

gpytorch/module.py
mshvartsman: I think it works!

gpytorch/module.py
```diff
@@ -262,6 +262,20 @@ def setting_closure(module, val):
                 )
             closure = param_or_closure
 
+        if prior is not None:
+            hyperparameter_shape = closure(self).shape
+            prior_shape = prior.shape()
```
Balandat, line 267: Did this work at any point? Is this meant to be `event_shape`?

Alternatively we would give `Prior`s a `shape` property that is just `self._batch_shape + self._event_shape`.

Balandat (commented):
I'm running into a ton of test failures when rebasing on master. Did this largely work in the past (except for the `SmoothBoxPrior`?). If so, there may have been some other changes since that caused issues with this. 

> Separately: I added a guard in case prior is getting passed as None, which sometimes happens. Is this intentional?
I don't think so?

mshvartsman:
Yeah, I thought it worked except the box prior stuff, so let me rebase and revisit.

PR #1521, opened 2021-03-04
Add a Check for TriangularLazyTensors in inv_quad_logdet

wjmaddox:
Adds a check to see if the root decomposition has already been computed and if that returned a `TriangularLazyTensor` for `inv_quad_logdet`. 

Enables the internals to use the new root decompositions after rows/columns have been updated like in #1499, which correspondingly makes some BO acquisition functions easier to implement (like botorch #724).

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -23,7 +23,7 @@ from ..utils.cholesky import psd_safe_cholesky
 from ..utils.deprecation import _deprecate_renamed_methods
 from ..utils.errors import CachingError
 from ..utils.getitem import _compute_getitem_size, _convert_indices_to_tensors, _is_noop_index, _noop_index
-from ..utils.memoize import add_to_cache, cached, get_from_cache, pop_from_cache
+from ..utils.memoize import _is_in_cache_ignore_all_args, add_to_cache, cached, get_from_cache, pop_from_cache
 from ..utils.pivoted_cholesky import pivoted_cholesky
 from ..utils.warnings import NumericalWarning
 from .lazy_tensor_representation_tree import LazyTensorRepresentationTree
@@ -1031,8 +1031,7 @@ class LazyTensor(ABC):
             # if the root decomposition has already been computed and is triangular we can use it instead
             # of computing the cholesky.
             will_need_cholesky = True
-            # TODO: shouldn't this be _is_in_cache_ignore_args
-            if hasattr(self, "_memoize_cache") and "root_decomposition" in [x[0] for x in self._memoize_cache.keys()]:
+            if _is_in_cache_ignore_all_args(self, "root_decomposition"):
                 root = self.root_decomposition().root
                 if isinstance(root, TriangularLazyTensor):
                     cholesky = CholLazyTensor(root)
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -23,7 +23,7 @@ from ..utils.cholesky import psd_safe_cholesky
 from ..utils.deprecation import _deprecate_renamed_methods
 from ..utils.errors import CachingError
 from ..utils.getitem import _compute_getitem_size, _convert_indices_to_tensors, _is_noop_index, _noop_index
-from ..utils.memoize import _is_in_cache_ignore_args, add_to_cache, cached, get_from_cache, pop_from_cache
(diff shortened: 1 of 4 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -1028,7 +1028,14 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
             from .chol_lazy_tensor import CholLazyTensor
             from .triangular_lazy_tensor import TriangularLazyTensor
 
-            cholesky = CholLazyTensor(TriangularLazyTensor(self.cholesky()))
+            # if the root decomposition has already been computed and is triangular we can use it instead
+            # of computing the cholesky.
+            if _is_in_cache_ignore_args(self, "root_decomposition"):
+                root = self.root_decomposition().root
+                if isinstance(root, TriangularLazyTensor):
```
Balandat, line 1035: mhh looksl like we'd get an `UnboundLocalError` if this is not true?

Balandat (changes requested):

gpytorch/lazy/lazy_tensor.py
Balandat: which is a reasonably passive aggressive way of saying that we should probably have test coverage here :)

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1028,7 +1028,17 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
             from .chol_lazy_tensor import CholLazyTensor
             from .triangular_lazy_tensor import TriangularLazyTensor
 
-            cholesky = CholLazyTensor(TriangularLazyTensor(self.cholesky()))
+            # if the root decomposition has already been computed and is triangular we can use it instead
+            # of computing the cholesky.
+            will_need_cholesky = True
+            # TODO: shouldn't this be _is_in_cache_ignore_args
+            if hasattr(self, "_memoize_cache") and "root_decomposition" in [x[0] for x in self._memoize_cache.keys()]:
```
Balandat, line 1035: This should be in `_is_in_cache` b/c it uses a `@cached(name="root_decomposition")` decorator. The reason being that there is a `method` kwarg to `root_decomposition`. I guess we could also always return the root that was computed under the first call of this method, the gotcha then is that if you call it with `None` first and then, say, `lanczos`, you'll always get the cholesky root even though you eplixitly passed in lanczos

wjmaddox: Good point, although it doesn't make sense (to me at least), to be overwriting the "cholesky" method with something like a triangular low rank update to a root decomposition, so this check should stay as it is for now then?

Balandat: Hmm yeah though I'd like to push this into the `gpytorch.utils.memoize` module as some function, I'm a little uneasy about doing very specific things to the cache outside the cache helper functions. Any checking for `_memoize_cache` attributes or calling any functions with leading `_` from `gpytorch.utils.memoize` shouldn't really happen outside that module, too much potential for things going wrong. 

So we probably want to rename  `_get_from_cache_ignore_args` to `get_from_cache_ignore_args` and add another method to the module (maybe call it `get_from_cache_regardless_of_kwarg` or something like that) that implements what you do here and just picks out the first element in the named cache corresponding to the method). 

I also think we can be a little more pythonic regarding the cache, i.e. just try getting it and then except the `CachingError` if not available.

wjmaddox: So that's what's already being done in `_get_from_cache_ignore_args` and fails because of the method + other kwargs. I'm just attempting to pull out a/any root decomposition here, but I agree that this should probably be a method rather than a suspect one-liner.

Balandat: So `_get_from_cache_ignore_args` is kind of a challenging semantic collision here - `_get_from_cache_ignore_args` retrieves from cache what has been *put in ignoring* the args to the function, while what we want to do here is get something from cache that has been put in *with* function args, but we want to ignore these functions args for the purpose of retrieving it.

wjmaddox: `_is_in_cache_ignore_all_args` should resolve the semantics issue.

Balandat (approved):

PR #1529, opened 2021-03-16
Fix handling of batched tensors in cat_rows

Balandat:
This fixes the issue discussed in https://github.com/cornellius-gp/gpytorch/pull/1499#discussion_r594720889: Prior to this PR things break if a batched set of additional rows/cols are added to a LazyTensors via `cat_rows`. This is required for batched fantasizing. This PR also adds unit tests to cover this setting.

The unit test for the batched version with a low-rank root currently fails due to what looks like numerical issues.

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1594,7 +1594,8 @@ class LazyTensor(ABC):
         from .chol_lazy_tensor import CholLazyTensor
         from .root_lazy_tensor import RootLazyTensor
 
-        # # no need to do extra work if we already have a cached cholesky decomposition
+        # no need to do extra work if we already have a cached cholesky decomposition
+        # TODO: Resolve issues related to triangular tensor instantiation
         # if _is_in_cache_ignore_args(self, "cholesky"):
         #     res = self.cholesky()
         #     return CholLazyTensor(res)
@@ -1667,7 +1668,8 @@ class LazyTensor(ABC):
         if self.shape[-2:].numel() == 1:
             return RootLazyTensor(1 / self.evaluate().sqrt())
 
-        # # no need to do extra work if we already have a cached cholesky decomposition
+        # no need to do extra work if we already have a cached cholesky decomposition
+        # TODO: Resolve issues related to triangular tensor instantiation \
         # if _is_in_cache_ignore_args(self, "cholesky"):
         #     res = self.cholesky()
         #     return CholLazyTensor(res).inverse()
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -840,7 +840,8 @@ class LazyTensor(ABC):
             else:
                 # solve with least squares
                 if R.ndim > 2:
-                    # temporary hack around PyTorch limitation, this is no good...
+                    # TODO: Replace with `torch.linalg.lstsq` (introduced in #49093) once we have a Pytorch >=1.9 req.
+                    # temporary hack around this limitation for now
                     new_inv_root = (torch.pinverse(R) @ Q.transpose(-1, -2)).transpose(-1, -2)
                 else:
                     new_inv_root = torch.lstsq(Q.transpose(-2, -1), R).solution.transpose(-2, -1)
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -775,7 +775,6 @@ class LazyTensor(ABC):
             :obj:`LazyTensor`: concatenated lazy tensor with the new rows and columns.
         """
         from . import lazify
-        from .batch_repeat_lazy_tensor import BatchRepeatLazyTensor
         from .cat_lazy_tensor import CatLazyTensor
         from .root_lazy_tensor import RootLazyTensor
         from .triangular_lazy_tensor import TriangularLazyTensor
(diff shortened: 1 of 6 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -838,7 +839,11 @@ def cat_rows(self, cross_mat, new_mat, generate_roots=True, **root_decomp_kwargs
                 new_inv_root = torch.triangular_solve(Q.transpose(-2, -1), R, upper=True).solution.transpose(-2, -1)
             else:
                 # solve with least squares
-                new_inv_root = torch.lstsq(Q.transpose(-2, -1), R).solution.transpose(-2, -1)
+                if R.ndim > 2:
+                    # temporary hack around PyTorch limitation, this is no good...
```
vishwakftw, line 843: Batching is supported in PyTorch master. This is not appropriately reflected in the PyTorch docs, but is supported per my reading of ATen internals.

The input shapes should satisfy the following criteria:
1. `A` should be of dimension >= 2
2. `b` should be of dimension >= 1
3. the difference between `A.dim` and `b.dim` should be 0 and 1
4. if the difference between `A.dim` and `b.dim` is 1, then `b` is unsqueezed in the last dimension i.e., `b.unsqueeze(-1)`.
5. let the shapes of `A` and `b` after step 4 be `A_shape` and `b_shape`. Then `A_shape = b_shape`.

Balandat: Ah ok so this is supported in `torch.linalg.lstsq`, but not in `torch.lstsq`.

Balandat: Also, it's not available in 1.8, so we'll need to wait until we can bump pytorch to 1.9 for this.

jacobrgardner:
For an R coming from a QR decomposition is there a reason I'd need a pseudoinverse at all? I am clearly missing something from not being quite up to speed on the previous PR @Balandat

Balandat:
> For an R coming from a QR decomposition is there a reason I'd need a pseudoinverse at all?

Well we need the pseudoinverse of `new_inv_root` but I guess you're right that computing the pseudoinverse of `R` doesn't make much sense. We can just construct it directly using the QR decomposiiton's components. Let me modify this.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1609,6 +1594,11 @@ def root_decomposition(self, method: Optional[str] = None):
         from .chol_lazy_tensor import CholLazyTensor
         from .root_lazy_tensor import RootLazyTensor
 
+        # # no need to do extra work if we already have a cached cholesky decomposition
+        # if _is_in_cache_ignore_args(self, "cholesky"):
+        #     res = self.cholesky()
+        #     return CholLazyTensor(res)
```
Balandat, line 1600: this is sth that I think makes sense. It currently causes some issues though w/ handling `DiagLazyTensors` as proper special cases of `TriangularLazyTensor`. Will have to revisit this some other time.

Balandat:
OK so I kind of went down the rabbit hole why working on this. `cat_rows` was pretty hard to parse b/c of overloaded notation, so I refactored the whole thing. Also fixed some potential issues (e.g. we used `TriangularLazyTensor` without actually being sure that the tensor we used was triangular). 

Introduces a `stable_pinverse` util to clean things up a bit that uses `stable_qr` under the hood. As Jake said there was no reason to compute any pseudoinverses of `R` matrices from a QR decomposition.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -775,73 +775,58 @@ def cat_rows(self, cross_mat, new_mat, generate_roots=True, **root_decomp_kwargs
...
-            # we know L is triangular, so inverting is a triangular solve agaist the identity
-            # we don't need the batch shape here, thanks to broadcasting
-            Eye = torch.eye(new_root.shape[-2], device=new_root.device, dtype=new_root.dtype)
-            new_inv_root = torch.triangular_solve(Eye, new_root, upper=False)[0]
-            new_inv_root = new_inv_root.transpose(-1, -2)
+        E = self.root_decomposition(**root_decomp_kwargs).root
+        m, n = E.shape[-2:]
+        R = self.root_inv_decomposition().root.evaluate()  # this is fast if L is triangular
+        F = B_ @ R
+        G = lazify(D - F.matmul(F.transpose(-2, -1))).root_decomposition().root.evaluate()
```
wjmaddox, line 808: Personally, I find `lower_left` and `schur_root` to be a bit more human comprehensible than `F` and `G` respectively. Along with also doing a better job of pointing out what exactly is going on.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -775,73 +775,58 @@ def cat_rows(self, cross_mat, new_mat, generate_roots=True, **root_decomp_kwargs
...
+        num_fant = G.size(-2)
+        new_root = torch.zeros(*batch_shape, m + num_fant, n + num_fant, device=E.device, dtype=E.dtype)
+        new_root[..., :m, :n] = E.evaluate()
+        new_root[..., m:, : F.shape[-1]] = F
+        new_root[..., m:, n : (n + G.shape[-1])] = G
+
+        if isinstance(E, TriangularLazyTensor) and isinstance(G, TriangularLazyTensor):
+            # make sure these are actually upper triangular
+            if getattr(E, "upper", False) or getattr(G, "upper", False):
+                raise NotImplementedError
```
wjmaddox, line 820: Can this just be resolved with a transpose? Perhaps something like `new_root = TriangularLazyTensor(new_root, upper=False)`?

wjmaddox (commented):
Just a couple of quick comments but otherwise the algebra looks the same to me.

gpytorch/lazy/lazy_tensor.py
Balandat: The main reason of for changing this was to be consistent with the comments and to dedup some double usage of notation for different things.  I can get back to those var names and add more comments if that is desired.

gpytorch/lazy/lazy_tensor.py
Balandat: Well so we assume from the start that we're going to build a block lower triangular matrix. Here we're in the case where `E` or `G` are triangular, but if either of them is upper triangular then there is no way the whole matrix is going to be. Does this make sense? I don't think we currently would ever get into that situation but I figured better safe than sorry here.

wjmaddox: Ahh yeah, good point. The proper thing to do is probably to use the transpose before we even get to this point then, but I think it's probably unlikely to hit it.

Balandat: yeah, we can fix it when we run into it, at least it's breaking loudly

gpleiss (approved):

gpleiss:
@Balandat lgtm!

PR #1534, opened 2021-03-17
Set the default method as symeig for KroneckerProductLazyTensor to compute diag

qingfeng10:
Latest GPyTorch master changed the default method from `symeig` to `lanczos` to compute diagonalization when data exceeds max_cholesky_size. However lanczos causes sizing mismatch for KroneckerProductLazyTensor, thus Higher-Order GP fitting fails.  Here I change the default method back to `symeig` for KroneckerProductLazyTensor.

### DIFF
--- a/gpytorch/lazy/kronecker_product_lazy_tensor.py
+++ b/gpytorch/lazy/kronecker_product_lazy_tensor.py
@@ -138,6 +138,11 @@ class KroneckerProductLazyTensor(LazyTensor):
                 raise RuntimeError("Diag works on square matrices (or batches)")
         return _kron_diag(*self.lazy_tensors)
 
+    def diagonalization(self, method: Optional[str] = None):
+        if method is None:
+            method = "symeig"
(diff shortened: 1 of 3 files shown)


gpytorch/lazy/kronecker_product_lazy_tensor.py
```diff
@@ -138,6 +138,11 @@ def diag(self):
                 raise RuntimeError("Diag works on square matrices (or batches)")
         return _kron_diag(*self.lazy_tensors)
 
+    def diagonalization(self, method: Optional[str] = None):
+        if method is None:
+            method = "symeig"
+        return super().diagonalization(method=method)
```
Balandat, line 144: @wjmaddox I think this is the right thing to do as a default for Kronecker, what do you think?

Balandat: confirmed offline that yes

Balandat (approved):

PR #1537, opened 2021-03-19
Changes the Hadamard Multitask notebook to avoid confusing new users

4aHxKzD:
Like discussed in #1474 (where @gpleiss agrees in this [comment](https://github.com/cornellius-gp/gpytorch/discussions/1474#discussioncomment-485653)) I changed the method used to create the tensor of indices that is used as an input to the IndexKernel.

I also took the liberty of adding more information to the IndexKernel's docstring.


gpytorch/kernels/index_kernel.py
```diff
@@ -28,7 +28,9 @@ class IndexKernel(Kernel):
             Set if the MultitaskKernel is operating on batches of data (and you want different
             parameters for each batch)
         :attr:`rank` (int):
-            Rank of :math:`B` matrix.
+            Rank of :math:`B` matrix. Controls the degree of
+            correlation between the outputs. With a rank of 1 the
+            outputs share same shape only differing in scale.
```
gpleiss, line 33: What do you mean by this shape/scale comment?

4aHxKzD: I'm trying to explain what is stated in the paper "Kernels for Vector-Valued Functions: a Review". In Figure 2 the last sentence (referring to the case when the rank is 1) says:

>  Samples are identical except for scale.

and in the next page, right before section 4.2.4, when discussing the case where the ranks are different (even though it is for the LMC kernel), they use the following wording:

>  samples from different outputs have different shapes

That is why I used the words "shape" and "scale".

Maybe you can word it a little better, I'm not a native speaker.

4aHxKzD: Would

> With a rank of 1 the outputs are identical except for a scaling factor.

sound better?

gpleiss: Yeah that sounds better.

gpleiss (approved):

PR #1545, opened 2021-03-29
Fantasy Strategy for SKI

wjmaddox:
Work in progress implementation of https://arxiv.org/pdf/2103.01454.pdf which implements a constant time fantasy strategy for SKI. I'll only be editing the caches for `InterpolatedPredictionStrategy` and not also implementing the special marginal log likelihood in this PR (can add if necessary later).

ToDos:

- [x] Correct mean and covariance caches
- [x] Fantasy strategy
- [x] Revert back to old mean and covariance caches if fantasy strategy is not called. 
- [x] Update unit tests as necessary to ensure downstream botorch compatibility (e.g. unit tests for ski fantasy model)

Original codebase is at: https://github.com/wjmaddox/onlinegp . 

cc @samuelstanton


gpytorch/models/exact_prediction_strategies.py
```diff
@@ -397,64 +456,35 @@ def mean_cache(self):
     @property
     @cached(name="covar_cache")
     def covar_cache(self):
```
gpleiss, line 458: I like the cleanup here :)

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -365,28 +364,88 @@ def _exact_predictive_covar_inv_quad_form_root(self, precomputed_cache, test_tra
...
+        fant_strat = self.__class__(
+            train_inputs=full_inputs,
+            train_prior_dist=self.train_prior_dist.__class__(full_mean, full_covar),
+            train_labels=full_targets,
+            likelihood=fant_likelihood,
         )
+        add_to_cache(fant_strat, "interp_inner_prod", new_wmat)
+        add_to_cache(fant_strat, "interp_response_cache", new_interp_response_cache)
+
+    def prepare_dense_wmat(self, covar=None):
```
gpleiss, line 400: Just so I'm clear here - this method is only called with `fant_fant_covar`, so we can expect these matrices to be small. Is that correct?

gpleiss (changes requested):
Everything looks good (once unit tests pass). It'd be good to:
- Add a unit test that makes sure that these functions are actually called - something like what we have [for RFF kernels here](https://github.com/cornellius-gp/gpytorch/blob/48207896e9396036b493b81939d9f8589596392d/test/kernels/test_additive_and_product_kernels.py#L314)
- Add a unit test that makes fantasy observations for SKI kernels

gpytorch/models/exact_prediction_strategies.py
wjmaddox: It also has to be done on the training distribution so it could be a dense matrix of shape `m x n`.

wjmaddox:
I threw in a bunch of forced cholesky decompositions to fix the numerical instability issues, which is going to have the unfortunate side effect of breaking large scale SKI until this is resolved. Solution is to only compute these caches on calling `get_fantasy_model` and to revert to the old ones otherwise.

test/examples/test_kissgp_kronecker_product_regression.py
```diff
@@ -96,6 +96,7 @@ def test_kissgp_gp_mean_abs_error(self):
             likelihood.eval()
 
             test_preds = likelihood(gp_model(test_x)).mean
+            print(test_preds.shape, test_y.shape)
```
gpleiss, line 99:
```suggestion
```

gpleiss:
```suggestion
```

gpleiss (approved):

PR #1546, opened 2021-03-31
Add sample method to SmoothedBoxPrior

spectraldani:
This is done by sampling the distribution as a mixture of a Gaussian distribution with a Uniform dist. In this patch, the space cost of this function is O(3n), where n is the number of elements sampled. I haven't yet figured out a way to implement `rsample`, but for my use case, it was not necessary.


gpytorch/priors/smoothed_box_prior.py
```diff
@@ -71,3 +71,20 @@ def _log_prob(self, x):
         # x = "distances from box`"
         X = ((x - self._c).abs_() - self._r).clamp(min=0)
         return (self.tails.log_prob(X) - self._M).sum(-1)
+
+    def sample(self, sample_shape=torch.Size()):
```
gpleiss, line 75: I think this could be a `rsample` method - `a` and `b` can be differentiated through given your implementation. (Not that I see the use case right now, but it could be a nice future-proof addition).

gpleiss (changes requested):
Thanks @spectraldani - this is great! Could I get you to add a unit test to `test/priors/test_smoothed_box_prior.py`? A very simple unit test would draw 50000 samples from your prior and assert that some percentage of them lie between `a` and `b`. See https://github.com/cornellius-gp/gpytorch/blob/48207896e9396036b493b81939d9f8589596392d/gpytorch/test/lazy_tensor_test_case.py#L514 for an example.

gpytorch/priors/smoothed_box_prior.py
spectraldani: Indeed, I got hung up on the `torch.bernoulli` for some reason and couldn't see that the method could've been `rsample` from the start!

spectraldani:
>Could I get you to add a unit test to `test/priors/test_smoothed_box_prior.py`? A very simple unit test would draw 50000 samples from your prior and assert that some percentage of them lie between `a` and `b`. See
> 
> https://github.com/cornellius-gp/gpytorch/blob/48207896e9396036b493b81939d9f8589596392d/gpytorch/test/lazy_tensor_test_case.py#L514
> 
> for an example.

Sure! I just need to create a new file with a new `TestCase` class, right? I'm still learning how the test cases are structured in the repo, so I wanted to be extra sure that that how things are done 😅

gpleiss:
> I'm still learning how the test cases are structured in the repo, so I wanted to be extra sure that that how things are done

No worries! Rather than creating a new `TestCase`, I would just add a `def test_sample(self):` method to this file: https://github.com/cornellius-gp/gpytorch/blob/master/test/priors/test_smoothed_box_prior.py

gpleiss (approved):

PR #1564, opened 2021-04-12
Add default setting values to docs.

### LINKED ISSUE #1533: [Docs] Settings defaults mismatched in source code and documentation
# 📚 Documentation/Examples
Separating out from #1526 

The defaults shown in the settings don't consistently match the source code.

### For example:  
settings.memory_efficient is [True in the documentation](https://docs.gpytorch.ai/en/stable/settings.html#gpytorch.settings.memory_efficient), but [False in the corresponding source code](https://docs.gpytorch.ai/en/stable/_modules/gpytorch/settings.html#memory_efficient).

gpleiss:
[Fixes #1533]

### DIFF
--- a/gpytorch/settings.py
+++ b/gpytorch/settings.py
@@ -558,7 +558,7 @@ class memory_efficient(_feature_flag):
     Pros: memory efficient, faster on CPU
     Cons: slower on GPUs with < 10000 inducing points
 
-    (Default: True)
+    (Default: False)
     """
 
     _default = False
--- a/gpytorch/settings.py
+++ b/gpytorch/settings.py
@@ -183,6 +183,8 @@ class skip_posterior_variances(_feature_flag):
     forward pass. If this is on, the returned gpytorch MultivariateNormal will have a
     ZeroLazyTensor as its covariance matrix. This allows gpytorch to not compute
     the covariance matrix when it is not needed, speeding up computations.
+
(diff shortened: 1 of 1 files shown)


gpytorch/settings.py
```diff
@@ -518,6 +557,8 @@ class memory_efficient(_feature_flag):
     Whether or not to use Toeplitz math with gridded data, grid inducing point modules
     Pros: memory efficient, faster on CPU
     Cons: slower on GPUs with < 10000 inducing points
+
+    (Default: True)
     """
 
     _default = False
```
Balandat, line 564: `True == False`?

gpleiss: we have bigger problems if it does...

gpytorch/settings.py
```diff
@@ -518,6 +557,8 @@ class memory_efficient(_feature_flag):
     Whether or not to use Toeplitz math with gridded data, grid inducing point modules
     Pros: memory efficient, faster on CPU
     Cons: slower on GPUs with < 10000 inducing points
+
+    (Default: True)
```
gpleiss, line 561:
```suggestion
    (Default: False)
```

gpytorch/settings.py
Balandat: `P=NP`?

PR #1575, opened 2021-04-15
Removed unnecessary line in FirstSincKernel forward

joacorapela:
I believe Line 117 in examples/00_Basic_Usage/Implementing_a_custom_Kernel.ipynb is not functional, since it is not assigning the result of diff.where(diff == 0, torch.as_tensor(1e-20) to diff.

In addition, I think this line is not necessary since method _dist in class Distance uses clamp_min_(1e-30).


gpleiss:
Hmm... it doesn't seem like that line is doing anything, but it's probably good to keep a check that prevents divide-by-zero errors. Maybe we should change it to `diff = diff.where(diff == 0, torch.as_tensor(1e-20)`?

PR #1576, opened 2021-04-15
fix lazy low rank root addition bugs

mwlon:
It appears this old broken code has become an issue after some recent should-be fixes opened up new code paths.

I have verified that the test fails on master. See the issue https://github.com/cornellius-gp/gpytorch/issues/1577

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -898,7 +898,7 @@ class LazyTensor(ABC):
         # compute p = M B and take its SVD
         pvector = current_inv_root.matmul(low_rank_mat)
         # USV^T = p; when p is a vector this saves us the trouble of computing an orthonormal basis
-        U, S, _ = torch.svd(pvector, some=False)
+        U, S, _ = torch.svd(pvector.evaluate(), some=False)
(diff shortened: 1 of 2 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -898,6 +898,8 @@ def add_low_rank(
         # compute p = M B and take its SVD
         pvector = current_inv_root.matmul(low_rank_mat)
         # USV^T = p; when p is a vector this saves us the trouble of computing an orthonormal basis
+        if isinstance(pvector, LazyTensor):
+            pvector = pvector.evaluate()
```
Balandat, line 902: would suggest using `delazify` from `lazy.lazy_tensor` for this

Balandat (approved):
Thanks for putting up this fix - lgtm, maybe you can use `delazify` as suggested?

gpytorch/lazy/lazy_tensor.py
gpleiss:
```suggestion
        pvector = delazify(pvector)
```

gpleiss (approved):

Issue #1580, opened 2021-04-19
[Bug] SumPredictionStrategy fails or gives wrong results whenever covariance matrix is lazy computed


mwlon:
# 🐛 Bug

According to the comments on `DefaultPredictionStrategy._exact_predictive_covar_inv_quad_form_root`:

> Computes :math:`K_{X^{*}X} S` given a precomputed cache
> Where :math:`S` is a tensor such that :math:`SS^{\top} = (K_{XX} + \sigma^2 I)^{-1}`

This checks out with the implementation, which does a root inverse operation to compute the cache. However, `SumPredictionStrategy._exact_predictive_covar_inv_quad_form_root` current sums the inverse root for each kernel term, which it not valid; (sum_i A_i)^(-1/2) != sum(A_i^(-1/2)). I think it will also be faster to do it the correct way, since taking the inverse root is the expensive operation, and we should only need to do it once.

Plus, `SumPredictionStrategy` assumes that if the input tensor is lazy, then the cached data is a tuple, which is (other than a hack since the type is usually tensor), invalid when the cache was computed from a non-lazy tensor. This can lead to very confusing size mismatch errors like the one below.

## To reproduce

** Code snippet to reproduce **
```
from gpytorch.kernels import LinearKernel, MaternKernel, AdditiveKernel, ScaleKernel
from gpytorch.models import ExactGP
from gpytorch.mlls import ExactMarginalLogLikelihood
from gpytorch.likelihoods.gaussian_likelihood import GaussianLikelihood
import torch
import gpytorch
from gpytorch import settings as gpt_settings
from gpytorch.priors import GammaPrior

k0 = MaternKernel(active_dims=[0])
k1 = LinearKernel(active_dims=[1])
kernel = AdditiveKernel(k0, k1)


class ExactGpModel(ExactGP):
    def __init__(self, train_x, train_y, likelihood, covar_module):
        super(ExactGpModel, self).__init__(train_x, train_y, likelihood)
        self.mean_module = gpytorch.means.ConstantMean()
        self.covar_module = covar_module

    def forward(self, x):
        mean_x = self.mean_module(x)
        covar_x = self.covar_module(x)
        return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)

n_dims = 5
base_n = 500
likelihood = GaussianLikelihood()
x = torch.randn((base_n, n_dims))
y = torch.randn(base_n)
[...]

PR #1582, opened 2021-04-21
Sum prediction strategy fix

mwlon:
I verified the new test fails on master. See https://github.com/cornellius-gp/gpytorch/issues/1580 for more details

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -15,7 +15,6 @@ from ..lazy import (
     MatmulLazyTensor,
     NonLazyTensor,
     RootLazyTensor,
-    SumLazyTensor,
     ZeroLazyTensor,
     delazify,
     lazify,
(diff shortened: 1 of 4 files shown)


gpytorch/models/exact_prediction_strategies.py
```diff
@@ -651,47 +650,7 @@ def exact_predictive_covar(self, test_test_covar, test_train_covar):
 
 
 class SumPredictionStrategy(DefaultPredictionStrategy):
```
gpleiss, line 652: Perhaps we should just get rid of `SumPredictionStrategy` at this point, if it's going to be a no-op? I am open to this: this code is old and stale, and it's probably best to use the default prediction strategy at this point.

gpleiss (changes requested):

gpytorch/models/exact_prediction_strategies.py
mwlon: Sound good to me, will adjust

gpleiss (approved):

Issue #1583, opened 2021-04-21
[Bug] LazyTensor root_inv_decomposition can fail under certain conditions


mwlon:
# 🐛 Bug

For instance, if `method='cholesky'` and there is a cache miss, `root_inv_decomposition` will throw a nonsensical error saying `local variable 'inv_root' referenced before assignment`.

## To reproduce

** Code snippet to reproduce **
The issue is flaky, hard to reproduce, but the code path causing it is pretty obvious. See below.

** Stack trace/error message **
```
/code/venvs/venv/lib/python3.8/site-packages/gpytorch/module.py in __call__(self, *inputs, **kwargs)
     26 
     27     def __call__(self, *inputs, **kwargs):
---> 28         outputs = self.forward(*inputs, **kwargs)
     29         if isinstance(outputs, list):
     30             return [_validate_module_outputs(output) for output in outputs]

/code/venvs/venv/lib/python3.8/site-packages/gpytorch/mlls/exact_marginal_log_likelihood.py in forward(self, function_dist, target, *params)
     60         # Get the log prob of the marginal distribution
     61         output = self.likelihood(function_dist, *params)
---> 62         res = output.log_prob(target)
     63         res = self._add_other_terms(res, params)
     64 

/code/venvs/venv/lib/python3.8/site-packages/gpytorch/distributions/multivariate_normal.py in log_prob(self, value)
    145     def log_prob(self, value):
    146         if settings.fast_computations.log_prob.off():
--> 147             return super().log_prob(value)
    148 
    149         if self._validate_args:

/code/venvs/venv/lib/python3.8/site-packages/torch/distributions/multivariate_normal.py in log_prob(self, value)
    206             self._validate_sample(value)
    207         diff = value - self.loc
--> 208         M = _batch_mahalanobis(self._unbroadcasted_scale_tril, diff)
    209         half_log_det = self._unbroadcasted_scale_tril.diagonal(dim1=-2, dim2=-1).log().sum(-1)
    210         return -0.5 * (self._event_shape[0] * math.log(2 * math.pi) + M) - half_log_det

/code/venvs/venv/lib/python3.8/site-packages/gpytorch/distributions/multivariate_normal.py in _unbroadcasted_scale_tril(self)
     55         if self.islazy and self.__unbroadcasted_scale_tril is None:
     56             # cache root decoposition
---> 57             ust = delazify(self.lazy_covariance_matrix.cholesky())
[...]

PR #1584, opened 2021-04-21
simplify and fix root inverse decomposition logic

mwlon:
See https://github.com/cornellius-gp/gpytorch/issues/1583 for more details

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1700,7 +1700,7 @@ class LazyTensor(ABC):
             # we know L is triangular, so inverting is a simple triangular solve agaist the identity
             # we don't need the batch shape here, thanks to broadcasting
             Eye = torch.eye(L.shape[-2], device=L.device, dtype=L.dtype)
-            Linv = torch.triangular_solve(Eye, L, upper=False)[0]
+            Linv = torch.triangular_solve(Eye, L, upper=False).solution
             res = lazify(Linv.transpose(-1, -2))
             inv_root = res
         elif method == "lanczos":
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1677,11 +1677,19 @@ class LazyTensor(ABC):
         if self.shape[-2:].numel() == 1:
             return RootLazyTensor(1 / self.evaluate().sqrt())
 
-        # no need to do extra work if we already have a cached cholesky decomposition
-        # TODO: Resolve issues related to triangular tensor instantiation \
-        # if _is_in_cache_ignore_args(self, "cholesky"):
-        #     res = self.cholesky()
-        #     return CholLazyTensor(res).inverse()
+        # check to see if we have already run lanczos for a diagonalization
+        try:
+            evals, evecs = get_from_cache(self, "diagonalization")
+            F = evecs * evals.clamp(1e-7).sqrt().reciprocal().unsqueeze(-2)
+            return RootLazyTensor(F)
+        except CachingError:
+            pass
+
+        if not self.is_square:
+            raise RuntimeError(
+                "root_inv_decomposition only operates on (batches of) square (symmetric) LazyTensors. "
+                "Got a {} of size {}.".format(self.__class__.__name__, self.size())
+            )
 
         if method is None:
             if (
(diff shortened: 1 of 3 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -1692,36 +1700,21 @@ def root_inv_decomposition(self, initial_vectors=None, test_vectors=None, method
             else:
                 method = "lanczos"
 
+        # no need to do extra work if we already have a cached cholesky decomposition
```
mwlon, line 1703: I think these comments and any commented code here is obsolete. Shall I go ahead and delete them?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1677,11 +1677,19 @@ def root_inv_decomposition(self, initial_vectors=None, test_vectors=None, method
...
 
-        # no need to do extra work if we already have a cached cholesky decomposition
-        # TODO: Resolve issues related to triangular tensor instantiation \
-        # if _is_in_cache_ignore_args(self, "cholesky"):
-        #     res = self.cholesky()
-        #     return CholLazyTensor(res).inverse()
+        # check to see if we have already run lanczos for a diagonalization
+        try:
+            evals, evecs = get_from_cache(self, "diagonalization")
+            F = evecs * evals.clamp(1e-7).sqrt().reciprocal().unsqueeze(-2)
```
Balandat, line 1683: Equivalent but more clear what's happening

```suggestion
            F = evecs * evals.clamp_min(1e-7).sqrt().reciprocal().unsqueeze(-2)
```

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1692,36 +1700,21 @@ def root_inv_decomposition(self, initial_vectors=None, test_vectors=None, method
...
-            evals, evecs = get_from_cache(self, "diagonalization")
-            F = evecs * evals.clamp(1e-7).sqrt().reciprocal().unsqueeze(-2)
-            return RootLazyTensor(F)
-        except CachingError:
-            pass
+            L = delazify(self.cholesky())
+            # we know L is triangular, so inverting is a simple triangular solve agaist the identity
+            # we don't need the batch shape here, thanks to broadcasting
+            Eye = torch.eye(L.shape[-2], device=L.device, dtype=L.dtype)
+            Linv = torch.triangular_solve(Eye, L, upper=False)[0]
```
Balandat, line 1713:
```suggestion
            Linv = torch.triangular_solve(Eye, L, upper=False).solution
```

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1677,11 +1677,19 @@ def root_inv_decomposition(self, initial_vectors=None, test_vectors=None, method
...
+            F = evecs * evals.clamp(1e-7).sqrt().reciprocal().unsqueeze(-2)
+            return RootLazyTensor(F)
+        except CachingError:
+            pass
+
+        if not self.is_square:
+            raise RuntimeError(
+                "root_inv_decomposition only operates on (batches of) square (symmetric) LazyTensors. "
+                "Got a {} of size {}.".format(self.__class__.__name__, self.size())
+            )
```
Balandat, line 1692: let's move this up to before the cache check

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1677,11 +1677,19 @@ def root_inv_decomposition(self, initial_vectors=None, test_vectors=None, method
         if self.shape[-2:].numel() == 1:
             return RootLazyTensor(1 / self.evaluate().sqrt())
 
-        # no need to do extra work if we already have a cached cholesky decomposition
-        # TODO: Resolve issues related to triangular tensor instantiation \
-        # if _is_in_cache_ignore_args(self, "cholesky"):
-        #     res = self.cholesky()
-        #     return CholLazyTensor(res).inverse()
+        # check to see if we have already run lanczos for a diagonalization
```
Balandat, line 1680: we should probably check the cache for both diagonalization and for cholesky, so that we don't end up unnecessarily computing things?

mwlon:
@Balandat I think this should address your comments

gpleiss (approved):
LGTM. I'll let @Balandat give it the final okay.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1590,9 +1583,30 @@ def diagonalization(self, method: Optional[str] = None):
...
 
+    def _choose_root_method(self, method: Optional[str] = None) -> str:
+        # Better inform which root_decomposition or root_inv_decomposition
+        # method to use based on available caches and matrix size.
+        if _is_in_cache_ignore_all_args(self, "symeig"):
+            return "symeig"
+        if _is_in_cache_ignore_all_args(self, "diagonalization"):
+            return "diagonalization"
+        if _is_in_cache_ignore_all_args(self, "lanczos"):
+            return "lanczos"
```
Balandat, line 1599: I like how you pulled out this functionality to keep things tidy. 

One concern is that this may introduce a bit of a gotcha, since it's not really transparent to the user that the `method` selection will be overridden silently (I guess that was the case prior to this change as well, though). That is, say you compute the root via Lanczos first, and are dissatisfied with the quality of the approximation and thus want to use `symeig` instead. If you just call `root_decomposition` with the `symeig` method you just going to get lanczos back from the cache without knowing.

One thing we could do is add a `force_method` arg to this method (and to `root_decomposition`) that used the desired method regardless of what's in cache. An alternative would be to issue a warning if the desired method does not end up getting selected due to other roots being readily available in cache. 

Curious to hear what @jacobrgardner, @gpleiss, @wjmaddox think about this.

Balandat (commented):
Overall this looks great. One comment about silently overriding user choices that we should further discuss.

gpytorch/lazy/lazy_tensor.py
wjmaddox: At least from my experience, this logic is a pretty strong gotcha.  Returning lanczos decompositions when you're intending to specify `cholesky` can result in some pretty sub-optimal behavior. I think there's some downstream code (SKI fantasy strategy) that could end up having both lanczos and cholesky roots being computed at different times, and there I know that the cholesky is definitely required at a certain time. Maybe just wrapping this code in a if method is None tag will clean it up in the default setting?

mwlon: It's an interesting point because
* In the case of a 1x1 matrix we're ignoring method and just doing 1/sqrt(self), and I think we want to keep that behavior
* The existing implementation actually checked the `diagonalization` cache before some of the methods, so we'll be changing that behavior
I agree we should use largely use method when specified though. I'll change the logic to do that.

Balandat: > In the case of a 1x1 matrix we're ignoring method and just doing 1/sqrt(self), and I think we want to keep that behavior

Definitely. I don't think we need to do lanczos to take square roots. We could add a `babylonian` method maybe :)

mwlon: So are we good to go with this PR?

PR #1586, opened 2021-04-23
Use torch.linalg.cholesky_ex instead of torch.cholesky

Balandat:
This can *significantly* speed up `psd_safe_cholesky` due to cutting out the pytorch error-handling middle man, achieving ~2,000 X reduction in wall time: https://github.com/pytorch/pytorch/pull/56724#issuecomment-825397299.

This also allows us to add jitter to the specific batch elements for which the decomp failed (rather than indiscriminately to all).

This requires https://github.com/pytorch/pytorch/pull/56724 that hasn't landed yet but will be part of 1.9. Either way, I implemented this in a backward-compatible fashion so this will work with older pytorch versions as well.

### DIFF
--- a/gpytorch/utils/cholesky.py
+++ b/gpytorch/utils/cholesky.py
@@ -45,13 +45,13 @@ except ImportError:
     # Fall back to torch.cholesky - this can be more than 3 orders of magnitude slower!
     # TODO: Remove once PyTorch req. is >= 1.9
 
-    def _psd_safe_cholesky(A, upper=False, out=None, jitter=None, max_tries=3):
+    def _psd_safe_cholesky(A, out=None, jitter=None, max_tries=3):
         # Maybe log
         if settings.verbose_linalg.on():
             settings.verbose_linalg.logger.debug(f"Running Cholesky on a matrix of size {A.shape}.")
 
         try:
-            L = torch.cholesky(A, upper=upper, out=out)
+            L = torch.cholesky(A, out=out)
             return L
         except RuntimeError as e:
             isnan = torch.isnan(A)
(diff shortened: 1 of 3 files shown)


gpytorch/utils/cholesky.py
```diff
@@ -8,51 +8,97 @@
...
+            :attr:`jitter` (float, optional):
+                The jitter to add to the diagonal of A in case A is only p.s.d. If omitted,
+                uses settings.cholesky_jitter.value()
+            :attr:`max_tries` (int, optional):
+                Number of attempts (with successively increasing jitter) to make before raising an error.
+        """
+    L = _psd_safe_cholesky(A, out=out, jitter=jitter, max_tries=max_tries)
+    if upper:
+        if out is not None:
+            out = out.transpose_(-1, -2)
```
Balandat, line 101: I guess the assignment is kind of pointless but oh well

gpytorch/utils/cholesky.py
```diff
@@ -8,51 +8,97 @@
...
+                See torch.cholesky
+            :attr:`out` (Tensor, optional):
+                See torch.cholesky
+            :attr:`jitter` (float, optional):
+                The jitter to add to the diagonal of A in case A is only p.s.d. If omitted,
+                uses settings.cholesky_jitter.value()
+            :attr:`max_tries` (int, optional):
+                Number of attempts (with successively increasing jitter) to make before raising an error.
+        """
+    L = _psd_safe_cholesky(A, out=out, jitter=jitter, max_tries=max_tries)
```
vishwakftw, line 98: You could pass `upper` here to avoid the transposing below.

Balandat: so the reason I'm doing it this way is that `cholesky_ex` does not support the `upper` arg, so I wouldn't be able to have `_psd_safe_cholesky` have the same signature for both `cholesky_ex` and `cholesky` internals. This ensures that `_psd_safe_cholesky` is functionally equivalent whether or not `cholesky_ex` is available or not

Balandat:
@jacobrgardner can we get this in?`cholesky_ex` has been merged into master, and it would be really nice to use it.

gpytorch/utils/cholesky.py
vishwakftw: Oh I thought `cholesky_ex` supported the `upper` arg. Strange...

Balandat:
hmm looks like I need to fix a number of mock call counts in a backward compatible way.

gpleiss (approved):

PR #1593, opened 2021-04-26
Periodic kernel handles multi-dimensional inputs

### LINKED ISSUE #835: Periodic kernel for multivariate input
# 🐛 Bug

Implementation of PeriodicKernel is not positive definite in case of multivariate input
Here is how I checked it:

```python
import torch
from gpytorch import kernels

X = torch.randn(1000, 2, dtype=torch.float64)
kernel = kernels.PeriodicKernel().double()
with torch.no_grad():
    K = kernel(X, X).evaluate()
    eig = torch.symeig(K)[0]

    negative_eig = eig[eig < 0]
    assert torch.allclose(negative_eig, torch.zeros_like(negative_eig))
```

```
AssertionError                            Traceback (most recent call last)
<ipython-input-2-1d738d7c73d9> in <module>
      9 
     10     negative_eig = eig[eig < 0]
---> 11     assert torch.allclose(negative_eig, torch.zeros_like(negative_eig))

AssertionError: 
```

## Expected Behavior

Positive definite kernel must have non-negative eigenvalues.

##
The implementation is correct for one-dimensional input. For multidimensional input it should be product of one-dimensional kernels, however, it calculates the norm first and applies sine and exponent to the norm which is not correct: see eq. (47) on page 15 of
> MacKay, D. J. (1998). Introduction to Gaussian processes. NATO ASI Series F Computer and Systems Sciences, 168, 133-166.

## System information

- gpytorch=0.3.4
- torch=1.1.0
- Ubuntu 19.04

### LINKED ISSUE #1615: [Bug] Wrong formulation of the PeriodicKernel
# 🐛 Bug

The current formulation of the periodic kernel is the following:

However, the original formula proposed [Mackay, 1998](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.81.1927&rep=rep1&type=pdf) (Eq. 47) is the following:

The constant is not -2 or -0.5 is not really important (it will have an impact on the derivative only). However, the original formulation is taking the the sum of the sine squared of the difference. Currently, the formulation is the sine squared of the sum of the difference. Note that the confusion could come from the fact that in the literature, the kernel is used for 1D signal and thus there is no sum, thus both formulations are equivalent.

In addition, looking at the implementation, it seems that the distance computed between the samples is norm-2 distance and not the norm-1 distance.

The current formulation make it that the kernel is not PSD. In addition, there is still another bug with a missing square over the length scale as reported in https://github.com/cornellius-gp/gpytorch/issues/1020.

Note: I found this issue because I am currently working on a similar issues in scikit-learn:
https://github.com/scikit-learn/scikit-learn/issues/19343
https://github.com/scikit-learn/scikit-learn/pull/20070

## To reproduce

```python
import numpy as np
import torch
from gpytorch.kernels import PeriodicKernel

r = np.linspace(0, 1, num=4)
train_x, train_y = np.meshgrid(r, r)
X = np.stack((train_x.flatten(), train_y.flatten()), axis=-1)

periodicity, length_scale=1.5, 0.5
kernel_pytorch = PeriodicKernel(
    period=periodicity, length_scale=length_scale
)
K_pytorch = kernel_pytorch(torch.from_numpy(X))

np.linalg.eigh(K_pytorch.numpy())[0]
```

```
array([-1.27815495, -1.21435487, -1.21435487, -0.93566031, -0.75456704,
       -0.24301193, -0.24301193,  0.47435512,  0.72370254,  1.58773803,
        1.58773803,  2.42356711,  2.67356606,  2.67356606,  3.10565507,
        6.63322787])
```

There is some negative eigenvalues.

## Expected Behavior

We are investigating the issue in scikit-learn and already implemented a naive approach: https://github.com/scikit-learn/scikit-learn/pull/20070 for tracking the bug. With the code of the PR, there is no negative eigenvalues:

```python
from sklearn.gaussian_process.kernels import ExpSineSquared

kernel_sklearn = ExpSineSquared(
    periodicity=periodicity, length_scale=length_scale,
)
K_sklearn = kernel_sklearn(X)
np.linalg.eigh(K_sklearn)[0]
```

```
array([0.88422032, 0.92031869, 0.92031869, 0.95789078, 0.95961332,
       0.95961332, 0.99716727, 0.99716727, 0.99878962, 0.99878962,
       1.03787671, 1.03787671, 1.0414347 , 1.08219069, 1.08219069,
       1.12454163])
```

On the process, we also worked with the implementation of tensorflow-probability that follow the equation of the paper, and we have consistent results:

```python
from tensorflow_probability import math

kernel_tfp = math.psd_kernels.ExpSinSquared(
   period=periodicity, length_scale=length_scale
)
K_tfp = kernel_tfp.matrix(X, X)
np.linalg.eigh(K_tfp.numpy())[0]
``` 

```
array([0.8842203 , 0.92031866, 0.92031866, 0.95789075, 0.9596133 ,
       0.9596133 , 0.9971673 , 0.9971673 , 0.9987896 , 0.9987896 ,
       1.0378767 , 1.0378767 , 1.0414346 , 1.0821906 , 1.0821906 ,
       1.1245416 ], dtype=float32)
```

## System information

- GPyTorch version: `1.4.1`
- PyTorch version: `1.8.1+cpu`
- Linux 5.8.0-50-generic #56-Ubuntu SMP Mon Apr 12 17:18:36 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

tawe141:
Periodic kernel now uses eq. 47 in David Mackay's Introduction to Gaussian Processes, found here: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.81.1927&rep=rep1&type=pdf

with one small difference: it doesn't do different period lengths or lengthscales for each feature. Fixes #835 

If I were to add in ARD, would this be a different PR or can I add it here? Sorry, new to this whole contributing to open source thing, so let me know what works best for everyone. Thanks for your patience :)

Fixes #1615

### DIFF
--- a/gpytorch/kernels/periodic_kernel.py
+++ b/gpytorch/kernels/periodic_kernel.py
@@ -16,9 +16,8 @@ class PeriodicKernel(Kernel):
 
         \begin{equation*}
             k_{\text{Periodic}}(\mathbf{x_1}, \mathbf{x_2}) = \exp \left(
-            -\frac{1}{2} \sum_i \left(
-            \frac{\sin \left( \frac{\pi}{\lambda} (\mathbf{x_{1,i}} - \mathbf{x_{2,i}}\right)}{p}
-            \right)^2
+            -2 \sum_i
+            \frac{\sin ^2 \left( \frac{\pi}{p} (\mathbf{x_{1,i}} - \mathbf{x_{2,i}} ) \right)}{\lambda}
             \right)
         \end{equation*}
 
@@ -27,9 +26,10 @@ class PeriodicKernel(Kernel):
     * :math:`p` is the period length parameter.
     * :math:`\lambda` is a lengthscale parameter.
 
-    Equation is taken from [David Mackay's Introduction to Gaussian Processes equation 47]
+    Equation is based on [David Mackay's Introduction to Gaussian Processes equation 47]
     (http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.81.1927&rep=rep1&type=pdf)
-    albeit without feature-specific lengthscales and period lengths.
+    albeit without feature-specific lengthscales and period lengths. The exponential
+    coefficient was changed and lengthscale is not squared to maintain backwards compatibility
 
     .. note::
 
@@ -120,7 +120,7 @@ class PeriodicKernel(Kernel):
         else:
             raise ValueError("x.dims() must be 2 or 3")
         diff = diff * math.pi / self.period_length.unsqueeze(-1)
-        res = diff.sin().div(self.lengthscale.unsqueeze(-1)).pow(2).sum(dim=-1).mul(-0.5).exp_()
+        res = diff.sin().pow(2).sum(dim=-1).div(self.lengthscale).mul(-2.0).exp_()
         if diag:
             res = res.squeeze(0)
         return res
--- a/gpytorch/kernels/periodic_kernel.py
+++ b/gpytorch/kernels/periodic_kernel.py
@@ -107,10 +107,18 @@ class PeriodicKernel(Kernel):
         self.initialize(raw_period_length=self.raw_period_length_constraint.inverse_transform(value))
 
     def forward(self, x1, x2, diag=False, **params):
-        x1_ = x1.div(self.period_length)
-        x2_ = x2.div(self.period_length)
-        diff = self.covar_dist(x1_, x2_, diag=diag, **params)
-        res = torch.sin(diff.mul(math.pi)).pow(2).mul(-2 / self.lengthscale).exp_()
+        # x1_ = x1.div(self.period_length)
+        # x2_ = x2.div(self.period_length)
+        # diff = self.covar_dist(x1_, x2_, diag=diag, **params)
+        if x1.dim() == 3 and x2.dim() == 3:  # batched operation
+            diff = x1.unsqueeze(2) - x2.unsqueeze(1)
+        elif x1.dim() == 2 and x2.dim() == 2:
+            diff = x1.unsqueeze(1) - x2.unsqueeze(0)
+        else:
+            raise ValueError("x.dims() must be 2 or 3")
+        diff = diff * math.pi / self.period_length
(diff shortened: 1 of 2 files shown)


gpytorch/kernels/periodic_kernel.py
```diff
@@ -107,10 +113,14 @@ def _set_period_length(self, value):
         self.initialize(raw_period_length=self.raw_period_length_constraint.inverse_transform(value))
 
     def forward(self, x1, x2, diag=False, **params):
-        x1_ = x1.div(self.period_length)
-        x2_ = x2.div(self.period_length)
-        diff = self.covar_dist(x1_, x2_, diag=diag, **params)
-        res = torch.sin(diff.mul(math.pi)).pow(2).mul(-2 / self.lengthscale).exp_()
+        if x1.dim() == 3 and x2.dim() == 3:  # batched operation
```
gpleiss, line 116: This code will fail if you have a multi-batch GP (4 or more tensor dimensions). It's better to do `diff = x1.unsqueeze(-2) - x2.unsqueeze(-3)`.

gpytorch/kernels/periodic_kernel.py
```diff
@@ -107,10 +113,14 @@ def _set_period_length(self, value):
...
+            diff = x1.unsqueeze(2) - x2.unsqueeze(1)
+        elif x1.dim() == 2 and x2.dim() == 2:
+            diff = x1.unsqueeze(1) - x2.unsqueeze(0)
+        else:
+            raise ValueError("x.dims() must be 2 or 3")
+        diff = diff * math.pi / self.period_length.unsqueeze(-1)
```
gpleiss, line 122: I've found that things tend to be more mathematically stable when you divide by the period length before you compute the difference. Can we go back to using L110-111?

gpleiss (changes requested):
Hey @tawe141 sorry for the slow review! Thanks for the contribution - just a few small changes and then we can merge :)

gpleiss (approved):

jacobrgardner:
Also Fixes #1615

Issue #1286, opened 2020-09-13
Bayesian GPs with pyro (NUTS) - example notebook crashes when jit_compile=True


mihai-spire:
I'm trying to run your [fully Bayesian GP example](https://docs.gpytorch.ai/en/v1.2.0/examples/01_Exact_GPs/GP_Regression_Fully_Bayesian.html). 

The notebook runs OK as-is. As you may expect, sampling is much slower when i increase the size of the training dataset. I've tried to enable `jit` compilation in the `pyro` NUTS sampler:
```python
nuts_kernel = NUTS(pyro_model, adapt_step_size=True, jit_compile=True)
```

After this change, the NUTS sampler crashes:
```
Warmup:   0%|          | 0/300 [00:00, ?it/s]
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-56cc00113944> in <module>()
     26 nuts_kernel = NUTS(pyro_model, adapt_step_size=True, jit_compile=True)
     27 mcmc_run = MCMC(nuts_kernel, num_samples=num_samples, warmup_steps=warmup_steps, disable_progbar=smoke_test)
---> 28 mcmc_run.run(train_x, train_y)

24 frames
/usr/local/lib/python3.6/dist-packages/pyro/poutine/messenger.py in _context_wrap(context, fn, *args, **kwargs)
      9 def _context_wrap(context, fn, *args, **kwargs):
     10     with context:
---> 11         return fn(*args, **kwargs)
     12 
     13 

/usr/local/lib/python3.6/dist-packages/pyro/infer/mcmc/api.py in run(self, *args, **kwargs)
    378         with optional(pyro.validation_enabled(not self.disable_validation),
    379                       self.disable_validation is not None):
--> 380             for x, chain_id in self.sampler.run(*args, **kwargs):
    381                 if num_samples[chain_id] == 0:
    382                     num_samples[chain_id] += 1

/usr/local/lib/python3.6/dist-packages/pyro/infer/mcmc/api.py in run(self, *args, **kwargs)
    167             for sample in _gen_samples(self.kernel, self.warmup_steps, self.num_samples, hook_w_logging,
    168                                        i if self.num_chains > 1 else None,
--> 169                                        *args, **kwargs):
    170                 yield sample, i  # sample, chain_id
    171             self.kernel.cleanup()

[...]

PR #1594, opened 2021-04-27
Fix pyro sampling

### LINKED ISSUE #1578: [Bug] Fully Bayesian Inference with Pyro Does Not Work
# 🐛 Bug

Right now, the way our Bayesian inference with pyro works is that we loop over all `Parameter`s with priors, call `pyro.sample` and load the sample into the parameter value. Here's an extremely simplified model that is equivalent to what we do that does NOT work:
```python
def pyro_model2(x, y):
    raw_lengthscale = pyro.sample('raw_lengthscale', Normal(0, 1))
    raw_outputscale = pyro.sample('raw_outputscale', Normal(0, 1))

    # This is not an okay way to load the results of pyro.sample statements into a Module
    covar_module.initialize(**{'raw_outputscale': raw_outputscale, 'base_kernel.raw_lengthscale': raw_lengthscale})
    
    covar_x = covar_module(x).evaluate()
    pyro.sample("obs", pyro.distributions.MultivariateNormal(torch.zeros(y.size(-1)), covar_x + torch.eye(y.size(-1))), obs=y)
```
See the attached python notebook that demonstrates that this leads to completely wrong gradients for the potential energy, so HMC is entirely broken.

The best way I've found to fix this is to use the (possibly deprecated?) `pyro.random_module` primitive. Here's a pyro model using a full GPyTorch GP that gets correct potential derivatives:
```python
def pyro_model5(x, y):
    priors= {
        'covar_module.base_kernel.raw_lengthscale': Normal(0, 1).expand([1, 1]),
        'covar_module.raw_outputscale': Normal(0, 1),
        'likelihood.noise_covar.raw_noise': Normal(0, 1).expand([1]),
        'mean_module.constant': Normal(0, 1),
   }
    fn = pyro.random_module("model", model, prior=priors)
    sampled_model = fn()
    
    output = sampled_model.likelihood(sampled_model(x))
    pyro.sample("obs", output, obs=y)
```
Now, we could definitely just replace our existing GPyTorch interface to basically do the above instead for any parameter that has a prior registered. The only problem is that I don't know if this will let us place priors over functions of `Parameters` (i.e., if we want to place a prior over the `lengthscale` rather than the `raw_lengthscale`).

Thoughts? I feel like the above is *almost* there, but it'd be great to still support placing priors over derived values of the parameters.

cc/ @rmgarnett

also cc/ @fritzo @eb8680 @martinjankowiak in case any of the pyro devs hopefully have a thought about how to do this properly that I'm just missing?

[Sampling Bug.ipynb.txt](https://github.com/cornellius-gp/gpytorch/files/6328496/Sampling.Bug.ipynb.txt)

jacobrgardner:
The implementation of this fix gets a bit nasty but has the advantage that changes to the existing sampling interface are minimal as a result, e.g.

```python
def pyro_model(x, y):
    model.pyro_sample_from_prior()
    output = model(x)
    pyro.sample("obs", output, obs=y)
```

becomes:

```python
def pyro_model(x, y):
    sampled_model = model.pyro_sample_from_prior()
    output = sampled_model(x)
    pyro.sample("obs", output, obs=y)
```

Given that none of the code I touched gets used anywhere in the package except specifically when using pyro for HMC, it's probably okay.

@rmgarnett @Balandat @sdaulton 

Fixes #1578


sdaulton:
Thanks for fixing this! Running the example notebook with `jit_compile=True` still crashes (https://github.com/cornellius-gp/gpytorch/issues/1286). Any idea what the issue is there?

sdaulton:
Also, when using `load_state_dict()` as in the last cell of the example notebook, we need to set the `batch_shape` of each module according to the `batch_shape` of its loaded parameters

jacobrgardner:
Yeah, I noticed the `jit_compile` error. It now looks to be much more tame though and is just some shape bug somewhere. Should be easier to track down. My plan is to get this merged in so that the integration actually works and then track that down.

On a more practical note, a huge part of why the pyro integration was slow before was because with incorrect gradients the adaptive step size portion of HMC kept shrinking the step size until it would take forever to draw each sample -- note that with incorrect gradients the sampler was still a "valid" MCMC sampler probably, it just had a terrible proposal distribution.

There's a pretty solid chance that a lot of your perf issues were caused by #1578 and not #1286 (and actually, the investigation that led to #1578 initially started as investigating a perf issue). On some examples, the pyro integration is easily 100x+ faster after this PR.

sdaulton:
Yes, I have tested your changes and it is way faster. I have a pure pytorch + pyro GP implementation that I am benchmarking this gpytorch + pyro fix against. The model fits are the same, but the gpytorch version is ~5x slower than the pytorch version (some of this is due to python overhead with lazy tensors, etc), but it would be awesome to make this as fast as possible and jit_compile should help to some extent

gpleiss (approved):

PR #1608, opened 2021-05-04
Fix IndexKernel for non-square cases and mixed discrete-continuous inputs

mshvartsman:
1. Currently IndexKernel doesn't work if the two inputs (`i1` and `i2`) have mismatching index shapes, and this looks like a broadcasting bug to me, hopefully a minor fix. 
2. Currently IndexKernel is incompatible with using `ard_num_dims` and `active_dims` to split a single input into continuous and discrete kernels (via additive or multiplicative kernel, or some other wrapper). Not sure if this is intended functionality, but it'd be nice to have. The explicit cast fixes that.

### DIFF
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -85,7 +85,7 @@ class IndexKernel(Kernel):
 
         i1, i2 = i1.long(), i2.long()
         covar_matrix = self._eval_covar_matrix()
-        batch_shape = _mul_broadcast_shape(i1.shape[:-2], self.batch_shape)
+        batch_shape = _mul_broadcast_shape(i1.shape[:-2], i2.shape[:-2], self.batch_shape)
 
         res = InterpolatedLazyTensor(
             base_lazy_tensor=covar_matrix,
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -83,7 +83,7 @@ class IndexKernel(Kernel):
 
     def forward(self, i1, i2, **params):
 
-        i1, i2 = torch.LongTensor(i1.long()), torch.LongTensor(i2.long())
+        i1, i2 = i1.long(), i2.long()
         covar_matrix = self._eval_covar_matrix()
         batch_shape = _mul_broadcast_shape(i1.shape[:-2], self.batch_shape)
         index1_shape = batch_shape + i1.shape[-2:]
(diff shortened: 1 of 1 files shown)


gpytorch/kernels/index_kernel.py
```diff
@@ -82,13 +82,16 @@ def covar_matrix(self):
         return res
 
     def forward(self, i1, i2, **params):
+
+        i1, i2 = torch.LongTensor(i1.long()), torch.LongTensor(i2.long())
```
Balandat, line 86: You mentioned explicit cast, but I'm nor sure I understand why you have the `LongTensor` here. This is a nullop on the CPU and breaks on the GPU. 
```suggestion
        i1, i2 = i1.long(), i2.long()
```

gpytorch/kernels/index_kernel.py
```diff
@@ -82,13 +82,16 @@ def covar_matrix(self):
...
-        index_shape = batch_shape + i1.shape[-2:]
+        index1_shape = batch_shape + i1.shape[-2:]
+        index2_shape = batch_shape + i2.shape[-2:]
 
         res = InterpolatedLazyTensor(
             base_lazy_tensor=covar_matrix,
-            left_interp_indices=i1.expand(index_shape),
-            right_interp_indices=i2.expand(index_shape),
+            left_interp_indices=i1.expand(index1_shape),
+            right_interp_indices=i2.expand(index2_shape),
```
Balandat, line 95: not used, may as well save the two lines

```
            left_interp_indices=i1.expand(batch_shape + i1.shape[-2:]),
            right_interp_indices=i2.expand(batch_shape + i2.shape[-2:])
```

gpytorch/kernels/index_kernel.py
```diff
@@ -82,13 +82,16 @@ def covar_matrix(self):
         return res
 
     def forward(self, i1, i2, **params):
+
+        i1, i2 = torch.LongTensor(i1.long()), torch.LongTensor(i2.long())
         covar_matrix = self._eval_covar_matrix()
         batch_shape = _mul_broadcast_shape(i1.shape[:-2], self.batch_shape)
```
Balandat, line 88:
```suggestion
        batch_shape = _mul_broadcast_shape(i1.shape[:-2], i2.shape[:-2], self.batch_shape)
```

gpytorch/kernels/index_kernel.py
mshvartsman: Oops, you're right. That's just me not knowing pytorch, apparently :).

Balandat (approved):
This lgtm. Ideally we'd have some units to test this behavior, but I guess we can add these in a separate PR if we really want to (@mshvartsman does, I am sure).

PR #1609, opened 2021-05-05
Working keops bypass for RBF

jacobrgardner:
The KeOps bypass logic in the KeOps RBF kernel doesn't work if `diag=True` since the lambda function is explicitly created with `diag=False`. In fixing it I also moved the logic into `forward` since there's not much value in creating a KeOpsLazyTensor out of the result if we're not going to use KeOps as far as I can tell.

The Matern kernel is fine.

### DIFF
--- a/gpytorch/kernels/keops/rbf_kernel.py
+++ b/gpytorch/kernels/keops/rbf_kernel.py
@@ -26,15 +26,6 @@ try:
             )
 
         def covar_func(self, x1, x2, diag=False):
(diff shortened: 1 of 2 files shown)


jacobrgardner:
Oops, this is still broken -- only worked on one test. I'll reopen when I have a better fix.

jacobrgardner:
Now fixed in the same way that Matern kernel handles this case.

Balandat (approved):

PR #1611, opened 2021-05-06
Do not instantiate autograd functions

Balandat:
`apply` is a static method, let's use it as such. In anticipation of https://github.com/pytorch/pytorch/pull/57357


Balandat:
@jacobrgardner, @gpleiss can we merge this in ? This is raising a gazillion deprecation warnings on pytorch master

gpleiss (approved):

Issue #1634, opened 2021-06-03
[Bug] module.initialize does not work with module lists (additive kernels)


jpchen:
[`module.initialize`](https://docs.gpytorch.ai/en/stable/module.html?#gpytorch.Module.initialize) does not seem to handle module lists properly eg when an additive kernel is used. This seems to be because of the [recursive `initialize` call](https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/module.py#L90) and `ModuleLists` do not have an `initialize` method as `gpytorch.Modules` do. One dirty solution might be to include the index into the namespace and index in the event of a module list.

cc: @caozhen-bupt

Example:
<details>

```python
In [32]: 
    ...: class ExactGPModel(gpytorch.models.ExactGP):
    ...:     def __init__(self, train_x, train_y, likelihood):
    ...:         super(ExactGPModel, self).__init__(train_x, train_y, gpytorch.likelihoods.GaussianLikelihood())
    ...:         self.covar_module = gpytorch.kernels.RBFKernel()
    ...:         self.additive = gpytorch.kernels.RBFKernel() + gpytorch.kernels.RBFKernel()
    ...:
    ...:     def forward(self, x):
    ...:         pass
    ...:
    ...: # initialize likelihood and model
    ...: likelihood = gpytorch.likelihoods.GaussianLikelihood()
    ...: model = ExactGPModel(train_x, train_y, likelihood)

In [33]: model.initialize(**{'covar_module.lengthscale': torch.tensor(1.)})  # this is ok

In [34]: model.additive
Out[34]:
AdditiveKernel(
  (kernels): ModuleList(
    (0): RBFKernel(
      (raw_lengthscale_constraint): Positive()
    )
    (1): RBFKernel(
      (raw_lengthscale_constraint): Positive()
    )
  )
)

In [35]: model.initialize(**{'additive.kernels.0.lengthscale': torch.tensor(1.)})
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-30-38d6b8203c68> in <module>
----> 1 model.initialize(**{'additive.kernels.0.lengthscale': torch.tensor(1.)})

~/miniconda3/envs/bm/lib/python3.7/site-packages/gpytorch/module.py in initialize(self, **kwargs)
     88             if "." in name:
     89                 module, name = self._get_module_and_name(name)
---> 90                 module.initialize(**{name: val})
     91             elif not hasattr(self, name):
[...]

PR #1635, opened 2021-06-03
Fix initialize bug for additive kernels.

jpchen:
Addresses #1634. 

cc: @Balandat

### DIFF
--- a/gpytorch/module.py
+++ b/gpytorch/module.py
@@ -87,7 +87,11 @@ class Module(nn.Module):
                 val = float(val)
             if "." in name:
                 module, name = self._get_module_and_name(name)
-                module.initialize(**{name: val})
+                if isinstance(module, nn.ModuleList):
+                    idx, name = name.split(".", 1)
+                    module[int(idx)].initialize(**{name: val})
+                else:
+                    module.initialize(**{name: val})
             elif not hasattr(self, name):
                 raise AttributeError("Unknown parameter {p} for {c}".format(p=name, c=self.__class__.__name__))
             elif name not in self._parameters and name not in self._buffers:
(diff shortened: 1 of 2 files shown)


gpytorch/module.py
```diff
@@ -87,7 +87,11 @@ def initialize(self, **kwargs):
                 val = float(val)
             if "." in name:
                 module, name = self._get_module_and_name(name)
-                module.initialize(**{name: val})
+                if isinstance(module, nn.ModuleList):
+                    idx, name = name.split(".", 1)
```
Balandat, line 91: Should we handle the case of an empty `ModuleList` in the model (in that case you'd get a `TypeError` from trying to unpack a single element)? Might be safe to ignore since one generally wouldn't have that as a key in the first place...

gpytorch/module.py
```diff
@@ -87,7 +87,11 @@ def initialize(self, **kwargs):
...
+                    module[int(idx)].initialize(**{name: val})
+                else:
+                    module.initialize(**{name: val})
```
Balandat, line 94: why not just 
```suggestion
                    module[int(idx)].initialize(name=val)
                else:
                    module.initialize(name=val)
```

gpytorch/module.py
jpchen: I was thinking that `ModuleLists` are organically created from composing kernels (and so wouldn't be empty) but sure it makes sense to support that. This also makes the assumption that `ModuleList`s contains `Modules` which can contain other `ModuleList`s but `ModuleList`s aren't directly nested.

Balandat: I smell recursion...

jpchen: Oops duh I wrote my test wrong; nested module lists will have a `kernels` in its namespace so it yep it will do the right thing. Will update with the empty ModuleList case.

gpytorch/module.py
jpchen: This treats "name" as the key in the kwarg vs the string the variable `name` contains.

Balandat: argh brainfart

gpytorch/module.py
Balandat: yeah I guess that's what I mean by 

> Might be safe to ignore since one generally wouldn't have that as a key in the first place...

Balandat (approved):

PR #1640, opened 2021-06-08
_cholesky should use default jitter

wjmaddox:
Currently, `LazyTensor._cholesky` isn't using jitter when being passed into the psd_safe_cholesky method. This should fix that, making `_cholesky` more stable.


Balandat (commented):
So this should already be doing the right thing: https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/utils/cholesky.py#L34-L35

I faintly remember that the reason we did not add this to the call had sth to do with backward compatibility of the jitter arg https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/utils/cholesky.py#L34-L35 ? I might be wrong though

wjmaddox:
Weird, I wasn't able to set the tolerance earlier until I changed it to be the PR.

wjmaddox:
Closing this for now.

Issue #1673, opened 2021-06-24
[Bug] InvQuadLogDet with SumKroneckerLazyTensor doesn't work


wjmaddox:
# 🐛 Bug

solves + logdets fail when using `inv_quad_logdet` on a SumKroneckerLazyTensor. I'll put up a PR fixing it soon hopefully.

## To reproduce

** Code snippet to reproduce **
```python
import torch
from gpytorch.kernels import MaternKernel
from gpytorch.lazy import SumKroneckerLazyTensor, KroneckerProductLazyTensor, DiagLazyTensor

kernels = [MaternKernel()] * 3
x = [torch.randn(20, 1)] * 3
kplt1 = KroneckerProductLazyTensor(*[k(xx) for xx, k in zip(x, kernels)])

kplt2 = KroneckerProductLazyTensor(*[DiagLazyTensor(torch.rand(20))]*3)

sum_kplt = kplt1 + kplt2

rhs = torch.randn(sum_kplt.shape[-1], 1)

sum_kplt.inv_matmul(rhs) # works

sum_kplt.inv_quad_log_det(inv_quad_rhs=rhs, logdet = True) # error
```

** Stack trace/error message **
```
/mnt/xarfuse/uid-228567/f0320543-seed-nspid4026531836_cgpid17430139-ns-4026531840/gpytorch/lazy/lazy_tensor.py in inv_quad_logdet(self, inv_quad_rhs, logdet, reduce_inv_quad)
   1277         func = InvQuadLogDet.apply
   1278 
-> 1279         inv_quad_term, logdet_term = func(
   1280             self.representation_tree(),
   1281             self.dtype,

/mnt/xarfuse/uid-228567/f0320543-seed-nspid4026531836_cgpid17430139-ns-4026531840/gpytorch/functions/_inv_quad_log_det.py in forward(ctx, representation_tree, dtype, device, matrix_shape, batch_shape, inv_quad, logdet, probe_vectors, probe_vector_norms, *args)
    155         t_mat = None
    156         if ctx.logdet and settings.skip_logdet_forward.off():
--> 157             solves, t_mat = lazy_tsr._solve(rhs, preconditioner, num_tridiag=num_random_probes)
    158 
    159         else:

ValueError: too many values to unpack (expected 2)
```

## Expected Behavior

This should work, was attempting to write a specific type of high rank, tensorized Gaussian likelihood. 

## System information

**Please complete the following information:**
gpytorch 1.4.2
pytorch master

[...]

PR #1674, opened 2021-06-24
inv_quad_logdet method for sumKronecker

wjmaddox:
Fix for #1673. Had to do some weird things with the unit test because I think it's proper to allow `linear_cg` to be called in this situation (solve itself doesn't use cg, but the sub-solve does).

Edit: Also realized that previously rank > 1 multi-task gaussian likelihoods were not returning the specialized `SumKroneckerLazyTensor`, which is a pretty big performance hit (CG vs structured e-decomp solves / logdets).

### DIFF
--- a/test/lazy/test_sum_kronecker_lazy_tensor.py
+++ b/test/lazy/test_sum_kronecker_lazy_tensor.py
@@ -22,7 +22,7 @@ class TestSumKroneckerLazyTensor(LazyTensorTestCase, unittest.TestCase):
     seed = 0
     should_call_lanczos = True
     should_call_cg = False
-    skip_slq_tests = True
+    skip_slq_tests = False
 
     def create_lazy_tensor(self):
         a = torch.tensor([[4, 0, 2], [0, 3, -1], [2, -1, 3]], dtype=torch.float)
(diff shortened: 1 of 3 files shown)


test/lazy/test_sum_kronecker_lazy_tensor.py
```diff
@@ -47,3 +47,19 @@ def evaluate_lazy_tensor(self, lazy_tensor):
             lazy_tensor.lazy_tensors[1].lazy_tensors[0].tensor, lazy_tensor.lazy_tensors[1].lazy_tensors[1].tensor
         )
         return res1 + res2
+
+    def test_inv_quad_logdet(self):
+        # mock call cg here
+        self.__class__.should_call_cg = True
+        super().test_inv_quad_logdet()
+        self.__class__.should_call_cg = False
```
Balandat, line 55: Might this lead to issues with test concurrency if you manually mock this here?  How about the following (also below)?
```suggestion

    @mock.patch(
        "gpytorch.lazy.test_sum_kronecker_lazy_tensor.should_call_cg",
        new_callable=PropertyMock, 
        return_value=True,
    ):
    def test_inv_quad_logdet(self):
        super().test_inv_quad_logdet()
```

Balandat (approved):
lgtm thanks for the patch. Minor comment about mocking

test/lazy/test_sum_kronecker_lazy_tensor.py
wjmaddox: This approach sadly doesn't quite work, will deal with concurrency issues if they exist later on.

Issue #1680, opened 2021-06-30
[Feature Request] Sparse lazy tensor for SKI wights 


ymohit:
# 🚀 Feature Request

Currently, SKI weights (typically, written as W) are handled through two matrices containing indices and values separately. Also, special routines are written in utils to handle MVMs with them. These two things make it suitable for a new lazy tensor. 

## Motivation
As a result of this feature, the SKI code will be more readable. 

**Are you willing to open a pull request?** (We LOVE contributions!!!)
I'm happy to give it a try but as of now I'm not sure about two things: a) if at all such an abstraction will be computationally efficient (or even more efficient than utils.sparse.bdsmm procedure)? b) I'm not able to understand how batch shape is organized in this project and in particular in these functions. Would it be possible to have a sparse SKI weight in the batch made if such a lazy tensor is created?

wjmaddox:
Hi, I recently did a bunch of work using the matrices `W` in SKI, and this would be pretty helpful. For GPs, I think we generally consider solely the type of operations of `W K W' v` which is served by `InterpolatedLazyTensor` but for online learning with GPs, I know that one might want better access to `W` (e.g. `W v` or `W'W v`) instead...

Responding directly to a) if done properly, it should be give or take as computationally efficient as the current set of operations (maybe a bit slower due to slightly more overhead, outweighed by potential downstream usefulness) and b) lazy tensors should support batch mode when implemented correctly, in general, all up to the last dimension is considered batch dimensions. I could help support an attempt for constructing a SparseLazyTensor.

To develop a `fantasy_strategy` (e.g. conditioning a SKI model on new inputs), I ended up actually densifying the `W` matrices [here](https://github.com/cornellius-gp/gpytorch/blob/b237019940620b1f8546a28b71b632599684769f/gpytorch/models/exact_prediction_strategies.py#L374), but a lazy tensor would probably also have been useful :).

Re your second question, we need gradients of values for Bayesian optimization where one typically uses gradient based optimization with respect to input data points at test time (or even occasionally the training data itself). See [botorch repo](https://github.com/pytorch/botorch).

ymohit:
Thanks for responding @wjmaddox. I believe even W'W is likely to be sparse in many scenarios (e.g., https://arxiv.org/abs/2101.11751) which further suggests we could have lazy tensor even for that. 

Help on support SparseLazyTensor would be super useful. Thanks a lot!

[...]

PR #1682, opened 2021-07-02
Planning sparse lazy tensor

ymohit:
This is in response to #1680 and following the suggestion of @wjmaddox. As of now, the plan is as follows:

- [ ] Create sparse lazy tensor that can hold values and indices much like scipy sparse csr and/or coo matrices. In this step, we will stick to batch mode and other generic gpytorch features while giving the functionality of scipy. 
- [ ] Extend the above sparse lazy tensor to handle values and keys of SKI weight matrix W. This shall be a better alternative to using InterpolatedLazyTensor in typical GP SKI models.
- [ ]  We will visit methods ([factorized approach](https://arxiv.org/abs/2101.11751) and [online gp](https://arxiv.org/abs/2103.01454)) that require setting up W'Wv  and set up efficient MVM and updates routine for them.


jacobrgardner:
Part of the reason we implemented a lot of how we handle W ourselves is because back in like PyTorch 0.2 when we first implemented SKI, the sparse tensor support in native PyTorch was pretty slow.

It's still not a guarantee that just implementing W as a sparse tensor in native torch will be faster (or possible -- back then they didn't support batch matmuls), because W has special additional structure, but it could be worth a look.

If it's still better to special case W, I agree we should have SparseLazyTensor that wraps PyTorch native sparse tensors, and perhaps a subclass that deals with W specifically. 

Finally, I wonder if this effort should be done on the linear_operator package we're trying to pull out of GPyTorch, because SparseLazyTensors aren't necessarily positive definite. This could be a good motivation to finally finish up a minimal version of that package and make it a dependency for GPyTorch.

ymohit:
Thanks for commenting @jacobrgardner. 

I'll try to run some simple tests that shall allow us to compare current implementation with torch sparse case. 

For now, I'll push this work here only as I'm not able to see how gpytorch depends on the linera_operator library. Feel free to drop any pointer if I'm missing any infomation.

wjmaddox:
Linear_operator lives here: https://github.com/cornellius-gp/linear_operator and as you can see it's currently an inconsistently updated clone of `gpytorch.lazy` but the goal is to have `linear_operator` be a (more) stand-alone linear algebra package that supports generic linear operators beyond just those useful for GPs.

Balandat:
Ultimately the hope is that the basic infrastructure for this will be implemented in pytorch proper (as it is for scipy / Tensorflow): https://github.com/pytorch/pytorch/issues/28341.

PR #1709, opened 2021-07-24
Fix SGPR prediction speed with no fast_computations

jacobrgardner:
With `fast_computations` off, our `InvMatmul` doesn't call `_solve` it calls `cholesky` regardless of the lazy tensor, so `LowRankRootAddedDiagLazyTensor` needs to override `inv_matmul` in addition to `inv_quad_logdet`

Addresses #1708


Balandat (approved):
lgtm. Should we add a test to validate we're not calling Cholesky in this case?

jacobrgardner:
Modified `test_sgpr_regression` to verify that `cholesky` is only ever called on `m x m` matrices, rather than `n x n`. Also simplified the test file a bit since it predates our new pattern of having the cuda test and CPU test share code.

Issue #1713, opened 2021-07-28
[Bug] stable_pinverse fails when not full column rank


wjmaddox:
# 🐛 Bug

`stable_pinverse` fails if some columns are exactly zero. 

## To reproduce

** Code snippet to reproduce **
```python
import torch
from gpytorch.utils import stable_qr, stable_pinverse

A = torch.randn(30, 5)
A = torch.cat((A, torch.zeros(30, 25)), dim=-1)

stable_pinverse(A) 
```

** Stack trace/error message **
```
~/Documents/GitHub/wjm_gpytorch/gpytorch/utils/pinverse.py in stable_pinverse(A)
     13         # skinny (or square) matrix
     14         Q, R = stable_qr(A)
---> 15         return torch.triangular_solve(Q.transpose(-1, -2), R).solution
     16     else:
     17         # fat matrix

RuntimeError: triangular_solve_cpu: U(6,6) is zero, singular U.
```

## Expected Behavior

What's happening is that the jittering in stable_qr doesn't work if the diagonals are exactly zero (torch.sign returns zero)

## System information

**Please complete the following information:**
- gpytorch master
- pytorch 1.8.1

## Additional context

Will put up a fix for stable_qr

PR #1714, opened 2021-07-28
Force add to zero sign in stable_qr

wjmaddox:
Resolution to #1713. 

Forcibly sets zero sign returned values and sets them to one so we can add jitter to them. This enforces a full rank `R`.


Balandat (approved):
lgtm, thanks

PR #1725, opened 2021-08-13
Add dtype registry for symeig

### LINKED ISSUE #1213: [Feature Request] User-modifiable registry for methods to be run in fp64
# 🚀 Feature Request

Create an internal of registry of methods top be run in fp64 that can be user modified. We could accomplish this without directly patching e.g. `torch.symeig` / `torch.cholesky` by defining `gpytorch.symeig` / `gpytorch.cholesky` that get used internally. This would avoid hard coding fp64 casts everywhere.

## Motivation

See discussion here: https://github.com/cornellius-gp/gpytorch/pull/1105#discussion_r453872265

wjmaddox:
This PR adds a new setting for a dtype registry for symeig (not cholesky, but wouldn't be too difficult to set it there).

Resolves #1213. 

Relatedly, should we put up more unit tests for settings?

### DIFF
--- a/gpytorch/settings.py
+++ b/gpytorch/settings.py
@@ -671,7 +671,7 @@ class skip_logdet_forward(_feature_flag):
     pass will skip certain computations (i.e. the logdet computation), and will therefore
     be improper estimates.
 
-    If you're using SGD (or a varient) to optimize parameters, you probably
+    If you're using SGD (or a variant) to optimize parameters, you probably
     don't need an accurate MLL estimate; you only need accurate gradients. So
     this setting may give your model a performance boost.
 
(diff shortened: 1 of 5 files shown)


gpytorch/settings.py
```diff
@@ -681,6 +681,17 @@ class skip_logdet_forward(_feature_flag):
     _default = False
 
 
+class symeig_dtype(_value_context):
```
Balandat, line 684: Seems like this could lead to an inflation of value contexts if we do this for every operation. Would it be possible to group these together (similar to `fast_computations`?)

wjmaddox: Yes, ultimately, but from what I see now, it's likely the only value context that does computations in double. I could be wrong though.

wjmaddox: Hmm.. it does look like some choleskies are stored / performed in double, particularly these https://github.com/cornellius-gp/gpytorch/blob/4685d383f5cb28a39c8ef2b564881e2f0aa32aa7/gpytorch/variational/variational_strategy.py#L71 .

Balandat: If it's not too much work I'd advocate for generating an umbrella `settings.dtypes` registry so we can do things like `settings.dtypes.cholesky` or `settings.dtypes.symeig` from the start rather than introducing this and then having to deprecate it. But it that's annoying/difficult we can also merge this in as is.

Balandat (approved):

wjmaddox:
Added in a unit test for the symeig dtype hook within the `LazyTensor.test_symeig` construction. Just want to double-check that a) this is the implementation you wanted and b) that unit test is fine.

gpytorch/settings.py
```diff
@@ -681,6 +681,40 @@ class skip_logdet_forward(_feature_flag):
     _default = False
 
 
+class _linalg_symeig(_value_context):
```
Balandat, line 684: Suggest calling this `_linalg_dtype_symeig` (and `_linalg_dtype_*` more generally) for clarity. g

gpytorch/test/lazy_tensor_test_case.py
```diff
@@ -754,51 +755,58 @@ def test_sqrt_inv_matmul_no_lhs(self):
...
-        for idx in itertools.product(*[range(b) for b in evals_actual.shape[:-1]]):
-            eval_i = evals_actual[idx]
-            if torch.unique(eval_i.detach()).shape[-1] == eval_i.shape[-1]:  # detach to avoid pytorch/pytorch#41389
-                self.assertAllClose(evecs_abs[idx], evecs_actual_abs[idx], **self.tolerances["symeig"])
-            else:
-                any_evals_repeated = True
+        dtype_names = ["double", "float"]
+        dtypes = {"double": torch.double, "float": torch.float}
+        for name in dtype_names:
+            dtype = dtypes[name]
```
Balandat, line 761:
```suggestion
        dtypes = {"double": torch.double, "float": torch.float}
        for name, dtype in dtypes.items():
```

PR #1726, opened 2021-08-13
Add half and float calls to lazy tensor

wjmaddox:
Adds half and float calls to lazy tensor, mimicking the `.double()` calls that were already implemented.

```python
lazy_tensor.half()
```

### DIFF
--- a/gpytorch/lazy/interpolated_lazy_tensor.py
+++ b/gpytorch/lazy/interpolated_lazy_tensor.py
@@ -386,6 +386,13 @@ class InterpolatedLazyTensor(LazyTensor):
         else:
             return super(InterpolatedLazyTensor, self).diag()
 
+    def double(self, device_id=None):
+        # We need to ensure that the indices remain integers.
+        new_lt = super().double(device_id=device_id)
+        new_lt.left_interp_indices = new_lt.left_interp_indices.type(torch.int64)
+        new_lt.right_interp_indices = new_lt.right_interp_indices.type(torch.int64)
+        return new_lt
+
     def matmul(self, tensor):
         # We're using a custom matmul here, because it is significantly faster than
         # what we get from the function factory.
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1960,6 +1960,16 @@ class LazyTensor(ABC):
 
         return res
 
+    def type(self, dtype):
+        if dtype == torch.float:
+            return self.float()
+        elif dtype == torch.double:
+            return self.double()
+        elif dtype == torch.half:
+            return self.half()
+        else:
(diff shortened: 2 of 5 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -1924,6 +1960,16 @@ def transpose(self, dim1, dim2):
 
         return res
 
+    def type(self, dtype):
```
jacobrgardner, line 1963: I suppose at this point we could also implement `to` if we wanted to, since now we support `.cuda`, `.cpu`, and the dtype casts.

wjmaddox: Yeah, i could probably set that up in another PR. Currently, the `.to(...)` method has a bit different signature than the pytorch one so it'll require a bit of work.

Balandat (commented):
probably want to add a unit test to the base LT test case?

wjmaddox:
This commit resolves #1730, figured it was best to do inside of here.

gpytorch/lazy/interpolated_lazy_tensor.py
```diff
@@ -386,6 +386,13 @@ def diag(self):
         else:
             return super(InterpolatedLazyTensor, self).diag()
 
+    def double(self, device_id=None):
+        # We need to ensure that the indices remain integers.
+        new_lt = super().double(device_id=device_id)
+        new_lt.left_interp_indices = new_lt.left_interp_indices.type(torch.int64)
+        new_lt.right_interp_indices = new_lt.right_interp_indices.type(torch.int64)
```
Balandat, line 393: Would we ever want to move ints to a FP dtype? Otherwise we can just not move those over in the `float` and `double` calls (could add an optional `include_int` kwarg if we really wanted to allow forcing that). Then we wouldn't have to do this check here (which is likely a gotcha for new types).

Balandat (commented):
There is a lot of code duplication here. Can we refactor this so that the different convenience methods just call call `to` internally? Then you'd have to implement that only once.

gpytorch/lazy/interpolated_lazy_tensor.py
wjmaddox: I don't think we'd ever want to allow these to be non-integer valued due to the explicit indexiing that `InterpolatedLazyTensor` uses. Even `.half/float/double()` etc is really a misnomer because `InterpolatedLazyTensor` will include two different tensor types by default (integers and floating point values).

wjmaddox: Not passing these over to double precision integers also produces errors as well, so this type of type hack has to be performed in my opinion.

Balandat (approved):
Thanks for deduping the dup here. lgtm

PR #1737, opened 2021-08-25
Add sample method for LKJ correlation prior

### LINKED ISSUE #1690: Trying to sample from the LKJCovariancePrior
Trying to declare and LKJ prior and sample from it however it doesn't like the event_shape of the sd_prior:

```
import torch
from gpytorch.priors import LKJCovariancePrior, GammaPrior

sd_prior = GammaPrior(torch.tensor([1.0]),torch.tensor([1.0]), validate_args=True)
cov_prior = LKJCovariancePrior(n=10, eta=1, sd_prior=sd_prior)
```
```

  File "<ipython-input-40-1ca20a6657f1>", line 5, in <module>
    cov_prior = LKJCovariancePrior(n=10, eta=1, sd_prior=sd_prior)

  File "/home/vr308/anaconda3/lib/python3.8/site-packages/gpytorch/priors/lkj_prior.py", line 111, in __init__
    raise ValueError("sd_prior must have event_shape 1 or n")

ValueError: sd_prior must have event_shape 1 or n

```
event_shape is not a settable attribute and by default it is empty torch.Size([]) for a prior over scalar - is this a bug?

Thanks.

wjmaddox:
Adds support for sampling from the LKJ correlation prior. Wasn't sure if the sampled distribution should return correlation or covariance matrices, so I made it return correlation matrices as a default.

Example:
```python
from gpytorch.priors import LKJPrior

prior = LKJPrior(50, 0.3)

samples = prior.rsample(torch.Size((256,)))
samples.shape #256 x 50 x 50

samples.diagonal(dim2=-2, dim1=-1) # all ones
```

Resolves #1690 and supports multiple batch dimensions. 

TODOs:
- [x] verification that this is the correct sampling algorithm
- [x] unit tests


gpytorch/priors/lkj_prior.py
```diff
@@ -61,6 +61,47 @@ def log_prob(self, X):
...
+            R[..., m, m] = (1.0 - y).pow(0.5)
+
+        sample_as_covariance = R.matmul(R.transpose(-1, -2))
+
+        if not return_covariance:
+            # we seem to need to back-convert the resulting matrix into a correlation matrix
+            inverse_diagonal = torch.diag_embed(
+                torch.diagonal(sample_as_covariance, dim1=-2, dim2=-1).clamp(min=1e-6).pow(-0.5)
+            )
+            return inverse_diagonal.matmul(sample_as_covariance).matmul(inverse_diagonal)
```
Balandat, line 98: No need to do the `diag_embed` here and to a matmul with a lot of zeros. You should be doing sth like the following (the unsqueezes might be wrong but you get the idea). 

```suggestion
            inverse_sqrt_diagonal = torch.diagonal(sample_as_covariance, dim1=-2, dim2=-1).clamp(min=1e-6).pow(-0.5)
            return inverse_diagonal.unsqueeze(-1) * sample_as_covariance * inverse_diagonal.unsqueeze(-2)
```

gpytorch/priors/lkj_prior.py
```diff
@@ -89,6 +130,10 @@ def log_prob(self, X):
         log_diag_sum = torch.diagonal(X, dim1=-2, dim2=-1).log().sum(-1)
         return self.C + (self.eta - 1) * 2 * log_diag_sum
 
+    def rsample(self, sample_shape):
+        res = super()._rsample(sample_shape=sample_shape, return_covariance=False)
+        return res.cholesky()
```
Balandat, line 135: This seems like unnecessary work. IIUC in `_rsample` of `LKJPrior` you're essentially sampling an upper triangular Cholesky factor `R`. Can you just make a `_rsample_cholesky` method that does that (and then call this here), and then have the forming of the covariance/correlation matrix happen in `LKJPrior.rsample`?

gpytorch/priors/lkj_prior.py
```diff
@@ -130,6 +175,22 @@ def log_prob(self, X):
...
+        # create an empty diagonal matrix
+        empty_diag_mat = torch.zeros(
+            *sample_shape,
+            self.correlation_prior.n,
+            self.correlation_prior.n,
+            device=base_correlation.device,
+            dtype=base_correlation.dtype,
+        )
+        # and add to the batch dimension
+        marginal_sds = marginal_sds.unsqueeze(-1) + empty_diag_mat
```
Balandat, line 190: What exactly are you trying to do here? I don't think `marginal_sds` will be diagonal after this. See comment above how to implement a covar-to-corr or corr-to-covar scaling more efficiently by avoiding matmuls.

Balandat (changes requested):

Balandat:
(I haven't checked the sampling algorithm)

wjmaddox:
Yeah, I'm unable to find a good reference for the sampling algorithm, need to find one still.

gpytorch/priors/lkj_prior.py
wjmaddox: I'm trying to embed the marginal sds in as a diagonal. Hopefully this isn't too boneheaded.

gpytorch/priors/lkj_prior.py
wjmaddox: yeah, the issue is that the cholesky returned is the cholesky of the covariance matrix we're sampling, not the correlation matrix corresponding to that covariance. off the top of my head, i dont think there's a way to re-use the cholesky under that type of transformation.

wjmaddox:
Ended up deciding to check marginal means and variances using the formulas given here https://distribution-explorer.github.io/multivariate_continuous/lkj.html. The tolerances are quite loose especially for the variance, but look decent as you yank up the samples.

wjmaddox:
Unit test failure looks unrelated?

gpytorch/priors/lkj_prior.py
```diff
@@ -61,6 +61,58 @@ def log_prob(self, X):
...
+        R[..., 0, 1] = r12
+        R[..., 1, 1] = (1 - r12.pow(2)).pow(0.5)
+
+        for m in range(1, N):
+            alpha = alpha - 0.5
+            y = Beta(m / 2, alpha).rsample(sample_shape)
+
+            # uniform on a hypersphere
+            z = torch.randn(*sample_shape, m, device=self.eta.device, dtype=self.eta.dtype)
+            z = z / z.pow(2.0).sum(-1, keepdim=True).pow(0.5)
```
Balandat, line 86:
```suggestion
            z = z / torch.linalg.norm(z, dim=-1, keepdim=True)
```

gpytorch/priors/lkj_prior.py
```diff
@@ -61,6 +61,58 @@ def log_prob(self, X):
...
+
+            R[..., :m, m] = y.pow(0.5).unsqueeze(-1) * z
+            R[..., m, m] = (1.0 - y).pow(0.5)
+
+        # the sample R is drawn from as the cholesky of a covariance matrix, not a correlation
```
Balandat, line 91:
```suggestion
        # the sample R is drawn from the Cholesky decomposition of a covariance matrix, not a correlation
```

gpytorch/priors/lkj_prior.py
```diff
@@ -61,6 +61,58 @@ def log_prob(self, X):
...
+                inverse_sqrt_diagonal.unsqueeze(-1) * sample_as_covariance * inverse_sqrt_diagonal.unsqueeze(-2)
+            )
+
+            # and then return the cholesky of the correlation matrix
+            if return_cholesky:
+                return correlation_mat.cholesky()
+            else:
+                return correlation_mat
+
+    def rsample(self, sample_shape=torch.Size([])):
```
Balandat, line 113: no need to instantiate empty lists for this
```suggestion
    def rsample(self, sample_shape=torch.Size()):
```

gpytorch/priors/lkj_prior.py
```diff
@@ -61,6 +61,58 @@ def log_prob(self, X):
         log_diag_sum = psd_safe_cholesky(X, upper=True).diagonal(dim1=-2, dim2=-1).log().sum(-1)
         return self.C + (self.eta - 1) * 2 * log_diag_sum
 
+    def _rsample(self, sample_shape=None, return_covariance=False, return_cholesky=False):
```
Balandat, line 64: The different return types here is a bit of an antipattern. Let's factor out the functionality for drawing Cholesky decomps of covariance matrices as a functional helper into sth like `gpytorch/utils/stats.py`? Then we can just call that here. (also something that would be good in a pytorch probability lib...)

gpytorch/priors/lkj_prior.py
```diff
@@ -89,6 +141,9 @@ def log_prob(self, X):
         log_diag_sum = torch.diagonal(X, dim1=-2, dim2=-1).log().sum(-1)
         return self.C + (self.eta - 1) * 2 * log_diag_sum
 
+    def rsample(self, sample_shape=torch.Size([])):
```
Balandat, line 144:
```suggestion
    def rsample(self, sample_shape=torch.Size()):
```

gpytorch/priors/lkj_prior.py
```diff
@@ -107,7 +162,10 @@ def __init__(self, n, eta, sd_prior, validate_args=False):
             raise ValueError("sd_prior must be an instance of Prior")
         if not isinstance(n, int):
             raise ValueError("n must be an integer")
-        if sd_prior.event_shape not in {torch.Size([1]), torch.Size([n])}:
+        # bug-fix event shapes if necessary
```
Balandat, line 165: Is this basically a hack around #1317/#1318?

gpytorch/priors/lkj_prior.py
```diff
@@ -130,6 +188,14 @@ def log_prob(self, X):
         log_prob_sd = self.sd_prior.log_prob(marginal_sd)
         return log_prob_corr + log_prob_sd
 
+    def rsample(self, sample_shape=torch.Size([])):
```
Balandat, line 191:
```suggestion
    def rsample(self, sample_shape=torch.Size()):
```

gpytorch/priors/lkj_prior.py
wjmaddox: No, this is actually a hack around a difference in event shapes between exponential family distributions in `torch.distributions` and the event shape in `SmoothedBoxPrior`. I think it's a bug in torch.distributions.

gpytorch/priors/lkj_prior.py
wjmaddox: I'm not sure I follow here? The cholesky decomp of the covariance matrix is actually what's drawn in the first place, the other types end up doing various transforms of it.

Balandat: Yep indeed - I think that drawing cholesky decomps of the covariance matrix is a generally useful functionality that we should factor out rather than having hidden in this "private" `_rsample` method. Then we can just call that new method here and do the transforms if necessary.

By "antipattern" I just meant that the object `_rsample` returns depends on the input values - typically you'd consider having a different method for these but I can see how having 4 different `_rsample_*_*_` methods (e.g. `_rsample_covar_factor` being a bit much.

wjmaddox: We do already have four separate methods. How about refactoring the second half of that function into the various `rsample` methods by type then?

Balandat: Hmm what do you mean "We do already have four separate methods"? As part of this method? Or as part of the different Prior types? I think logically what `_rsample` does should correspond 1:1 to what the prior represents, as is the case now (e.g. if you have the `LKJCovariancePrior` then `_rsample` needs to return a sampled covar and not a factor...).

wjmaddox: I mean by part of the different prior types.

wjmaddox: Well at least the cholesky factor and the full matrix are themselves possible to reach from various prior types. I should probably code up a `LKJBaseCovariancePrior` actually to return the unscaled covariance matrices however.

gpytorch/priors/lkj_prior.py
wjmaddox: Here's the distinction:

```python
from gpytorch.priors import GammaPrior, SmoothedBoxPrior
prior1 = SmoothedBoxPrior(0.1, 1.0)
prior1.event_shape # torch.Size([1])

prior2 = GammaPrior(3.0, 5.0)
prior2.event_shape # torch.Size([])
```

Balandat: I see. Yeah so I believe all of `torch.distributions` will have an empty event shape if initialized with scalars. So I'm wondering if we should change `SmoothedBoxPrior` to do the same.

gpytorch/priors/lkj_prior.py
```diff
@@ -1,93 +1,72 @@
...
-        if any(s != self.n for s in X.shape[-2:]):
-            raise ValueError("Correlation matrix is not of size n={}".format(self.n.item()))
-        if not _is_valid_correlation_matrix(X):
-            raise ValueError("Input is not a valid correlation matrix")
-        log_diag_sum = psd_safe_cholesky(X, upper=True).diagonal(dim1=-2, dim2=-1).log().sum(-1)
-        return self.C + (self.eta - 1) * 2 * log_diag_sum
+    def rsample(self, sample_shape=torch.Size()):
+        # mocking the sample call here
+        # TODO: determine why torch.distributions.LKJCholesky only implements sample
+        return super().sample(sample_shape=sample_shape)
```
Balandat, line 41: It seems like the proper thing to do here would be to overwrite the `sample` method and then raise a `NotImplementedError` for `rsample`. Or is there anywhere in the code where we're calling `rsample` where we really would be ok with calling `sample`?

test/priors/test_lkj_prior.py
```diff
@@ -46,11 +44,12 @@ def test_lkj_prior_log_prob_cuda(self):
     def test_lkj_prior_batch_log_prob(self, cuda=False):
         device = torch.device("cuda") if cuda else torch.device("cpu")
         prior = LKJPrior(2, torch.tensor([0.5, 1.5], device=device))
+        dist = LKJCholesky(2, torch.tensor([0.5, 1.5], device=device))
 
         S = torch.eye(2, device=device)
-        self.assertTrue(approx_equal(prior.log_prob(S), torch.tensor([-1.86942, -0.483129], device=S.device)))
+        self.assertTrue(approx_equal(prior.log_prob(S), dist.log_prob(S.cholesky())))
```
Balandat, line 50: `.cholesky` will raise a `DeprecationWarning`. Can you use `torch.linalg.cholesky` instead?

gpytorch/priors/lkj_prior.py
wjmaddox: This was really a placeholder for checking into why https://github.com/pytorch/pytorch/pull/48798 uses sample instead of rsample (and also the name of the PR in the first place :) ).

wjmaddox:
Switched all methods over to `...sample` from `...rsample` because I still haven't figured out why `torch.distributions.LKJ..` only implements `sample`.

test/priors/test_lkj_prior.py
```diff
@@ -184,19 +191,31 @@ def test_lkj_covariance_prior_batch_log_prob(self, cuda=False):
...
 
     def test_lkj_covariance_prior_batch_log_prob_cuda(self):
         if torch.cuda.is_available():
             with least_used_cuda_device():
                 self.test_lkj_covariance_prior_batch_log_prob(cuda=True)
 
+    def test_lkj_prior_sample(self):
+        prior = LKJCovariancePrior(2, 0.5, sd_prior=SmoothedBoxPrior(exp(-1), exp(1)))
+        random_samples = prior.sample(torch.Size((6,)))
+        # only need to check that this is a PSD matrix
```
Balandat, line 213: Did you want to check symmetry as well? I think `eigh` will produce (wrong) results even if the matrix is not symmetric

Balandat (commented):
> Switched all methods over to ...sample from ...rsample because I still haven't figured out why torch.distributions.LKJ.. only implements sample.

See https://github.com/pytorch/pytorch/issues/69281

wjmaddox:
Thanks, I'd lost track of that issue, can edit the code to note switching to rsample if that goes in.

test/priors/test_lkj_prior.py
wjmaddox: Everything ought to be symmetric by construction, but probably a good idea.

Balandat: Yup, but the construction happens in your code :)

Balandat (approved):

PR #1746, opened 2021-09-08
Refactor to Arguments to more closely match torch.tensor

wjmaddox:
Refactors lazy_tensor.to's arguments to have both dtype and device arguments. Resolves the current failing error:

```python
import torch
from gpytorch.lazy import KroneckerProductLazyTensor, CatLazyTensor, lazify

cat_lt = CatLazyTensor(
    lazify(torch.randn(50, 50)),
    lazify(torch.randn(5, 50))
)

kplt = KroneckerProductLazyTensor(cat_lt, cat_lt)
kplt.half()
```

which has error message:
```
     84 
     85         # Helpers for _getitem
---> 86         cat_dim_sizes = torch.tensor([t.size(dim) for t in lazy_tensors], device=output_device)
     87         cat_dim_cum_sizes = torch.zeros(len(lazy_tensors) + 1, dtype=torch.long, device=output_device)
     88         torch.cumsum(cat_dim_sizes, dim=-1, out=cat_dim_cum_sizes[1:])

TypeError: tensor(): argument 'device' must be torch.device, not torch.dtype
```

@Balandat  I thought I'd caught this before but my previous PR didn't quite catch it.

### DIFF
--- a/gpytorch/lazy/cat_lazy_tensor.py
+++ b/gpytorch/lazy/cat_lazy_tensor.py
@@ -3,7 +3,7 @@
 import torch
 
 from .. import settings
-from ..utils.broadcasting import _matmul_broadcast_shape, _mul_broadcast_shape
+from ..utils.broadcasting import _matmul_broadcast_shape, _mul_broadcast_shape, _to_helper
 from ..utils.deprecation import bool_compat
 from ..utils.getitem import _noop_index
 from .lazy_tensor import LazyTensor, delazify
@@ -366,22 +366,12 @@ class CatLazyTensor(LazyTensor):
 
     def to(self, *args, **kwargs):
         """
-        Returns a new CatLazyTensor with device as the output_device
+        Returns a new CatLazyTensor with device as the output_device and dtype
+        as the dtype.
         Warning: this does not move the LazyTensors in this CatLazyTensor to
-        device
+        device.
         """
-        dtype = kwargs.pop("dtype", None)
-        device = kwargs.pop("device", None)
-
-        if dtype is None:
-            dtype_list = [x for x in args if type(x) is torch.dtype]
-            if len(dtype_list) > 0:
-                dtype = dtype_list[0]
-
-        if device is None:
-            device_list = [x for x in args if type(x) is torch.device]
-            if len(device_list) > 0:
-                device = device_list[0]
+        device, dtype = _to_helper(*args, **kwargs)
 
         new_kwargs = {**self._kwargs, "output_device": device}
         res = self.__class__(*self._args, **new_kwargs)
--- a/gpytorch/lazy/cat_lazy_tensor.py
+++ b/gpytorch/lazy/cat_lazy_tensor.py
@@ -364,15 +364,24 @@ class CatLazyTensor(LazyTensor):
     def device_count(self):
         return len(set(self.devices))
 
-    def to(self, device_id):
+    def to(self, device=None, dtype=None):
         """
-        returns a new CatLazyTensor with device_id as the output_device
+        returns a new CatLazyTensor with device as the output_device
         Warning: this does not move the LazyTensors in this CatLazyTensor to
-        device_id
+        device
         """
+        if type(device) is torch.dtype and dtype is None:
+            dtype = device
+            device = self.output_device
+
         new_kwargs = dict(self._kwargs)
-        new_kwargs["output_device"] = device_id
-        return self.__class__(*self._args, **new_kwargs)
+        new_kwargs["output_device"] = device
+        res = self.__class__(*self._args, **new_kwargs)
+
+        if dtype is not None:
+            res = res.type(dtype)
+
+        return res
 
     def all_to(self, device_id):
         """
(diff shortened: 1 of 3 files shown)


gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -364,15 +364,24 @@ def devices(self):
...
         """
-        returns a new CatLazyTensor with device_id as the output_device
+        returns a new CatLazyTensor with device as the output_device
         Warning: this does not move the LazyTensors in this CatLazyTensor to
-        device_id
+        device
         """
+        if type(device) is torch.dtype and dtype is None:
+            dtype = device
+            device = self.output_device
```
Balandat, line 375: This is kind of funky. Is this what pytorch does?

gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -364,15 +364,24 @@ def devices(self):
...
+
         new_kwargs = dict(self._kwargs)
```
Balandat, line 377: this should always be a dict, right?

gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -364,15 +364,24 @@ def devices(self):
...
-        new_kwargs["output_device"] = device_id
-        return self.__class__(*self._args, **new_kwargs)
+        new_kwargs["output_device"] = device
```
Balandat, line 378: in which case you could do `new_kwargs = {**self._kwargs, "output_device": device}`

gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -364,15 +364,24 @@ def devices(self):
...
+        res = self.__class__(*self._args, **new_kwargs)
+
+        if dtype is not None:
+            res = res.type(dtype)
```
Balandat, line 382: This seems like it might be wasteful - We could instead add an `output_dtype` kwarg to the constructor and then handle things there?

gpytorch/lazy/cat_lazy_tensor.py
wjmaddox: Not quite, it's left as `*args, **kwargs` from what I can tell here: https://pytorch.org/docs/master/generated/torch.Tensor.to.html#torch.Tensor.to . I could replace with that type of inferrring instead?

gpytorch/lazy/cat_lazy_tensor.py
wjmaddox: I'd assume so, will take a closer look.

gpytorch/lazy/cat_lazy_tensor.py
wjmaddox: Wasteful as in defining too many new objects?

Balandat: Yeah. Probably not a big deal, but ideally we could construct `res` above directly with the right `dtype` rather than calling `.type` on it.

gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -364,15 +364,32 @@ def devices(self):
     def device_count(self):
         return len(set(self.devices))
 
-    def to(self, device_id):
+    def to(self, *args, **kwargs):
         """
-        returns a new CatLazyTensor with device_id as the output_device
+        Returns a new CatLazyTensor with device as the output_device
```
Balandat, line 369: Let's update this and include dtype? And maybe be a bit more clear about how the `*args / **kwargs` are interpreted here?

Balandat: Ah I see you did this below - just copy paste here or get rid of the docstring altogether?

gpytorch/lazy/lazy_tensor.py
```diff
@@ -1859,25 +1859,41 @@ def symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional["LazyTens
             pass
         return self._symeig(eigenvectors=eigenvectors)
 
-    def to(self, device_id):
+    def to(self, *args, **kwargs):
         """
-        A device-agnostic method of moving the lazy_tensor to the specified device.
+        A device-agnostic method of moving the lazy_tensor to the specified device or dtype.
+        Note that we do NOT support non_blocking or other `torch.to` options other than
+        device and dtype.
```
Balandat, line 1866: Should we error out if those are passed? Or log a warning?

gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -364,15 +364,32 @@ def devices(self):
...
+
+        if dtype is None:
+            dtype_list = [x for x in args if type(x) is torch.dtype]
+            if len(dtype_list) > 0:
+                dtype = dtype_list[0]
+
+        if device is None:
+            device_list = [x for x in args if type(x) is torch.device]
+            if len(device_list) > 0:
+                device = device_list[0]
```
Balandat, line 384: feels like you could make a little type helper for this and then call it here and below to dedup code

Balandat (approved):

gpytorch/lazy/lazy_tensor.py
wjmaddox: They're all silent options?

wjmaddox: Mention that it ignores currently. Might be a good thing for linear operator to ultimately support though.

PR #1752, opened 2021-09-14
add generate_inv_roots option to cat_rows

sdaulton:
This allows for adding rows to the covariance and updating the root decomposition without updating the root inv decomposition, which can be relatively costly due to the required QR decomposition


wjmaddox:
lgtm, is there a reason why we wouldn't want to automatically generate these?

also, it seems logically weird to be able to not generate the root but to generate the inverse root, so add a warning for that case?

wjmaddox (approved):

sdaulton:
> is there a reason why we wouldn't want to automatically generate these?

In the case that we just want to append new rows to a covariance matrix and then sample from the augmented covariance matrix, we only need the root decomposition (not the root inv decomposition). This is useful in BO for optimizing acquisition functions such as NEI and NEHVI that require repeated (potentially thousands of times) evaluating and sampling from the posterior over f(X_baseline, X) for different X.

> also, it seems logically weird to be able to not generate the root but to generate the inverse root, so add a warning for that case?

Good point. I'll add this

Issue #1765, opened 2021-09-21
[Question] Reduce memory requirement in batch variational GPs with different evaluation times


flcello:
Hi,

I have two questions concerning the reduction of memory usage in variational GPs, specifically Cox processes:

**First question:** I'm considering a Cox process like in [this example](https://docs.gpytorch.ai/en/stable/examples/07_Pyro_Integration/Cox_Process_Example.html), running on a GPU. I noticed that calling the model does not use much memory, but calling the guide does. The memory usage of calling the guide grows with the number of arrivals or quadrature points passed to it. With approximately 30 000 arrivals my GPU memory is used up. I know that I could do subsampling, but is there another way to reduce memory consumption? I assume that the problem is not the data itself because it's just a shape [30 000] tensor that should easily fit in memory. I mean something like using a different variational distribution or so.

**Second (and more important) question:** Now I have different datasets and want to train one Cox process per dataset. However, this has to be done in a single Pyro model because I have additional latent variables that are global in the sense that they are the same for every dataset. To accomplish this I tried to use the `batch_shape` argument in the variational strategy, distribution and mean / covariance modules, as described in the [docs](https://docs.gpytorch.ai/en/stable/examples/04_Variational_and_Approximate_GPs/SVGP_Multitask_GP_Regression.html) (but without the `IndependentMultitaskVariationalStrategy` wrapper). The forward method then takes a tensor of evaluation times and samples from all batch GPs at these times. But in the Cox process example, having different datasets means that the different GPs also need to be evaluated at different arrival times. My solution so far was to simply concatenate the arrival times from all datasets, evaluate each batch GP at all times, and disentangle the values afterwards. However, I noticed that this drastically limits the number of datasets I can use: With my solution I have a total of `(batch_size * n_arrivals_all_datasets_concatenated)` GP evaluations, resulting in very high memory usage. It would be desirable to be able to evaluate the batch GPs at different arrival times, resulting in only `(n_arrivals_dataset_1 + n_arrivals_dataset_2 + ... + n_arrivals_batch_size)` evaluations and therefore also large memory savings. Is there a way to do this? If not, are you planning to implement something like this in the future?

I would be very happy to get this solved, because I enjoy using GPyTorch a lot!

wjmaddox:
[...]

Issue #1743, opened 2021-09-03
[Question] Implementing multi-output multi-task approximate GP


fleskovar:
I am looking into implementing a model that produces multiple correlated output for multiple tasks (multi-task-multi-output - MTMO). For this type of model, I assume that the input tensor has the shape n x d+1 (d inputs plus the additional task index) while the output tensor has the shape n x o (where o is the number of correlated outputs). Additionally,  all outputs are observed simultaneously but not all tasks are. The training data for this mode would look like this:
```python
train_x = torch.cat([torch.rand(n_training, 1), (torch.rand(n_training,1)>0.5).float()], dim=1)
train_y = torch.stack([
    torch.sin(5.5 * train_x[:,0])* train_x[:,0]**2 * train_x[:,-1] + torch.cos(5.5 * train_x[:,0])* train_x[:,0]**2  * (1-train_x[:,-1]),
    torch.cos(5.5 * train_x[:,0])* train_x[:,0]*2 * train_x[:,-1] + torch.sin(5.5 * train_x[:,0])* train_x[:,0]*2 * (1-train_x[:,-1])
], axis = -1)
```
For the exact GP case, the model looks like this:

```python
class MultiOutputMultiTaskGP(ExactGP,):

    def __init__(
        self,
        train_X: Tensor,
        train_Y: Tensor,
        likelihood: MultitaskGaussianLikelihood = None,
        rank = None,
        lik_rank = None,
    ) -> None:

        num_tasks = train_Y.shape[-1]
        batch_shape, ard_num_dims = train_X.shape[:-2], train_X.shape[-1]
        
        if lik_rank is None:
            lik_rank = rank
                       
        self._validate_tensor_args(X=train_X, Y=train_Y)

        if likelihood is None:
            likelihood = MultitaskGaussianLikelihood(
                num_tasks=num_tasks, 
                rank=lik_rank if lik_rank is not None else 0,
            )

        super(MultiOutputMultiTaskGP, self).__init__(train_X, train_Y, likelihood)
        self._rank = rank if rank is not None else num_tasks

        self.mean_module = MultitaskMean(
            ConstantMean(), num_tasks=num_tasks
        )
        
        self.data_kernel = MaternKernel()
        self.task_kernel = IndexKernel(num_tasks=len(torch.unique(train_X[..., -1])))
        self.output_kernel = IndexKernel(num_tasks=num_tasks)
        
        self.to(train_X)
        
[...]

Issue #1433, opened 2021-01-15
[Docs] Example with IndexKernel in variational setting


srn284:
I wish to conduct multitask variational classification with IndexKernel or perhaps a custom variant of it: a covariance factor for inputs for same task and a covariance fac for those coming from different tasks. It'd be great to use this kind of kernel for a way to learn information generally - especially when data is not seen and coming to other tasks. Lastly, the 

Anyways, I am not sure this is possible with the current IndexKernel but please any help would help me a lot! Also, all tasks don't get all input points, so can we can do some masking on the MLL after (re `https://github.com/cornellius-gp/gpytorch/issues/1366`) or should it be dealt with within kernel construction?

What would be the best solution to go about this? I'm a bit lost and I've tried a couple things and none of it work; not sure if this requires changing IndexKernel. 

If it helps, I'm trying to recreate the model here: [https://drive.google.com/file/d/1-_YTW0dmtGv8A3NnS75QWEHXpBKrw30e/view?usp=sharing](url) and full, original paper here: [http://proceedings.mlr.press/v80/alaa18a/alaa18a-supp.pdf](url). Thank you very much for your time and efforts.

gpleiss:
Hmm... I think that using the index kernel in conjunction with SVGP is a bit challenging. At the moment, SVGP models are only able to handle a single input to the forward function.

The best solution is probably to follow the [multi-output SVGP example](https://docs.gpytorch.ai/en/stable/examples/04_Variational_and_Approximate_GPs/SVGP_Multitask_GP_Regression.html) - and then write a custom likelihood that selects the appropriate task output for each `x` value.

srn284:
Thanks for the response @gpleiss. Really appreciate it despite your busy schedule.

Indeed, I have made a custom masked MLL and ELBO to deal with the input x from different tasks. And it works great! 

Additionally, I wanted to set up index based covariance as in the paper shared above. Do you have any leads on how to _not_ use IndexKernel and _not_ pass in indices (that's difficult for SVGP indeed, as you have stated) but develop kernels for each index/task? 

If that can be constructed, since I mask out the loss at the end, I probably would avoid backprop for those tasks/indices not relevant for a given x?

Thank you very much.

gpleiss:
> Indeed, I have made a custom masked MLL and ELBO to deal with the input x from different tasks.

If you think this is something that would be useful to other users, we'd happily accept a PR.

[...]

PR #1769, opened 2021-09-30
LMC multitask-SVGP models can output a single task per input.

### LINKED ISSUE #1285: [Question] Multi-output SVGP with irregular inputs?
Is there a way of training a multi-output variational GP model with irregular inputs, as in the  [Hadamard Multitask GP Regression](https://docs.gpytorch.ai/en/v1.2.0/examples/03_Multitask_Exact_GPs/Hadamard_Multitask_GP_Regression.html) example notebook?

Thanks,
Jimmy

gpleiss:
If one defines a ApproximateGP model with a LMCVariationalStrategy,
there are now two different options for return types:

1. Calling `model(x)` will return a `... x N x num_tasks`
MultitaskMultivariateNormal distribution
1. Calling `model(x, task_indices=i)` will return a `... x N`
MultivariateNormal distribution, where `i` corresponds to the selected
task index for each input.

[Closes #1285, #1433]
[Addresses #1743, #1765]


gpleiss:
TODO:

- [ ] Update docs
- [ ] Update LMC tutorial

gpytorch/variational/lmc_variational_strategy.py
```diff
@@ -120,28 +144,84 @@ def variational_params_initialized(self):
...
+        r"""
+        Computes the variational (or prior) distribution
+        :math:`q( \mathbf f \mid \mathbf X)` (or :math:`p( \mathbf f \mid \mathbf X)`).
+        There are two modes:
+
+        1.  Compute **all tasks** for all inputs.
+            If this is the case, the :attr:`task_indices` attribute should be None.
+            The return type will be a (... x N x num_tasks)
+            :class:`~gpytorch.distributions.MultitaskMultivariateNormal`.
+        2.  Compute **one tasks** per inputs.
```
jacobrgardner, line 157:
```suggestion
        2.  Compute **one task** per inputs.
```

gpytorch/variational/lmc_variational_strategy.py
```diff
@@ -2,12 +2,34 @@
...
+    """
+    Given a list of indices for ... x N datapoints,
+      select the row from lmc_coefficient that corresponds to each datapoint
+
+    lmc_coefficients: torch.Tensor ... x num_latents x ... x num_tasks
+    indices: torch.Tesnor ... x N
+    """
+    batch_shape = _mul_broadcast_shape(lmc_coefficients.shape[:-1], indices.shape[:-1])
+
+    # We will use the left_t_interp helper to do the indexing
```
jacobrgardner, line 24:
```suggestion
    # We will use the left_interp helper to do the indexing
```

jacobrgardner (approved):
Looks good to me, and I actually need this PR for something :-)

flcello:
As far as I understand it there is still some overhead in the standard `VariationalStrategy` because the conditional is still computed for all tasks and the relevant task is only selected afterwards in `IndependentMultitaskVariationalStrategy`. Is that right? Would it theoretically be possible to change that?

PR #1802, opened 2021-10-27
Added Type hints and exceptions in kernels

patel-zeel:
As per the [discussion](https://github.com/cornellius-gp/gpytorch/discussions/1779#discussioncomment-1526068):

* Added and standardized type hints to all kernel constructors.
* Raise ```TypeError``` if a hyperparameter prior is not of type ```gpytorch.priors.Prior```.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,8 @@
+docutils==0.17.1
 nbformat
 ipython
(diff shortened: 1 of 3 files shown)


patel-zeel:
Thank you, @gpleiss for merging the PR.

PR #1813, opened 2021-11-08
Don't evaluate kernel in `LazyEvaluatedKernelTensor._unsqueeze_batch`

valtron:
Related to pytorch/botorch#971.


gpleiss (approved):
I think this should be fine - especially if all the unit tests pass.

Issue #1827, opened 2021-11-16
[Bug] GetItem in BatchRepeated LazyEvaluatedKernelTensor Fails


wjmaddox:
# 🐛 Bug

This is seemingly directly caused by #1813 and causes downstream botorch errors (especially https://github.com/pytorch/botorch/issues/980). I've tracked it down to the

```python
from gpytorch.kernels import MaternKernel
from gpytorch.lazy import KroneckerProductLazyTensor, BatchRepeatLazyTensor

kernel = MaternKernel()
BatchRepeatLazyTensor(kernel(torch.randn(30)), torch.Size((10,))).diag()
```

produces an indexing error.

## To reproduce

** Stack trace/error message **
```
IndexError                                Traceback (most recent call last)
~/Documents/GitHub/gpytorch/gpytorch/lazy/lazy_evaluated_kernel_tensor.py in _getitem(self, row_index, col_index, *batch_indices)
    111         try:
--> 112             x2 = x2[(*batch_indices, col_index, dim_index)]
    113         # We're going to handle multi-batch indexing with a try-catch loop

IndexError: too many indices for tensor of dimension 2

During handling of the above exception, another exception occurred:

RuntimeError                              Traceback (most recent call last)
/var/folders/ms/lbkrq70x7lnbnztclff1ln6r0000gn/T/ipykernel_8975/568402408.py in <module>
      6 
      7 kernel = MaternKernel()
----> 8 BatchRepeatLazyTensor(kernel(torch.randn(30)), torch.Size((10,))).diag()

~/Documents/GitHub/gpytorch/gpytorch/lazy/lazy_tensor.py in diag(self)
   1065 
   1066         row_col_iter = torch.arange(0, self.matrix_shape[-1], dtype=torch.long, device=self.device)
-> 1067         return self[..., row_col_iter, row_col_iter]
   1068 
   1069     def dim(self):

~/Documents/GitHub/gpytorch/gpytorch/lazy/lazy_tensor.py in __getitem__(self, index)
   2144                 self, (*batch_indices, row_index, col_index)
   2145             )
-> 2146             res = self._get_indices(row_index, col_index, *batch_indices)
   2147         else:
   2148             res = self._getitem(row_index, col_index, *batch_indices)

~/Documents/GitHub/gpytorch/gpytorch/lazy/batch_repeat_lazy_tensor.py in _get_indices(self, row_index, col_index, *batch_indices)
     81 
     82         # Now call the sub _get_indices method
[...]

PR #1828, opened 2021-11-17
Fix `LazyEvaluatedKernelTensor._unsqueeze_batch`

valtron:
This is in response to #1813, which caused some bugs as the indexing code in lazy tensors did not support `None` indexing.

Addresses:
- #1827
- pytorch/botorch#980
- pytorch/botorch#976

Tests are passing locally* for GPyTorch (with this patch) and BoTorch (with patch pytorch/botorch#976).

*Except for some CUDA tests that failed before, and are unrelated

### DIFF
--- a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
@@ -68,8 +68,16 @@ class LazyEvaluatedKernelTensor(LazyTensor):
             # Now we know that x1 and x2 are slices
             # Let's make sure that the slice dimensions perfectly correspond with the number of
             # outputs per input that we have
-            row_start, row_end, row_step = row_index.start, row_index.stop, row_index.step
-            col_start, col_end, col_step = col_index.start, col_index.stop, col_index.step
+            row_start, row_end, row_step = (
+                row_index.start,
+                row_index.stop,
+                row_index.step,
+            )
+            col_start, col_end, col_step = (
+                col_index.start,
+                col_index.stop,
+                col_index.step,
+            )
             if row_step is not None or col_step is not None:
                 return self.evaluate_kernel()._getitem(row_index, col_index, *batch_indices)
             if (
@@ -127,7 +135,7 @@ class LazyEvaluatedKernelTensor(LazyTensor):
             new_kernel = self.kernel.__getitem__(batch_indices)
 
         # Now construct a kernel with those indices
-        return self.__class__(x1, x2, kernel=new_kernel, last_dim_is_batch=self.last_dim_is_batch, **self.params)
+        return self.__class__(x1, x2, kernel=new_kernel, last_dim_is_batch=self.last_dim_is_batch, **self.params,)
 
     def _matmul(self, rhs):
         # This _matmul is defined computes the kernel in chunks
@@ -148,7 +156,7 @@ class LazyEvaluatedKernelTensor(LazyTensor):
             res = []
             for sub_x1 in sub_x1s:
                 sub_kernel_matrix = lazify(
-                    self.kernel(sub_x1, x2, diag=False, last_dim_is_batch=self.last_dim_is_batch, **self.params)
+                    self.kernel(sub_x1, x2, diag=False, last_dim_is_batch=self.last_dim_is_batch, **self.params,)
                 )
                 res.append(sub_kernel_matrix._matmul(rhs))
 
@@ -177,7 +185,7 @@ class LazyEvaluatedKernelTensor(LazyTensor):
             sub_x1.requires_grad_(True)
             with torch.enable_grad(), settings.lazily_evaluate_kernels(False):
                 sub_kernel_matrix = lazify(
-                    self.kernel(sub_x1, x2, diag=False, last_dim_is_batch=self.last_dim_is_batch, **self.params)
+                    self.kernel(sub_x1, x2, diag=False, last_dim_is_batch=self.last_dim_is_batch, **self.params,)
                 )
             sub_grad_outputs = tuple(sub_kernel_matrix._quad_form_derivative(sub_left_vecs, right_vecs))
             sub_kernel_outputs = tuple(sub_kernel_matrix.representation())
@@ -230,7 +238,7 @@ class LazyEvaluatedKernelTensor(LazyTensor):
 
     def _transpose_nonbatch(self):
         return self.__class__(
-            self.x2, self.x1, kernel=self.kernel, last_dim_is_batch=self.last_dim_is_batch, **self.params
+            self.x2, self.x1, kernel=self.kernel, last_dim_is_batch=self.last_dim_is_batch, **self.params,
         )
 
     def add_jitter(self, jitter_val=1e-3):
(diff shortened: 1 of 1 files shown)


gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -185,7 +185,18 @@ def _get_indices(self, row_index, col_index, *batch_indices):
 
     def _getitem(self, row_index, col_index, *batch_indices):
         indices = [*batch_indices, row_index, col_index]
-        cat_dim_indices = indices[self.cat_dim]
+        cat_dim = self.cat_dim % len(self.shape)
```
Balandat, line 188: Is this something we'd want to do at a higher level (e.g. upon init)?

gpytorch/lazy/cat_lazy_tensor.py
```diff
@@ -185,7 +185,18 @@ def _get_indices(self, row_index, col_index, *batch_indices):
...
+            if idx is None:
+                cat_dim += 1
+                continue
+            if i == cat_dim:
+                break
+        cat_dim_indices = indices[cat_dim]
+        out_cat_dim = cat_dim - sum(isinstance(idx, int) for idx in indices[:cat_dim])
+        num_tensor_indices = sum(torch.is_tensor(idx) for idx in indices)
+        if num_tensor_indices > 0:
+            out_cat_dim -= num_tensor_indices - 1
```
Balandat, line 199: I *think* I understand what you're doing here, but it's not at all easy to follow. Could you please add some comments here to explain what's going on?

Balandat (commented):
Awesome, thanks for submitting this fix. I have some questions as to exactly what you're doing in the code (see inline). I also think that @gpleiss of @jacobrgardner better take a look at this as well since this again goes into the indexing weeds.

gpytorch/lazy/cat_lazy_tensor.py
valtron: `cat_dim` is actually normalized to a negative index in the init, and here I'm making it positive again for convenience.

When I originally looked at this code, I thought the reason `cat_dim` is stored as a negative index is to avoid having to do adjustments to it when when indexing through the cat... but on second thought, adjustments would still be needed, e.g.:

```
M = zeros(size = (2, 3, 4, 5, 6))
C = cat([M, M], dim = -4)
C[:,:,0,:,:] == cat([M[:,:,0,:,:], M[:,:,0,:,:]], dim = -3) # -4 -> -3
```

(I haven't had time to take a deeper look since making this PR.)

gpytorch/lazy/cat_lazy_tensor.py
valtron: The first part (189 - 194) is adjusting `cat_dim` to take into account unsqueezed dimensions. Line 196 is adjusting the `cat_dim` of the returned lazy cat tensor to take into account dimensions that were removed because of `int` indexing. And lines 197 - 199 adjust it for tensor indexes [1]: if `indices` contains `n > 0` tensor indices, those dimensions are removed and a dimension is added to the beginning [2].

... except turns out indexing is more complicated than I originally thought.
[1] Firstly, here I'm not handling the case when the shape of the tensor index has len() > 1. The number of dimensions added is actually the number of dimensions of (the broadcast of) all the tensor indices. 😱
[2] Apparently (this is really weird), if _all_ tensor indices are in adjacent dimensions (`int` indexing is a type of tensor indexing, just with shape `()`), the new dimensions are added to in place of the removed dimensions. If they're not adjacent, the new dimensions are added to the beginning: 😱😱
```python
import numpy as np
A = np.zeros((4,5,6,7,8))
idxs = [[0,0],[0,0],[0,0]]
print(A.shape)
# (4, 5, 6, 7, 8)
print(A[:,idxs,idxs,idxs,:].shape)
# (4, 3, 2, 8)
print(A[:,idxs,:,idxs,idxs].shape)
# (3, 2, 4, 6)
```

So, this will take more work to work properly. I'll leave adding comments until the end.

gpytorch/lazy/lazy_tensor.py
```diff
@@ -2092,6 +2092,9 @@ def __getitem__(self, index):
         index = tuple(torch.tensor(idx) if isinstance(idx, list) else idx for idx in index)
         index = tuple(idx.item() if torch.is_tensor(idx) and not len(idx.shape) else idx for idx in index)
 
+        # Increase ndimension by the number of dimensions unsqueezed via `None` indexing
+        ndimension += sum(idx is None for idx in index)
+
```
gpleiss, line 2097: I think here it would make sense to call the `self._unsqueeze_batch` for each of the `None` indexes. I think this should be all there is to this PR?

valtron: Because of https://github.com/cornellius-gp/gpytorch/pull/1813/files that would be infinite recursion. But you are right, in that rather than trying to support `None` indexing in lazy tensors, it might be easier (with smaller chance of breaking things) to implement a non-evaluating `LazyEvaluatedKernelTensor._unsqueeze_batch` directly. (The reason I punted it off to `__getitem__` was because I didn't know `None` indexing wasn't supported.) I tried something basic [1] just on the off chance it worked, it doesn't though. Kind of stuck right now...

[1] 
```python
indexes = (slice(None),) * dim + (None,)
return self.__class__(self.x2, self.x1, kernel=self.kernel[indexes], ...)
```

Balandat: @gpleiss, @wjmaddox any further thoughts here? What's currently on master still breaks the botorch HighOrderGP tests, which I'd like to fix...

valtron:
@Balandat @gpleiss Hey, I forgot about this for a few days. This new way of doing `_unsqueeze_batch` seems to work.

Balandat (commented):
This lgtm, also it fixes the issues on the botorch end. If you can just fix the formatting issue with black I think this is good to go.

Balandat:
The difficulties with black here are probably b/c the pre-commit hooks are using a pretty old version and what you have locally is different. Let me see if I can update this in another PR.

valtron:
I tried using 19.10b0 as well, no luck -_-

Balandat:
I just put up #1842 that changes a whole swath of things and upgraded versions, with that the 21.11b1 formatting should work.

gpleiss (approved):

gpleiss:
LGTM! Let me see if I can fix the listing issue

valtron:
Thanks @gpleiss! I'm curious, what did you do? For me `black` still reformats it to what github's CI rejects.

```
~/Desktop/a/gpytorch ((HEAD detached at origin/patch-1))
$ py -m black --version
python -m black, version 19.10b0

~/Desktop/a/gpytorch ((HEAD detached at origin/patch-1))
$ py -m black -l 120 --target-version=py36 gpytorch/lazy/lazy_evaluated_kernel_tensor.py
reformatted gpytorch\lazy\lazy_evaluated_kernel_tensor.py
All done! ✨ 🍰 ✨
1 file reformatted.

~/Desktop/a/gpytorch ((HEAD detached at origin/patch-1))
$ git diff
diff --git a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
index 016edda9..5c0e3f4c 100644
--- a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
@@ -135,7 +135,13 @@ class LazyEvaluatedKernelTensor(LazyTensor):
             new_kernel = self.kernel.__getitem__(batch_indices)

         # Now construct a kernel with those indices
-        return self.__class__(x1, x2, kernel=new_kernel, last_dim_is_batch=self.last_dim_is_batch, **self.params,)
+        return self.__class__(
+            x1,
+            x2,
+            kernel=new_kernel,
+            last_dim_is_batch=self.last_dim_is_batch,
+            **self.params,
+        )
(etc.)
```

Balandat:
I think this was merged failing, I fixed it as part of #1842

PR #1836, opened 2021-11-22
Implement variance reduction in SLQ logdet backward pass.

gpleiss:
Based on ["Reducing the Variance of Gaussian Process Hyperparameter Optimization with Preconditioning"](https://arxiv.org/pdf/2107.00243.pdf) by Wenger et al., 2021.

When using iterative methods (i.e. CG/SLQ) to compute the log determinant, the forward pass currently computes:
`logdet K \approx logdet P + SLQ( P^{-1/2} K P^{-1/2} )`,
where `P` is a preconditioner, and `SLQ` is a stochastic estimate of the log determinant. If the preconditioner is a good approximation of K, then this forward pass can be seen as a form of variance reduction.

In this PR, we apply this same variance reduction strategy to the backward pass. We compute the backward pass as:
`d logdet(K)/dtheta \approx d logdet(P)/dtheta + d SLQ/dtheta`

TODOs:
- [x] Implement pivoted cholesky as a torch.autograd.Function, so that we can compute backward passes through it.
- [x] Redo `inv_quad_logdet` function to apply variance reduction in the forward and backward passes.


gpleiss:
@jacobrgardner @JonathanWenger ready for review

gpleiss:
(actually ready for review now. I just fixed broken tests.)

gpytorch/functions/_inv_quad_logdet.py
```diff
@@ -0,0 +1,225 @@
...
+        # Get info about matrix
+        ctx.dtype = lazy_tsr.dtype
+        ctx.device = lazy_tsr.device
+        ctx.matrix_shape = lazy_tsr.matrix_shape
+        ctx.batch_shape = lazy_tsr.batch_shape
+
+        # Probe vectors
+        if probe_vectors is None or probe_vector_norms is None:
+            num_random_probes = settings.num_trace_samples.value()
+            if settings.deterministic_probes.on():
```
jacobrgardner, line 79: Should we deprecate deterministic_probes? Variance reduction more properly solved the problem this was originally trying to, and as far as I can tell this isn't used anywhere.

gpytorch/lazy/identity_lazy_tensor.py
```diff
@@ -0,0 +1,188 @@
...
+from ..utils.memoize import cached
+from .diag_lazy_tensor import ConstantDiagLazyTensor
+from .lazy_tensor import LazyTensor
+from .triangular_lazy_tensor import TriangularLazyTensor
+
+
+class IdentityLazyTensor(ConstantDiagLazyTensor):
+    def __init__(self, diag_shape, batch_shape=torch.Size([]), dtype=None, device=None):
+        """
+        Diagonal lazy tensor. Supports arbitrary batch sizes.
```
jacobrgardner, line 19:
```suggestion
        Identity matrix lazy tensor. Supports arbitrary batch sizes.
```

gpytorch/lazy/identity_lazy_tensor.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Args:
+            :attr:`diag` (Tensor):
+                A `b1 x ... x bk x n` Tensor, representing a `b1 x ... x bk`-sized batch
+                of `n x n` diagonal matrices
```
jacobrgardner, line 24:
```suggestion
                of `n x n` identity matrices
```

gpytorch/lazy/identity_lazy_tensor.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        one = torch.tensor(1.0, dtype=dtype, device=device)
+        super(TriangularLazyTensor, self).__init__(
```
jacobrgardner, line 27:
```suggestion
        super().__init__(
```

gpytorch/lazy/identity_lazy_tensor.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def _transpose_nonbatch(self):
+        return self
+
+    def abs(self):
+        return self
+
+    def inverse(self):
+        return self
+
```
jacobrgardner, line 126: exp could return self and log could return a ZeroLazyTensor I guess?

gpytorch/lazy/added_diag_lazy_tensor.py
```diff
@@ -88,7 +88,7 @@ def _preconditioner(self):
             return self.preconditioner_override(self)
```
jacobrgardner, line 88: These changes are probably breaking for anyone who was using preconditioner_override before. @wjmaddox

wjmaddox: I don't think I have anything in the works that's actually using the preconditioner_override. But, is this necessary?

jacobrgardner:
@gpleiss something seems off about how this is computing the preconditioner log determinant now. We're still computing it efficiently using the QR decomposition in the `init_cache` methods on AddedDiagLT, and then it looks like we discard that efficient computation and call `logdet` on `precond_lt` which could be as bad as calling Cholesky on the full n x n preconditioner matrix, right? PsdSumLazyTensor doesn't override logdet? 

Entirely possible I just missed the relevant code here...

jacobrgardner:
Ideally, this would be the logdet value we'd return in the forward pass:
https://github.com/cornellius-gp/gpytorch/blob/70b79c2d6beca0762e18929102cdee580089e876/gpytorch/lazy/added_diag_lazy_tensor.py#L147

gpytorch/functions/_inv_quad_logdet.py
gpleiss: Done. The deprecation warning links to this PR.

gpytorch/lazy/identity_lazy_tensor.py
gpleiss: Done.

gpytorch/lazy/added_diag_lazy_tensor.py
gpleiss: I reverted back to the original preconditioner return values (preconditioner_function, precond_lt, logdet_p) to address @jacobrgardner 's other comments. Does this resolve this issue as well?

jacobrgardner: Yes, preconditioner_override should work fine again with the revert

jacobrgardner (approved):
Looks solid to me now 👍

gpleiss:
I just profiled this PR on the KeOps example notebook - just to double check. It is just as fast as what is on is on master.

PR #1842, opened 2021-11-30
Update pre-commit config

Balandat:
This started off with a benign black version upgrade but ended up upgrading pretty much all of the pre-commits.


.github/workflows/linting.yml
```diff
@@ -21,12 +21,12 @@ jobs:
...
-        pip install flake8==3.7.9 flake8-print==3.1.4 pre-commit
+        pip install flake8==4.0.1 flake8-print==4.0.0 pre-commit
         pre-commit install
         pre-commit run seed-isort-config || true
     - name: Run linting
       run: |
         flake8
     - name: Run pre-commit checks
       run: |
-        SKIP=flake8 pre-commit run --files test/**/*.py gpytorch/**/*.py
```
gpleiss, line 32: We can keep the `SKIP=flake8` flag in because we run flake8 above. I think the reason we run `flake8` separately is because the flake8 command in `pre-commit` does not include `flake8-print`, and so it does not catch print statements.

gpleiss (approved):
Very picky, but could we lock down the flake8 and flake8-print versions in setup.py? https://github.com/cornellius-gp/gpytorch/blob/master/setup.py#L65

gpleiss (approved):

PR #1850, opened 2021-12-03
Remove `_psd_safe_cholesky` that uses `torch.linalg.cholesky`

valtron:
`torch>=1.9` since https://github.com/cornellius-gp/gpytorch/commit/452443a000c6e043c11618c49479a8bb37af3c80, so https://github.com/cornellius-gp/gpytorch/blob/7289e7d43fc74f8e8dcfc3325fb2871ef5e828a0/gpytorch/utils/cholesky.py#L51-L90 can be removed. `CHOLESKY_METHOD` can also be removed, since (AFAIK) it's not part of the public API (https://docs.gpytorch.ai/en/stable/search.html?q=CHOLESKY_METHOD).

### DIFF
--- a/gpytorch/utils/cholesky.py
+++ b/gpytorch/utils/cholesky.py
@@ -23,9 +23,7 @@ def _psd_safe_cholesky(A, out=None, jitter=None, max_tries=3):
 
     isnan = torch.isnan(A)
     if isnan.any():
-        raise NanError(
-            f"cholesky_cpu: {isnan.sum().item()} of {A.numel()} elements of the {A.shape} tensor are NaN."
-        )
+        raise NanError(f"cholesky_cpu: {isnan.sum().item()} of {A.numel()} elements of the {A.shape} tensor are NaN.")
 
     if jitter is None:
         jitter = settings.cholesky_jitter.value(A.dtype)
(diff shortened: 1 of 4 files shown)


gpytorch/utils/cholesky.py
```diff
@@ -8,102 +8,57 @@
...
+                    The tensor to compute the Cholesky decomposition of
+            :attr:`upper` (bool, optional):
+                    See torch.cholesky
+            :attr:`out` (Tensor, optional):
+                    See torch.cholesky
+            :attr:`jitter` (float, optional):
+                    The jitter to add to the diagonal of A in case A is only p.s.d. If omitted,
+                    uses settings.cholesky_jitter.value()
+            :attr:`max_tries` (int, optional):
+                    Number of attempts (with successively increasing jitter) to make before raising an error.
```
Balandat, line 61: what happened here?

Balandat (approved):
This lgtm, thanks for cleaning this up.  Only comment is about the docstring indents.

gpytorch/utils/cholesky.py
valtron: Just `black`.

Balandat: This doesn't seem like a change that black should suggest? Does black complain when this is undone?

valtron: OK, I see what's going on now, and this probably explains the weird problems I had with the CI in the other PR.

Locally I have filters on smudge/clean to convert spaces to tabs. `black` saw these tabs and replaced with 8 spaces. Otherwise, `black` leaves leading spaces in docstrings unchanged.

Fixed.

gpleiss (approved):

Issue #1854, opened 2021-12-07
[Bug] remove your test folder from pkg


ngam:
# 🐛 Bug

## To reproduce

see this issue for reference: https://github.com/conda-forge/botorch-feedstock/issues/9

** Code snippet to reproduce **
```
conda install botorch gpytorch
```

** Stack trace/error message **
```
// Paste the bad output here!
ClobberWarning: This transaction has incompatible packages due to a shared path.
  packages: conda-forge/noarch::gpytorch-1.5.1-pyhd8ed1ab_0, conda-forge/noarch::botorch-0.2.2-py_0
  path: 'lib/python3.9/site-packages/test/optim/__pycache__/__init__.cpython-39.pyc'

```

## System information

**Please complete the following information:**
- 
- 
- 
- 

## Additional context

You can add `recursive-exclude test *` or something to your `MANIFEST.in`. As a workaround, we have simply force-deleted the tests folder on the conda-forge's botorch side for now, pending a fix from you.

xref: https://github.com/pytorch/botorch/issues/1019

jacobrgardner:
Fixed by #1855

PR #1855, opened 2021-12-08
Do not include tests in package

Balandat:
Addresses #1854

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,7 @@ setup(
     },
     license="MIT",
     classifiers=["Development Status :: 4 - Beta", "Programming Language :: Python :: 3"],
-    packages=find_packages(),
+    packages=find_packages(exclude=["test", "test.*"]),
     python_requires=">=3.6",
     install_requires=install_requires,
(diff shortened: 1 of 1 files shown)


jacobrgardner (approved):

hadim:
This fix does not seem to be sufficient to remove the test folder from the pypi source package. In consequence, it is also within the conda package at https://anaconda.org/conda-forge/gpytorch/1.6.0/download/noarch/gpytorch-1.6.0-pyhd8ed1ab_0.tar.bz2

The best way to check is to build the pypi package locally and look into it whether the `test` folder is still here. You could try using the `MANIFEST.in` file (https://packaging.python.org/en/latest/guides/using-manifest-in/).

hadim:
Well I just realized that this PR is 6 days old and the 1.6.0 release is 9 days old so maybe the fix works but has not been propagated to the pypi release.

Balandat:
This fix should work, at least it worked for botorch. You're right that this has not made it into the 1.6.0 release yet.

PR #1874, opened 2021-12-27
Add Fantasy Strategy for Variational GPs

wjmaddox:
Implements the online variational conditioning (OVC) procedure described in https://papers.nips.cc/paper/2021/hash/325eaeac5bef34937cfdc1bd73034d17-Abstract.html for Gaussian models. Broadly, this enables fantasy models and closed form updates to variational GPs (especially SVGPs) as more data comes in, for example, in the streaming setting or during lookaheads in Bayes opt.

Currently, this PR will only support the procedure for Gaussian responses (hooking up a Newton iteration strategy for any 1d likelihood is a good next PR). 

ToDos:
- [x] test out current version of strategy
- [x] basic unit tests
- [x] tutorial notebook (edit: needs some word-smithing)
- [x] unit tests in examples
- [x] documentation doesn't fail

cc @samuelstanton


gpleiss:
I'll review this early next week!

gpleiss (changes requested):
I don't love the idea of calling `self.model.mean_module` and `self.model.covar_module` in the fantasy model function - because it's not guaranteed that people will follow this naming convention. However, I'm not sure what the best solution is here either...

gpytorch/test/variational_test_case.py
```diff
@@ -272,3 +285,62 @@ def test_training_all_batch_zero_mean(self):
...
+            constant_mean=constant_mean,
+            num_inducing=num_inducing,
+        )
+
+        fant_model = self._fantasy_iter(model, likelihood, data_batch_shape, self.cuda, num_fant=num_fant)
+        self.assertTrue(isinstance(fant_model, gpytorch.models.ExactGP))
+        self.assertTrue(fant_model.prediction_strategy is not None)
+        for key in fant_model.prediction_strategy._memoize_cache.keys():
+            if key[0] == "mean_cache":
+                break
```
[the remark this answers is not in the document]
wjmaddox: Yeah, I was hoping that I could figure out a better `get_from_cache_ignore_all_args` function there but this felt like the best at the moment.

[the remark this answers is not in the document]
gpleiss: For testing purposes, I think this is fine.

gpytorch/models/approximate_gp.py
```diff
@@ -76,6 +76,30 @@ def pyro_model(self, input, beta=1.0, name_prefix=""):
         return super().pyro_model(input, beta=beta, name_prefix=name_prefix)
 
     def get_fantasy_model(self, inputs, targets, **kwargs):
```
samuelstanton, line 78: I can see why it might be convenient to always have `get_fantasy_model` return an `ExactGP`, regardless of the original model class, but it might be worth considering naming this something else, reserving `get_fantasy_model` for the version of OVC that returns a variational GP (in other words consider requiring that `get_fantasy_model` always returns an instance of the original class).

gpytorch/variational/variational_strategy.py
```diff
@@ -84,6 +98,65 @@ def prior_distribution(self):
...
+        cov_diff = -1.0 * cov_diff
+
+        # K^{1/2}
+        Kmm = self.model.covar_module(self.inducing_points)
+        Kmm_root = Kmm.cholesky()
+
+        # D_a = (S^{-1} - K^{-1})^{-1} = S + S R^{-1} S
+        # note that in the whitened case R = I - S, unwhitened R = K - S
+        # we compute (R R^{T})^{-1} R^T S for stability reasons as R is probably not PSD.
+        eval_lhs = var_cov.evaluate()
```
samuelstanton, line 132: `eval_lhs` is a rather confusing name since `S` is on the right.

gpytorch/variational/variational_strategy.py
```diff
@@ -84,6 +98,65 @@ def prior_distribution(self):
...
+        eval_rhs = cov_diff.transpose(-1, -2).matmul(eval_lhs)
+        inner_term = cov_diff.matmul(cov_diff.transpose(-1, -2))
+        # TODO: flag the jitter here
+        inner_solve = inner_term.add_jitter(1e-3).inv_matmul(eval_rhs, eval_lhs.transpose(-1, -2))
+        inducing_covar = var_cov + inner_solve
+
+        inducing_covar = Kmm_root.matmul(inducing_covar).matmul(Kmm_root.transpose(-1, -2))
```
samuelstanton, line 139: Matches appendix C.3 in the OVC paper.

gpytorch/variational/unwhitened_variational_strategy.py
```diff
@@ -58,6 +62,58 @@ def prior_distribution(self):
...
+
+        # R = K - S
+        Kmm = self.model.covar_module(self.inducing_points)
+        res = Kmm - var_cov
+
+        cov_diff = res
+
+        # D_a = (S^{-1} - K^{-1})^{-1} = S + S R^{-1} S
+        # note that in the whitened case R = I - S, unwhitened R = K - S
+        # we compute (R R^{T})^{-1} R^T S for stability reasons as R is probably not PSD.
```
samuelstanton, line 92: seems like the next 10 lines or so should be a utility function, rather than duplicating the same code across different strategies.

gpytorch/models/approximate_gp.py
wjmaddox: That was a thought I originally had but it requires the reasonably unstable updates to `m`, `S` in order to return its own model class itself rather than the exactGP. Although potentially lower overhead in the future to implement new fantasization strategies.

gpytorch/variational/_variational_strategy.py
```diff
@@ -97,6 +122,113 @@ def kl_divergence(self):
             kl_divergence = torch.distributions.kl.kl_divergence(self.variational_distribution, self.prior_distribution)
         return kl_divergence
 
+    @cached(name="inducing_model")
+    def inducing_model(self):
+        with torch.no_grad():
+            inducing_noise_covar, inducing_mean = self.pseudo_points
```
samuelstanton, line 128: like my other comment, I would be careful about overloading term 'inducing', really you're converting/transforming the variational GP into an equivalent ExactGP. Something like `exact_transform` may be better. Make sure this cache is getting cleared whenever the underlying variational GP changes.

gpytorch/variational/_variational_strategy.py
```diff
@@ -97,6 +122,113 @@ def kl_divergence(self):
...
+            inducing_points = self.inducing_points.detach()
+            if inducing_points.ndim < inducing_mean.ndim:
+                inducing_points = inducing_points.expand(*inducing_mean.shape[:-2], *inducing_points.shape)
+            # TODO: add flag for conditioning into SGPR after building fantasy strategy for SGPR
+            new_covar_module = deepcopy(self.model.covar_module)
+
+            # update inducing mean if necessary
+            inducing_mean = inducing_mean.squeeze() + self.model.mean_module(inducing_points)
```
samuelstanton, line 136: `inducing_mean` --> `pseudo_target_mean`

gpytorch/variational/_variational_strategy.py
```diff
@@ -97,6 +122,113 @@ def kl_divergence(self):
...
+
+            inducing_exact_model = _BaseExactGP(
+                inducing_points,
+                inducing_mean,
+                mean_module=deepcopy(self.model.mean_module),
```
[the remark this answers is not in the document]
samuelstanton: Seems like this is playing the same role as this line

https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/models/exact_gp.py#L223

But here we want to copy some of the attributes of one class into a completely different class. Not sure there is a general way to do this without assuming the attribute names.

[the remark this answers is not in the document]
samuelstanton: would be good to add an informative error message if the attributes are missing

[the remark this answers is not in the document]
wjmaddox: The solution we're going to try here is to require either `mean / covar` to be so named in the model or to require it in the kwargs, which gives the added benefit of some amount of fantasizing through updated hypers.

gpytorch/variational/_variational_strategy.py
```diff
@@ -97,6 +122,113 @@ def kl_divergence(self):
...
+        # first we construct an exact model over the inducing points with the inducing covariance
+        # matrix
+        inducing_exact_model = self.inducing_model()
+
+        # then we update this model by adding in the inputs and pseudo targets
+        # if inputs.shape[-2] == 1 or targets.shape[-1] != 1:
+        #     targets = targets.unsqueeze(-1)
+        # put on a trailing bdim for bs of 1
+        # finally we fantasize wrt targets
+        fantasy_model = inducing_exact_model.get_fantasy_model(inputs, targets, **kwargs)
```
samuelstanton, line 217: If the original likelihood is non-Gaussian there should be a Laplace approximation here.

gpytorch/variational/unwhitened_variational_strategy.py
wjmaddox: Hmm.. only the last set of calls actually are (e.g. creting the `pseudo_target_covar`)

gpytorch/variational/_variational_strategy.py
wjmaddox: Went with `amortized_exact_gp` which should be clear the type of perspective we're going for.

wjmaddox:
I'm still not sure what the docs issue is.

Balandat:
It says 

> /home/docs/checkouts/readthedocs.org/user_builds/gpytorch/checkouts/1874/docs/source/examples/08_Advanced_Usage/SVGP_Model_Updating.ipynb: WARNING: document isn't included in any toctree

gpytorch/variational/_variational_strategy.py
```diff
@@ -97,6 +123,144 @@ def kl_divergence(self):
...
+        an exact GP model that incorporates the inputs and targets alongside the variational model's inducing
+        points and targets.
+
+        Currently, instead of directly updating the variational parameters (and inducing points), we instead
+        return an ExactGP model rather than an updated variational GP model. This is done primarily for
+        numerical stability.
+
+        Unlike the ExactGP's call for get_fantasy_model, we enable options for mean_module and covar_module
+        that allow specification of the mean / covariance.
+
```
gpleiss, line 204: Can you add the expected arguments for the doc string? It should mention the `mean_module`/`covar_module` behavior.

gpytorch/variational/_variational_strategy.py
```diff
@@ -97,6 +123,144 @@ def kl_divergence(self):
...
+                ". Only GaussianLikelihoods are currently supported.",
+            )
+        # we assume that either the user has given the model a mean_module and a covar_module
+        # or that it will be passed into the get_fantasy_model function. we check for these.
+        if mean_module is None:
+            mean_module = getattr(self.model, "mean_module", None)
+            if mean_module is None:
+                raise ModuleNotFoundError(
+                    "Either you must provide a mean_module as input to get_fantasy_model",
+                    "or it must be an attribute of the model.",
```
gpleiss, line 231: Could this be a little more explicit? "or it must be an attribute of the model named `mean_module`." or something like that.

gpytorch/variational/_variational_strategy.py
```diff
@@ -97,6 +123,144 @@ def kl_divergence(self):
...
+                )
+        if covar_module is None:
+            covar_module = getattr(self.model, "covar_module", None)
+            if covar_module is None:
+                # raise an error
+                raise ModuleNotFoundError(
+                    "Either you must provide a covar_module as input to get_fantasy_model",
+                    "or it must be an attribute of the model.",
```
gpleiss, line 239: Same thing here "or it must be an attribute of the model named `covar_module`." or something like that.

gpleiss (changes requested):
A few tiny doc string/error message fixes, but then I think it's good to go!

wjmaddox:
Sorry for taking a while to clean that up, but hopefully the wording is a lot more clear now.

gpleiss (approved):

PR #1876, opened 2022-01-05
Make it possible to pickle simple kernels and likelihoods

dme65:
We sometimes need to pickle (simple) GPyTorch models, but this doesn't work when using priors since `register_prior` creates local lambdas. This PR makes it possible to pickle simple kernels such as Matern, RBF, Scale as well as the Gaussian likelihood.


gpytorch/kernels/kernel.py
```diff
@@ -172,7 +172,7 @@ def __init__(
                 if not isinstance(lengthscale_prior, Prior):
                     raise TypeError("Expected gpytorch.priors.Prior but got " + type(lengthscale_prior).__name__)
                 self.register_prior(
-                    "lengthscale_prior", lengthscale_prior, lambda m: m.lengthscale, lambda m, v: m._set_lengthscale(v)
+                    "lengthscale_prior", lengthscale_prior, self._lengthscale_param, self._lengthscale_closure
```
Balandat, line 175: I guess this should also work here?
```suggestion
                    "lengthscale_prior", lengthscale_prior, "lengthscale", self._lengthscale_closure
```

gpytorch/kernels/scale_kernel.py
```diff
@@ -81,11 +81,17 @@ def __init__(
             if not isinstance(outputscale_prior, Prior):
                 raise TypeError("Expected gpytorch.priors.Prior but got " + type(outputscale_prior).__name__)
             self.register_prior(
-                "outputscale_prior", outputscale_prior, lambda m: m.outputscale, lambda m, v: m._set_outputscale(v)
+                "outputscale_prior", outputscale_prior, self._outputscale_param, self._outputscale_closure
```
Balandat, line 84: same here
```suggestion
                "outputscale_prior", outputscale_prior, "outputscale", self._outputscale_closure
```

gpytorch/likelihoods/noise_models.py
```diff
@@ -26,10 +26,16 @@ def __init__(self, noise_prior=None, noise_constraint=None, batch_shape=torch.Si
 
         self.register_parameter(name="raw_noise", parameter=Parameter(torch.zeros(*batch_shape, num_tasks)))
         if noise_prior is not None:
-            self.register_prior("noise_prior", noise_prior, lambda m: m.noise, lambda m, v: m._set_noise(v))
+            self.register_prior("noise_prior", noise_prior, self._noise_param, self._noise_closure)
```
Balandat, line 29: and here
```suggestion
            self.register_prior("noise_prior", noise_prior, "noise", self._noise_closure)
```

Balandat (commented):
Nice, thanks! Very useful.

gpytorch/kernels/kernel.py
dme65: I don't think this works (making this change causes 31 tests to fail). Note that `_lengthscale_param` returns `m.lengthscale` and not `self.lengthscale`.

Balandat: Hmm not sure I understand why it doesn’t, seems like the closure definition here should do the same? https://github.com/cornellius-gp/gpytorch/blob/32cde571f8ee0bf8d1a523b23e51cce57aaa5045/gpytorch/module.py#L235

Either way, we can figure that out later (or I’m missing sth obvious). Accepting  this as is.

Balandat (approved):

gpytorch/kernels/kernel.py
dme65: Sounds good! Looks like merging is blocked though until the docs can be built successfully. Looking at https://readthedocs.org/projects/gpytorch/builds/15684237/ it looks like the issue is a warning that is unrelated to this PR.

dme65:
Ok, fixed the doc building issue by pinning `nbsphinx==0.8.7`. Looks like they released 0.8.8 on Dec 31, 2021.

Issue #1911, opened 2022-02-09
[Feature Request] `__rmatmul__` for LazyTensors


j-wilson:
# 🚀 Feature Request

Implement `__rtmatmul__` for LazyTensors.

## Motivation

LazyTensors must be converted to vanilla Tensors when used by functions that perform matrix multiplies between Tensors (left) and LazyTensors (right). For example, if `A` is a lazy and `K` is not then expressions such as `A @ K @ A.T` will fail.

**Describe the solution you'd like**
I do not know what challenges present themselves here. This may be as simple as adding something like

```
def __rmatmul__(self, other: Tensor) -> Tensor:
    retvalT = self.transpose(-1, -2).matmul(other)
    return retvalT.transpose(-1, -2) if other.ndim > 1 else retvalT
```

**Are you willing to open a pull request?**
Potentially, yes. It would be good to determine the scope of the PR first however.

jacobrgardner:
I think an implementation that basically does the above (although of course also transpose other) would be a perfectly reasonable first pass at this.

The only thing to note is that lazy tensors that represent symmetric positive definite matrices (which is the dominant use case in gpytorch of course) don't even need to have `_transpose_nonbatch` implemented, since obviously `self.transpose_nonbatch(-2, -1)` would just be the same LT as `self`.

That said, I just looked through all of the LTs again, and basically every one I see either implements `_transpose_nonbatch` or is already actually explicitly assuming the matrix is symmetric, e.g., https://github.com/cornellius-gp/gpytorch/blob/4df30488b2e16b46519cb9d34758745c7e2dbb61/gpytorch/lazy/toeplitz_lazy_tensor.py#L52-L53

so the quick implementation may be all we need here.

jacobrgardner:
@j-wilson Yeah, both the implementation and the basic test looks pretty good to me :+1:

PR #1932, opened 2022-03-07
LazyTensor.rmatmul

j-wilson:
Add support for right multiplication by a lazy tensor; see #1911.

### DIFF
--- a/gpytorch/lazy/lazy_tensor.py
+++ b/gpytorch/lazy/lazy_tensor.py
@@ -1408,6 +1408,23 @@ class LazyTensor(ABC):
 
         return Matmul.apply(self.representation_tree(), other, *self.representation())
 
+    def rmatmul(self, other):
+        """
+        Multiplies a matrix by self.
+
+        Args:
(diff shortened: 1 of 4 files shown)


gpytorch/lazy/lazy_tensor.py
```diff
@@ -2307,6 +2307,11 @@ def _symeig(self, eigenvectors: bool = False) -> Tuple[Tensor, Optional["LazyTen
     def __matmul__(self, other):
         return self.matmul(other)
 
+    def __rmatmul__(self, other: Tensor) -> Tensor:
+        if other.ndim == 1:
+            return self.transpose(-1, -2).matmul(other)
+        return self.transpose(-1, -2).matmul(other.transpose(-1, -2)).transpose(-1, -2)
+
```
Balandat, line 2314: I wonder whether it make sense to just introduce a `rmatmul` method on the base lazy tensor that implements this as done here, and then have `__rmatmul__` just call that. That would make it easier to special-case right matmuls for certain lazy tensors by overwriting `rmatmul` and would follow the same API pattern as with most other calls.

jacobrgardner (approved):
Looks good to me now

PR #1970, opened 2022-04-08
Pin pyro to <=1.8 in CI workflows

Balandat:
Pyro 1.8.1 requires torch 1.11, which caused issues during install in the CI workflows.


dme65 (approved):
Thanks! Let's remember to go back and unpin Pyro once the torch 1.11 issues have been resolved.

PR #1973, opened 2022-04-10
Require py>=3.7 and pyro==1.8

Balandat:
It's about time


Balandat:
Failure here is that newer py requires newer black but not new enough to not break b/c of some recent click upgrade. #1972 bumps the black version accordingly to fix this. We can merge this PR in and then fast follow with #1972.

PR #1976, opened 2022-04-13
Increase tolerance on _test_rmatmul of TestLazyEvaluatedKernelTensorBatch

Balandat:
This caused some flaky tests. Should be safe to increase this..


dme65 (approved):
LGTM!

PR #1992, opened 2022-04-24
Make it possible to pickle constant mean with prior

dme65:
If we try to pickle a constant mean with a prior we end up with an `AttributeError: Can't pickle local object 'Module.register_prior.<locals>.closure'` error. The issue is that pickle doesn't seem to play well with specifying parameters by name (string). This PR makes it possible to pickle a constant mean with a simple prior (related: #1876 which made similar changes to some of the kernels to make it possible to pickle them).


Balandat (approved):
thanks, lgtm!

PR #1998, opened 2022-04-28
Fixed documentation of CG's tolerance parameter

JonathanWenger:
# In a Nutshell
The `tolerance` parameter of `linear_cg` was incorrectly documented as controlling the maximum residual, while the iteration stops when the average residual is below the `tolerance`. 

https://github.com/cornellius-gp/gpytorch/blob/306d7332d74b0d337552911d69d97d939991d031/gpytorch/utils/linear_cg.py#L285-L291

### DIFF
--- a/gpytorch/utils/contour_integral_quad.py
+++ b/gpytorch/utils/contour_integral_quad.py
@@ -21,7 +21,7 @@ def contour_integral_quad(
     shift_offset=0,
 ):
     r"""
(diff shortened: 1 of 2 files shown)


JonathanWenger:
One could argue, that the tolerance should bound the maximum residual as described in the documentation.

gpleiss (approved):

PR #2007, opened 2022-05-11
Special case some BlockDiagLazyTensor matmul ops

Balandat:
This avoids creating MatMulLazyTensor objects in some cases where matmuls are trivial.


gpytorch/lazy/block_diag_lazy_tensor.py
```diff
@@ -115,6 +115,18 @@ def inv_quad_logdet(self, inv_quad_rhs=None, logdet=False, reduce_inv_quad=True)
             logdet_res = logdet_res.view(*logdet_res.shape).sum(-1)
         return inv_quad_res, logdet_res
 
+    def matmul(self, other):
+        from .diag_lazy_tensor import DiagLazyTensor
```
dme65, line 119: Why isn't this imported at the top of the file? Does it cause a circular import or something?

dme65 (approved):
Not super familiar with some of the lazy tensor internals, but the special cases look reasonable to me. Do we have enough test coverage for the special cases introduced here?

gpytorch/lazy/block_diag_lazy_tensor.py
Balandat: Yes, since `BlockDiagLazyTensor` is imported top-of-file in `diag_lazy_tensor.py` (see below).

Balandat:
> Do we have enough test coverage for the special cases introduced here?

Good question - They ran successfully downstream in BoTorch where this was used, but it probably makes sense to check test coverage in the units again.

gpleiss:
Some unit tests would be helpful here, especially for the LinearOperator refactor.

gpleiss (changes requested):
Looks good, but a few unit tests would be good for the LinearOperator refactor. I can add those if you want.

Balandat:
Yeah I have some local changes to add tests for matmuls with `DiagLazyTensor` and `BlockDiagLazyTensor` but I'll need a bit of time to clean this up. maybe I'll fix the issues there and push them.

Balandat:
Added unit tests. Making sure those pass actually required special casing some logic for lazy tensors in `InterpolatedLazyTensor` - matmuls with diagonal matrices can be done very easily simply by multiplying the interpolation values (someone should double check this though...). There are probably a number of other places where things could be better optimized...

Balandat:
We should probably also more generally think about whether we want the interface to support rmatmuls with lazy tensors as first-class operations (my vote is a strong yes). But we should probably just do this as part of the linear_operator implementation  / refactor.

PR #2009, opened 2022-05-17
Clamp noise in `FixedGaussianNoise` via `min_fixed_noise` context manager

saitcakmak:
Introduces a `min_fixed_noise` context manager that is used to clamp the `noise` values in `FixedGaussianNoise` to a minimum value dependent on the `dtype`. This helps eliminate some numerical issues we face in downstream when using too small (0) noise. 

Test plan: Units.

cc @Balandat, @mpolson64, @gpleiss


gpytorch/settings.py
```diff
@@ -420,6 +420,21 @@ class max_cg_iterations(_value_context):
...
 
 
+class min_fixed_noise(_dtype_value_context):
+    """
+    The minimum noise value that can be used in :obj:`~gpytorch.likelihoods.FixedNoiseGaussianLikelihood`.
+    If the supplied noise values are smaller than this, they are rounded up and a warning is raised.
+
+    - Default for `float`: 1e-6
+    - Default for `double`: 1e-10
+    - Default for `half`: 1e-3
```
Balandat, line 430: this doesn't seem to agree with the values below?

Balandat: also 1e-8 is pretty small... seems like this wouldn't resolve the issues we've seen with numerical issues creeping up in the backward pass?

saitcakmak: 1e-8 seemed to work for our particular case, but swapped to 1e-6 / 1e-4 for more conservative values.

test/likelihoods/test_gaussian_likelihood.py
```diff
@@ -84,6 +85,13 @@ def test_fixed_noise_gaussian_likelihood(self, cuda=False):
             obs_noise = 0.1 + torch.rand(5, device=device, dtype=dtype)
             out = lkhd(mvn, noise=obs_noise)
             self.assertTrue(torch.allclose(out.variance, 1 + obs_noise))
+            # test noise smaller than min_fixed_noise
+            expected_min_noise = settings.min_fixed_noise.value(dtype)
+            noise[:2] = 0
+            lkhd = FixedNoiseGaussianLikelihood(noise=noise)
```
Balandat, line 91: Should we also check that the warning is raised here? Not super critical though.

Balandat (approved):

PR #2028, opened 2022-06-03
Lower Precision CG Solver

AndPotap:
Added our CG solver that is more numerically stable and that incorporates re-orthogonalization.


gpleiss:
Hi @AndPotap this looks awesome! We're in the process of moving all of the numerical utilities into GitHub.com/cornellius-gp/linear_operator, so could you open up this PR in that repo instead?

AndPotap:
Sure!

wjmaddox:
Just for reference, this is based off of @AndPotap and I's recent paper on low precision CG for GPs, available here: https://openreview.net/forum?id=S3NOX_Ij9xc

PR #2039, opened 2022-06-14
Fix handling of prior terms in ExactMarginalLogLikelihood

saitcakmak:
This fixes an issue originally discovered in pytorch/botorch#1259. The bug relates to handling of prior terms in https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/mlls/exact_marginal_log_likelihood.py#L42-L43

When training batched model of `n_batch` batches, the `res` term (which is the log probability of observations under the prior) has shape `n_batch`. The prior term `prior.log_prob(closure(module))` also has a shape that is `n_batch x (potential other dimensions)`. When we add the sum of all prior terms via `res.add_` we end up adding each prior term to each one of the terms in `res`, when they should only be added to the term corresponding to the same batch. This is corrected by replacing the `sum()` with a series of sums that reduce the prior term to have the same `ndim` as `res`.

Test plan:
Units

cc @Balandat


saitcakmak:
This seems to be the same issue as #1318, so that should also get fixed.

gpytorch/mlls/exact_marginal_log_likelihood.py
```diff
@@ -39,8 +39,12 @@ def _add_other_terms(self, res, params):
             res = res.add(added_loss_term.loss(*params))
 
         # Add log probs of priors on the (functions of) parameters
+        res_ndim = res.ndim
         for name, module, prior, closure, _ in self.named_priors():
-            res.add_(prior.log_prob(closure(module)).sum())
+            prior_term = prior.log_prob(closure(module))
```
Balandat, line 44: How about 
```
prior_term = prior.log_prob(closure(module))
res.add_(prior_term.view(*res.ndim, -1).sum(dim=-1))
```

Balandat (approved):
Thanks!

gpytorch/mlls/exact_marginal_log_likelihood.py
saitcakmak: Oops, forgot about this for a while. Updated!

PR #2042, opened 2022-06-15
Fix sample from prior for ConstantMean

dme65:
The use of `fill_` doesn't play well with the shape of the priors.

### DIFF
--- a/gpytorch/means/constant_mean.py
+++ b/gpytorch/means/constant_mean.py
@@ -18,7 +18,9 @@ class ConstantMean(Mean):
         return m.constant
 
     def _constant_closure(self, m, value):
-        return m.constant.data.fill_(value)
+        if not torch.is_tensor(value):
+            value = torch.as_tensor(value).to(self.constant)
+        self.initialize(constant=value.reshape(self.constant.shape))
 
     def forward(self, input):
         if input.shape[:-2] == self.batch_shape:
--- a/gpytorch/means/constant_mean.py
+++ b/gpytorch/means/constant_mean.py
@@ -20,7 +20,7 @@ class ConstantMean(Mean):
     def _constant_closure(self, m, value):
         if not torch.is_tensor(value):
             value = torch.as_tensor(value).to(self.constant)
-        self.initialize(constant=value.reshape(self.constant.shape))
+        m.initialize(constant=value.reshape(self.constant.shape))
 
     def forward(self, input):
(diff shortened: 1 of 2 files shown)


test/means/test_constant_mean.py
```diff
@@ -11,27 +11,33 @@
...
+        else:
+            prior = NormalPrior(torch.zeros(self.__class__.batch_shape), torch.ones(self.__class__.batch_shape))
+        mean = self.create_mean(prior=prior)
         self.assertEqual(mean.mean_prior, prior)
         pickle.loads(pickle.dumps(mean))  # Should be able to pickle and unpickle with a prior
-        mean._constant_closure(mean, 1.234)
-        self.assertAlmostEqual(mean.constant.item(), 1.234)
+        value = prior.sample()
+        mean._constant_closure(mean, value)
+        self.assertTrue((mean.constant.data == value.reshape(mean.constant.data.shape)).all())
```
Balandat, line 29: Use `torch.equal` here?

test/means/test_constant_mean.py
```diff
@@ -11,27 +11,33 @@
...
-        return ConstantMean()
+    batch_shape = None
+
+    def create_mean(self, prior=None):
+        return ConstantMean(prior=prior, batch_shape=torch.Size([]))
 
     def test_prior(self):
-        prior = NormalPrior(0.0, 1.0)
-        mean = ConstantMean(prior=prior)
+        if self.__class__.batch_shape is None:
```
Balandat, line 20: Any particular reason why you're accessing this via the class rather than via the instance (`self.batch_shape`)?

dme65: I'm doing the same as what is done in the base class. It may not be necessary though, will check.

gpytorch/means/constant_mean.py
```diff
@@ -18,7 +18,9 @@ def _constant_param(self, m):
         return m.constant
 
     def _constant_closure(self, m, value):
-        return m.constant.data.fill_(value)
+        if not torch.is_tensor(value):
+            value = torch.as_tensor(value).to(self.constant)
```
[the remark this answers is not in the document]
dme65: I'm just following what is done for other closures, for example: https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/kernels/kernel.py#L262.

PR #2061, opened 2022-07-11
Fix rendering of Google-style docstrings

adamjstewart:
Many docstrings use Google-style conventions and require the [sphinx.ext.napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) extension to render correctly. 

For example, [gpytorch.kernels.PeriodicKernel](https://docs.gpytorch.ai/en/stable/kernels.html#periodickernel) uses Python-style `:param:` lists and renders correctly, but [gpytorch.kernels.ScaleKernel](https://docs.gpytorch.ai/en/stable/kernels.html#scalekernel) uses Google-style `Args:` lists and is not rendered correctly. This PR fixes that: [before](https://docs.gpytorch.ai/en/stable/kernels.html#scalekernel), [after](https://gpytorch--2061.org.readthedocs.build/en/2061/kernels.html#scalekernel).

Fixing this uncovered a lot of other issues in the docs. `:attr:` should only be used to cross-reference class attributes, not to define or cross-reference function parameters.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -80,6 +80,7 @@ release = version
 extensions = [
     "sphinx.ext.coverage",
     "sphinx.ext.mathjax",
+    'sphinx.ext.napoleon',
     "sphinx.ext.viewcode",
     "sphinx.ext.githubpages",
     "sphinx.ext.autodoc",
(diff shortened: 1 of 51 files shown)


Balandat (approved):
Nice, thanks for this!

PR #2063, opened 2022-07-12
torch.triangular_solve: fix deprecation warning

adamjstewart:
Every time I run my code with PyTorch 1.12, I see the following UserWarning printed:
```
~/lib/python3.9/site-packages/gpytorch/lazy/triangular_lazy_tensor.py:136: UserWarning: torch.triangular_solve is deprecated in favor of torch.linalg.solve_triangularand will be removed in a future PyTorch release.
torch.linalg.solve_triangular has its arguments reversed and does not return a copy of one of the inputs.
X = torch.triangular_solve(B, A).solution
should be replaced with
X = torch.linalg.solve_triangular(A, B). (Triggered internally at  ~/src/aten/src/ATen/native/BatchLinearAlgebra.cpp:2189.)
  res = torch.triangular_solve(right_tensor, self.evaluate(), upper=self.upper).solution
```
This PR fixes that by using `torch.linalg.triangular_solve` if it exists and falling back to `torch.solve_triangular` if it does not.

This isn't the cleanest solution. Maybe we should introduce a `compat.py` file to create a wrapper around this? If/when you drop PyTorch 1.10 support this will no longer be needed.

### DIFF
--- a/gpytorch/functions/_pivoted_cholesky.py
+++ b/gpytorch/functions/_pivoted_cholesky.py
@@ -112,16 +112,11 @@ class PivotedCholesky(Function):
             # Compute (Krows * L^{-T}) - the (pivoted) result of Pivoted Cholesky
             if hasattr(torch.linalg, "solve_triangular"):
                 # PyTorch 1.11+
-                res_pivoted = torch.cat(
-                    [L, torch.linalg.solve_triangular(L, Krows[..., m:, :].transpose(-1, -2), upper=False).transpose(-1, -2)],
-                    dim=-2,
-                )
+                solution = torch.linalg.solve_triangular(L, Krows[..., m:, :].transpose(-1, -2), upper=False)
             else:
                 # PyTorch 1.10
-                res_pivoted = torch.cat(
-                    [L, torch.triangular_solve(Krows[..., m:, :].transpose(-1, -2), L, upper=False)[0].transpose(-1, -2)],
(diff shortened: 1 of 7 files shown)


adamjstewart:
P.S. There are a couple other deprecation warnings visible in the [test logs](https://github.com/cornellius-gp/gpytorch/runs/7307322191?check_suite_focus=true):
```
/opt/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/torch/functional.py:478: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at ../aten/src/ATen/native/TensorShape.cpp:2966.)
  return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
...
/opt/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/torch/testing/_deprecated.py:35: FutureWarning: torch.testing.assert_allclose() is deprecated since 1.12 and will be removed in 1.14. Use torch.testing.assert_close() instead. For detailed upgrade instructions see https://github.com/pytorch/pytorch/issues/61844.
  warnings.warn(msg, FutureWarning)
```
You can actually test for these using `pytest -Werror` if you want to have the tests fail on these kinds of issues. See https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#controlling-warnings for more details.

PR #2073, opened 2022-07-20
Fix missing tilde in docs

adamjstewart:
The tilde in "ε ~ 𝑁(...)" was missing. Solution is to use the `\sim` character in LaTeX. See [before](https://docs.gpytorch.ai/en/latest/likelihoods.html#gpytorch.likelihoods.Likelihood) vs. [after](https://gpytorch--2073.org.readthedocs.build/en/2073/likelihoods.html#gpytorch.likelihoods.Likelihood).

Many of these equations aren't wrapped in math mode so aren't being rendered at all...


gpleiss (approved):

gpleiss:
Awesome, thanks @adamjstewart !

PR #2080, opened 2022-07-26
Upgrade to modern Sphinx

adamjstewart:
The docs are currently pinned to an ancient version of Sphinx from 2019. This PR updates the docs to support the latest version of Sphinx.

I'm working on a bunch of other doc fixes and this seemed like the best place to start so as to avoid tracking down known Sphinx issues that were fixed in a later release.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -5,3 +5,4 @@ sphinx
 sphinx_rtd_theme
 nbsphinx
 m2r2
+pyro-ppl
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,8 +1,7 @@
-docutils==0.17.1
 nbformat
 ipython
 ipykernel
-sphinx==1.8.5
+sphinx
 sphinx_rtd_theme
-nbsphinx==0.8.7
-m2r
+nbsphinx
+m2r2
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -85,7 +85,7 @@ extensions = [
     "sphinx.ext.githubpages",
     "sphinx.ext.autodoc",
     "nbsphinx",
-    "m2r",
+    "m2r2",
 ]
 
 # Disable docstring inheritance
@@ -107,7 +107,7 @@ master_doc = "index"
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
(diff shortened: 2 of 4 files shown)


docs/requirements.txt
```diff
@@ -1,8 +1,7 @@
-docutils==0.17.1
 nbformat
 ipython
 ipykernel
-sphinx==1.8.5
+sphinx
 sphinx_rtd_theme
-nbsphinx==0.8.7
-m2r
```
adamjstewart, line 8: m2r has been abandoned, m2r2 is an API-compatible fork of m2r that supports Sphinx 3+

docs/source/conf.py
```diff
@@ -107,7 +107,7 @@ def find_version(*file_paths):
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = "en"
```
adamjstewart, line 110: `language = None` is no longer valid in Sphinx 5, either set it to `"en"` or remove the setting.

docs/source/conf.py
```diff
@@ -208,11 +208,3 @@ def find_version(*file_paths):
...
     )
 ]
-
-
-# Taken from https://github.com/pyro-ppl/pyro/blob/dev/docs/source/conf.py#L213
-# @jpchen's hack to get rtd builder to install latest pytorch
-# See similar line in the install section of .travis.yml
-if "READTHEDOCS" in os.environ:
-    os.system("pip install torch==1.10.0+cpu -f https://download.pytorch.org/whl/torch_stable.html")
-    os.system("pip install pyro-ppl")
```
adamjstewart, line 218: I'm hoping that adding `requirements.txt` to `.readthedocs.yaml` makes this obsolete.

gpleiss: We will still need to install pyro somewhere.

setup.py
```diff
@@ -59,7 +59,7 @@ def find_version(*file_paths):
     install_requires=install_requires,
     extras_require={
         "dev": ["black", "twine", "pre-commit"],
-        "docs": ["ipython", "ipykernel", "sphinx<3.0.0", "sphinx_rtd_theme", "nbsphinx", "m2r"],
+        "docs": ["ipython", "ipykernel", "sphinx", "sphinx_rtd_theme", "nbsphinx", "m2r2"],
```
gpleiss, line 62: While we're cleaning up the dependencies, we could probably get rid of the docs `extras_require`, since we exclusively use `docs/requirements.txt`. So I would propose just deleting this line.

gpleiss (changes requested):
Thanks @adamjstewart for the cleanup! Could you make the two changes I requested, and then I'll merge this!

gpleiss (approved):

gpleiss:
Thanks @adamjstewart !

PR #2082, opened 2022-07-27
Add constant_constraint to ConstantMean

### LINKED ISSUE #2074: [Question] Setting a constraint for the ConstantMean()
Hi - I am trying to set a GreaterThan() constraint for the constant of the mean_module -- which I thought is trival following the examples from the docs, but unfortunately, the training doesn't seem to obey it. 

```
    
    def __init__(self, train_x, train_y, likelihood, kernel):
        self.mean_module = gpytorch.means.ConstantMean(constant_constraint=gpytorch.constraints.GreaterThan(0.5))
        self.covar_module = kernel

    def forward(self, x):
        mean_x = self.mean_module(x)
```
 ```
    model = ExactGPModel(train_x, train_y, likelihood, kernel).double().cuda()
    model.mean_module.register_constraint('constant', gpytorch.constraints.GreaterThan(0.5))
   
```
Anything missing here?

Upon some digging, I noticed that the constant mean is slightly different to the other hyperparameters which are tensors while the constant mean is a Parameter -- could that be it?

```
model.likelihood.noise
tensor([0.2515], device='cuda:0', dtype=torch.float64, grad_fn=<AddBackward0>)
```

```
model.mean_module.constant
Parameter containing:
tensor([0.0770], device='cuda:0', dtype=torch.float64, requires_grad=True)


```

### PR DESCRIPTION
**Pull request by gpleiss**
Previously, ConstantMean only supported a prior on the constant (not a constraint). This PR adds a `constant_constraint` kwarg to ConstantMean.

**Backwards compatibility**
There are two non-backwards compatible changes to ConstantMean, though there is a state_dict hook to ensure that previous state dicts can be loaded:
- The `constant` parameter is renamed to `raw_constant`.
- The shape of `raw_constant` is now `*batch_shape`, rather than `*batch_shape x 1`.

[Fixes #2074]

### DIFF
--- a/gpytorch/means/constant_mean.py
+++ b/gpytorch/means/constant_mean.py
@@ -1,29 +1,111 @@
 #!/usr/bin/env python3
 
+import warnings
+from typing import Any, Optional
+
 import torch
 
-from ..utils.broadcasting import _mul_broadcast_shape
+from ..constraints import Interval
+from ..priors import Prior
+from ..utils.warnings import OldVersionWarning
 from .mean import Mean
 
 
+def _ensure_updated_strategy_flag_set(
+    state_dict, prefix, local_metadata, strict, missing_keys, unexpected_keys, error_msgs
+):
+    if prefix + "constant" in state_dict:
+        constant = state_dict.pop(prefix + "constant").squeeze(-1)  # Remove deprecated singleton dimension
+        state_dict[prefix + "raw_constant"] = constant
+        warnings.warn(
+            "You have loaded a GP model with a ConstantMean  from a previous version of "
+            "GPyTorch. The mean module parameter `constant` has been renamed to `raw_constant`. "
+            "Additionally, the shape of `raw_constant` is now *batch_shape, whereas the shape of "
+            "`constant` was *batch_shape x 1. "
+            "We have updated the name/shape of the parameter in your state dict, but we recommend that you "
+            "re-save your model.",
+            OldVersionWarning,
+        )
+
+
 class ConstantMean(Mean):
-    def __init__(self, prior=None, batch_shape=torch.Size(), **kwargs):
+    r"""
+    A (non-zero) constant prior mean function, i.e.:
+
+    .. math::
+        \mu(\mathbf x) = C
+
+    where :math:`C` is a learned constant.
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: gpytorch/means/constant_mean.py [hunk 1]
```diff
@@ -1,29 +1,111 @@
...
+        \mu(\mathbf x) = C
+
+    where :math:`C` is a learned constant.
+
+    :param constant_prior: Prior for constant parameter :math:`C`.
+    :type constant_prior: ~gpytorch.priors.Prior, optional
+    :param constant_constraint: Constraint for constant parameter :math:`C`.
+    :type constant_constraint: ~gpytorch.priors.Interval, optional
+    :param batch_shape: The batch shape of the learned constant(s) (default: []).
+    :type batch_shape: torch.Size, optional
```
Balandat, line 45: consistency nit
```suggestion
    :type constant_prior: ~gpytorch.priors.Prior, optional.
    :param constant_constraint: Constraint for constant parameter :math:`C`.
    :type constant_constraint: ~gpytorch.priors.Interval, optional.
    :param batch_shape: The batch shape of the learned constant(s) (default: []).
    :type batch_shape: torch.Size, optional.
```

gpytorch/means/constant_mean.py
```diff
@@ -1,29 +1,111 @@
...
+
+    :var torch.Tensor constant: :math:`C` parameter
```
Balandat, line 47:
```suggestion
    :var torch.Tensor constant: :math:`C` parameter.
```

gpytorch/means/constant_mean.py
```diff
@@ -1,29 +1,111 @@
...
+                    "The kwarg `prior` for ConstantMean has been renamed to `constant_prior`, and will be deprecated.",
+                    DeprecationWarning,
+                )
+                constant_prior = constant_prior_deprecated
+            else:  # Weird edge case where someone set both `prior` and `constant_prior`
+                warnings.warn(
+                    "You have set both the `constant_prior` and the deprecated `prior` arguments for ConstantMean. "
+                    "`prior` is deprecated, and will be ignored.",
+                    DeprecationWarning,
+                )
```
Balandat, line 73: I think this is weird enough that we may just want to throw an error here?

Balandat (approved):

PR #2091, opened 2022-08-07
Run unittests with and without the faiss library.

gpleiss:
faiss is now an optional package for GPyTorch, to be used in conjunction with the
VNNGP model (#2026) and hopefully more nearest neighbor GP methods in the future.
This PR ensures that our test suite tests the functionality of GPyTorch with and
without the faiss library.


gpleiss:
I renamed the test suite here slightly (from `with-pyro` to `with-extras`), which is why the PR says "some checks haven't completed yet." We'll have to update the settings once this is merged.

Balandat (approved):
gtmtl

PR #2027, opened 2022-06-03
LazyTensor -> LinearOperator

gpleiss:
The LazyTensor functionality is moving to a separate repository (GitHub.com/cornellius-gp/linear_operator).

This PR deprecates LazyTensor and instead uses the LinearOperator implementation.

(Note that this PR does not change any GPyTorch code; other than removing gpytorch/lazy/*_lazy_tensor.py and all other related files. A separate PR will make changes to the code that rename all *LazyTensor references and utilize the torch_function functionality of LinearOperator).

## New requirements
- Python 3.8 (was 3.7)
- Torch 1.11 (was 1.9/1.10)

PR #2093, opened 2022-08-08
Meta constructor for `BlockDiagLazyTensor` to fix its `matmul` in special case

SebastianAment:
`BlockDiagLazyTensor`'s `matmul` method fails with a dimension mismatch, when `BlockDiagLazyTensor` is constructed with a `DiagLazyTensor` as the `base_lazy_tensor` argument. We hit this case in a BoTorch test, and I added it to the `TestBlockDiagLazyTensorMetaClass` below.

@Balandat and I thought that intercepting the construction of the `BlockDiagLazyTensor` and returning a vanilla `DiagLazyTensor` in this case would be a nice solution, instead of introducing another special case to `matmul`. This PR implements this feature via the `_MetaBlockDiagLazyTensor` constructor.

### DIFF
--- a/gpytorch/lazy/block_diag_lazy_tensor.py
+++ b/gpytorch/lazy/block_diag_lazy_tensor.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from abc import ABCMeta
 from typing import Optional, Tuple
 
 import torch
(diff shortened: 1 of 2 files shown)


gpleiss:
Hey @SebastianAment could you re-open this PR in the GitHub.com/cornellius-gp/linear_operator repo? We're about to switch over to using that, and deprecating all of gpytorch's lazy tensor stuff. (Either later today or tomorrow morning, [this PR](https://github.com/cornellius-gp/gpytorch/pull/2027) is finally going to be merged in!)

gpytorch/lazy/block_diag_lazy_tensor.py
```diff
@@ -9,8 +10,26 @@
...
+# BlockDiagLazyTensor(base_lazy_tensor, block_dim=-3) to return a DiagLazyTensor
+# if base_lazy_tensor is a DiagLazyTensor itself
+class _MetaBlockDiagLazyTensor(ABCMeta):
+    def __call__(cls, base_lazy_tensor, block_dim=-3):
+        from .diag_lazy_tensor import DiagLazyTensor
+        if cls is BlockDiagLazyTensor and isinstance(base_lazy_tensor, DiagLazyTensor):
+            if not block_dim == -3:
+                raise NotImplementedError(
+                    f"Passing a base_lazy_tensor of type DiagLazyTensor to the constructor"
+                    "of BlockDiagLazyTensor with block_dim = {block_dim} != -3 is not supported."
```
Balandat, line 23: Got the f on the wrong line here. Also forgot a space :) 
```suggestion
                    "Passing a base_lazy_tensor of type DiagLazyTensor to the constructor "
                    f"of BlockDiagLazyTensor with block_dim = {block_dim} != -3 is not supported."
```

test/lazy/test_block_diag_lazy_tensor.py
```diff
@@ -67,6 +67,31 @@ def evaluate_lazy_tensor(self, lazy_tensor):
                     actual[i, k, j * 4 : (j + 1) * 4, j * 4 : (j + 1) * 4] = blocks[i, k, j]
         return actual
 
+class TestBlockDiagLazyTensorMetaClass(unittest.TestCase):
+    def test_metaclass_constructor(self):
+        k, n = 3, 5 # number of blocks, block size
```
Balandat, line 72: This only tests this without any batch dimension. Could you also add a unit test that checks this for the batched case?

SebastianAment:
Thank you for your feedback! I incorporated it and am moving this PR [over here](https://github.com/cornellius-gp/linear_operator/pull/4).

PR #2111, opened 2022-08-30
Enable automatic versioning via setuptools_scm

gpleiss:
Based on https://github.com/cornellius-gp/linear_operator/pull/2


docs/source/conf.py
```diff
@@ -63,7 +65,7 @@ def find_version(*file_paths):
 author = "Cornellius GP"
 
 # The short X.Y version
-version = find_version("gpytorch", "__init__.py")
+version = find_version("..", "..", "gpytorch", "version.py")
```
Balandat, line 68: did you want to do this relative to the current dir by using `__file__` to make it robust to where this is run from?

Balandat (approved):

docs/source/conf.py
gpleiss: @Balandat  that happens on Line 32:

```python
def find_version(*file_paths):
     # ...
     try:
         with io.open(os.path.join(os.path.dirname(__file__), *file_paths), encoding="utf8") as fp:
             # ...
```

PR #2112, opened 2022-08-30
Fix LMCVariationalStrategy example in docs

adamjstewart:
The first argument to `VariationalStrategy` should be the model, in this case `self`.

The `latent_dim` argument to `LMCVariationalStrategy` must be negative.


gpleiss (approved):

gpleiss:
Thanks @adamjstewart !

Issue #1652, opened 2021-06-11
[Feature Request] Input Transformations


Balandat:
# 🚀 Feature Request

Essentially, upstream BoTorch's `InputTransformation` from https://github.com/pytorch/botorch/blob/master/botorch/models/transforms/input.py to GPyTorch. This allows to add either fixed or learnable transformations that are automatically applied when training models. 

## Motivation

This allows to do things like normalize inputs but also to combine the GP with a learnable transformation. This simplifies model setup. We currently have this in BoTorch and essentially apply the transform in the `forward` methods.

## Additional context
We recently worked on having input transformations that can change the shape of the input https://github.com/pytorch/botorch/pull/819, which caused some headaches for how to best set this up without a ton of boilerplate code. We were hoping to do this in the `__call__` rather than forward method, but this collides with some of GPyTorch's assumptions. Moving this functionality upstream into gpytorch would allow us to solve these challenges more organically.

**Describe alternatives you've considered**
You could do this as we do it right now, but one would have to add boilerplate transformation code to every implementation of `forward`.

Balandat:
cc @sdaulton @saitcakmak

saitcakmak:
I'll just add concrete examples of what we tried and where it fails.

Apply one-to-many transforms at `model.forward()`: For a `batch x q x d`-dim input these return `batch x new_q x m`-dim output. This fails at the reshape operation here: https://github.com/cornellius-gp/gpytorch/blob/a0d8cd2d379742fd2c72a22fe9fcc16e43b3d843/gpytorch/models/exact_prediction_strategies.py#L43

Define a wrapper around `ExactGP.__call__` and apply the transforms before calling `ExactGP.__call__`: Leads to https://github.com/cornellius-gp/gpytorch/blob/a0d8cd2d379742fd2c72a22fe9fcc16e43b3d843/gpytorch/models/exact_gp.py#L256
We could maybe get around this by wrapping the call with `with debug(False)`, but that also breaks tests, so probably not a good idea.

Current proposal at pytorch/botorch#819 is to apply the transforms at `model.forward` at the training time and at `posterior` call in `eval` mode to get around these issues (and do this at each models `forward` and each `posterior`). This is admittedly not a good design, so upstreaming the input transforms would make it much cleaner.

gpleiss:
@Balandat and @saitcakmak is this the interface you have in mind?

```python
class MyGP(gpytorch.models.ExactGP):
[...]

PR #2114, opened 2022-08-31
[RFC] Move input transforms to GPyTorch

saitcakmak:
This diff presents a minimal implementation of input transforms in GPyTorch, as requested in #1652. This should be viewed together with pytorch/botorch#1372. The input transforms themselves are currently implemented in https://github.com/pytorch/botorch/blob/cdd668d18b2a7e35bed09b7a2b2fca40e5fd2067/botorch/models/transforms/input.py

What this does:
* Moves the `transform_inputs` from BoTorch `Model` to GPyTorch `GP` class, with some modifications to explicitly identify whether given inputs are train or test inputs.
* Modifies the `InputTransform.forward` call to use `is_training_input` argument instead of `self.training` check to apply the transforms that have `transform_on_train=True`.
* Removes `preprocess_transform` method since this is no-longer needed.
* For `ExactGP` models, it transforms both train and test inputs in `__call__`. For `train_inputs` it always uses `is_training_input=True`. For generic `inputs`, it uses `is_training_input=self.training` which signals that these are training inputs when the model is in `train` mode, and that these are test inputs when the model is in `eval` mode.
* For `ApproximateGP` models, it applies the transform to `inputs` in `__call__` using `is_training_input=self.training`. This again signifies whether the given inputs are train or test inputs based on the mode of the model. Note that this NEVER transforms `inducing_points`, thus fixes the previous bug with `inducing_points` getting transformed in `train` but not getting transformed in `eval`. It is expected that the user will define inducing points in the appropriate space (mostly the normalized space / unit cube).
* For BoTorch `SingleTaskVariationalGP`, it moves the `input_transform` attribute down to `_SingleTaskVariationalGP`, which is the actual `ApproximateGP` instance. This makes the transform accessible from GPyTorch.

What this doesn't do:
* It doesn't do anything about `DeterministicModel`s. Those will still need to deal with their own transforms, which is not implemented here. If we make `Model` inherit from `GP`, we can keep the existing setup with very minimal changes.
* It does not clean up the call sites for `self.transform_inputs`. This is just made into a no-op and the clean-up is left for later.
* It does not upstream the abstract `InputTransform` classes to GPyTorch. That'll be done if we decide to go forward with this design.
* It does not touch `PairwiseGP`. `PairwiseGP` has some non-standard use of input transforms, so it needs an audit to make sure things still work fine.
* I didn't look into `ApproximateGP.fantasize`. This may need some changes similar to `ExactGP.get_fantasy_model`.
* It does not support `PyroGP` and `DeepGP`.

### DIFF
--- a/gpytorch/models/gp.py
+++ b/gpytorch/models/gp.py
@@ -1,7 +1,14 @@
 #!/usr/bin/env python3
 
+from torch import Tensor
+
 from ..module import Module
 
 
 class GP(Module):
-    pass
+    def apply_input_transforms(self, X: Tensor, is_training_input: bool) -> Tensor:
+        input_transform = getattr(self, "input_transform", None)
+        if input_transform is not None:
+            return input_transform(X=X, is_training_input=is_training_input)
+        else:
+            return X
--- a/gpytorch/models/approximate_gp.py
+++ b/gpytorch/models/approximate_gp.py
@@ -105,4 +105,5 @@ class ApproximateGP(GP, _PyroMixin):
     def __call__(self, inputs, prior=False, **kwargs):
         if inputs.dim() == 1:
             inputs = inputs.unsqueeze(-1)
+        inputs = self.apply_input_transforms(X=inputs, is_training_input=self.training)
         return self.variational_strategy(inputs, prior=prior, **kwargs)
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -210,7 +210,11 @@ class ExactGP(GP):
         except KeyError:
             fantasy_kwargs = {}
 
-        full_output = super(ExactGP, self).__call__(*full_inputs, **kwargs)
+        # Prediction strategy should have transformed train inputs.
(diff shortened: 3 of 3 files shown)


saitcakmak:
cc @wjmaddox, @gpleiss, @Balandat

gpleiss (commented):
This looks great to me, though we should add some unit tests and docs.

gpytorch/models/gp.py
```diff
@@ -1,7 +1,14 @@
 #!/usr/bin/env python3
 
+from torch import Tensor
+
 from ..module import Module
 
 
 class GP(Module):
-    pass
+    def apply_input_transforms(self, X: Tensor, is_training_input: bool) -> Tensor:
```
Balandat, line 9: Is there a reason to not name this arg `is_training` rather than `is_training_input`?

saitcakmak: Just being verbose and differentiating the model being in training and the inputs being the `train_inputs` or otherwise being treated as such.

PR #2125, opened 2022-09-07
Autogenerate parameter types in documentation from python typehints

gpleiss:
For example: if the method is...

```python
def forward(self, x1: Tensor, x2: Optional[Tensor]) -> Union[Tensor, LinearOperator]:
    r"""
    Does the forward thing.

    :param x1: The x1 arg
    :param x2: The x2 arg
    :return: The forward stuff.
    """
    # ...
```

The resulting documentation will include the appropriate types:

```
Parameters:
  - x1 (torch.Tensor) - The x1 arg
  - x2 (torch.Tensor, optional) - The x2 arg

Return type: torch.Tensor or LinearOperator

Returns: The forward stuff.
```

This (hopefully) should prevent a lot of duplicate effort on our end.

This PR also refactors the gpytorch.kernels.Kernel and gpytorch.distributions.MultivariateNormal docs to utilize auto-typing.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,6 +4,7 @@ ipython
 ipykernel
 sphinx
 sphinx_rtd_theme
+sphinx_autodoc_typehints
 nbsphinx
 m2r2
 pyro-ppl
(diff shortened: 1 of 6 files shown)


Balandat (approved):
can't say I reviewed in detail, but generally lgtm!

PR #2131, opened 2022-09-08
Include linear operator in installation instructions

saitcakmak:
Since latest features may rely on latest features in linear operator.


gpleiss:
Good call. Thanks @saitcakmak !

PR #2132, opened 2022-09-09
fix custom dtype_value_context setting

sdaulton:
Previously, setting custom values in `dtype_value_context` was broken because `dtype_value_context.value` requires a `dtype` argument and no `dtype` argument was provided in the calls in `dtype_value_context.__init__`.


Balandat (approved):
Oh wow this seems like a pretty bad bug. Thanks for adding the tests!

PR #2136, opened 2022-09-13
Expose jitter

hughsalimbeni:
* Exposes jitter in variational_strategy
* Added a test that checks variational vs exact gp with high precision comparison. This is not possible without setting jitter to a very small value

One small issue is that the previous behaviour had both `1e-3` and `1e-4` added in various places. I've only found one test that breaks with 1e-4 for both, so I've set this one to 1e-3 via the new exposed `jitter_val` arg. 

Happy to do the other variational strategies, and move the defaults to the settings.


gpytorch/variational/variational_strategy.py
```diff
@@ -12,18 +12,19 @@
     SumLinearOperator,
     TriangularLinearOperator,
 )
-from linear_operator.utils.cholesky import psd_safe_cholesky
```
jacobrgardner, line 15: I think we keep this import instead of the changed one -- all of the linear algebra internals (gpytorch.utils, gpytorch.lazy) were separated out into their own package.

jacobrgardner (approved):
lgtm other than the lint errors and the import question!

gpytorch/variational/variational_strategy.py
```diff
@@ -12,18 +12,19 @@
...
 
 from ..distributions import MultivariateNormal
 from ..settings import _linalg_dtype_cholesky, trace_mode
-from ..utils.errors import CachingError
+from ..utils.cholesky import psd_safe_cholesky
+from ..utils.errors import CachingError, NotPSDError
 from ..utils.memoize import cached, clear_cache_hook, pop_from_cache_ignore_args
 from ..utils.warnings import OldVersionWarning
 
+# from ._variational_strategy import _VariationalStrategy
```
Balandat, line 26: what's up with this?

gpytorch/variational/variational_strategy.py
```diff
@@ -167,7 +170,7 @@ def forward(self, x, inducing_points, inducing_values, variational_inducing_cova
         # Covariance terms
         num_induc = inducing_points.size(-2)
         test_mean = full_output.mean[..., num_induc:]
-        induc_induc_covar = full_covar[..., :num_induc, :num_induc].add_jitter()
+        induc_induc_covar = full_covar[..., :num_induc, :num_induc].add_jitter(self.jitter_val) # 1as 1e-4
```
Balandat, line 173: what is this comment here?

hughsalimbeni: Removed

gpytorch/variational/variational_strategy.py
hughsalimbeni: Reverted

gpytorch/variational/variational_strategy.py
```diff
@@ -73,11 +73,13 @@ class VariationalStrategy(_VariationalStrategy):
         https://www.repository.cam.ac.uk/handle/1810/278022
     """
 
-    def __init__(self, model, inducing_points, variational_distribution, learn_inducing_locations=True):
+    def __init__(
+        self, model, inducing_points, variational_distribution, learn_inducing_locations=True, jitter_val=1e-4
```
gpleiss, line 77:
```suggestion
        self, model, inducing_points, variational_distribution, learn_inducing_locations=True, jitter_val=None
```

gpytorch/variational/variational_strategy.py
```diff
@@ -73,11 +73,13 @@ class VariationalStrategy(_VariationalStrategy):
...
+    ):
         super().__init__(model, inducing_points, variational_distribution, learn_inducing_locations)
         self.register_buffer("updated_strategy", torch.tensor(True))
         self._register_load_state_dict_pre_hook(_ensure_updated_strategy_flag_set)
-
+        self.jitter_val = jitter_val
```
gpleiss, line 82:
```suggestion
        if jitter_val is None:
            self.jitter_val = cholesky_jitter.value()
        else:
            self.jitter_val = jitter_val
```

gpleiss: This way we can default to the setting that we use for exact GPs, which the user can specify with a context manager.

(We'll also have to modify line 22 to import `cholesky_jitter`:

```python
from ..settings import _linalg_dtype_cholesky, cholesky_jitter, trace_mode
```

I can't seem to do this through "suggested changes" though... ☹️)

hughsalimbeni:
In the last commit I found one more in the grid interpolation strategy, but this time it breaks with anything smaller than 1e-3, so I've left explicit for now. 

The default is now changed from 1e-3 to 1e-6 (or 1e-8 for float64), which is quite a large change. Might this cause problems for some users?

gpleiss:
> The default is now changed from 1e-3 to 1e-6 (or 1e-8 for float64), which is quite a large change. Might this cause problems for some users?

I imagine this should be okay - we had 1e-3 or 1e-4 in our ExactGP code until we moved to the context manager, and (AFAIK) it hasn't introduced too many numerical issues. We can always tighten this (or make a separate setting for variational models) if we see trouble down the line).

Thanks @hughsalimbeni !

hughsalimbeni:
Odd that the `test_decoupled_svgp_regression` test failed as it was fine on my machine. I've changed to `1e-4`. Hope it is ok now

jacobrgardner:
Interesting also that it's failing against pytorch master only, but not pytorch's current stable release. Maybe 1e-6 is problematically small for variational GPs in fp32 and we should make a `settiings.variational_cholesky_jitter` with a default of 1e-4? For what it's worth, I've empirically noticed that K_uu matrices can be pretty artificially numerically sensitive with learned inducing locations, since optimization can sometimes put a few inducing points essentially directly on top of each other.

PR #2141, opened 2022-09-19
slight improvements to IndexKernel for the rank = 0 (diagonal) case

rmgarnett:
This PR tweaks `IndexKernel` to have slightly better performance/behavior in the corner case of `rank  = 0`, where the underlying covariance matrix is diagonal. Although this case is atypical from the point of view of multitask GPs, I have found this construction useful as a component of larger covariance structures.

Two changes are proposed:
- the kernel only registers the `covar_factor` parameter (the low-rank component when `rank > 0`) when `rank > 0`
- in the `rank == 0` case, the `covar_matrix()` method returns a `DiagLinearOpeator` rather than a `PsdSumLinearOperator(RootLinearOperator, DiagLinearOperator)`

### DIFF
--- a/gpytorch/kernels/index_kernel.py
+++ b/gpytorch/kernels/index_kernel.py
@@ -65,9 +65,11 @@ class IndexKernel(Kernel):
         if var_constraint is None:
             var_constraint = Positive()
 
-        self.register_parameter(
-            name="covar_factor", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks, rank))
-        )
+        self.rank = rank
+        if self.rank > 0:
+            self.register_parameter(
+                name="covar_factor", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks, self.rank))
+            )
         self.register_parameter(name="raw_var", parameter=torch.nn.Parameter(torch.randn(*self.batch_shape, num_tasks)))
         if prior is not None:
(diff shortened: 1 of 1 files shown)


Balandat (commented):
In principle it seems reasonable enough to me to support this. 

However, looking at `forward` and `_eval_covar_matrix` the code as is will cause some issues. Namely in `_eval_covar_matrix` (called from `forward`) we will get an `AttributeError` when we try this from `self.covar_factor`. Which means that we should also add a unit test case to cover the `rank==0` behavior. 

Taking a step back, if we handled that correctly, we probably shouldn't return an `InterpolatedLinearOperator` in `forward` - or at least we'd want to special case `InterpolatedLinearOperator` to handle a diagonal `base_linear_op` (it doesn't seem that this is currently the case). 


edge

rmgarnett:
I've added a branch to `_eval_covar_matrix()` to handle the diagonal case to partially address @Balandat's comments.

gpleiss (approved):
LGTM, but I'll let @Balandat make the final approval.

Balandat (approved):
lgtm. I think at least some of the special casing of using a `DiagLinearOperator` is actually happening inside `InterpolatedLinearOperator` so we'll probably get some gains there already.

Balandat:
I guess I would still like to see this covered by a unit test though :)

PR #2163, opened 2022-10-19
Migrate trace_mode to linear_operators and disable conditional control flow

feynmanliang:
We would like to [use gpytorch in our NUTS sampler](https://github.com/facebookresearch/beanmachine/blob/main/tests/ppl/experimental/gp/inference_test.py#L61-L70). While the linked code works, it does not utilize JIT compilation (`use_nnc`) because of control flow.

I understand `trace_mode` was originally used to disable custom pytorch functions back when JIT did not support compiling through them. This PR (along with https://github.com/cornellius-gp/linear_operator/pull/27) revives the setting for the AoT JIT compiler (https://pytorch.org/functorch/nightly/aot_autograd.html).

 * Moves `settings.trace_mode` from `gpytorch` to `linear_operators`
 * Skips conditional control flow when `trace_mode` is true, currently:
   * PSD check / jitter in `linear_operators.utils.cholesky`
   * Diagonal `x1_eq_x2` check in `gpytorch.kernels.kernel`

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -319,7 +319,10 @@ class Kernel(Module):
             x1 = x1.transpose(-1, -2).unsqueeze(-1)
             x2 = x2.transpose(-1, -2).unsqueeze(-1)
 
-        x1_eq_x2 = torch.equal(x1, x2)
+        if settings.trace_mode.on():
+            x1_eq_x2 = False
+        else:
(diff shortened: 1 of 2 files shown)


gpleiss:
We'll put this PR on hold until https://github.com/cornellius-gp/linear_operator/pull/27 is merged in and we make a new LinearOperator release (this will happen soon).

PR #2140, opened 2022-09-16
Avoid evaluating kernel when adding jitter

dannyfriar:
The memory usage is much higher as a result of evaluating the kernel on this line.

Is there any reason for not using the parent class method to do this?

PR #2185, opened 2022-11-03
Avoid evaluating kernel in `expand_batch`

dannyfriar:
This is a very similar case to #2140.

Not necessarily expecting this to be merged as is but just curious to see if anything breaks with this tiny change. This also results in materialising the NxN matrix in some cases.


gpleiss (approved):
LGTM. I'll wait till the `add_jitter` PR is merged before merging this.

dannyfriar:
@gpleiss what is the expected release cadence for GPyTorch? I just wanted to get an idea of when I can expect this to be available in pypi.

Thanks!

PR #2189, opened 2022-11-08
Avoid evaluating kernel when adding jitter

### LINKED ISSUE #1826: `LazyEvaluatedKernelTensor.add_jitter` evaluates the kernel
The `add_jitter` method of the `LazyEvaluatedKernelTensor` class explicitly evaluates the kernel: https://github.com/cornellius-gp/gpytorch/blob/c58dc3a77c7be7ee4434484e359a0bbc1c2d27f6/gpytorch/lazy/lazy_evaluated_kernel_tensor.py#L237

Is there a particular reason for doing so? Adding the jitter in a lazy way as implemented by the parent class `LazyTensor` ([here](https://github.com/cornellius-gp/gpytorch/blob/c58dc3a77c7be7ee4434484e359a0bbc1c2d27f6/gpytorch/lazy/lazy_tensor.py#L730)) would drastically decrease the memory consumption.

### LINKED ISSUE #1951: Improving memory requirements for latent function inference (low-level Pyro-GPyTorch integration)
I have been working on a program performing latent function inference via the low-level Pyro-GPyTorch integration. The code is really similar to the example available in the [documentation](https://docs.gpytorch.ai/en/v1.5.1/examples/07_Pyro_Integration/Pyro_GPyTorch_Low_Level.html) but it is supposed to train on hundreds of thousands of samples.

During training, the memory requirements increase quadratically with the number of training data used. This has been pinned down to the `forward` method of the `VariationalStrategy` class. As pointed out in #1826, the `add_jitter` method of the `LazyEvaluatedKernelTensor` class evaluates the covariance matrix on the training samples and thus allocates the corresponding memory.

It seems to me that the full matrix does not need to be evaluated. Indeed, only the `mean` and the `stddev` of the variational distribution are used:
https://github.com/cornellius-gp/gpytorch/blob/aa0ff73cbd065a3a29b893c89bdab5497da63298/gpytorch/models/pyro/_pyro_mixin.py#L16-L17

As the output of the variational strategy is a lazy distribution, the latter involves only the diagonal of the covariance matrix:

https://github.com/cornellius-gp/gpytorch/blob/aa0ff73cbd065a3a29b893c89bdab5497da63298/gpytorch/distributions/multivariate_normal.py#L249

- Modules that will be modified:

In [`gpytorch.lazy.lazy_evaluated_kernel_tensor.py`](https://github.com/cornellius-gp/gpytorch/blob/c58dc3a77c7be7ee4434484e359a0bbc1c2d27f6/gpytorch/lazy/lazy_evaluated_kernel_tensor.py#L236-L237) use the lazy implementation of `add_jitter` from the `LazyTensor` class:

```diff
    def add_jitter(self, jitter_val=1e-3):
-        return self.evaluate_kernel().add_jitter(jitter_val)
+        return super(LazyEvaluatedKernelTensor, self).add_jitter(jitter_val)
```

- Impact on speed:

The spatial complexity of the algorithm drops from being quadratic to linear w.r.t. the
number of training data. Benefits are visible also in terms of throughput. Using the [example in the documentation](https://docs.gpytorch.ai/en/v1.5.1/examples/07_Pyro_Integration/Pyro_GPyTorch_Low_Level.html) as reference:

time_mem

The above plots report the time and memory required to perform 10 training steps of a  GP model on an increasing number of training data (`NSamp`).  Details and code available [here](https://github.com/pierocor/gpytorch_test).

gpleiss:
This significantly reduces memory requirements for variational models.

This PR requires adding an expand method to kernels, which expands the size of the parameters to match specified batch dimensions.
This makes it possible to call Kernel.__getitem__ with broadcasting indices that may introduce new batch dimensions.

This PR also requires bumping the linear_operator version from 0.1.1 to 0.2.0

[Fixes #1826]
[Fixes #1951]


environment.yml
```diff
@@ -5,4 +5,4 @@ channels:
 dependencies:
   - pytorch
   - scikit-learn
-  - linear_operator>0.1.1
+  - linear_operator>0.2.0
```
Balandat, line 8: Did you mean
```suggestion
  - linear_operator>=0.2.0
```

gpytorch/kernels/kernel.py
```diff
@@ -352,6 +352,49 @@ def covar_dist(
...
+    def expand(self, *sizes: Union[torch.Size, Tuple[int, ...]]) -> Kernel:
+        r"""
+        Constructs a new kernel where the lengthscale (and other kernel parameters)
+        are expanded to match the batch dimension determined by `sizes`.
+
+        :param sizes: The batch shape of the new tensor
+        """
+        # Type checking
+        if len(sizes) == 1 and hasattr(sizes, "__iter__"):
+            new_batch_shape = torch.Size([*(sizes[0])])
```
Balandat, line 364: Does this not work? If you're checking the `len` then this wouldn't work with generators anyway, no? Or is this some hack around type checking weirdness?
```suggestion
            new_batch_shape = torch.Size(sizes[0])
```

gpytorch/kernels/kernel.py
```diff
@@ -352,6 +352,49 @@ def covar_dist(
 
         return res
 
+    def expand(self, *sizes: Union[torch.Size, Tuple[int, ...]]) -> Kernel:
```
Balandat, line 355: I know this is the abstract kernel and doesn't have non-batch shapes, but should we still call this `expand_batch` to be unambiguous?

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -124,6 +124,18 @@ def _getitem(self, row_index, col_index, *batch_indices):
             num_outs_per_in_rows = num_outs_per_in
             num_outs_per_in_cols = num_outs_per_in
 
+        # We will be running the __getitem__ command on x1, x2, and the kernel parameters
+        # Since kernels can broadcast, x1, x2, and the kernel parameters may not have all of the batch dimensions
+        #   that are being indexed by the __getitem__ operation
```
Balandat, line 129:
```suggestion
        # that are being indexed by the __getitem__ operation
```

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -176,36 +188,28 @@ def _getitem(self, row_index, col_index, *batch_indices):
...
-            x1 = x1.expand(*([1] * (len(batch_indices) - self.x1.dim() + 2)), *self.x1.shape)
+            x1 = x1.expand(*batch_shape, *x1.shape[-2:])
             x1 = x1[(*batch_indices, row_index, dim_index)]
 
         # Call x2[*batch_indices, col_index, :]
         try:
             x2 = x2[(*batch_indices, col_index, dim_index)]
         # We're going to handle multi-batch indexing with a try-catch loop
-        # This way - in the default case, we can avoid doing expansions of x1 which can be timely
+        # This way - in the default case, we can avoid doing expansions of x2 which can be timely
```
Balandat, line 198: What does "timely" mean in this context?

gpytorch/kernels/kernel.py
gpleiss: That's a good idea. I'll update the PR.

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
Balandat: Hmm I figure it's correct after thinking through it, but my brain always wants to read "costly in terms of time" rather than "timely". Maybe it's just my ESL...

Balandat:
@gpleiss can I ask you to hold off merging this in for a bit? I'm seeing some downstream failure in botorch from this, will have to take a closer look at that one.

PR #2204, opened 2022-11-28
Improving `_sq_dist` when `x1_eq_x2`

SebastianAment:
This commit moves the `adjustment` of `x2` into the `else` branch if `x1_eq_x2 == True`, since it is not necessary in this case.


SebastianAment:
The doc build failure is caused by `ipython3` not being found. Seems to be an unrelated issue.

PR #2205, opened 2022-11-29
Deprecating `postprocess`

SebastianAment:
The `postprocess` logic in `Distance` and `Kernel` adds a lot of lines of code, even though it is only used by three call sites (`RBFKernel`, `RBFKernelGrad`, and `RQKernel`) and can be very efficiently put at those call sites. This commit mostly removes the `postprocess` arguments, but keeps enough (e.g. the `Distance` class) to support legacy code, and adds `_sq_dist` and `_dist` as module level functions. If we decide to go fully deprecate `Distance` as well, a further 17 lines could be removed.

### DIFF
--- a/gpytorch/kernels/kernel.py
+++ b/gpytorch/kernels/kernel.py
@@ -23,44 +23,55 @@ from ..module import Module
 from ..priors import Prior
 
 
-def default_postprocess_script(x):
-    return x
+def sq_dist(x1, x2, x1_eq_x2=False):
+    # TODO: use torch squared cdist once implemented: https://github.com/pytorch/pytorch/pull/25799
+    adjustment = x1.mean(-2, keepdim=True)
+    x1 = x1 - adjustment
+
+    # Compute squared distance matrix using quadratic expansion
+    x1_norm = x1.pow(2).sum(dim=-1, keepdim=True)
+    x1_pad = torch.ones_like(x1_norm)
+    if x1_eq_x2 and not x1.requires_grad and not x2.requires_grad:
+        x2, x2_norm, x2_pad = x1, x1_norm, x1_pad
+    else:
+        x2 = x2 - adjustment  # x1 and x2 should be identical in all dims except -2 at this point
+        x2_norm = x2.pow(2).sum(dim=-1, keepdim=True)
+        x2_pad = torch.ones_like(x2_norm)
+    x1_ = torch.cat([-2.0 * x1, x1_norm, x1_pad], dim=-1)
+    x2_ = torch.cat([x2, x2_pad, x2_norm], dim=-1)
+    res = x1_.matmul(x2_.transpose(-2, -1))
+
+    if x1_eq_x2 and not x1.requires_grad and not x2.requires_grad:
+        res.diagonal(dim1=-2, dim2=-1).fill_(0)
+
+    # Zero out negative values
+    return res.clamp_min_(0)
 
 
+def dist(x1, x2, x1_eq_x2=False):
+    # TODO: use torch cdist once implementation is improved: https://github.com/pytorch/pytorch/pull/25799
+    res = sq_dist(x1, x2, x1_eq_x2=x1_eq_x2)
+    return res.clamp_min_(1e-30).sqrt_()
+
+
+# only necessary for legacy purposes
 class Distance(torch.nn.Module):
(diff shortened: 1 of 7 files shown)


gpytorch/kernels/kernel.py
```diff
@@ -320,37 +320,17 @@ def covar_dist(
...
-            self.distance_module = Distance(dist_postprocess_func)
-
         if diag:
             # Special case the diagonal because we can return all zeros most of the time.
             if x1_eq_x2:
-                res = torch.zeros(*x1.shape[:-2], x1.shape[-2], dtype=x1.dtype, device=x1.device)
-                if postprocess:
-                    res = dist_postprocess_func(res)
-                return res
+                return torch.zeros(*x1.shape[:-2], x1.shape[-2], dtype=x1.dtype, device=x1.device)
```
SebastianAment, line 328: In terms of complexity reduction, this block of code in `covar_dist` benefits most.

gpytorch/kernels/kernel.py
```diff
@@ -23,44 +23,51 @@
 from ..priors import Prior
 
 
-def default_postprocess_script(x):
-    return x
-
-
+def _sq_dist(x1, x2, x1_eq_x2=False):
```
SebastianAment, line 26: I saw two examples of these functions being used in BoTorch, so we could have these as public as well. I don't know to what extent other users are relying on `Distance`, but we could remove it from this commit without an internal change in the functionality, and the required changes to BoTorch would be minimal.

gpytorch/kernels/kernel.py
```diff
@@ -23,44 +23,51 @@
...
-        if x1_eq_x2 and not x1.requires_grad and not x2.requires_grad:
-            x2_norm, x2_pad = x1_norm, x1_pad
-        else:
-            x2_norm = x2.pow(2).sum(dim=-1, keepdim=True)
-            x2_pad = torch.ones_like(x2_norm)
-        x1_ = torch.cat([-2.0 * x1, x1_norm, x1_pad], dim=-1)
-        x2_ = torch.cat([x2, x2_pad, x2_norm], dim=-1)
-        res = x1_.matmul(x2_.transpose(-2, -1))
+        if postprocess is not None:
+            warnings.warn("The `postprocess` argument will be deprecated.", DeprecationWarning)
```
Balandat, line 62: :)
```suggestion
            warnings.warn("The `postprocess` argument is deprecated.", DeprecationWarning)
```

Balandat: Could add a link to this PR here for people to understand why.

gpytorch/kernels/kernel.py
```diff
@@ -320,37 +320,17 @@ def covar_dist(
...
-
-        elif square_dist:
-            res = self.distance_module._sq_dist(x1, x2, postprocess, x1_eq_x2)
+                res = torch.linalg.norm(x1 - x2, dim=-1)  # 2-norm by default
+                return res.pow(2) if square_dist else res
         else:
-            res = self.distance_module._dist(x1, x2, postprocess, x1_eq_x2)
-
-        return res
+            return (_sq_dist if square_dist else _dist)(x1, x2, x1_eq_x2)
```
Balandat, line 333: I find this kind of hard to read, but that might be personal taste
```suggestion
            distfunc = _sq_dist if square_dist else _dist
            return distfunc(x1, x2, x1_eq_x2)
```

Balandat (approved):
In general this makes a lot of sense to me, but I'd like to also hear @gpleiss's or @jacobrgardner 's take on this change.

gpytorch/kernels/kernel.py
SebastianAment: Done.

gpytorch/kernels/kernel.py
```diff
@@ -148,15 +159,15 @@ class Kernel(Module):
     def __init__(
         self,
         ard_num_dims: Optional[int] = None,
-        batch_shape: Optional[torch.Size] = torch.Size([]),
+        batch_shape: Optional[torch.Size] = None,
```
SebastianAment, line 162: Making `flake8` happy.

gpleiss (approved):

gpleiss:
I'm very much for this! LGTM

PR #2229, opened 2022-12-15
Let `LazyEvaluatedKernelTensor` recall the grad state at instantiation

SebastianAment:
If a `LazyEvaluatedKernelTensor` is created in a `no_grad` context, the grad state is not propagated to the point of tensor evaluation, which can lead to counter-intuitive behavior. See the following:
```python
import torch
import gpytorch

d = 3
b = 2
X = torch.randn(b, d)
X.requires_grad = True
with torch.no_grad():
    k = RBFKernel()
    K = k(X)
M = K.to_dense()
print(M.requires_grad)  # this is true on master because the evaluation of k(X) is delayed 
```
This commit makes `LazyEvaluatedKernelTensor` recall the Torch's grad state upon instantiation using the newly introduced `recall_grad_state` decorator. As a consequence, the last line of the previous code snippet returns `requires_grad = False`, as expected.

### DIFF
--- a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python3
 
+import functools
+from typing import Callable
+
 import torch
 from linear_operator import LinearOperator, to_linear_operator
 from linear_operator.utils.getitem import _noop_index
@@ -7,6 +10,22 @@ from linear_operator.utils.getitem import _noop_index
 from .. import beta_features, settings
 from ..utils import deprecation
 from ..utils.memoize import cached
+
+
+def recall_grad_state(method: Callable) -> Callable:
+    """Decorator for LazyEvaluatedKernelTensor's methods to put their execution
+    inside the same grad state as during the instantiation of the lazy object.
+    This makes the lazy tensor object behave in the same way as a regular tensor
+    with respect to the grad state.
+    """
+
+    @functools.wraps(method)
+    def wrapped(self, *args, **kwargs):
+        with torch.set_grad_enabled(self._is_grad_enabled):
+            output = method(self, *args, **kwargs)
+        return output
+
+    return wrapped
 
 
 class LazyEvaluatedKernelTensor(LinearOperator):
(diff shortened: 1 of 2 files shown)


gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -9,6 +12,18 @@
 from ..utils.memoize import cached
 
 
+def recall_grad_state(method: Callable) -> Callable:
+    @functools.wraps(method)
```
Balandat, line 16: maybe add a docstring here?

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -9,6 +12,18 @@
...
+    def wrapped(self, *args, **kwargs):
+        prev = torch.is_grad_enabled()
+        torch.set_grad_enabled(self._is_grad_enabled)
+        output = method(self, *args, **kwargs)
+        torch.set_grad_enabled(prev)
```
Balandat, line 21: How about 
```suggestion
        with torch.set_grad_enabled(self._is_grad_enabled):
            output = method(self, *args, **kwargs)
```

Balandat (approved):
This looks great to me. @gpleiss, @jacobrgardner any concerns with this?

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
SebastianAment: Clean!

test/lazy/test_lazy_evaluated_kernel_tensor.py
```diff
@@ -159,6 +159,16 @@ def test_half(self):
         lazy_tensor.kernel.raw_lengthscale_constraint.transform = lambda x: x + 0.1
         self._test_half(lazy_tensor)
 
+    def test_grad_state(self):
```
SebastianAment, line 162: Added a test case to track this behavior.

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
Balandat: download

SebastianAment:
The `run_unit_tests (master, with-extras)` workflow has been in "expected" status for a long time. Does that need separate approval to run?

Balandat:
I don't think so? maybe som infra issue? I restarted the test jobs, let's see if that resolves it.

Balandat:
I can’t really see why this PR would cause this one test to not report so let me merge it in.

Issue #1409, opened 2020-12-31
[Question] What is the difference between "model(test_x)" and "likelihood(model(test_x))" for prediction? 


mohammad-saber:
Hi, 
I am trying regression with GPyTorch following the example in the [documentation](https://docs.gpytorch.ai/en/stable/examples/01_Exact_GPs/Simple_GP_Regression.html). 

In the example, there are two ways to get mean and variance of prediction. 

1)

`f_preds = model(test_x)`

2)

`y_preds = likelihood(model(test_x))`

I noticed that the mean of prediction is the same when I use `model(test_x)` and `likelihood(model(test_x))`, but variance is different. I didn't understand the reason. 
More generally, I didn't understand the difference between `model(test_x)` and `likelihood(model(test_x))`. 

Could you please guide me on this issue?

jacobrgardner:
In the regression setting, the former does not include the likelihood noise (it is the latent variance) and the latter does include it.

mohammad-saber:
Thank you. 

Is likelihood noise similar to the noise that we can add to the kernel?  For example, when I define a kernel as below:

RBFKernel() + WhiteNoiseKernel()

I checked the list of kernels. I couldn't find "WhiteNoiseKernel()". It seems that GPyTorch adds noise automatically to the kernel. Is my understanding correct?

gpleiss:
Philosophically, it is different (is the "noise" due to your data observations or due to your uncertainty about your model). Practically, it is the same.

`WhiteNoiseKernel` has been replaced by [FixedNoiseGaussianLikelihood](https://docs.gpytorch.ai/en/stable/likelihoods.html#fixednoisegaussianlikelihood). This adds different observational noise for each data point. The standard GaussianLikelihood adds the same observational noise to each data point.

ginward:
> In the regression setting, the former does not include the likelihood noise (it is the latent variance) and the latter does include it.

Will the prediction performance be different between `model(test_x) `and `likelihood(model(test_x))`? 
For example, will `likelihood(model(test_x)) `be better?

gpleiss:
> Will the prediction performance be different between model(test_x) and likelihood(model(test_x))?
For example, will likelihood(model(test_x)) be better?

Let `X, y` be the training data, and let `x*, y*` be a test point.

[...]

PR #2230, opened 2022-12-16
Doc Update for Posterior Model Distribution and Posterior Predictive Distribution

varunagrawal:
As per the discussion in #1409.


examples/01_Exact_GPs/Simple_GP_Regression.ipynb
```diff
@@ -274,7 +274,21 @@
...
+    "\n",
+    "In contrast, `likelihood(model(test_x))` gives us the posterior predictive distribution `p(y* | x*, X, y)` which is the probability distribution over the predicted output value. This includes the _likelihood noise_ which is the noise in your observation (e.g. due to noisy sensor). Recall in our problem setup \n",
+    "\n",
+    "$$\n",
+    "\\begin{align}\n",
+    "y &= \\sin(2\\pi x) + \\epsilon\n",
+    "\\end{align}\n",
+    "$$\n",
+    "\n",
+    "where 𝜖 is the likelihood noise for each observation. Since the posterior predictive distribution includes the likelihood noise, it will provide better prediction performance.\n",
```
Balandat, line 289: > Since the posterior predictive distribution includes the likelihood noise, it will provide better prediction performance

This seems like an odd phrasing here - what's "better"? I think we're just providing predictions for two different things - over the latent function value vs. over the observations. Maybe we can rephrase this here / be more precise about this?

varunagrawal: @Balandat I was paraphrasing @gpleiss. I'll update this.

Balandat (approved):
Thanks for making this clearer!

Balandat:
@gpleiss I keep running into the "run_unit_tests (master, with-extras)" workflow to report the following status without ever resolving: "Expected — Waiting for status to be reported". Any ideas what's going on there? Force-merging this one for now.

PR #2234, opened 2022-12-23
Add `device` property to `Kernel`s, add unit tests

Balandat:
This is kinda useful. At times. Maybe.


gpytorch/kernels/kernel.py
```diff
@@ -261,6 +261,15 @@ def batch_shape(self) -> torch.Size:
     def batch_shape(self, val: torch.Size):
         self._batch_shape = val
 
+    @property
+    def device(self) -> Optional[torch.device]:
+        if self.has_lengthscale:
+            return self.lengthscale.device
+        else:
+            for param in self.parameters():
+                return param.device
```
esantorella, line 270: - What if `self.lengthscale` and parameters are on different devices? Can they be? Can other tensor attributes of a Kernel be on different devices?
- How about `self.active_dims` (when not None), since all Kernels should have that?

esantorella (approved):

esantorella:
What's the use case for this? It seems potentially misleading without checks that all of the kernel's attribute tensors are on the same device. Maybe that's why there's no `torch.nn.Module.device`. Or if that isn't an issue, maybe this should be a method of the superclass (gpytorch Module)?

gpytorch/kernels/kernel.py
Balandat: Good point - this is modeled after the code in the `dtype` property below. I guess the standing assumption has been here that all the parameters (of which lenghtscale is one) are on the same device / have the same dtype. I can simplify this and raise an error if this assumption is violated.

Balandat:
The use case is basically if we need to conveniently get "the" device/dtype of a kernel object (e.g. for creating new tensors in computations in other places). 

> It seems potentially misleading without checks that all of the kernel's attribute tensors are on the same device

That is true, I can raise a clear error if this is called on something that has heterogeneous devices/dtypes


> Maybe that's why there's no torch.nn.Module.device. Or if that isn't an issue, maybe this should be a method of the superclass (gpytorch Module)?

So since modules can nest other modules it's true that generally we can't be sure (neither should we) that everything lives on the same device or has the same dtype. I think so far this has largely be the implicit assumption though, so let me raise the errors as mentioned above. I think the discussion as to whether this should live on the general module level is one we can have separately from this specific PR.

gpytorch/kernels/kernel.py
Balandat: I retained the behavior if `has_lengthscale = True` for now to not cause any unforeseen BC issues. But I'm raising an error if parameters / devices are heterogeneous between the parameters.

PR #2235, opened 2022-12-24
Mean and kernel functions for first and second derivatives

ankushaggarwal:
Added new classes to work with derivatives: 

1. Second derivative RBF kernel called RBFKernelGradGrad (without mixed second derivatives)
2. Three new mean functions:
  a. ConstantMeanGradGrad
  b. LinearMeanGrad
  c. LinearMeanGradGrad


gpytorch/kernels/rbf_kernel_gradgrad.py
```diff
@@ -0,0 +1,157 @@
...
+
+            # Apply a perfect shuffle permutation to match the MutiTask ordering
+            pi1 = torch.arange(n1 * (2*d + 1)).view(2*d + 1, n1).t().reshape((n1 * (2*d + 1)))
+            pi2 = torch.arange(n2 * (2*d + 1)).view(2*d + 1, n2).t().reshape((n2 * (2*d + 1)))
+            K = K[..., pi1, :][..., :, pi2]
+
+            return K
+
+        else:
+            raise RuntimeError("diag=True not implemented yet")
```
gpleiss, line 143: This needs to be implemented, otherwise you could get errors when e.g. preconditioning.

gpleiss (changes requested):
This mostly looks good! Diag should be implemented though.

There are linting errors - can you make sure that you pass the precommit hooks (see our contribution guidelines in the readme).

Additionally, please make sure that you add this kernel and these means to the documentation, and be sure that the docs compile for you locally (without warnings).

Thanks @ankushaggarwal !
Thanks @

ankushaggarwal:
Thanks @gpleiss I have made hook fixes using pre-commit and added the new kernel and means to the doc files. However I cannot compile them locally (I am getting different errors, such as version.py not found etc.

I will look at implementing the diagonal option.

ankushaggarwal:
I found an issue, as I had been working with an older version of gpytorch that used lazy tensor with which the gpytorch.kernels.RBFKernelGradGrad worked correctly. I changed it to KroneckerProductLinearOperator as per the new version (commit [9bf9c3b](https://github.com/cornellius-gp/gpytorch/pull/2235/commits/9bf9c3ba1f57181122239503490731c3752e5b85)) and it gives wrong result. Any idea why this could be the case?

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
```
Balandat, line 1: Would you mind splitting out the hook update into a separate PR? It's good practice and easier to figure out if things go wrong if this is not intermingled with feature additions.

ankushaggarwal: @Balandat I am not too familiar with the hook updates. Do you mean the changes in .pre-commit-config.yaml or changes to all the files by pre-commit (in commit [c18e6d8](https://github.com/cornellius-gp/gpytorch/pull/2235/commits/c18e6d8e13356c9f9b17bf5e03e3877938bb6c07))

gpytorch/kernels/rbf_kernel_gradgrad.py
ankushaggarwal: I have implemented it

ankushaggarwal (commented):
I have implemented the diag option for rbf_kernel_gradgrad

.pre-commit-config.yaml
gpleiss: I'm not sure why we need to change the `pre-commit-hooks` file at all. Can we get rid of these changes?

gpytorch/kernels/rbf_kernel_gradgrad.py
```diff
@@ -0,0 +1,170 @@
...
+    and :math:`\mathbf{x_2}`.
+
+    See :class:`gpytorch.kernels.Kernel` for descriptions of the lengthscale options.
+
+    .. note::
+
+        This kernel does not have an `outputscale` parameter. To add a scaling parameter,
+        decorate this kernel with a :class:`gpytorch.kernels.ScaleKernel`.
+
+    Args:
```
gpleiss, line 22: Can you update the doc string to be in the standard sphinx format? (See [the Kernel base class]( https://github.com/cornellius-gp/gpytorch/blob/a2b5fd8c1f164001e8ce018eddc0b86d332cacf6/gpytorch/kernels/kernel.py#L79) for an example)

gpytorch/means/constant_mean_gradgrad.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python3
+
+import torch
+
+from .mean import Mean
+
+
+class ConstantMeanGradGrad(Mean):
+    def __init__(self, prior=None, batch_shape=torch.Size(), **kwargs):
```
gpleiss, line 9: Please add a doc string and type hints.

gpytorch/means/linear_mean_grad.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+
+import torch
+
+from .mean import Mean
+
+
+class LinearMeanGrad(Mean):
+    def __init__(self, input_size, batch_shape=torch.Size(), bias=True):
```
gpleiss, line 9: Doc string and type hints

gpytorch/means/linear_mean_gradgrad.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+
+import torch
+
+from .mean import Mean
+
+
+class LinearMeanGradGrad(Mean):
+    def __init__(self, input_size, batch_shape=torch.Size(), bias=True):
```
gpleiss, line 9: Doc string and type hints.

gpleiss (changes requested):
Thank you for adding the `diag` functionality, but this PR still needs unit tests. Please also add docstrings and type hints to all new classes and methods.

.pre-commit-config.yaml
ankushaggarwal: Yes, removed it.

gpytorch/kernels/rbf_kernel_gradgrad.py
ankushaggarwal: Updated

gpytorch/means/constant_mean_gradgrad.py
ankushaggarwal: Added

gpytorch/means/linear_mean_grad.py
ankushaggarwal: Added

gpytorch/means/linear_mean_gradgrad.py
ankushaggarwal: Added

gpleiss (approved):

gpleiss:
Thanks @ankushaggarwal !

Issue #2211, opened 2022-11-30
[Bug] Use of zip leads to silent incomplete evaluation of IndependentModelList


saitcakmak:
# 🐛 Bug

Due to the use of `zip`, when fewer than # of models inputs are passed to `IndependentModelList.__call__/forward` (or even `fantasize`), it will only evaluate the first # of inputs models and ignore the rest. 

## To reproduce

This extracted from the unit tests for `IndependentModelList`, which also falls into this bug.

** Code snippet to reproduce **
```python
import gpytorch
from gpytorch.models import IndependentModelList


class ExactGPModel(ExactGP):
    def __init__(self, train_x, train_y, likelihood):
        super().__init__(train_x, train_y, likelihood)
        self.mean_module = gpytorch.means.ConstantMean()
        self.covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.RBFKernel())

    def forward(self, x):
        mean_x = self.mean_module(x)
        covar_x = self.covar_module(x)
        return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)


N_PTS = 5


def create_test_data():
    return torch.randn(N_PTS, 1)


def create_likelihood_and_labels():
    likelihood = gpytorch.likelihoods.GaussianLikelihood()
    labels = torch.randn(N_PTS) + 2
    return likelihood, labels


def create_model():
    data = create_test_data()
    likelihood, labels = create_likelihood_and_labels()
    return ExactGPModel(data, labels, likelihood)

models = [create_model() for _ in range(2)]
model = IndependentModelList(*models)
model.eval()
model(torch.rand(3))
# This outputs [MultivariateNormal(loc: torch.Size([3]))], which is the evaluation of the first model only.

model(torch.rand(3), torch.rand(3))
# This outputs [MultivariateNormal(loc: torch.Size([3])), MultivariateNormal(loc: torch.Size([3]))], which is both models.

```

## Expected Behavior

Either error out when the # of inputs does not match # of models, or repeat the input if only a single tensor input is given. 
I'd be happy to patch this if we agree on a solution.

## System information

**Please complete the following information:**
-  GPyTorch: Latest
-  PyTorch: Latest
-  OS: CentOS8

## Additional context
Originally surfaced in https://github.com/pytorch/botorch/issues/1467

Balandat:
[...]

PR #2249, opened 2023-01-11
Fix silently ignored arguments in IndependentModelList

saitcakmak:
Fixes https://github.com/cornellius-gp/gpytorch/issues/2211 by introducing a `_length_safe_zip` that checks that the arguments are of equal length.

### DIFF
--- a/gpytorch/models/model_list.py
+++ b/gpytorch/models/model_list.py
@@ -7,6 +7,7 @@ from torch.nn import ModuleList
 
 from gpytorch.likelihoods import LikelihoodList
 from gpytorch.models import GP
+from gpytorch.utils.generic import length_safe_zip
 
 
 class AbstractModelList(GP, ABC):
@@ -47,7 +48,7 @@ class IndependentModelList(AbstractModelList):
 
     def forward(self, *args, **kwargs):
         return [
-            model.forward(*args_, **kwargs) for model, args_ in _length_safe_zip(self.models, _get_tensor_args(*args))
+            model.forward(*args_, **kwargs) for model, args_ in length_safe_zip(self.models, _get_tensor_args(*args))
         ]
 
     def get_fantasy_model(self, inputs, targets, **kwargs):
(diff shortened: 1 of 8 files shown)


gpytorch/models/model_list.py
```diff
@@ -91,8 +98,26 @@ def train_targets(self):
...
-            yield arg
+            arg_list.append(arg)
+    return arg_list
+
+
+def _length_safe_zip(*args):
+    """Python's `zip(...)` with checks to ensure the arguments have
+    the same number of elements.
+
+    NOTE: This does not work with generators.
```
Balandat, line 114: An alternative would be to convert the elements that do not implement `__len__` into a list automatically. But I guess that could be problematic if they are long / take up a bunch of memory that we don't want to allocate.

gpytorch/models/model_list.py
```diff
@@ -91,8 +98,26 @@ def train_targets(self):
...
         if torch.is_tensor(arg):
-            yield (arg,)
+            arg_list.append((arg,))
         else:
-            yield arg
+            arg_list.append(arg)
+    return arg_list
+
+
+def _length_safe_zip(*args):
```
Balandat, line 110: Should this live in some shared utility? Like in `gpytorch/utils/generic.py` or the like?

Balandat (approved):

gpytorch/models/model_list.py
saitcakmak: Yeah, I thought about that. For generators specifically, we would have to update the `args` directly, since once you call `list(generator)`, the generator is consumed and cannot be reused. The issue with that is `args` is a tuple, so we would have to first make that into a list, then iterate over its elements and update them. Eliminating the generator args seemed easier to me for this particular use case. If there are other possible use cases for this across the codebase, then it might be worth it though.

saitcakmak:
@Balandat I made it into a generic util as suggested and used it in some other places where similar issues could arise.

Balandat (approved):
lgtm - we can wait until the pytorch PR that messed up the tests gets unlanded and retry them? Or could merge if this is urgent for some reason.

saitcakmak:
No urgency to merge this. We can wait on the CI to land

PR #2255, opened 2023-01-17
gpytorch.settings.variational_cholesky_jitter can be set dynamically.

### LINKED ISSUE #2244: [Bug] settings.variational_cholesky_jitter does not work in run time
# 🐛 Bug

## To reproduce

Using `settings.variational_cholesky_jitter` in the model run time does not change the actual jitter value 

```python
from gpytorch.models import ApproximateGP
from gpytorch.variational import CholeskyVariationalDistribution
from gpytorch.variational import VariationalStrategy
from gpytorch import settings

class GPModel(ApproximateGP):
    def __init__(self, inducing_points):
        variational_distribution = CholeskyVariationalDistribution(inducing_points.size(0))
        variational_strategy = VariationalStrategy(self, inducing_points, variational_distribution, learn_inducing_locations=True)
        super(GPModel, self).__init__(variational_strategy)

    def forward(self, x):
        mean_x = self.mean_module(x)

train_x = torch.randn(10)
inducing_points = train_x
model = GPModel(inducing_points=inducing_points)

with settings.variational_cholesky_jitter(float_value=1e-2):
    print(model.variational_strategy.jitter_val)
```

The printed output is 0.0001, which is the default jitter value. 

## Expected Behavior

Expected value is 1e-2. 

## Additional context
In fact, the jitter value can be set through `settings` when instantiating the model: 
```
with settings.variational_cholesky_jitter(float_value=1e-2):
    model = GPModel(inducing_points=inducing_points)
    print(model.variational_strategy.jitter_val)
```
The printed output is 1e-2, which is expected. However, such way of setting variational_cholesky_jitter value is not consistent with other settings, e.g. `settings.cholesky_jitter`.

gpleiss:
Previously, this context manager was only used when VariationalStrategy modules were initialized. With this change,
gpytorch.settings.variational_cholesky_jitter will dynamically change the jitter value (for variational models already in use), unless the user specifies a `jitter_val` in the VariationalStrategy constructor.

In addition, this PR adds type hintsd to a majority of the VariationalStrategy modules.

[Fixes #2244]

### DIFF
--- a/gpytorch/variational/lmc_variational_strategy.py
+++ b/gpytorch/variational/lmc_variational_strategy.py
@@ -1,8 +1,11 @@
 #!/usr/bin/env python3
 
+from typing import Optional, Union
+
 import torch
 from linear_operator.operators import KroneckerProductLinearOperator, RootLinearOperator
 from linear_operator.utils.interpolation import left_interp
+from torch import LongTensor, Tensor
 
 from .. import settings
 from ..distributions import MultitaskMultivariateNormal, MultivariateNormal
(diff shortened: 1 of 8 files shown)


gpytorch/variational/lmc_variational_strategy.py
```diff
@@ -174,18 +179,15 @@ def __call__(self, x, task_indices=None, prior=False, **kwargs):
             :class:`~gpytorch.distributions.MultivariateNormal`.
 
         :param x: Input locations to evaluate variational strategy
-        :type x: torch.Tensor (... x N x D)
```
Balandat, line 177: It's a bit unfortunate to remove the shape annotation here in favor of the type annotations. Can we keep it around for now? Hopefully down the road we can use shape annotations directly in the types as well (e.g. as in the works in https://github.com/cornellius-gp/linear_operator/pull/42).

Balandat (approved):
Very nice. 

You'll probably have to rebase and rerun the pre-commit hooks with my recent ufmt changes.

Issue #1554, opened 2021-04-07
[Bug] Spurious error raised while indexing a RBFKernelGrad LazyTensor


ankushaggarwal:
# 🐛 Bug

While indexing RBFKernelGrad, I get a spurious error

## To reproduce
```python
import torch
import gpytorch
x=torch.tensor([0.,1.,2.])
index=torch.tensor([True,False,True])
full_kernel = gpytorch.kernels.RBFKernel()(x)
full_kernel[:,index]
full_kernel[:,index][index,:]
print('Slicing worked fine for RBF kernel')

full_kernel = gpytorch.kernels.RBFKernelGrad()(x)
index=torch.cat([index,index])
full_kernel[:,index]
print('However, the next line raises an error for RBFGrad kernel')
full_kernel[:,index][index,:]
```

** Stack trace/error message **
```
Slicing worked fine for RBF kernel
However, it raises an error for RBFGrad kernel
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
     12 full_kernel[:,index]
     13 print('However, it raises an error for RBFGrad kernel')
---> 14 full_kernel[:,index][index,:]

/opt/anaconda3/lib/python3.7/site-packages/gpytorch/lazy/lazy_tensor.py in __getitem__(self, index)
   1730                 raise RuntimeError(
   1731                     "{}.__getitem__ failed! Expected a final shape of size {}, got {}. This is a bug with GPyTorch, "
-> 1732                     "or your custom LazyTensor.".format(self.__class__.__name__, expected_shape, res.shape)
   1733                 )
   1734 

RuntimeError: NonLazyTensor.__getitem__ failed! Expected a final shape of size torch.Size([6, 4]), got torch.Size([4, 4]). This is a bug with GPyTorch, or your custom LazyTensor.
```

## Expected Behavior
A [4,4] tensor size is expected, so I do not see how the `Expected a final shape of size torch.Size([6, 4])` is computed. Therefore, I believe the error should not be raised.

## System information
- GPyTorch Version 1.1.1
- PyTorch Version 1.8.0
- Computer OS MacOS 11.2.3

gpleiss:
Yeah this seems like a bug on our end.

PR #2272, opened 2023-02-12
Variational GP with derivatives and monotonic gp

ankushaggarwal:
I have implemented an Indexed version of `VariationalStrategy` that allows us to use derivative information. I have used it to create a monotonic GP with a composite likelihood. The new variational strategy is called `VariationalStrategyIndexed`, but it could be easily merged with the normal `VariationalStrategy` (I can do that if that is preffered). 

I have also provided an example on how to use it. One minor issue though, as raised in a bug report https://github.com/cornellius-gp/gpytorch/issues/1554, the slicing raises a spurious error. To suppress it, we have to set `gpytorch.linear_operator.settings.debug._default = False`. This pull request is related to https://github.com/cornellius-gp/gpytorch/discussions/1639#discussioncomment-1121582


gpleiss:
Sorry for the slow reply @ankushaggarwal . I'm a little hesitant about adding this PR for a few reasons. 1) I'm not sure how re-usable `VariationalStrategyIndex` is, and I think it's also too specific to add to `VariationStrategy` proper. 2) I'm not sure this is the best/correct way to use derivative information with variational models. Derivative observations qualify as an inter-domain GP, and for variational GPs its best to factorize over the derivative and non-derivative observations (see https://arxiv.org/pdf/2003.01115.pdf for more details).

ankushaggarwal:
> Sorry for the slow reply @ankushaggarwal . I'm a little hesitant about adding this PR for a few reasons. 1) I'm not sure how re-usable `VariationalStrategyIndex` is, and I think it's also too specific to add to `VariationStrategy` proper. 2) I'm not sure this is the best/correct way to use derivative information with variational models. Derivative observations qualify as an inter-domain GP, and for variational GPs its best to factorize over the derivative and non-derivative observations (see https://arxiv.org/pdf/2003.01115.pdf for more details).

Thank you @gpleiss for your response. No problem. Since we needed a monotonic GP which needs variational GP with Bernouille likelihood for derivatives (as pointed out by you in https://github.com/cornellius-gp/gpytorch/discussions/1639#discussioncomment-1121582). The current implementation by me works for us in practice, but we would be very much interested in the variational GPs that handle derivatives through factorization over the derivative and non-derivative observations. So please let me know when this gets implemented in gpytorch, and we would love to use this more suitable approach.

PR #2282, opened 2023-02-22
Use raw strings to avoid "DeprecationWarning: invalid escape sequence"

saitcakmak:
This currently produces the following warnings:
```
.../gpytorch/gpytorch/likelihoods/gaussian_likelihood.py:168: DeprecationWarning: invalid escape sequence \s
  """
.../gpytorch/gpytorch/mlls/leave_one_out_pseudo_likelihood.py:12: DeprecationWarning: invalid escape sequence \l
  """
.../gpytorch/gpytorch/priors/horseshoe_prior.py:14: DeprecationWarning: invalid escape sequence \s
  """Horseshoe prior.
.../gpytorch/gpytorch/variational/ciq_variational_strategy.py:257: DeprecationWarning: invalid escape sequence \m
```

PR #2295, opened 2023-03-10
Use raw strings to avoid "DeprecationWarning: invalid escape sequence"

saitcakmak:
This was fixed in https://github.com/cornellius-gp/gpytorch/pull/2282 but looks like it made its way back.


Balandat (approved):
thanks

PR #2297, opened 2023-03-10
Ensure PSD-safe factorization in constructor of `MultivariateNormal`

SebastianAment:
There was a [recent BoTorch issue](https://github.com/pytorch/botorch/pull/1716) that was caused by a positive semi-definite matrix being passed to `MultivariateNormal` as a `Tensor`, which causes the constructor to fail because PyTorch's constructor calls `cholesky` on the tensor. This commit upstreams the corresponding BoTorch PR to ensure that all covariance matrices are `LinearOperator` types, thereby triggering `_psd_safe_cholesky`, whenever `cholesky` is called.


gpytorch/distributions/multivariate_normal.py
```diff
@@ -42,27 +42,36 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
...
+        mean: Union[Tensor, LinearOperator],
+        covariance_matrix: Union[Tensor, LinearOperator],
+        validate_args: bool = False,
+    ):
+        self._islazy = True
+        # casting Tensor to DenseLinearOperator because the super constructor calls cholesky, which
+        # will fail if the covariance matrix is semi-definite, whereas DenseLinearOperator ends up
+        # calling _psd_safe_cholesky, which factorizes semi-definite matrices by adding to the diagonal.
+        if isinstance(covariance_matrix, Tensor):
+            self._islazy = False  # to allow _unbroadcasted_scale_tril setter
```
Balandat, line 56: It seems odd to have `_islazy` set to `True` if the covariance matrix is indeed a `LinearOperator`. I guess the "lazy" nomenclature is a bit outdated anyway with the move to `LinearOperator`.

gpytorch/distributions/multivariate_normal.py
```diff
@@ -42,27 +42,36 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
...
+                raise ValueError(f"Wrong shapes in {self._repr_sizes(mean, covariance_matrix)}")
+        self.loc = mean
+        self._covar = covariance_matrix
+        self.__unbroadcasted_scale_tril = None
+        self._validate_args = validate_args
+        batch_shape = torch.broadcast_shapes(self.loc.shape[:-1], covariance_matrix.shape[:-2])
+
+        event_shape = self.loc.shape[-1:]
+
+        # TODO: Integrate argument validation for LinearOperators into torch.distribution validation logic
```
Balandat, line 73: Do you mean changing the torch code to validate LinearOperator inputs? That might be somewhat challenging to do if we want to use LinearOperators there explicitly. What would work is to make changes in pure torch that would make it easier to use LinearOperator objects by means of the `__torch_function__` interface we define in LinearOperator.

Balandat (commented):
Thanks! An alternative to making everything into a LinearOperator would be upon receiving a dense tensor, rather than just storing that and passing it along, to compute the cholesky decomposition with jitted and then pass that as the scale_tril to the torch distribution. The downside of that is ofc that this would do a lot of compute upon construction of the object, potentially unnecessary. Another hack could be to mock some of the torch distribution code so that it users the psd-safe cholesky decomposition internally (though that seems very hacky and potentially problematic).

Issue #2301, opened 2023-03-13
[Feature Request] Add batch_shape property to GP models


Balandat:
# 🚀 Feature Request

Expose a `batch_shape` property on models. For Exact GPs or models that have all the training inputs, this can just return the batch shape of the input tensors. For other models this can be overwritten as needed

## Motivation

See https://github.com/pytorch/botorch/issues/1680#issuecomment-1466331781 - for some fantasizing functionality we'd like to know what the batch shape of the training data of the model is to appropriately fantasize based on model outputs. 

There are a lot of other use cases for such property, we currently have this defined on the botorch `Model` class and are using it in a number of places (check the botorch repo).

## Pitch

**Describe the solution you'd like**
Upstream the functionality from botorch.

**Describe alternatives you've considered**
Not doing this. The downside is that it makes things rather cumbersome and requires some workarounds. 

**Are you willing to open a pull request?** (We LOVE contributions!!!)
Sure

Balandat:
@gpleiss any thoughts on this?

gpleiss:
Yeah this is a great idea!

PR #2307, opened 2023-03-19
Add batch_shape property to GP model class

Balandat:
Implements #2301.

Outstanding TODOs:
- support for approximate GP models
- unit tests
- verify compatibility with the botorch setup of other models

### DIFF
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -72,6 +72,17 @@ class ExactGP(GP):
         self.prediction_strategy = None
 
     @property
+    def batch_shape(self) -> torch.Size:
+        r"""The batch shape of the model.
+
+        This is a batch shape from an I/O perspective, independent of the internal
+        representation of the model. For a model with `(m)` outputs, a
+        `test_batch_shape x q x d`-shaped input to the model in eval mode returns a
+        distribution of shape `broadcast(test_batch_shape, model.batch_shape) x q x (m)`.
+        """
+        return self.train_inputs[0].shape[:-2]
+
+    @property
     def train_targets(self):
         return self._train_targets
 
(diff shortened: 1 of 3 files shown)


gpleiss (approved):
LGTM!

examples/045_GPLVM/Gaussian_Process_Latent_Variable_Models_with_Stochastic_Variational_Inference.ipynb
```diff
@@ -137,14 +137,13 @@
...
     "        self.n = n\n",
-    "        self.batch_shape = torch.Size([data_dim])\n",
     "        \n",
     "        # Locations Z_{d} corresponding to u_{d}, they can be randomly initialized or \n",
     "        # regularly placed with shape (D x n_inducing x latent_dim).\n",
     "        self.inducing_inputs = torch.randn(data_dim, n_inducing, latent_dim)\n",
     "    \n",
     "        # Sparse Variational Formulation (inducing variables initialised as randn)\n",
-    "        q_u = CholeskyVariationalDistribution(n_inducing, batch_shape=self.batch_shape) \n",
+    "        q_u = CholeskyVariationalDistribution(n_inducing, batch_shape=torch.Size([data_dim])) \n",
```
gpleiss, line 146: Why aren't we using the `self.batch_shape` property here?

gpleiss (approved):
@Balandat should I merge this PR? Or do you still edits you want to make?

Balandat:
Let's hold off. I have been running into some issues with this approach (caused by the fact that in botorch the batch shape is not necessarily the same as the gpytorch-internal batch shape. So right now this would cause a bunch of headaches and bugs on our end rather than only resolving some. We'll have to rethink if this is the right approach.

PR #2323, opened 2023-04-20
Fix solve_triangular(Tensor, LinearOperator) not supported in VNNGP

Turakar:
See https://github.com/cornellius-gp/linear_operator/issues/53.


Balandat (approved):
lgtm

PR #2325, opened 2023-04-21
Metrics fixes and cleanup

JonathanWenger:
# In a Nutshell
This PR fixes the implementation of the mean standardized log loss to match its definition and adds some documentation.

# Detailed
The mean standardized log loss as defined by Rasmussen and Williams (2006) compares the average log loss to a trivial model based off of the training data summary statistics. This standardization was missing in the implementation.

Screenshot from 2023-04-21 19-31-05

 Further I added the standardized mean squared error to allow for better comparison across experiments.

### DIFF
--- a/gpytorch/metrics/metrics.py
+++ b/gpytorch/metrics/metrics.py
@@ -1,4 +1,5 @@
 from math import pi
+from typing import Optional
 
 import torch
 
@@ -12,7 +13,7 @@ def mean_absolute_error(
     test_y: torch.Tensor,
 ):
     """
-    Mean Absolute Error.
+    Mean absolute error.
     """
     combine_dim = -2 if isinstance(pred_dist, MultitaskMultivariateNormal) else -1
     return torch.abs(pred_dist.mean - test_y).mean(dim=combine_dim)
@@ -24,7 +25,7 @@ def mean_squared_error(
     squared: bool = True,
 ):
     """
-    Mean Squared Error.
+    Mean squared error.
     """
     combine_dim = -2 if isinstance(pred_dist, MultitaskMultivariateNormal) else -1
     res = torch.square(pred_dist.mean - test_y).mean(dim=combine_dim)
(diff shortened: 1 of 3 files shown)


gpytorch/metrics/metrics.py
```diff
@@ -33,29 +34,59 @@ def mean_squared_error(
...
+    loss_model = (0.5 * torch.log(2 * pi * f_var) + torch.square(test_y - f_mean) / (2 * f_var)).mean(dim=combine_dim)
+    res = loss_model
+
+    if train_y is not None:
+        data_mean = train_y.mean(dim=combine_dim)
+        data_var = train_y.var()
+        loss_trivial_model = (
+            0.5 * torch.log(2 * pi * data_var) + torch.square(test_y - data_mean) / (2 * data_var)
+        ).mean(dim=combine_dim)
+        res -= loss_trivial_model
```
Balandat, line 87: Let's avoid in-place ops to make autograd happier
```suggestion
        res = res - loss_trivial_model
```

test/metrics/test_metrics.py
```diff
@@ -126,6 +127,9 @@ def test_negative_log_predictive_density(self):
     def test_mean_standardized_log_loss(self):
         self._test_metric(mean_standardized_log_loss)
 
+    def test_standardized_mean_squared_error(self):
+        self._test_metric(standardized_mean_squared_error)
```
Balandat, line 131: Hmm this is a kind of interesting test in that it only executes the metric code, without really checking what the computation does. Not really in scope for this PR but this should probably do something more reasonable...

Balandat (approved):

test/metrics/test_metrics.py
JonathanWenger: Agreed. Treating this as out of scope for this PR for now, but created an issue for it: #2326 .

PR #2327, opened 2023-04-23
add Hamming IMQ kernel

samuelstanton:
Add one of the simpler discrete sequence kernels proposed here
https://arxiv.org/abs/2304.03775
and originally implemented here
https://github.com/AlanNawzadAmin/Kernels-with-guarantees/blob/main/kernels_w_guarantees/kernels.py#L298

Virtually all the current GPyTorch kernels are designed for continuous inputs, I think this could be a great addition to the arsenal. Will add unit tests soon but thought I'd open the PR for feedback.


gpytorch/kernels/hamming.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def _imq(self, dist: Tensor) -> Tensor:
+        return ((1 + self.alpha) / (self.alpha + dist)).pow(self.beta)
+
+    def forward(self, x1: Tensor, x2: Tensor, diag: bool = False, **params):
+        # GPyTorch is pretty particular about dimensions so we need to unflatten the one-hot encoding
+        x1 = x1.view(*x1.shape[:-1], -1, self.vocab_size)
+        x2 = x2.view(*x2.shape[:-1], -1, self.vocab_size)
+
+        x1_eq_x2 = torch.equal(x1, x2)
```
Balandat, line 129: This is kind of expensive - should we just do `x1 is x2` instead here? I don't really see a lot of situations in which we want to compute the forward pass on the same tensor but that not being the same object.

gpytorch/kernels/hamming.py
```diff
@@ -0,0 +1,151 @@
...
+                dist = x1.size(-2) - (x1 * x2).sum(dim=(-1, -2))
+                return self._imq(dist)
+
+        else:
+            dist = hamming_dist(x1, x2, x1_eq_x2)
+
+        return self._imq(dist)
+
+
+def hamming_dist(x1: Tensor, x2: Tensor, x1_eq_x2: Tensor) -> Tensor:
```
Balandat, line 146:
```suggestion
def hamming_dist(x1: Tensor, x2: Tensor, x1_eq_x2: bool) -> Tensor:
```

gpytorch/kernels/hamming.py
samuelstanton: that line was copied from here, should it be changed there as well?
https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/kernels/kernel.py#L339

samuelstanton: It looks like the unit tests are failing due to a circular import error, either from the prior or the constraint imports I used for type hints and default args. I can take them out but this behavior seems suboptimal

samuelstanton:
It looks like the unit tests are failing due to a circular import error, either from the prior or the constraint imports I used for type hints and default args. I can take them out but this behavior seems suboptimal

gpytorch/kernels/hamming.py
Balandat: I would advocate for that - @gpleiss, @jacobrgardner, any reason not to do this?

gpytorch/kernels/hamming.py
```diff
@@ -0,0 +1,150 @@
+from typing import Optional
+
+import torch
+from torch import nn, Tensor
+
+from ..constraints import Interval, Positive
+from ..kernels import Kernel
+from ..priors import Prior
```
Balandat, line 8: If you import from the specific module rather than the `__all__` defined in `__init__.py` you should avoid circular import issues here. 
```suggestion
from gpytorch.constraints.constraints import Interval, Positive
from gpytorch.kernels.kernel import Kernel
from gpytorch.priors.prior import Prior
```

samuelstanton:
added some unit tests, any other requested changes?

samuelstanton:
@gpleiss @jacobrgardner can we get this merged?

gpytorch/kernels/hamming.py
jacobrgardner: Maybe we revisit the x1_eq_x2 question in a second PR where we can make sure it all works?

jacobrgardner:
Generally looks good to me. Maybe we can add a docstring telling people they'll need to convert their sequence data to flattened one hot format, and a snippet of code that does so?  Right now this is only really made apparent in the unit test, which people won't go searching through.

samuelstanton:
not sure why the docs build is failing

jacobrgardner:
It looks like some version increase in one of the packages involved in docs building caused failures -- all PRs are currently failing to build docs.

gpleiss (approved):

PR #2342, opened 2023-05-13
[RFC, WIP]: Remove evaluate kernel

gpleiss:
With `KernelLinearOperator` defined in the linear operator package, `LazyEvaluatedKernelTensor` becomes kind of redundant.

This PR attempts to improve the lazy evaluation of kernels in GPyTorch, by doing the following:
- Remove `LazyEvaluatedKernelTensor`, instead using `KernelLinearOperator` (which is also used by KeOps kernels)
- Removing `evaluate_kernel` form GPyTorch (which will be redundant with `KernelLinearOperator`) and also removing it from the LinearOperator package.

cc/ @Balandat


jacobrgardner:
Aren't we going to take at least some amount of performance hit here on kernels you don't lazily evaluate when making predictions, because we don't get to do indexing before calling the kernel?

gpytorch/kernels/kernel.py
```diff
@@ -75,6 +76,45 @@ def _dist(self, x1, x2, x1_eq_x2=False, postprocess=False):
         return self._postprocess(res) if postprocess else res
 
 
+class _autograd_kernel_hack(object):
```
Balandat, line 79:
```suggestion
class _autograd_kernel_hack:
```

gpytorch/kernels/kernel.py
```diff
@@ -206,6 +246,37 @@ def __init__(
...
+    def _lazily_evaluate(self) -> bool:
+        r"""
+        Determines whether or not the kernel is lazily evaluated.
+
+        If False, kernel(x1, x2) produces a Tensor/LinearOperator where the covariance function has been evaluated
+        over x1 and x2.
+
+        If True, kernel(x1, x2) produces a KernelLinearOperator that delays evaluation of the kernel function.
+        The kernel function will only be evaluated when either
+            - An mathematical operation is performed on the kernel matrix (e.g. solves, logdets, etc.), or
```
Balandat, line 259:
```suggestion
            - A mathematical operation is performed on the kernel matrix (e.g. solves, logdets, etc.), or
```

gpytorch/kernels/kernel.py
```diff
@@ -517,11 +588,42 @@ def __call__(
...
+                num_outputs_per_input = self.num_outputs_per_input(x1_, x2_)
+                named_parameters = tuple(self.named_parameters())
+
+                if last_dim_is_batch:
+                    x1_ = x1_.transpose(-1, -2).unsqueeze(-1)
+                    x2_ = x2_.transpose(-1, -2).unsqueeze(-1)
+
+                if len(named_parameters):
+                    param_names, params = zip(*named_parameters)
+                    param_batch_shapes = [self.batch_shape] * len(params)
```
Balandat, line 601: So being able to use `self.batch_shape` here relies on the fact that all parameters have a fully explicit shape (rather than broadcasting over them)?

gpytorch/kernels/kernel.py
```diff
@@ -517,11 +588,42 @@ def __call__(
...
+                        covar_func=self._kernel_linear_operator_covar_func,
+                        num_outputs_per_input=num_outputs_per_input,
+                        param_batch_shapes=param_batch_shapes,
+                        param_names=param_names,
+                        **kwargs,
+                    )
+                else:
+                    res = KernelLinearOperator(
+                        x1_, x2_, covar_func=self.forward, num_outputs_per_input=num_outputs_per_input, **kwargs
+                    )
```
Balandat, line 623: We're pretty deep in the indentations here, might make sense to pull out some of the above into helper functions to make it clear what's going on here / easier to read the code

gpytorch/kernels/scale_kernel.py
```diff
@@ -75,7 +68,7 @@ def __init__(
             outputscale_constraint = Positive()
 
         self.base_kernel = base_kernel
-        outputscale = torch.zeros(*self.batch_shape) if len(self.batch_shape) else torch.tensor(0.0)
+        outputscale = torch.zeros(*self.batch_shape, 1, 1) if len(self.batch_shape) else torch.zeros(1, 1)
```
Balandat, line 71: Changing the parameter shapes for widely used kernels like this will likely result in a bunch of downstream backward compatibility issues. I'm not necessarily suggestion not to do this, but those changes will need to be properly documented / announced.

Balandat: btw, if we're doing this, we should probably combine this with updating the shapes of the priors (c.f. #1317) to make all of that consistent

gpytorch/kernels/kernel.py
gpleiss: Yes. I think this is generally true for all kernels in GPyTorch. We can add something to the documentation to be more explicit about this.

gpytorch/kernels/scale_kernel.py
gpleiss: I'm going back and forward about changing parameter shapes. On one hand, it could be quite the breaking change. On the other hand, it would create more consistency, and we should also be able to dramatically simplify many kernels as well.

@Balandat thoughts? I imagine this would have the biggest impact on BoTorch.

gpleiss:
> Aren't we going to take at least some amount of performance hit here on kernels you don't lazily evaluate when making predictions, because we don't get to do indexing before calling the kernel?

@jacobrgardner I don't think it'll lead to a loss in performance. This PR mostly merges LazyEvaluatedKernelTensor and KeOpsLinearOperator, since they share most of the same functionality.

gpytorch/kernels/scale_kernel.py
Balandat: Makes sense. I think fixing the parameter shape inconsistencies and making things consistent across the board would be good, and hopefully this would also get rid of some of the long standing bugs/issues. This would impact botorch, but if we coordinate on the releases we can make sure that the effect of this is minimized. It would likely generate some short term pain for some power users (of both gpytorch and botorch) with custom setups, but I think that short term pain is probably the long-term gains from consistency.,

gpleiss: @Balandat That sounds like a good plan. Here's what I'm thinking as a course of action:

- Make a PR that makes kernel parameters consistent sizes (and adds appropriate warnings/deprecations/etc)
- Address #1317
- Merge this PR
- (+ other breaking changes that we've wanted to address for a while)

After all of that, we make a major release, timed with a BoTorch release.

Issue #2340, opened 2023-05-11
[Bug] log noise prior is added twice in the MLL computation


LuisAugenstein:
# Noise prior is added twice when computing the ExactMarginalLogLikelihood

When I create a simple ExactGP Model with a noise prior I think the marginal log likelihood is computed wrong. 
In my opinion the MLL (or let's rather call it log posterior in this scenario as we have a noise prior) should be computed as follows

$\text{log } p(X, \sigma^2 \vert \boldsymbol{y}) = \text{log } p(\boldsymbol{y} \vert X, \sigma^2) + \text{log } p(\sigma^2)$

However, in my experiment it seems as if currently gpytorch adds the log noise prior twice instead of once.

$\text{log } p(X, \sigma^2 \vert \boldsymbol{y}) = \text{log } p(\boldsymbol{y} \vert X, \sigma^2) + 2 \ \text{log } p(\sigma^2)$

This is because the MarginalLogLikelihood module contains the likelihood module twice, i.e., directly via `self.likelihood` and indirectly through the model `self.model.likelihood`.
```python
class gpytorch.mlls.MarginalLogLikelihood(Module):
    def __init__(self, likelihood, model):
        self.likelihood = likelihood
        self.model = model
```

## To reproduce
```python
import torch
import gpytorch

torch.manual_seed(123)

class ExactGPModel(gpytorch.models.ExactGP):
    def __init__(self, train_x, train_y, likelihood):
        super(ExactGPModel, self).__init__(train_x, train_y, likelihood)
        self.mean_module = gpytorch.means.ConstantMean()
        self.covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.RBFKernel())

    def forward(self, x):
        mean_x = self.mean_module(x)
        covar_x = self.covar_module(x)
        return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)


N = 10
train_x = torch.randn(N)
train_y = torch.randn(N)
noise_prior = gpytorch.priors.NormalPrior(0, 1)
likelihood = gpytorch.likelihoods.GaussianLikelihood(noise_prior)
model = ExactGPModel(train_x, train_y, likelihood)

mll = gpytorch.mlls.ExactMarginalLogLikelihood(model.likelihood, model)
gp_prior = model(train_x)
marginal_log_likelihood = mll(gp_prior, train_y)

marginal_likelihood = likelihood(gp_prior)

[...]

PR #2355, opened 2023-05-26
Bugfix: double added log noise prior

LuisAugenstein:
There was an issue #2340 with the mll computation when using a noise prior. 
The `MarginalLogLikelihood` class contains the likelihood twice - once as a direct child and once as child of the model.
```python
class gpytorch.mlls.MarginalLogLikelihood(Module):
    def __init__(self, likelihood, model):
        self.likelihood = likelihood
        self.model = model
```
 Previously the `_add_other_terms()` method iterated over all submodules of the MarginalLogLikelihood class, resulting in the noise prior of the likelihood being added twice.
With the proposed fix the `_add_other_terms()` method now only iterates over the submodules of the model.

### DIFF
--- a/gpytorch/mlls/exact_marginal_log_likelihood.py
+++ b/gpytorch/mlls/exact_marginal_log_likelihood.py
@@ -40,7 +40,7 @@ class ExactMarginalLogLikelihood(MarginalLogLikelihood):
 
         # Add log probs of priors on the (functions of) parameters
         res_ndim = res.ndim
-        for name, module, prior, closure, _ in self.named_priors():
+        for name, module, prior, closure, _ in self.model.named_priors():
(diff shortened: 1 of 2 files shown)


gpleiss (approved):

Balandat:
Are there potential cases where the MLL may be computed on a likelihood that is not a child of the model? This fix would break that. Similarly, are there cases where there are other (non-likelihood-related) parameters with priors set on non-model children of the MLL? If none of these are potential considerations then great. If they are then maybe some deduplication based on the `id` of the prior would be a more general solution here?

gpleiss:
@Balandat I can't think of a scenario where there'd be a prior on something not model related, but we can revert this if there are any cases in botorch.

PR #2358, opened 2023-06-01
Fix some overly brittle deprecation warning tests

Balandat:
These are failing locally for me - may need to be tied to the linop version.


Balandat:
This has become irrelevant with 96bd347214222781cd1a0f56e00bd6609eb7ca0e

PR #2360, opened 2023-06-02
Remove num_outputs from IndependentModelList

saitcakmak:
`IndependentModelList.num_outputs` makes a strong assumption that the underlying models are single-output, so it returns the # of sub-models. However, `IndependentModelList` works perfectly fine with multi-output sub-models, leading to the `num_outputs` attribute returning the wrong value. Since GPyTorch models do not define a consistent `num_outputs` attribute, I don't see an easy way of correcting this.

This becomes an issue in BoTorch where `ModelListGP` has `IndependentModelList` ahead of `ModelList` in its MRO, which correctly defines `num_outputs` as the sum of `num_outputs` of the sub-models. Removing the property from `IndependentModelList` should allow us to use the correct definition from `ModelList`.

### DIFF
--- a/gpytorch/models/model_list.py
+++ b/gpytorch/models/model_list.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python3
 
-from abc import ABC, abstractproperty
+from abc import ABC
 
 import torch
 from torch.nn import ModuleList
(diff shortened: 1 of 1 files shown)


gpleiss (approved):

Balandat:
> Since GPyTorch models do not define a consistent num_outputs attribute, I don't see an easy way of correcting this.

Would it make sense to upstream some of the `num_outputs` conventions & machinery we use in botorch to gpytorch? I guess right now this wouldn't bring additional immediate benefits, but this could be useful for some folks.

PR #2375, opened 2023-07-09
Fix VNNGP with batches

LuhuanWu:
- Fix VNNGP in batch settings #2300 
- In addition, set default jitter value to 1e-3 and variational stddev init value to 1e-2. 
- Change initialization strategy
- Special-case behaviour when training_batch_size = M (ensure we only have one minibatch and that we compute the full KL)
- Remove the option to set `k == M`

### DIFF
--- a/gpytorch/variational/nearest_neighbor_variational_strategy.py
+++ b/gpytorch/variational/nearest_neighbor_variational_strategy.py
@@ -9,7 +9,6 @@ from linear_operator.operators import DiagLinearOperator, LinearOperator, Triang
 from linear_operator.utils.cholesky import psd_safe_cholesky
 from torch import LongTensor, Tensor
 
-from .. import settings
 from ..distributions import MultivariateNormal
 from ..models import ApproximateGP, ExactGP
 from ..module import Module
@@ -325,8 +324,7 @@ class NNVariationalStrategy(UnwhitenedVariationalStrategy):
         variational_inducing_covar = DiagLinearOperator(variational_covar_fisrtk)
 
         variational_distribution = MultivariateNormal(inducing_values, variational_inducing_covar)
-        with settings.max_preconditioner_size(0):
-            kl = torch.distributions.kl.kl_divergence(variational_distribution, prior_dist)  # model_batch_shape
+        kl = torch.distributions.kl.kl_divergence(variational_distribution, prior_dist)  # model_batch_shape
         return kl
 
     def _stochastic_kl_helper(self, kl_indices: Float[Tensor, "n_batch"]) -> Float[Tensor, "..."]:  # noqa: F821
--- a/gpytorch/variational/nearest_neighbor_variational_strategy.py
+++ b/gpytorch/variational/nearest_neighbor_variational_strategy.py
@@ -8,7 +8,7 @@ from linear_operator.operators import DiagLinearOperator, LinearOperator, Triang
 from linear_operator.utils.cholesky import psd_safe_cholesky
 from torch import LongTensor, Tensor
 
-from .. import settings 
+from .. import settings
 from ..distributions import MultivariateNormal
 from ..models import ApproximateGP, ExactGP
 from ..module import Module
@@ -63,7 +63,7 @@ class NNVariationalStrategy(UnwhitenedVariationalStrategy):
         VariationalDistribution object that represents the form of the variational distribution :math:`q(\mathbf u)`
     :param k: Number of nearest neighbors.
     :param training_batch_size: The number of data points that will be in the training batch size.
-    :param jitter_val: Amount of diagonal jitter to add for covariance matrix numerical stability. 
+    :param jitter_val: Amount of diagonal jitter to add for covariance matrix numerical stability.
     :param compute_full_kl: Whether to compute full kl divergence or stochastic estimate.
 
     .. _Wu et al (2022):
@@ -82,7 +82,7 @@ class NNVariationalStrategy(UnwhitenedVariationalStrategy):
         k: int,
         training_batch_size: int,
         jitter_val: Optional[float] = 1e-3,
-        compute_full_kl: Optional[bool] = False 
+        compute_full_kl: Optional[bool] = False,
     ):
         assert isinstance(
             variational_distribution, MeanFieldVariationalDistribution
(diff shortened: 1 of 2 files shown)


gpytorch/variational/nearest_neighbor_variational_strategy.py
```diff
@@ -87,8 +90,6 @@ def __init__(
         super().__init__(
             model, inducing_points, variational_distribution, learn_inducing_locations=False, jitter_val=jitter_val
         )
-        # Make sure we don't try to initialize variational parameters - because of minibatching
-        self.variational_params_initialized.fill_(1)
```
gpleiss, line 91: Why did this line get deleted?

gpytorch/variational/nearest_neighbor_variational_strategy.py
```diff
@@ -134,13 +138,25 @@ def __call__(self, x: Tensor, prior: bool = False, **kwargs: Any) -> Multivariat
...
+            # (Maybe) initialize variational distribution
+            if not self.variational_params_initialized.item():
+                prior_dist = self.prior_distribution
+                self._variational_distribution.variational_mean.data.copy_(prior_dist.mean)
+                self._variational_distribution.variational_mean.data.add_(
+                    torch.randn_like(prior_dist.mean), alpha=self._variational_distribution.mean_init_std
+                )
+                # initialize with a small variational stddev for quicker conv. of kl divergence
+                self._variational_distribution._variational_stddev.data.copy_(torch.tensor(1e-2))
+                self.variational_params_initialized.fill_(1)
```
gpleiss, line 158: Ahhh okay, I see you've added a new initialization scheme.

gpytorch/variational/nearest_neighbor_variational_strategy.py
```diff
@@ -266,78 +298,121 @@ def _firstk_kl_helper(self) -> Tensor:
         variational_inducing_covar = DiagLinearOperator(variational_covar_fisrtk)
 
         variational_distribution = MultivariateNormal(inducing_values, variational_inducing_covar)
-        kl = torch.distributions.kl.kl_divergence(variational_distribution, prior_dist)  # model_batch_shape
+        with settings.max_preconditioner_size(0):
```
gpleiss, line 301: Why is this setting necessary? Can you add a comment in code?

gpytorch/variational/nearest_neighbor_variational_strategy.py
```diff
@@ -359,5 +434,7 @@ def _compute_nn(self) -> "NNVariationalStrategy":
         with torch.no_grad():
             inducing_points_fl = self.inducing_points.data.float()
             self.nn_util.set_nn_idx(inducing_points_fl)
-            self.nn_xinduce_idx = self.nn_util.build_sequential_nn_idx(inducing_points_fl)
+            if self.k < self.M:
+                self.nn_xinduce_idx = self.nn_util.build_sequential_nn_idx(inducing_points_fl)
```
gpleiss, line 438: Do we have a test in code for the `k < M` case? If not, can you add one?

test/variational/test_nearest_neighbor_variational_strategy.py
```diff
@@ -115,7 +115,7 @@ def _training_iter(
         return output, loss
 
     def _eval_iter(self, model, cuda=False):
-        inducing_batch_shape = model.variational_strategy.inducing_points.shape[:-2]
+        inducing_batch_shape = model.variational_strategy._inducing_batch_shape
```
gpleiss, line 118: Can you add a unit test that runs VNNGP with batches? AFIAK, this unit test is still only running for non-batched VNNGP.

gpleiss (changes requested):

test/variational/test_nearest_neighbor_variational_strategy.py
LuhuanWu: I think line 194 `test_training_iteration_batch_model` does the batch model testing.

gpytorch/variational/nearest_neighbor_variational_strategy.py
LuhuanWu: Yes. In practice I found that the variational standard deviation tends to shrink towards 0 in the end given that inducing points = data points. If initialized with ones, the KL term is way larger than log likelihood term, taking a longer time to converge. Initializing with a smaller value speeds up the convergence.

gpytorch/variational/nearest_neighbor_variational_strategy.py
LuhuanWu: Yeah I think k = 3 in the test code, which is smaller than M. I could add a case when k = M later, which is missing currently.

gpytorch/variational/nearest_neighbor_variational_strategy.py
LuhuanWu: I was following the KL computation in `_variational_strategy`, see [this line](https://github.com/cornellius-gp/gpytorch/blob/d19f52e07e2cc5dd69d35457d0792072adf37574/gpytorch/variational/_variational_strategy.py#L160). What comment do you think is suitable here? Or do you suggest remove this line?

test/variational/test_nearest_neighbor_variational_strategy.py
gpleiss: @LuhuanWu why wasn't `test_training_iteration_batch_model` failing before this PR then?
Ideally we want to have a test case that (1) would've failed before this PR was added (capturing the behavior described in #2300) but (2) does not fail with the new code in this PR.

gpytorch/variational/nearest_neighbor_variational_strategy.py
gpleiss: For the future, it's good to list all of these little changes in the PR description as well :)

gpytorch/variational/nearest_neighbor_variational_strategy.py
gpleiss: Let's get rid of it for now, I don't think it's necessary. And we might want to remove it eventually from _variational_strategy, but it's not necessary for this PR.

test/variational/test_nearest_neighbor_variational_strategy.py
LuhuanWu: good point. I'll look into that.

gpleiss (approved):

PR #2376, opened 2023-07-12
Fix lazy kernel slicing when there are multiple outputs

douglas-boubert:
Fixes a bug when slicing a lazy kernel and the number of outputs per row or column is greater than one.

The master implementation incorrectly checks to see if x1 and x2 are both slices. This always fails and defaults to the super method.

This PR fixes that check to ensure that row_index and col_index are slices.

The changes on the other lines are necessary to catch situations where the start point or end point of the slicing is implicit.

### DIFF
--- a/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
+++ b/gpytorch/lazy/lazy_evaluated_kernel_tensor.py
@@ -169,13 +169,13 @@ class LazyEvaluatedKernelTensor(LinearOperator):
             # Let's make sure that the slice dimensions perfectly correspond with the number of
             # outputs per input that we have
             row_start, row_end, row_step = (
-                row_index.start if row_index.start is not None else 0,
-                row_index.stop if row_index.stop is not None else self.shape[-2],
+                row_index.start or 0,
+                row_index.stop or self.shape[-2],
                 row_index.step,
             )
             col_start, col_end, col_step = (
-                col_index.start if col_index.start is not None else 0,
-                col_index.stop if col_index.stop is not None else self.shape[-1],
+                col_index.start or 0,
+                col_index.stop or self.shape[-1],
                 col_index.step,
             )
             if row_step is not None or col_step is not None:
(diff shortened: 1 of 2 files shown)


gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -161,21 +161,21 @@ def _getitem(self, row_index, col_index, *batch_indices):
         # However - if we have multiple outputs per input, then the indices won't directly
         # correspond to the entries of row/col. We'll have to do a little pre-processing
         if num_outs_per_in_rows != 1 or num_outs_per_in_cols != 1:
-            if not isinstance(x1, slice) or not isinstance(x2, slice):
```
Balandat, line 164: ugh

gpytorch/lazy/lazy_evaluated_kernel_tensor.py
```diff
@@ -161,21 +161,21 @@ def _getitem(self, row_index, col_index, *batch_indices):
...
                 return self.evaluate_kernel()._getitem(row_index, col_index, *batch_indices)
 
             # Now we know that x1 and x2 are slices
             # Let's make sure that the slice dimensions perfectly correspond with the number of
             # outputs per input that we have
             row_start, row_end, row_step = (
-                row_index.start,
-                row_index.stop,
+                row_index.start if row_index.start is not None else 0,
+                row_index.stop if row_index.stop is not None else self.shape[-2],
```
Balandat, line 173: I like the explicitness but this would also work (same below)
```suggestion
                row_index.start or 0,
                row_index.stop or self.shape[-2],
```

Balandat (approved):
Great, thanks for the fix. Would you be able to add a unit test for this behavior as well?

douglas-boubert:
@Balandat, thanks for the review! I've simplified the None start/stop filling and added a unittest that covers the behaviour.

Balandat:
Looks like the linter / code formatter is unhappy, other than that this lgtm

douglas-boubert:
> Looks like the linter / code formatter is unhappy, other than that this lgtm

Ah, sorry for that! I've run ufmt on both files I touched and pushed the changes.

PR #2378, opened 2023-07-14
Minor patch to Matern covariances

j-wilson:
This PR is intended as a minor patch to ensure that `MaternCovariance` and `MaternKernel` both produce identical outputs in batched and non-batched settings. Currently, these methods produce slightly different batched vs non-batched outputs due to how a shift is handled internally. Examples are given below.

Pre-PR:
```
kernel = MaternKernel(nu=2.5)
X = torch.rand(2, 4, 2)
kernel(X).to_dense()[0] - kernel(X[0]).to_dense()
> tensor([[ 0.0000e+00,  0.0000e+00,  0.0000e+00,  1.1102e-16],
        [ 0.0000e+00,  0.0000e+00,  0.0000e+00,  0.0000e+00],
        [ 0.0000e+00,  0.0000e+00,  0.0000e+00,  0.0000e+00],
        [ 1.1102e-16, -2.2204e-16,  0.0000e+00,  0.0000e+00]],
       grad_fn=<SubBackward0>)
```

Post-PR:
```
kernel = MaternKernel(nu=2.5)
X = torch.rand(2, 4, 2)
kernel(X).to_dense()[0] - kernel(X[0]).to_dense()
> tensor([[0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]], grad_fn=<SubBackward0>)
```

### DIFF
--- a/gpytorch/kernels/matern_kernel.py
+++ b/gpytorch/kernels/matern_kernel.py
@@ -92,7 +92,7 @@ class MaternKernel(Kernel):
             or params.get("last_dim_is_batch", False)
             or trace_mode.on()
         ):
-            mean = x1.reshape(-1, x1.size(-1)).mean(0)[(None,) * (x1.dim() - 1)]
+            mean = x1.mean(dim=-2, keepdim=True)
 
             x1_ = (x1 - mean).div(self.lengthscale)
             x2_ = (x2 - mean).div(self.lengthscale)
--- a/gpytorch/functions/matern_covariance.py
+++ b/gpytorch/functions/matern_covariance.py
@@ -13,7 +13,7 @@ class MaternCovariance(torch.autograd.Function):
         # Subtract mean for numerical stability. Won't affect computations
         # because covariance matrix is stationary.
         needs_grad = any(ctx.needs_input_grad)
-        mean = x1.reshape(-1, x1.size(-1)).mean(0)[(None,) * (x1.dim() - 1)]
+        mean = x1.mean(dim=-2, keepdim=True)
         x1_ = (x1 - mean).div(lengthscale)
         x2_ = (x2 - mean).div(lengthscale)
         scaled_unitless_dist = dist_func(x1_, x2_).mul_(math.sqrt(2 * nu))


gpytorch/kernels/matern_kernel.py
```diff
@@ -92,7 +92,7 @@ def forward(self, x1, x2, diag=False, **params):
             or params.get("last_dim_is_batch", False)
             or trace_mode.on()
         ):
-            mean = x1.reshape(-1, x1.size(-1)).mean(0)[(None,) * (x1.dim() - 1)]
+            mean = x1.mean(dim=-2, keepdim=True)
```
Balandat, line 95: Hmm the behavior here is different in batch mode. The old code would take the mean across all elements in all the batches, and then return a tensor of shape `1 x ... x 1 x 1 x d`. The new code takes the mean on a per-batch level and returns a tensor of shape `b1 x b2 x ... bk x 1 x d`. That seems to be the more sane thing to do? I guess it doesn't really matter for correctness of the kernel computation since this cancels out, but it could be an issue if the scale of the data is very different between batches.

@gpleiss Any thoughts? Was the previous implementation on purpose?

j-wilson: If I'm not mistaken, the proposed changes should be more numerically stable (since shifts are now batch-specific). Further improvements could probably be made in this regard, but my primary motivation is ensuring that batched and non-batched computations are identical.

In general, I expect the difference here to be marginal. At the same time, downstream computations will never match if prior covariances differ when batched. This makes debugging more tedious. In my case, I was trying to figure out why a function of the posterior differed when evaluated in batch mode; and, it took me a while to realize that it was due to the kernel.

Balandat: Yeah that makes sense. I think this is probably as strict improvement both in stability and certainly in readability. So happy to merge this in.

Balandat (approved):

PR #2404, opened 2023-09-13
Fix latex of gradients in docs

jlperla:
See https://docs.gpytorch.ai/en/stable/means.html#constantmeangradgrad etc. for the docs.   I think `\nabla` is appropriate?


gpleiss:
Good catch, thanks @jlperla !

PR #2413, opened 2023-09-21
Fix KeOps regressions from #2296.

### LINKED ISSUE #2363: [Bug] KeOps LazyTensors are no longer compatible with linear-operator
# 🐛 Bug

Using the default KeOps example notebook no longer works as of the latest release of GPyTorch and linear-operator (perhaps this issue is better suited for the linear-operator github page, but it was encountered specifically with the GPyTorch KeOps integration)

## To reproduce

Run the KeOps_GP_Regression.ipynb notebook at /examples/02_Scalable_Exact_GPs/ (or with a smaller toy dataset)

```
-> 2977     raise TypeError("object of class {} cannot be made into a Tensor".format(obj.__class__.__name__))

TypeError: object of class LazyTensor cannot be made into a Tensor
```

## Expected Behavior

The linear-operator package should be able to perform operations with LazyTensor objects

## System information

- GPyTorch Version 1.11
- linear-operator version 0.5.0
- PyTorch Version 2.0.1
- Computer OS Linux

## Additional context
I know @gpleiss discussed integrating linear_operator with KeOps in my previous issue-- perhaps I missed some critical step for integrating the linear_operator with LazyTensor objects properly?

gpleiss:
KernelLinearOperator was throwing errors when computing the diagonal of a KeOps kernel. (This computation happens during preconditioning, which requires the diagonal of the already-formed kernel LinearOperator object.) This error was because KeopsLinearOperator.diagonal calls to_dense on the output of a batch kernel operation. However, to_dense is not defined for KeOps LazyTensors.

This PR is in some sense a hack fix to this bug (a less hack fix will require changes to KernelLinearOperator), but it is also a generally nice and helpful refactor that will improve KeOps kernels in general.

The fixes:
- KeOpsKernels now only define a forward function, that will be used both when we want to use KeOps and when we want to bypass it.
- KeOpsKernels now use a `_lazify_inputs` helper method, which (potentially) wraps the inputs as KeOpsLazyTensors, or potentially leaves the inputs as torch Tensors.
- The KeOps wrapping happens unless we want to bypass KeOps, which occurs when either (1) the matrix is small (below Cholesky size) or (2) when the use has turned off the `gpytorch.settings.use_keops` option (*NEW IN THIS PR*).

Why this is beneficial:
- KeOps kernels now follow the same API as non-KeOps kernels (define a forward method)
- The user now only has to define one forward method, that works in both the keops and non-keops cases
- The `diagonal` call in KeopsLinearOperator constructs a batch 1x1 matrix, which is small enough to bypass keops and thus avoid the current bug. (Hence why this solution is currently a hack, but could become less hacky with a small modification to KernelLinearOperator and/or the to_dense method in LinearOperator).

Other changes:
- Fix stability issues with the keops MaternKernel. (There were some NaN issues)
- Introduce a `gpytorch.settings.use_keops` feature flag.
- Clean up KeOPs notebook

[Fixes #2363]

### DIFF
--- a/gpytorch/kernels/keops/keops_kernel.py
+++ b/gpytorch/kernels/keops/keops_kernel.py
@@ -1,7 +1,8 @@
-from abc import abstractmethod
-from typing import Any
+import warnings
+from typing import Any, Tuple, Union
 
 import torch
+from linear_operator import LinearOperator
 from torch import Tensor
 
 from ... import settings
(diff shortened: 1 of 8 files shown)


gpytorch/kernels/keops/keops_kernel.py
```diff
@@ -1,48 +1,66 @@
...
+        x2_ = x2[..., None, :, :]
+        return x1_, x2_
+
+    def _use_keops(x1: Tensor, x2: Tensor) -> bool:
+        return False
+
     class KeOpsKernel(Kernel):
-        def __init__(self, *args: Any, **kwargs: Any):
-            raise RuntimeError("You must have KeOps installed to use a KeOpsKernel")
+        def __call__(self, *args: Any, **kwargs: Any) -> Union[LinearOperator, Tensor]:
```
jacobrgardner, line 61: Why are we being softer about this failure? If someone has gone to the trouble to import and try to instantiate a KeOpsKernel instead of the standard kernel, doesn't the hard stop make sense?

jacobrgardner (approved):
lgtm other than the question about whether not having keops installed should be a hard failure

PR #2423, opened 2023-10-17
Skip the warning in `gpytorch.lazy.__getattr__` if name starts with `_`

saitcakmak:
The current setup leads to the unittest `self.assertWarns...` to wrongly trigger a warning when entering `_AssertWarnsContext` (calls `getattr` on all modules: https://github.com/python/cpython/blob/main/Lib/unittest/case.py#L292-L294). Filtering out based on name starting with `_` eliminates the needless warning and keeps the correct behavior.

### DIFF
--- a/gpytorch/lazy/__init__.py
+++ b/gpytorch/lazy/__init__.py
@@ -66,11 +66,12 @@ __all__ = [
 
 
 def __getattr__(name: str) -> Any:
-    warnings.warn(
-        "GPyTorch will be replacing all LazyTensor functionality with the linear operator package. "
-        "Replace all references to gpytorch.lazy.*LazyTensor with linear_operator.operators.*LinearOperator.",
-        DeprecationWarning,
-    )
+    if not name.startswith("_"):
+        warnings.warn(
+            "GPyTorch will be replacing all LazyTensor functionality with the linear operator package. "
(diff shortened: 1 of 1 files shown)


esantorella:
Hm, this seems like a very roundabout way of preventing this warning from being raised in an `_AssertWarnsContext`. Could the check be changed from `if not name.startswith("_")` to `if name != "__warningregistry__"`? Failing that, it would be good to have a comment explaining why the underscore check exists.

Or maybe the warning should be somewhere else? Perhaps in [`deprecated_lazy_tensor`](https://github.com/cornellius-gp/gpytorch/blob/8429b0afaebf2628b68ad50405c194742a62b564/gpytorch/lazy/lazy_tensor.py#L29) or `deprecated_lazy_tensor.__init__`, so it's only raised when a `LazyTensor` is actually instantiated?

Balandat:
> These have been deprecated for over a year by now, so the right move might be to reap them altogether.

It'd be nice to get rid of the lazies and this special handling - cc @gpleiss

saitcakmak:
Should we land this and follow up with full deprecation later?

gpleiss:
Oops sorry for the slow reply. I think it makes sense to land this, and then we can go with full deprecation later.

In general, we should do a deprecation sweep and actually get rid of all of the things marked as deprecated. I think everything marked has deprecated has been deprecated for over a year at least.

gpleiss (approved):

PR #2452, opened 2023-12-04
Fixed `mean_cache` data extraction for dKG

yyexela:
Issue: dKG in BoTorch no longer works since I last ran it a few months ago. It appears to be from the `mean_cache` going from size `[num_fantasies, 1, 1, num_data]` to `[num_fantasies, num_fantasies, num_fantasies, num_data]`. What this change does is extract the relevant data of `mean_cache` while keeping the expected shape for the rest of the code (ie. `[num_fantasies, 1, num_data]`).

Not sure about a unit test for this since it depends on having BoTorch's `qKnowledgeGradient` class, among other dependencies.

See https://github.com/pytorch/botorch/pull/2137

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -339,7 +339,7 @@ class DefaultPredictionStrategy(object):
         # see https://github.com/cornellius-gp/gpytorch/pull/2317#discussion_r1157994719
         mean_cache = self.mean_cache
         if len(mean_cache.shape) == 4:
-            mean_cache = mean_cache.squeeze(1)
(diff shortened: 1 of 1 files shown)


yyexela:
Noticed main fails the same test so I'm assuming my code works and it's another problem

CedricTravelletti:
Any news on this PR? I confirm that there is interest in gradient-enabled GPs/BO for several application cases.

nrummel:
I am also hoping to have this change in the main branch, as I would like to use derivative information for Bayesian Optimization through BoTorch.

yyexela:
@gpleiss any way we can get this through? It's a simple fix that would allow people to run this botorch notebook, there's clearly an interest in the community to do work on dGPs (https://github.com/meta-pytorch/botorch/pull/2137).

PR #2457, opened 2023-12-11
`python` should also be a runtime dependency

jaimergp:
See https://github.com/conda/conda/issues/13412#issuecomment-1849600501
cc @Balandat


Balandat (approved):
Thanks!

Issue #2416, opened 2023-10-05
How to customise a finer-grained Hadamard GP?


ambigeV:
I have read the great example for both multitask GP and Hadamard GP, where the former one requires same input for all the tasks, and the latter one can accept distinct inputs for distinct tasks.

However, it is weird that for Hadamard GP, we can only setup the same mean function for all the tasks, the same RBF-kernel for all the tasks, and also lacks task dependent noise since only a single likelihood is used.

So my request or question is, how to preserve the nice property that Hadamard GP can accept distinct inputs but add into more features like the canonical multitask GP with task-dependent parameters?

Issue #877, opened 2019-09-23
[Docs]


sirBayesian:
# 📚 Documentation/Examples

Dear development team,

I am trying to use GPyTorch for two distinct research project, both dealing with multitask GP modeling.

1) With basis in our example on Hadamard multitask GPs (https://github.com/cornellius-gp/gpytorch/blob/master/examples/03_Multitask_GP_Regression/Hadamard_Multitask_GP_Regression.ipynb), I am succeeded to implement a multitask GP with the same likelihood noise variance for all outputs. My questions is now, how to define a distinct likelihood noise variance for each output? In my case, I am considering experiments conducted under different conditions at different labs, thus these may have different noise levels.

2) My basis for the second project is our example on Kronecker multitask GPs (https://github.com/cornellius-gp/gpytorch/blob/e50b9878a090b41eaf38f371cab2b938d5b2ebbc/examples/03_Multitask_GP_Regression/Multitask_GP_Regression.ipynb). The examples shows how to get a diagonal likelihood noise covariance matrix using the class “MultitaskGaussianLikelihood”. The likelihood covariance for this case is parameterized by a scalar “likelihood.raw_noise” and a vector “likelihood.noise_covar.raw_noise”. My question is in this regard how the covariance is parameterized; I presume that “likelihood.noise_covar.raw_noise” is the diagonal of the matrix, but how does “likelihood.raw_noise” relate to this?
In my case, I have 4 outputs and 14 inputs, and the corresponding learnt parameters (after transformation to output scale) are
“likelihood.raw_noise”: tensor([0.0270], grad_fn=<AddBackward0>)
“likelihood.noise_covar.raw_noise”: tensor([0.0275, 0.0387, 0.0177, 0.1554], grad_fn=<AddBackward0>)

I really like you software and would like to continue using it, also for NN modeling.

Thank you very much.

Best regards,
Sebastian

jacobrgardner:
@sirBayesian --

For 1, note that `IndexKernel` learns both `raw_var` and `covar_factor` parameters, so that the final task covariance matrix is `covar_factor * covar_factor.t() + raw_var * I`. The `raw_var` term here in particular effectively ends up being a noise variance for each output (e.g., the Kronecker product with the diagonal matrix or hadamard product with the subindexed matrix will result in this).

Effectively, the likelihood learns noise common to all tasks, and the `raw_var` term learns any "additional" noise.

[...]

PR #2481, opened 2024-02-23
Implement Hadamard Gaussian noise

TobyBoyne:
Addresses #877, #2416. Implements noise that is homoskedastic in inputs, but task-dependent.

Existing multi-task noises (https://docs.gpytorch.ai/en/latest/likelihoods.html#multi-dimensional-likelihoods) only work where all tasks are observed for each input. In the 'Hadamard' setting, where each input corresponds to exactly one task, there is no existing implementation that supports a different noise for each task.

See the updated `Hadamard Multitask GP Regression` for usage.

Unit tests, documentation, type hinting will all also be provided soon!


TobyBoyne:
Two areas that likely need improving:
- The implementation of `_shaped_noise_covar` probably needs some work. I'm not sure if there is some `linear_operator` magic to make this neater?
- The syntax can be a little questionable. For example, I don't like the fact that the task indexes need to be in a list in `mll(output, full_train_y, [full_train_i])`. It is also not obvious that the implementation requires the task indexes to be passed as the last argument to the model, since they are accessed by `params[0][-1]`. At least some type checking is required here?

Also, I should credit @jpfolch for help with this implementation!

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+
+from ..distributions import base_distributions, MultivariateNormal
+from ..likelihoods import _GaussianLikelihoodBase
+from .noise_models import MultitaskHomoskedasticNoise
+
+
+class HadamardGaussianLikelihood(_GaussianLikelihoodBase):
+    r"""
+    Likelihood for input-wise homo-skedastic noise, and task-wise
+    hetero-skedastic, i.e. we learn a different (constant) noise level for each fidelity.
```
Balandat, line 17:
```suggestion
    Likelihood for input-wise homoskedastic and task-wise heteroskedastic noise,
    i.e. we learn a different (constant) noise level for each fidelity.
```

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Args:
+        num_of_tasks : number of tasks in the multi output GP
+        noise_prior : any prior you want to put on the noise
+        noise_constraint : constraint to put on the noise
```
Balandat, line 22:
```suggestion
        num_of_tasks: Number of tasks in the multi-output GP.
        noise_prior: Prior for the noise.
        noise_constraint: Constraint on the noise value.
```

Balandat: `batch_shape` is missing here

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    def __init__(
+        self,
+        num_tasks,
+        noise_prior=None,
+        noise_constraint=None,
+        batch_shape=torch.Size(),
```
Balandat, line 30: can you add type annotations here?

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+    Args:
+        num_of_tasks : number of tasks in the multi output GP
+        noise_prior : any prior you want to put on the noise
+        noise_constraint : constraint to put on the noise
+    """
+
+    def __init__(
+        self,
+        num_tasks,
+        noise_prior=None,
```
Balandat, line 28: So we can only define a single prior for all tasks here - I think that's ok but probably worth mentioning this limitation in the docstring.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+    def raw_noise(self) -> torch.Tensor:
+        return self.noise_covar.raw_noise
+
+    @raw_noise.setter
+    def raw_noise(self, value: torch.Tensor) -> None:
+        self.noise_covar.initialize(raw_noise=value)
+
+    def _shaped_noise_covar(self, base_shape: torch.Size, *params: Any, **kwargs: Any):
+        # params contains task indexes
+        task_idxs = params[0][-1]
```
Balandat, line 60: Language nit
```suggestion
        task_idcs = params[0][-1]
```

Balandat: As you said, some runtime typechecking to make sure these are indeed the task indices would probably be good here.

Balandat: Also, if you can introduce some comments with the shapes of the intermediate tensors / matrices in this function that would be helpful.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+        mask = DiagLinearOperator(diag)
+        return (noise_base_covar_matrix @ mask).sum(dim=-3)
+
+    def forward(
+        self,
+        function_samples: torch.Tensor,
+        *params: Any,
+        **kwargs: Any,
+    ) -> base_distributions.Normal:
+        noise = self._shaped_noise_covar(function_samples.shape, *params, **kwargs).diag()
```
Balandat, line 74: It's a bit unfortunate that IIUC `self._shaped_noise_covar` always returns the full covariance but that you're calling `diag()` on it here - ideally we can have a way to just compute the diagonal here.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+        return base_distributions.Normal(function_samples, noise.sqrt())
+
+    def marginal(self, function_dist: MultivariateNormal, *params: Any, **kwargs: Any) -> MultivariateNormal:
+        mean, covar = function_dist.mean, function_dist.lazy_covariance_matrix
+        noise_covar = self._shaped_noise_covar(mean.shape, *params, **kwargs).squeeze(0)
```
Balandat, line 79: Why is the `.squeeze(0)` necessary here? Does it work for general batch shapes?

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _shaped_noise_covar(self, base_shape: torch.Size, *params: Any, **kwargs: Any):
+        # params contains task indexes
+        task_idxs = params[0][-1]
+        noise_base_covar_matrix = self.noise_covar(*params, shape=base_shape, **kwargs)
+
+        all_tasks = torch.arange(self.num_tasks)[:, None]
+        diag = torch.eq(all_tasks, task_idxs.mT)
+        mask = DiagLinearOperator(diag)
+        return (noise_base_covar_matrix @ mask).sum(dim=-3)
```
Balandat, line 66: Could you use a [`MaskedLinearOperator`](https://github.com/cornellius-gp/linear_operator/blob/2789d185f9d8b7f6e93bc6c350c031d5ad82a9fd/linear_operator/operators/masked_linear_operator.py#L12) here? Not sure if that would necessarily be better (seems like it wouldn't help with the `diag` comment below b/c of [this](https://github.com/cornellius-gp/linear_operator/blob/2789d185f9d8b7f6e93bc6c350c031d5ad82a9fd/linear_operator/operators/masked_linear_operator.py#L79)).

examples/03_Multitask_Exact_GPs/Hadamard_Multitask_GP_Regression.ipynb
```diff
@@ -29,7 +29,22 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
Balandat, line 32: ideally we could avoid emitting this unrelated & unnecessary warning

examples/03_Multitask_Exact_GPs/Hadamard_Multitask_GP_Regression.ipynb
```diff
@@ -108,7 +123,7 @@
     "        \n",
     "        return gpytorch.distributions.MultivariateNormal(mean_x, covar)\n",
     "\n",
-    "likelihood = gpytorch.likelihoods.GaussianLikelihood()\n",
+    "likelihood = gpytorch.likelihoods.HadamardGaussianLikelihood(num_tasks=2)\n",
```
Balandat, line 126: Rather than changing the tutorial to only use task-specific noises, it would be good to instead show both cases and discuss the pros and cons of both approaches in the tutorial. Ideally you could produce the output you shared in https://github.com/pytorch/botorch/issues/2765#issuecomment-2715182824

We could also try to highlight a potential failure mode where in the very low data regime andwith uniform noise levels across tasks, estimating task-specific noises results in worse performance due to the less parsimonious model.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
TobyBoyne: From my understanding, if `self._shaped_noise_covar` returns a LinearOperator, then calling `.diagonal` should only calculate the diagonal (as in `_MultiTaskGaussianLikelihoodBase.forward()`). I'm not quite sure if my `_shaped_noise_covar` function does return a correct LinearOperator that shows this behaviour. I'll look into `MaskedLinearOperator` as you mentioned below.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
TobyBoyne: Would it not be possible to have a multi-dimensional prior on your task noises? I had assumed that this will work by-default with the current implementation, but I will confirm this behaviour works as intended.
```python
noise_prior=LogNormalPrior(
    loc=torch.tensor([0., 1.]), 
    scale=torch.tensor([1., 1.])
)
```

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
TobyBoyne: Is there a "canonical" way to check that these are the task indices? I will check shape and dtype, but I wasn't sure if there was a way to guarantee that `task_idcs` contains what we want.

TobyBoyne:
Thanks for the thorough review, @Balandat! I've left a couple of comments and will update the PR soon to reflect your suggestions.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
Balandat: hmm you're right, that should work with a multi-dimensional prior (but not with different types of priors on different tasks)

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
Balandat: I don't think there is, unfortunately

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
Balandat: > I'm not quite sure if my _shaped_noise_covar function does return a correct LinearOperator that shows this behaviour

I was worried since you're using `.sum(dim=-3)` in  `self._shaped_noise_covar`, will that return a `LinearOperator` in your case?

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
TobyBoyne: The `squeeze` here is due to the shape of `MultitaskHomoskedasticNoise.forward`. It seems to add an extra dimension than I would expect, as the output is `(1, num_tasks, num_data, num_data)`. I have changed this to `.squeeze(-4)`, and moved it to where it is actually relevant for clarity.

```python
base_shape = torch.Size([40])
HomoskedasticNoise()(shape=base_shape).shape # (40, 40)
MultitaskHomoskedasticNoise(num_tasks=2)(shape=base_shape).shape # (1, 2, 40, 40)
```

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
TobyBoyne: I have checked, and yes - the function does return a `LinearOperator`, since `sum`ming a `LinearOperator` will return another `LinearOperator` here.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
TobyBoyne: I don't think it can be used here. In my code, `diag`, which is the diagonal of the mask matrix, has shape `(num_tasks, num_data)`. However, it looks like `MaskedLinearOperator` would only allow for a mask of shape `(num_data,)`.

This could feasibly be rewritten as a comprehension, summing over `num_tasks` different `MaskedLinearOperator`, but I don't think it would help with the readability of the code, which does already return a `LinearOperator`.

TobyBoyne:
Hi @Balandat, I've pushed some changes that should address some of the concerns raised:
- Added tests to check behaviour in batched settings
- Rearranged notebook to discuss pros and cons of task-specific noise in appendix
- Added the requested type hints and shape comments

I'm not sure why the readthedocs check is failing - seems to be having trouble importing `matplotlib` just for my notebook?

Hope you have a weekend! Please let me know if there are any other changes to be made.

Balandat (commented):
Thanks, code looks great. The notebook has no outputs included though? Can you re-run this and make sure that it includes the figures so that users can see the plots without having to run them themselves (and so they render on GitHub)? 

> I'm not sure why the readthedocs check is failing - seems to be having trouble importing matplotlib just for my notebook?

Not sure what this is - could be transient, let's just try again maybe it'll work on another push.

TobyBoyne:
Hi @Balandat, were there any other changes for this PR that need to be completed before it can be merged?

Balandat (approved):
No, this just fell off my radar. Thanks!

Issue #2264, opened 2023-02-02
[Bug]  Pickling error with `MultitaskGPModel`


charelstoncrabb:
# 🐛 Bug

I am getting the following error when trying to pickle the model object that is constructed in the [exact multi-task tutorial](https://docs.gpytorch.ai/en/stable/examples/03_Multitask_Exact_GPs/Multitask_GP_Regression.html):

```
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
Cell In[7], line 1
----> 1 pickle.dump(model, open('gpytorch_mtgp.pkl','wb'))

PicklingError: Can't pickle <function TriangularLinearOperator.to_dense at 0x7fe95395a8b0>: it's not the same object as linear_operator.operators.triangular_linear_operator.TriangularLinearOperator.to_dense


```

## To reproduce

Directly copy and pasted this tutorial into a local notebook:  https://docs.gpytorch.ai/en/stable/examples/03_Multitask_Exact_GPs/Multitask_GP_Regression.html

Then appending the following cell will produce the above error:
```
import pickle
pickle.dump(model, open('gpytorch_mtgp.pkl','wb'))
```

## Expected Behavior

Note that pickling the `likelihood` object does seem to work fine/as expected

## System information

OS: ubuntu 22.04
python: 3.8.10
gpytorch version: 1.9.1
torch version: 1.13.1+cu117

Balandat:
The issue is the following `@cached` decorator on `TriangularLinearOperator.to_dense`: https://github.com/cornellius-gp/linear_operator/blob/main/linear_operator/operators/triangular_linear_operator.py#L133

Removing that caching lets you pickle the model. 

@gpleiss, @SebastianAment any reason to use `@cached` here? Presumably the underlying `_tensor.to_dense()` would already use caching if that makes sense there.

PR #2482, opened 2024-02-27
Allow pickle of IndexKernels

TobyBoyne:
Currently, index kernels cannot be pickled. This PR moves an anonymous function to the module-level to allow pickling.

Current behaviour is that index kernels *with a prior* throw an error when pickling.
```python
from gpytorch.kernels import IndexKernel
from gpytorch.priors import LKJPrior
import pickle

kernel = IndexKernel(2, prior=LKJPrior(2, 1))
pickle.dumps(kernel)
```
This throws
` > AttributeError: Can't pickle local object 'IndexKernel.__init__.<locals>.<lambda>'`

There isn't an open issue for this in particular, but there is a request for pickle-able kernels. (#907, #2264)


Balandat (approved):
Thanks

PR #2488, opened 2024-03-02
Update multivariate_normal.py to use loc instead of mean

chrisyeh96:
Update MultivariateNormal to better match torch.distributions.MultivariateNormal

Also, add typing annotation that allows `loc` to be a LinearOperator


Balandat:
Thanks. I expect this to cause some backward compatibility problems though, e.g. [here](https://github.com/pytorch/botorch/blob/3e34a4fc35d5b01f7b6e256a92cc4b351d361cbe/test/acquisition/test_analytic.py#L402) in BoTorch. Not sure if it's worth the trouble of changing this in a way that deprecates the old arg name, but that's something you could do if you feel strongly about the renaming.

Issue #2468, opened 2024-01-14
[Bug] Fantasy Model Speeds up after 2048 datapoints


naefjo:
# 🐛 Bug

Hey all!
We have an online learning setting where we add datapoints to the GP using the `get_fantasy_model` method. I noticed that there is a significant performance increase (in terms of computation time) when the dataset crosses the 2048 datapoint threshold which I guess can be traced back to #1224 (specifically [these lines](https://github.com/cornellius-gp/linear_operator/blob/a00c08e1b85b6eb8499a7b64083058df8d2a00ab/linear_operator/utils/qr.py#L14-L21) in `linear_operator`). Is this still up to date? 

I tried recreating the computation time graph in #1224 but I am getting slightly different results where the crossover point is already at around 100 datapoints as compared to 2000. The test was performed in a google colab with a Tesla T4 runtime and locally on a Surface Book 2 with similar results (haven't gotten around to testing this on an RTX 4090 but I assume it will look similar).
<img width="521" alt="image" src="https://github.com/cornellius-gp/gpytorch/assets/54582454/f095f915-e96b-4e6d-8788-6dd1b5275723">

I suppose it makes sense to update this value or to create a setting where this parameter can be changed as it is hardware dependent right? I am happy propose a PR for this if you want:)

## To reproduce

Code snippet used to generate the plot
```python
import torch
from time import perf_counter
import matplotlib.pyplot as plt

cpu_times = []
gpu_times = []
tensor_sizes = [10,100,200,400,600,800,1000,1500,1750,2000,2500]
for size in tensor_sizes:
  A = torch.rand((size,size))
  time_now = perf_counter()
  torch.qr(A.cpu())
  cpu_times.append(perf_counter()-time_now)
  time_now = perf_counter()
  torch.qr(A.cuda())
  gpu_times.append(perf_counter()-time_now)

plt.semilogy(tensor_sizes,cpu_times, 'x-', label="cpu")
plt.semilogy(tensor_sizes,gpu_times, 'x-', label="gpu")
plt.legend()
plt.show()
```

## Expected Behavior
- `get_fantasy_model` should not have a sudden decrease in computation time at 2048 datapoints
- i.e. qr decomposition logic should prefer faster option

## System information
gpytorch version 1.11
torch version 2.1.0
ubuntu 22.04

naefjo:
Closing this as #2494 was merged

PR #2494, opened 2024-03-11
Bug: Exploit Structure in get_fantasy_strategy

naefjo:
Hello :)

This PR is related to #2468 and cornellius-gp/linear_operator#93.

The `DefaultPredictionStrategy`'s `get_fantasy_model` updates the gram matrix with new datapoints and updates the `lik_train_train_covar`'s `root_decomposition` and `root_inv_decomposition` caches by passing them to the constructor. However by using `to_dense` in lines 214-215, the caches in the `__init__` on line 69 and 72 respectively are constructed with `root` and `inv_root` of type `torch.tensor` which in [`RootLinearOperator.__init__`](https://github.com/cornellius-gp/linear_operator/blob/a00c08e1b85b6eb8499a7b64083058df8d2a00ab/linear_operator/operators/root_linear_operator.py#L19) will assign a `DenseLinearOperator` to `self.root` since `to_linear_operator` defaults to `DenseLinearOperator` if provided with a `torch.tensor`.

As a result in [`LinearOperator.cat_rows`](https://github.com/cornellius-gp/linear_operator/blob/a00c08e1b85b6eb8499a7b64083058df8d2a00ab/linear_operator/operators/_linear_operator.py#L1265), the object `E` will be of type `DenseLinearOperator` which in turn will fail the check for for triangular matrices [here](https://github.com/cornellius-gp/linear_operator/blob/a00c08e1b85b6eb8499a7b64083058df8d2a00ab/linear_operator/operators/_linear_operator.py#L1281). This once again leads to a `stable_pinverse` with QR decomposition instead of exploiting a fast triangular solve to compute the inverse.

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -10,7 +10,6 @@ from linear_operator.operators import (
     AddedDiagLinearOperator,
     BatchRepeatLinearOperator,
     ConstantMulLinearOperator,
-    DenseLinearOperator,
     InterpolatedLinearOperator,
     LinearOperator,
     LowRankRootAddedDiagLinearOperator,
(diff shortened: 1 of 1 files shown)


fteufel:
Thanks for finding this! I've been confused myself for a few weeks why `get_fantasy_model` wasn't speeding things up compared to just recomputing caches, but couldn't figure it out.

Can confirm this makes things faster.

jacobrgardner (approved):
Oof, thanks for catching this!

PR #2495, opened 2024-03-13
Fix flaky SVGP classification test

gpleiss:


Balandat (approved):
thanks!

Balandat:
@gpleiss looks like we have some zombie tests running forever here?

PR #2511, opened 2024-04-18
`ConstantKernel`

SebastianAment:
This commit adds `ConstantKernel`, moving the inference of additive constants into the kernel. 

Status quo: By default both GPyTorch and BoTorch are currently using a `ConstantMean` that is inferred during the optimization of the marginal likelihood. Because the mean parameter does not show up in the log determinant, it is un-regularized, which can lead to some counter-intuitive values for the constant mean, especially when the other parameters are constrained or subject to a non-trivial prior.

The inference of the posterior mean of the constant using the kernel approach is carried out with the usual linear algebraic formulation, which is more robust. See the following for an illustrative example of what can go wrong with the current `ConstantMean` approach, and how the `RBF + Constant` kernel behaves as expected.

Screenshot 2024-03-24 at 7 40 14 PM

### DIFF
--- a/gpytorch/kernels/constant_kernel.py
+++ b/gpytorch/kernels/constant_kernel.py
@@ -1,0 +1,123 @@
+#!/usr/bin/env python3
+
+from typing import Optional, Tuple
+
+import torch
+from torch import Tensor
+
+from ..constraints import Interval, Positive
+from ..priors import Prior
+from .kernel import Kernel
+
+
+class ConstantKernel(Kernel):
+    """
+    Constant covariance kernel for the probabilistic inference of constant coefficients.
+
+    ConstantKernel represents the prior variance `k(x1, x2) = var(c)` of a constant `c`.
+    The prior variance of the constant is optimized during the GP hyper-parameter
+    optimization stage. The actual value of the constant is computed (implicitly) using
+    the linear algebraic approaches for the computation of GP samples and posteriors.
+
+    The constant kernel `k_constant` is most useful as a modification of an arbitrary
+    base kernel `k_base`:
+    1) Additive constants: The modification `k_base + k_constant` allows the GP to
+    infer a non-zero asymptotic value far from the training data, which generally
+    leads to more accurate extrapolation. Notably, the uncertainty in this constant
+    value affects the posterior covariances through the posterior inference equations.
+    This is not the case when a constant prior mean is not used, since the prior mean
+    does not show up the posterior covariance and is regularized by the log-determinant
+    during the optimization of the marginal likelihood.
+    2) Multiplicative constants: The modification `k_base * k_constant` allows the GP to
+    modulate the variance of the kernel `k_base`, and is mathematically identical to
+    `ScaleKernel(base_kernel)` with the same constant.
+    """
+
+    has_lengthscale = False
+
+    def __init__(
+        self,
+        batch_shape: Optional[torch.Size] = None,
+        constant_prior: Optional[Prior] = None,
+        constant_constraint: Optional[Interval] = None,
+        active_dims: Optional[Tuple[int, ...]] = None,
+    ):
+        """Constructor of ConstantKernel.
+
+        Args:
+            batch_shape: The batch shape of the kernel.
+            constant_prior: Prior over the constant parameter.
+            constant_constraint: Constraint to place on constant parameter.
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -82,6 +82,7 @@ def find_version(*file_paths):
             "nbclient<=0.7.3",
             "nbformat<=5.8.0",
             "nbsphinx<=0.9.1",
+            "lxml_html_clean",
```
SebastianAment, line 85: Orthogonal dependency fix, see e.g. https://github.com/assafelovic/gpt-researcher/pull/424/files for a similar fix.

gpytorch/kernels/constant_kernel.py
```diff
@@ -0,0 +1,119 @@
...
+class ConstantKernel(Kernel):
+    """
+    Constant covariance kernel for the probabilistic inference of constant coefficients.
+
+    ConstantKernel represents the prior variance `k(x1, x2) = var(c)` of a constant `c`.
+    The prior variance of the constant is optimized during the GP hyper-parameter
+    optimization stage. The actual value of the constant is computed (implicitly) using
+    the linear algebraic approaches for the computation of GP samples and posteriors.
+
+    The kernel (`k_constant`) is most useful as a modification of an arbitrary `k_base`:
```
Balandat, line 21:
```suggestion
    The kernel (`k_constant`) is most useful as a modification of an arbitrary
    kernel (`k_base`):
```

gpytorch/kernels/constant_kernel.py
```diff
@@ -0,0 +1,119 @@
...
+        constant_prior: Optional[Prior] = None,
+        constant_constraint: Optional[Interval] = None,
+        active_dims: Optional[Tuple[int, ...]] = None,
+    ):
+        """Constructor of ConstantKernel.
+
+        Args:
+            batch_shape: The batch shape of the kernel.
+            constant_prior: Prior over the constant parameter.
+            constant_constraint: Constraint to place on constant parameter.
```
Balandat, line 49: `active_dims` is missing here

gpytorch/kernels/constant_kernel.py
```diff
@@ -0,0 +1,119 @@
...
+                lambda m: m.constant,
+                lambda m, v: m._set_constant(v),
+            )
+
+        if constant_constraint is None:
+            constant_constraint = Positive()
+        self.register_constraint("raw_constant", constant_constraint)
+
+    @property
+    def constant(self) -> torch.Tensor:
```
Balandat, line 73: nit: import `Tensor` to make the type hints less verbose
```suggestion
    def constant(self) -> Tensor:
```

gpytorch/kernels/constant_kernel.py
```diff
@@ -0,0 +1,119 @@
...
+
+        Returns:
+            A (batch_shape x n1 x n2)-dim, resp. (batch_shape x n1)-dim, tensor of
+            constant covariance values if diag is False, resp. True.
+        """
+        if last_dim_is_batch:
+            x1 = x1.transpose(-1, -2).unsqueeze(-1)
+            x2 = x2.transpose(-1, -2).unsqueeze(-1)
+
+        dtype = torch.promote_types(x1.dtype, x2.dtype)
```
Balandat, line 108: nice

setup.py
Balandat: Hmm that link doesn't really have much information on what exactly this fixes / why this dependency is needed. Is there more info on this?

Balandat (approved):
lgtm, thanks!


FYI @gpleiss @jacobrgardner - IIRC we had a constant kernel in the past but removed it. @SebastianAment's results on the numerical issues in the MAP estimation due to the unregularized kernel are interesting and suggest that using such a kernel may be preferable to using a `ConstantMean`.

setup.py
SebastianAment: There was a change in some package required for the sphinx compilation recently, which outsourced part of the functionality into this new package (there's more info in the description tab of the linked PR). 

The doc compilation will fail for any upcoming commit without this, so this is independent of the other changes made here.

SebastianAment: Copying over the error message from the link, which was raised on this commit as well before I added the line:

ImportError: lxml.html.clean module is now a separate project lxml_html_clean.
Install lxml[html_clean] or lxml_html_clean directly.

PR #2512, opened 2024-04-19
Matern52 grad

m-julian:
Add Matern52Grad kernel and unit tests. I've attached a pdf with derivations below.

[matern52grad_derivations.pdf](https://github.com/cornellius-gp/gpytorch/files/15042820/matern52grad_derivations.pdf)


jacobrgardner (approved):
Read through the doc, and this all looks right to me!

PR #2513, opened 2024-04-20
DOC: fix broken URL in periodic_kernel.py

partev:
fix broken URL to an article

### DIFF
--- a/gpytorch/kernels/periodic_kernel.py
+++ b/gpytorch/kernels/periodic_kernel.py
@@ -78,7 +78,7 @@ class PeriodicKernel(Kernel):
         >>> covar = covar_module(x)  # Output: LazyVariable of size (2 x 10 x 10)
 
     .. _David Mackay's Introduction to Gaussian Processes equation 47:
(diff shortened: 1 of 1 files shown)


gpytorch/kernels/periodic_kernel.py
```diff
@@ -78,7 +78,7 @@ class PeriodicKernel(Kernel):
         >>> covar = covar_module(x)  # Output: LazyVariable of size (2 x 10 x 10)
 
     .. _David Mackay's Introduction to Gaussian Processes equation 47:
-        http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.81.1927&rep=rep1&type=pdf
+        https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=e045b76dc5daf9f4656ac10b456c5d1d9de5bc84
```
Balandat, line 81: How about using the pdf linked on the original author's page? 
```suggestion
        https://www.inference.org.uk/mackay/gpB.pdf
```

Balandat (changes requested):

gpytorch/kernels/periodic_kernel.py
partev: makes sense. I updated the commit. with https://inference.org.uk/mackay/gpB.pdf URL

PR #2541, opened 2024-06-26
Corrected configuration of ``exclude`` statements in ``pre-commit`` configuration

JonathanWenger:
**Description**
Currently the pre-commit configuration assumes glob matching in the [``exclude`` statements](https://pre-commit.com/#top_level-exclude), even though these are assumed to be regular expressions. This causes the exclude statements to match the wrong files and shows warnings when running ``pre-commit``.

```bash
[WARNING] The 'exclude' field in hook 'flake8' is a regex, not a glob -- matching '/*' probably isn't what you want here
```

This PR replaces the glob statements with regular expressions in the pre-commit configuration.

**Related Issues**
This seems to be a [common error](https://github.com/pre-commit/pre-commit/issues/1702) and similar PRs have been made to other libraries (e.g. https://github.com/openml/openml-python/pull/1129)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
     hooks:
     -   id: flake8
         args: [--config=setup.cfg]
-        exclude: ^(examples/*)|(docs/*)
+        exclude: ^(examples/.*)|(docs/.*)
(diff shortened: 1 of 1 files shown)


Balandat (approved):
thanks

PR #2543, opened 2024-07-02
Include jaxtyping to allow for Tensor/LinearOperator typehints with sizes.

gpleiss:
Using the same trick in LinearOperator, sized Tensor/LinearOperator typehints are automatically included in the documentation.


gpleiss:
It's worth noting that `jaxtyping` is already implicitly required by `gpytorch`, since it is a requirement of `linear_operator`. This PR simply makes it possible to get pretty formatting of jaxtype'd TypeHints in the gpytorch documentation, following what we have in the `linear_opeartor` documentation.

PR #2544, opened 2024-07-02
[Breaking Change] Remove last_dim_is_batch

gpleiss:
This PR (finally) bumps our PyTorch requirement to 2.0.

Most use cases are not affected. There are two breaking changes as part of this PR:

1. Remove `AdditiveStructureKernel` and `ProductStructureKernel`. The functionality for these kernels can be implemented manually, and there is a new example notebook showing how to do so.
2. Remove `NewtonGirardAdditiveKernel`. Again, the functionality of this kernel can be implemented manually, using the new `sum_interaction_terms` utility (which contains the the meat of the NewtonGirardAdditiveKernel anyways).


gpleiss:
Still todo: update example notebooks, and add new documentation on additive / product kernels.

esantorella:
How are you thinking about warning users about breaking changes -- e.g. will there be deprecation warnings? (I do see that this is for merge into the `develop` branch.) Losing `AdditiveStructureKernel` and `last_dim_is_batch` will cause a few downstream breakages.

gpleiss:
@esantorella good question, and this is a good question for other GPyTorch 2.0 breaking changes more broadly.

We could release a new minor release while we're still preparing 2.0 that adds deprecation warnings to notify users about breaking changes.

@jacobrgardner @JonathanWenger thoughts?

JonathanWenger:
I think a minor release that has deprecations makes a lot of sense. It's not a lot of extra work to add the deprecations, and it might prevent some issues coming our way right after the planned 2.0 release.

gpytorch/kernels/grid_kernel.py
```diff
@@ -1,179 +1,176 @@
...
+    :ivar ragged_grid: A concatenation of all grid projections
+    :type ragged_grid: Tensor (max(M_i) x D)
+    :ivar full_grid: A full representation of the grid
+    :type ragged_grid: Tensor (N x D)
 
     .. _Fast kernel learning for multidimensional pattern extrapolation:
         http://www.cs.cmu.edu/~andrewgw/manet.pdf
     """
 
+    # TODO: update doc
```
JonathanWenger, line 44: Is this leftover from work during the PR? If not could you specify what part of the docs needs to be updated so potentially somebody else can fix this in the future?

```suggestion
```

gpytorch/kernels/linear_kernel.py
```diff
@@ -85,12 +85,8 @@ def _set_variance(self, value: Union[float, Tensor]):
             value = torch.as_tensor(value).to(self.raw_variance)
         self.initialize(raw_variance=self.raw_variance_constraint.inverse_transform(value))
 
-    def forward(
-        self, x1: Tensor, x2: Tensor, diag: Optional[bool] = False, last_dim_is_batch: Optional[bool] = False, **params
-    ) -> LinearOperator:
+    def forward(self, x1: Tensor, x2: Tensor, diag: Optional[bool] = False, **params) -> LinearOperator:
```
JonathanWenger, line 88:
```suggestion
    def forward(self, x1: Tensor, x2: Tensor, diag: bool = False, **params) -> LinearOperator:
```
Confusingly the `Optional` type hint should only be used if `None` is a valid value, not if the argument is optional. See also the [documentation on `typing.Optional`](https://docs.python.org/3/library/typing.html#typing.Optional:

> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional.

gpytorch/variational/grid_interpolation_variational_strategy.py
```diff
@@ -30,6 +30,11 @@ class GridInterpolationVariationalStrategy(_VariationalStrategy):
...
     :param ~gpytorch.variational.VariationalDistribution variational_distribution: A
         VariationalDistribution object that represents the form of the variational distribution :math:`q(\mathbf u)`
+
+    :ivar grid: The grid of points that the inducing points are based on.
+        The grid is stored as a matrix, where each column corresponds to the
+        projection of the grid onto one dimension.
+    :type grid: torch.Tensor (M x D)
     """
 
     def __init__(self, model, grid_size, grid_bounds, variational_distribution):
```
JonathanWenger, line 40: We could add type hints to this, while we are at it, but I'll leave that up to you.

JonathanWenger (approved):
Looks all good! Some minor documentation could still be polished (see comments).

PR #2545, opened 2024-07-05
fix: replace deprecated scipy.integrate.cumtrapz with cumulative_trapezoid

natsukium:
`scipy.integrate.cumulative_trapezoid` was introduced in scipy 1.6.0 and `scipy.integrate.cumtrapz` was removed in scipy 1.14.0

https://scipy.github.io/devdocs/release/1.6.0-notes.html#scipy-integrate-improvements https://scipy.github.io/devdocs/release/1.14.0-notes.html#expired-deprecations

This PR will replace an obsolete method and allow us to keep up with the latest scipy.
Compatibility with scipy<1.6.0 will be lost, but I don't think it will be a problem even if we use python 3.8.

### DIFF
--- a/gpytorch/kernels/spectral_delta_kernel.py
+++ b/gpytorch/kernels/spectral_delta_kernel.py
@@ -52,7 +52,7 @@ class SpectralDeltaKernel(Kernel):
         """
         import numpy as np
         from scipy.fftpack import fft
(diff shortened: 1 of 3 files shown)


Balandat (approved):
thanks

PR #2547, opened 2024-07-07
use common notation for normal distribution N(\mu, \sigma^2)

partev:
also make it consistent with https://docs.gpytorch.ai/en/stable/examples/01_Exact_GPs/Simple_GP_Regression.html


Balandat (approved):
thanks

Issue #2550, opened 2024-07-11
[Bug] Some Prior parameters are missing from the state_dict


saitcakmak:
# 🐛 Bug

This is an issue we came across when switching from `GammaPrior` to `LogNormalPrior`. Some Priors have their parameters registered as buffers but others do not (and cannot, will get to it later). The Priors that register their parameters as buffers have them appear in the parent module's `state_dict` but others do not. Take `GammaPrior` as an example:
```
>>> from gpytorch.priors.torch_priors import GammaPrior
>>> GammaPrior(3.0, 6.0).state_dict()
OrderedDict([('concentration', tensor(3.)), ('rate', tensor(6.))])
```
Both parameters are there. We can save them with the rest of the `state_dict` and load them later to a different model. We can also compare two models based on their `state_dict`s. Let's check `LogNormalPrior`:
```
>>> from gpytorch.priors.torch_priors import LogNormalPrior
>>> LogNormalPrior(3.0, 6.0).state_dict()
OrderedDict()
```
The `state_dict` is empty!

### Why

I tried to fix this but ran into issues due to how these parameters are stored in the parent `Distribution`s. Many priors use `_bufferize_attributes` to convert attributes to buffers, which simply **deletes** the existing attribute and registers it as a buffer. Since `register_buffer` errors out if there is an attribute with the same name, the deletion step is necessary. 
Some priors like `LogNormalPrior` are subclasses of PyTorch `TransformedDistribution`. `TransformedDistributions` do not actually have parameters, they simply pass these down to the base distributions. To make the parameters easily accessible through the distribution instances, the parameters of the base distribution are exposed as a `property`. And `property` are bound attributes that cannot be deleted, thus cannot be replaced with a buffer.

## Expected Behavior

All prior parameters should appear in the `state_dict`.

## System information

**Please complete the following information:**
- GPyTorch: Latest main on July 9, 2024: 26997efe8be4c23f49ee277a8ba773e7037affc7
- PyTorch: 2.3.1
- Computer OS: macOS 14.5

cc @Balandat @hvarfner

Balandat:
Is this something that could be "fixex" on the pytorch side? For instance, could `LogNormalPrior.state_dict()` and `LogNormalPrior.load_state_dict()` be overwritten in such a way that it saves the state dict of the base distribution? (not sure if transforms can be stateful, if they can then one would have to handle that as well for the general case)

saitcakmak:
[...]

PR #2551, opened 2024-07-24
Added ability for priors of transformed distributions to have their p…

hvarfner:
A suggested fix to https://github.com/cornellius-gp/gpytorch/issues/2550.

The proposed solution adds a `_transformed_` buffer to all of the priors of `TransformedDistributions` (LogNormal, HalfNormal, HalfCauchy), as their original parameters cannot be saved to `state_dict`. Thus these parameters will be saved and loaded from state_dict, and sets the value of the parameters of the corresponding base distributions.

Added some tests for these methods as well, hopefully they are sufficient.

### DIFF
--- a/gpytorch/priors/prior.py
+++ b/gpytorch/priors/prior.py
@@ -1,10 +1,13 @@
 #!/usr/bin/env python3
 
 from abc import ABC
+from typing import Any, Mapping
 
+from torch.distributions import TransformedDistribution
 from torch.nn import Module
 
 from ..distributions import Distribution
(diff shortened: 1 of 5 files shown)


Balandat:
This seems like a reasonable workaround to me for the time being, though we should ideally see if saving/loading state dicts can be better supported on the pytorch side.

dme65 (approved):
This lgtm as a workaround for now, but I agree with @Balandat that we should try to address this on the PyTorch side in the long-term. Please take care of the linting before merging

PR #2559, opened 2024-08-07
Avoid unnecessary memory allocation for covariance downdate in SGPR prediction strategy

JonathanWenger:
**Problem**
Currently in the SGPR prediction strategy the downdate term in the predictive covariance (shape ``num_test x num_test``) is constructed densely in memory causing unnecessary memory overhead.

**Fix**
Replaced the downdate term with a ``linear_operator.MatmulLinearOperator``.


gpleiss:
@jacobrgardner yeah, we'll need at least one more pre-2.0 minor release to include new deprecation warnings. I'm changing this PR to be merged into `main` rather than `develop`.

Issue #2585, opened 2024-09-12
[Bug] Multitask GP Regression fails for LeaveOneOutPseudoLikelihood mll


sahakpetros:
# 🐛 Bug

## To reproduce
run [Multitask GP Regression](https://github.com/cornellius-gp/gpytorch/blob/main/examples/03_Multitask_Exact_GPs/Multitask_GP_Regression.ipynb) example but replace ExactMarginalLogLikelihood with LeaveOneOutPseudoLikelihood
** Code snippet to reproduce **
in [Multitask GP Regression](https://github.com/cornellius-gp/gpytorch/blob/main/examples/03_Multitask_Exact_GPs/Multitask_GP_Regression.ipynb) replace
```python
mll = gpytorch.mlls.ExactMarginalLogLikelihood(likelihood, model)
```
with
```python
mll = gpytorch.mlls.LeaveOneOutPseudoLikelihood(likelihood, model)
```

** Stack trace/error message **
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[5], line 20
     18 optimizer.zero_grad()
     19 output = model(train_x)
---> 20 loss = -mll(output, train_y)
     21 loss.backward()
     22 print('Iter %d/%d - Loss: %.3f' % (i + 1, training_iterations, loss.item()))

File ~/.venv/lib/python3.11/site-packages/gpytorch/module.py:31, in Module.__call__(self, *inputs, **kwargs)
     30 def __call__(self, *inputs, **kwargs) -> Union[Tensor, Distribution, LinearOperator]:
---> 31     outputs = self.forward(*inputs, **kwargs)
     32     if isinstance(outputs, list):
     33         return [_validate_module_outputs(output) for output in outputs]

File ~/.venv/lib/python3.11/site-packages/gpytorch/mlls/leave_one_out_pseudo_likelihood.py:62, in LeaveOneOutPseudoLikelihood.forward(self, function_dist, target, *params)
     60 identity = torch.eye(*L.shape[-2:], dtype=m.dtype, device=m.device)
     61 sigma2 = 1.0 / L._cholesky_solve(identity, upper=False).diagonal(dim1=-1, dim2=-2)  # 1 / diag(inv(K))
---> 62 mu = target - L._cholesky_solve((target - m).unsqueeze(-1), upper=False).squeeze(-1) * sigma2
     63 term1 = -0.5 * sigma2.log()
[...]

PR #2586, opened 2024-09-12
fix bug #2585

sahakpetros:
fix bug 2585 https://github.com/cornellius-gp/gpytorch/issues/2585


gpytorch/mlls/leave_one_out_pseudo_likelihood.py
```diff
@@ -54,6 +54,7 @@ def forward(self, function_dist: MultivariateNormal, target: Tensor, *params) ->
         :param torch.Tensor target: :math:`\mathbf y` The target values
         :param dict kwargs: Additional arguments to pass to the likelihood's forward function.
         """
+        target = target.flatten()
```
gpleiss, line 57: This will cause errors if `target` is a batch of targets.

Please also add unit tests, covering both the batch and the non-batch case. Ideally, your tests should catch the error that you are trying to fix. I.e. you should have unit tests that fail before you add your code change, but pass after you add your code change.

gpleiss (changes requested):

PR #2590, opened 2024-09-20
Hotfix: fix `.diagonal()` calls for keops kernel matrices.

### LINKED ISSUE #2589: [Bug] _diagonal() fails when covariance is computed using keops kernels
# 🐛 Bug

Calls to _diagnonal() error when keops kernels are used. This showed up when tring to use keops kernels in a Multitask/Multi-output approximate GP with SVI. But below is a MWE of the failure. 

## To reproduce

```python
import gpytorch
import torch

train_x = torch.rand(100,2)
cov1 = gpytorch.kernels.RBFKernel()(train_x)
cov2 = gpytorch.kernels.keops.RBFKernel()(train_x)

cov1._diagonal() # Returns a tensor of shape (100,)
cov2._diagonal() # RuntimeError: The kernel RBFKernel is not equipped to handle and diag. Expected size torch.Size([100]). Got size torch.Size([100, 100])

```

```
  File "<stdin>", line 1, in <module>
  File "/Users/anthonycorso/Workspace/gpytorch/gpytorch/utils/memoize.py", line 59, in g
    return _add_to_cache(self, cache_name, method(self, *args, **kwargs), *args, kwargs_pkl=kwargs_pkl)
  File "/Users/anthonycorso/Workspace/gpytorch/gpytorch/lazy/lazy_evaluated_kernel_tensor.py", line 25, in wrapped
    output = method(self, *args, **kwargs)
  File "/Users/anthonycorso/Workspace/gpytorch/gpytorch/lazy/lazy_evaluated_kernel_tensor.py", line 126, in _diagonal
    raise RuntimeError(
RuntimeError: The kernel RBFKernel is not equipped to handle and diag. Expected size torch.Size([100]). Got size torch.Size([100, 100])
```

## Expected Behavior

The output of `cov2_diagonal()` should match the output of `cov1._diagonal()`, as there should be no difference between built-in and keops kernels

## System information

- GPyTorch Version  = v1.12
- torch version = 2.4.0
- MacOS

## Additional context
The problem can be fixed by handling the `diag` keyword argument in the `forward` call of the keops kernels. For example 
```
    def forward(self, x1, x2, diag=False, **kwargs):
        x1_ = x1 / self.lengthscale
        x2_ = x2 / self.lengthscale
        # return KernelLinearOperator inst only when calculating the whole covariance matrix
        K = KernelLinearOperator(x1_, x2_, covar_func=_covar_func, **kwargs)    
        return K.diagonal() if diag else K   
```

I am happy to submit a PR with this change (and a similar change for the other keops kernels, but I'm not 100% sure this is the "right" solution or if there is a more elegant intervention.

gpleiss:
[Fixes #2589]

Note to @jacobrgardner : as part of the 2.0 refactor where we remove `LazyEvaluatedKernelTensor`, I want to add a method `_symmetric_diag` to all kernels that replaces the `diag=True` mode of current kernels.

When the kernel matrix isn't symmetric and we call diagonal (not sure if this ever happens in gpytorch, but technically we support it...) the best we can do is what we currently do in `KernelLinearOperator.diagonal`.

So this PR is more of a temporary hotfix that should cause any regressions in behavior, and I'll make sure diag is fast 2.0 with #2342


gpytorch/test/base_keops_test_case.py
```diff
@@ -42,6 +42,11 @@ def test_forward_x1_eq_x2(self, ard=False, use_keops=True, **kwargs):
                 k2 = kern2(x1, x1).to_dense()
                 self.assertLess(torch.norm(k1 - k2), 1e-4)
 
+                # Test diagonal
+                d1 = kern1(x1, x1).diagonal(dim1=-1, dim2=-2)
+                d2 = kern2(x1, x1).diagonal(dim1=-1, dim2=-2)
+                self.assertLess(torch.norm(d1 - d2), 1e-4)
```
Balandat, line 48: This only tests that what the `diagonal` calls returns are the same, not that they actually compute the diagonal. You should add sth like this here.
```suggestion
                self.assertLess(torch.norm(d1 - d2), 1e-4)
                self.assertTrue(torch.equal(k1.diag(), d1))
```

Balandat (approved):

PR #2593, opened 2024-09-27
Allow for dimension-specific variance in linear kernel

sdaulton:
This adds support for diagonal prior variances with heterogeneous variances across dimensions (rather than requiring that each dimension have the same variance).

This also removes offset and num_dimensions, which were deprecated before 1.0.


Balandat (approved):
This looks good to me, thanks!

Issue #2563, opened 2024-08-09
[Bug] Sign error in Matern52KernelGrad


ingolia:
# 🐛 Bug

I think the sign of the covariance between function values and first derivatives is flipped in `Matern52KernelGrad`.

On a simple test case, the sign of the function-to-derivative covariance terms for `Matern52KernelGrad` is flipped relative to `RBFKernelGrad`. 

Intuitively, the function value at -1 should be negatively correlated with the derivative at 0 while the function value at +1 should be positively correlated with the derivative at 0 (think about the finite difference formula for estimating the derivative at 0 expressed in terms of a kernel function). The signs on `RBFKernelGrad` match that expectation, the signs on `Matern52KernelGrad` do not.

Based on the description in the code, `# 2) First gradient block, cov(f^m, omega^n_d)` has a negative sign, but in the derivations PDF on the pull request, this term corresponds to equation (6) for $\frac{\partial k}{\partial x_i^n}$ which does not have a negative sign due to cancellation of the negatives in (∂k/∂r) and (∂r/∂x^n). Conversely, `# 3) Second gradient block, cov(omega^m_d, f^n)` says, `the - signs on -outer2 and -five_thirds cancel out` but I think this should not happen for (∂r/∂x^m) where x^m appears with a positive sign in the distance function r.

Finally, I tried to use this code to predict derivative values from a GP trained on function values for some simple functions and it predicts values that are correct, except the wrong sign.

The signs on the derivative-to-derivative covariance terms seem correct.

## To reproduce

```python
import torch

torch.set_printoptions(precision=4, sci_mode=False, linewidth=160)

from linear_operator import to_dense
from gpytorch.kernels import RBFKernel, RBFKernelGrad, MaternKernel, Matern52KernelGrad

matern = MaternKernel()
matern_grad = Matern52KernelGrad()

rbf = RBFKernel()
rbf_grad = RBFKernelGrad()

inputs = torch.tensor([[-1.0, 0.0], [0.0, 0.0], [1.0, 0.0]])

print(matern(inputs).to_dense().detach())
print(matern_grad(inputs).to_dense().detach())

print(rbf(inputs).to_dense().detach())
print(rbf_grad(inputs).to_dense().detach())
```

Running this code, Matérn52 gives
```
tensor([[ 1.0000,  0.0000,  0.0000,  0.3056,  0.5822,  0.0000,  0.0336,  0.0816,  0.0000],
...
```

whereas RBF gives
```
[...]

PR #2598, opened 2024-10-16
sign fix for gradient blocks

m-julian:
The sign of the first and second gradient blocks for the `Matern52KernelGrad` kernel was wrong. This PR should fix issue https://github.com/cornellius-gp/gpytorch/issues/2563. I've also updates the tests.


gpleiss:
@m-julian thanks for the PR!

Would you be able to add a derivation of the gradient in the docstring of the kernel?

m-julian:
@gpleiss I've added the equations for the gradient and Hessian blocks. I've also linked to the initial PR in the docstring that contains a pdf file with the full derivations. That pdf should be correct, I just messed up the signs in the initial implementation. I've also added the Matern52KernelGrad to the `kernels.rst` file, so it shows up in the documentation.

Issue #2604, opened 2024-11-08
[Bug] Models not JIT traceable/exportable to TorchScript after Fantasization


SaiAakash:
# 🐛 Bug

Fantasization / conditioning model on new data points renders the model unexportable to TorchScript/not traceable with JIT. Models cannot be JIT traced/exported to Torchscript once `get_fantasy_model` method is called.

## To reproduce

** Code snippet to reproduce **
```python
import math
import torch
import gpytorch

X = torch.linspace(0, 1, 100)
y = torch.sin(X * (2 * math.pi)) + torch.randn(X.size()) * 0.2

train_x, fantasy_x = X[:90], X[90:]
train_y, fantasy_y = y[:90], y[90:]


class ExactGPModel(gpytorch.models.ExactGP):
    def __init__(self, train_x, train_y, likelihood):
        super(ExactGPModel, self).__init__(train_x, train_y, likelihood)
        self.mean_module = gpytorch.means.ConstantMean()
        self.covar_module = gpytorch.kernels.ScaleKernel(gpytorch.kernels.RBFKernel())

    def forward(self, x):
        mean_x = self.mean_module(x)
        covar_x = self.covar_module(x)
        return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)


likelihood = gpytorch.likelihoods.GaussianLikelihood()
model = ExactGPModel(train_x, train_y, likelihood)

training_iter = 5
model.train()
likelihood.train()
optimizer = torch.optim.Adam(model.parameters(), lr=0.1)
mll = gpytorch.mlls.ExactMarginalLogLikelihood(likelihood, model)

for i in range(training_iter):
    optimizer.zero_grad()
    output = model(train_x)
    loss = -mll(output, train_y)
    loss.backward()
    optimizer.step()

model.eval()
X_test = torch.linspace(0, 1, 51)
pred = model(X_test)
fantasized_model = model.get_fantasy_model(fantasy_x, fantasy_y)

class MeanVarModelWrapper(torch.nn.Module):
    def __init__(self, gp):
        super().__init__()
        self.gp = gp

    def forward(self, x):
        output_dist = self.gp(x)
        return output_dist.mean, output_dist.variance


with torch.no_grad(), gpytorch.settings.fast_pred_var(), gpytorch.settings.trace_mode():
    fantasized_model.eval()
    pred = fantasized_model(X_test)  # Do precomputation
[...]

PR #2605, opened 2024-11-09
Detach `new_covar_cache` to enable JIT tracing of models after fantasization

SaiAakash:
This PR deals with #2604 . It's a one-liner where the `new_covar_cache` is detached so that models which are conditioned on new data/observations using `get_fantasy_model` can now be JIT traced/exported to TorchScript.


jacobrgardner:
> @gpleiss In all other places in GPyTorch, tensors are detached only when the `detach_test_caches` setting is on. I am wondering why in this case we have to detach when this setting is not on ?
> 
> Should it be something like,
> 
> ```
> if detach_test_caches.on():
>         new_covar_cache = new_lt.root_inv_decomposition().root.detach()
> else:
>         new_covar_cache = new_lt.root_inv_decomposition().root
> ```

@SaiAakash Yes, exactly -- we should only detach when the setting is on.

@gpleiss does `trace_mode` imply `detach_test_caches`? If no, it probably should I guess?

SaiAakash:
@jacobrgardner I have made the requested change. The cache is now detached only when the `detach_test_caches` setting is on.

Balandat (approved):

PR #2611, opened 2024-12-02
gpytorch.module: fix typing annotation

chrisyeh96:
This pull request makes 2 main changes:

1. I made a mistake in the typing annotations in #2608, which I am correcting in this pull request. Specifically, I realized that `MutableSet[str]` should be `MutableSet[Prior]`. This pull request fixes this mistake.

2. I updated the typing annotation for the `setting_closure` to have no return type (i.e., it returns None) to better match how setting closures are actually used throughout the GPyTorch codebase and examples. Previously, `setting_closure` was annotated with a "Self" return type, but the return value of `setting_closure` is never used anywhere in the GPyTorch codebase.


gpleiss (approved):

gpleiss:
Thanks @chrisyeh96 !

PR #2617, opened 2025-01-13
add offset parameter to linear kernel

sdaulton:
Without having an offset parameter, the linear model always has zero variance at the origin.

download - 2025-01-13T104233 326

Adding the offset parameter centers the point at which the model has zero variance. Note: one could just center the training inputs at zero to achieve this, but doing this automatically in the model is convenient.

download - 2025-01-13T104307 381

### DIFF
--- a/gpytorch/kernels/linear_kernel.py
+++ b/gpytorch/kernels/linear_kernel.py
@@ -67,6 +67,7 @@ class LinearKernel(Kernel):
             self.register_prior("variance_prior", variance_prior, lambda m: m.variance, lambda m, v: m._set_variance(v))
 
         self.register_constraint("raw_variance", variance_constraint)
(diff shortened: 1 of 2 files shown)


gpytorch/kernels/linear_kernel.py
```diff
@@ -67,6 +67,12 @@ def __init__(
             self.register_prior("variance_prior", variance_prior, lambda m: m.variance, lambda m, v: m._set_variance(v))
 
         self.register_constraint("raw_variance", variance_constraint)
+        self.register_parameter(
```
Balandat, line 70: Should we be concerned about backward compatibility for loading old state dicts? Or are we ok breaking this?

sdaulton: Fair question. I think a breaking change is probably fine here?

gpleiss: Since we initialize to zero it should probably be fine?

sdaulton:
Abandoning this in favor of https://github.com/pytorch/botorch/pull/2680

PR #2621, opened 2025-01-16
Support broadcasting batch shapes in MTMVN.from_independent_mvns

saitcakmak:
Adds support for broadcasting batch shapes in `MTMVN.from_independent_mvns`, allowing the mix of MVNs coming from batched and non-batched models. This is useful in BoTorch's `ModelListGP`, where we want to support mixing batched / ensemble models with non-batched models.


dme65 (approved):
LGTM!

PR #2623, opened 2025-01-17
Update MVN.expand() to support non-lazy MVN & reuse scale_tril where possible

saitcakmak:
The previous implementation of `MVN.expand()` would error out when MVN was not lazy (since `self._covar` is not defined) and would discard the `scale_tril`, requiring its re-computation down the line. This PR updates `MVN.expand()` to support both lazy and non-lazy MVN and to re-use `scale_tril` whenever possible.

Test Plan:
Added unittests for `MVN.expand()`.

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -158,7 +158,7 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
             )
             super(MultivariateNormal, new).__init__(loc=new_loc, scale_tril=new_scale_tril)
             # Set the covar matrix, since it is always available for GPyTorch MVN.
-            new.covariance_matrix = self.covariance_matrix
+            new.covariance_matrix = self.covariance_matrix.expand(batch_size + self.covariance_matrix.shape[-2:])
         return new
 
     def get_base_samples(self, sample_shape: torch.Size = torch.Size()) -> Tensor:
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -136,10 +136,28 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
         See :py:meth:`torch.distributions.Distribution.expand
         <torch.distributions.distribution.Distribution.expand>`.
         """
-        new_loc = self.loc.expand(torch.Size(batch_size) + self.loc.shape[-1:])
-        new_covar = self._covar.expand(torch.Size(batch_size) + self._covar.shape[-2:])
-        res = self.__class__(new_loc, new_covar)
-        return res
+        new_loc = self.loc.expand(batch_size + self.loc.shape[-1:])
+        if self.islazy:
+            new_covar = self._covar.expand(batch_size + self._covar.shape[-2:])
+            new = self.__class__(mean=new_loc, covariance_matrix=new_covar)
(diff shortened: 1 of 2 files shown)


gpytorch/distributions/multivariate_normal.py
```diff
@@ -136,10 +136,30 @@ def expand(self, batch_size: torch.Size) -> MultivariateNormal:
...
+        batch_size = torch.Size(batch_size)
+        new_loc = self.loc.expand(batch_size + self.loc.shape[-1:])
+        if self.islazy:
+            new_covar = self._covar.expand(batch_size + self._covar.shape[-2:])
+            new = self.__class__(mean=new_loc, covariance_matrix=new_covar)
+            if self.__unbroadcasted_scale_tril is not None:
+                # Reuse the scale tril if available.
+                new.__unbroadcasted_scale_tril = self.__unbroadcasted_scale_tril.expand(
+                    batch_size + self.__unbroadcasted_scale_tril.shape[-2:]
+                )
```
Balandat, line 149: very nice!

gpytorch/distributions/multivariate_normal.py
```diff
@@ -136,10 +136,30 @@ def expand(self, batch_size: torch.Size) -> MultivariateNormal:
...
+            # Constructing it from scale_tril will avoid unnecessary computation.
+            # Initialize using  __new__, so that we can skip __init__ and use scale_tril.
+            new = self.__new__(type(self))
+            new._islazy = False
+            new_scale_tril = self.__unbroadcasted_scale_tril.expand(
+                batch_size + self.__unbroadcasted_scale_tril.shape[-2:]
+            )
+            super(MultivariateNormal, new).__init__(loc=new_loc, scale_tril=new_scale_tril)
+            # Set the covar matrix, since it is always available for GPyTorch MVN.
+            new.covariance_matrix = self.covariance_matrix
```
Balandat, line 161: Don't we need to expand this one as well?

gpytorch/distributions/multivariate_normal.py
```diff
@@ -136,10 +136,30 @@ def expand(self, batch_size: torch.Size) -> MultivariateNormal:
...
+            new = self.__class__(mean=new_loc, covariance_matrix=new_covar)
+            if self.__unbroadcasted_scale_tril is not None:
+                # Reuse the scale tril if available.
+                new.__unbroadcasted_scale_tril = self.__unbroadcasted_scale_tril.expand(
+                    batch_size + self.__unbroadcasted_scale_tril.shape[-2:]
+                )
+        else:
+            # Non-lazy MVN is represented using scale_tril in PyTorch.
+            # Constructing it from scale_tril will avoid unnecessary computation.
+            # Initialize using  __new__, so that we can skip __init__ and use scale_tril.
```
Balandat, line 153: This seems slightly hairy - would an alternative here be to just add support for `MultivariateNormal` to be initialized directly from the scale_tril?

saitcakmak: Yeah, that'd be an option. I didn't want to touch it since it can potentially affect correctness of MVN. But it doesn't seem like `_covar` is directly used anywhere, so it may be safer than I initially though.

Balandat: I feel like we could also just make it such that we error out if both of them are provided.

PR #2624, opened 2025-01-17
Implement MVN.unsqueeze

saitcakmak:
Implements an `unsqueeze` method for MVN, that constructs a new MVN with the underlying tensors / linear operators unsqueezed along the given batch dimension. The choice of unsqueezing along the batch dimensions is consistent with the definition of `expand`.

`MVN.expand` allows us to start with `MVN.batch_shape = b2 x b1` and add additional batch dimensions to the left, so that `new_MVN.batch_shape = b3 x b2 x b1`. `unsqueeze`, when combined with `expand` will allow us to add batch dimensions in the middle, so that `new_MVN.batch_shape = b2 x b3 x b1 or b2 x b1 x b3`.

The use case for this is to match MVNs produced by two models of different batch shapes. Assume that `m1.batch_shape = mb2 x mb1` and `m2.batch_shape = mb1`. If we evaluate these two models with `X = xb1 x q x d`, we get `m1(X).batch_shape. = xb1 x mb2 x mb1` and `m2(X) = xb1 x mb1`. By calling `m2(X).unsqueeze(-2).expand(-1, mb2, -1)`, we can match the batch shapes of the two MVNs, allowing them to be combined into a single MTMVN using the `from_independent_mvns` method.

### DIFF
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -158,7 +158,7 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
             )
             super(MultivariateNormal, new).__init__(loc=new_loc, scale_tril=new_scale_tril)
             # Set the covar matrix, since it is always available for GPyTorch MVN.
-            new.covariance_matrix = self.covariance_matrix
+            new.covariance_matrix = self.covariance_matrix.expand(batch_size + self.covariance_matrix.shape[-2:])
         return new
 
     def get_base_samples(self, sample_shape: torch.Size = torch.Size()) -> Tensor:
--- a/gpytorch/distributions/multivariate_normal.py
+++ b/gpytorch/distributions/multivariate_normal.py
@@ -136,10 +136,28 @@ class MultivariateNormal(TMultivariateNormal, Distribution):
         See :py:meth:`torch.distributions.Distribution.expand
         <torch.distributions.distribution.Distribution.expand>`.
         """
-        new_loc = self.loc.expand(torch.Size(batch_size) + self.loc.shape[-1:])
-        new_covar = self._covar.expand(torch.Size(batch_size) + self._covar.shape[-2:])
-        res = self.__class__(new_loc, new_covar)
-        return res
+        new_loc = self.loc.expand(batch_size + self.loc.shape[-1:])
+        if self.islazy:
+            new_covar = self._covar.expand(batch_size + self._covar.shape[-2:])
+            new = self.__class__(mean=new_loc, covariance_matrix=new_covar)
+            if self.__unbroadcasted_scale_tril is not None:
+                # Reuse the scale tril if available.
+                new.__unbroadcasted_scale_tril = self.__unbroadcasted_scale_tril.expand(
+                    batch_size + self.__unbroadcasted_scale_tril.shape[-2:]
+                )
+        else:
+            # Non-lazy MVN is represented using scale_tril in PyTorch.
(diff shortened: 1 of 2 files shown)


gpytorch/distributions/multivariate_normal.py
```diff
@@ -136,10 +136,62 @@ def expand(self, batch_size: torch.Size) -> MultivariateNormal:
...
+        else:
+            # Non-lazy MVN is represented using scale_tril in PyTorch.
+            # Constructing it from scale_tril will avoid unnecessary computation.
+            # Initialize using  __new__, so that we can skip __init__ and use scale_tril.
+            new = self.__new__(type(self))
+            new._islazy = False
+            new_scale_tril = self.__unbroadcasted_scale_tril.unsqueeze(dim)
+            super(MultivariateNormal, new).__init__(loc=new_loc, scale_tril=new_scale_tril)
+            # Set the covar matrix, since it is always available for GPyTorch MVN.
+            new.covariance_matrix = self.covariance_matrix.unsqueeze(dim)
```
Balandat, line 193: A lot of this is duplicate from the `expand()` code above. Can we make a helper to reuse most of this? Or potentially consider allowing to instantiate a new MVN from the `scale_tril` directly as suggested on the other PR. Could be a class method `MultivariateNormal.from_scale_tril()` or the like if we don't want to change the `__init__()` signature.

gpytorch/distributions/multivariate_normal.py
```diff
@@ -136,10 +136,62 @@ def expand(self, batch_size: torch.Size) -> MultivariateNormal:
...
+    def unsqueeze(self, dim: int) -> MultivariateNormal:
+        r"""
+        Constructs a new MultivariateNormal with the batch shape unsqueezed
+        by the given dimension.
+        For example, if `self.batch_shape = torch.Size([2, 3])` and `dim = 0`, then
+        the returned MultivariateNormal will have `batch_shape = torch.Size([1, 2, 3])`.
+        If `dim = -1`, then the returned MultivariateNormal will have
+        `batch_shape = torch.Size([2, 3, 1])`.
+        """
+        # If dim is negative, get the positive equivalent.
```
Balandat, line 173: may also want to check that `dim` is valid (i.e. not 5 if there are only 2 batch dims)?

gpytorch/distributions/multivariate_normal.py
saitcakmak: They follow the same pattern but I wouldn't call them duplicates. You could make a helper that takes in the operation to apply to the tensors, but there are two separate operations for `expand` (for loc & covar), so I don't know if the resulting helper would improve the code readability.

>  Could be a class method MultivariateNormal.from_scale_tril() or the like if we don't want to change the __init__() signature.

This would avoid the issue of checking for `covar` & `scale_tril` compatibility in `__init__`. If we're not using `__init__`, would this just extract the `self.__new__` based construction used here into a separate method? I suppose this'd work for non-lazy `scale_tril` as is. We'd have to have a separate case for lazy, at which point we end up with a duplicate of `__init__` and I again question the added value of it over just keeping this as is.

Balandat: If we have a method exposed that allows constructing the MVN from `scale_tril` (either as part of an updated `__init__()` method or as a separate method) then I would expect that people will see and use this for other purposes going forward (but they most likely won't look for the logic for this somewhere deep in the code as in this PR. 

Not going to die on that hill, but the fact that we don't expose a natural way of doing this seems like a gap we ideally could address.

Balandat (approved):

Issue #2620, opened 2025-01-15
Possible typo in IndexKernel documentation


parrangoiz:
The `IndexKernel` [documentation](https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/kernels/index_kernel.py#L49-L50) says that `raw_var` is the element-wise log of the $`\mathbf{v}`$ vector. If I'm not mistaken it is actually the softplus of $`\mathbf{v}`$ instead, assuming the default constraint (`Positive`). Maybe a bit fastidious but this threw me for a spin when I was trying to manually reconstruct the task covariance matrix by hand from the factors and the diagonal.

gpleiss:
Good catch. Mind submitting a PR with the fix?

parrangoiz:
Sure no problem! First wanted to double check I was right since I'm new to the library.

PR #2625, opened 2025-01-17
Fix error in IndexKernel docstring

parrangoiz:
Fixes https://github.com/cornellius-gp/gpytorch/issues/2620


saitcakmak (approved):
Thanks!

PR #2626, opened 2025-01-22
Unpin jaxtyping

saitcakmak:
The required jaxtyping version conflicts with the requirements of other libraries. The code appears to be compatible with newer versions, so attempting to unpin to see if anything breaks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ readme = open("README.md").read()
 
 torch_min = "2.0"
 install_requires = [
-    "jaxtyping==0.2.19",
+    "jaxtyping",
     "mpmath>=0.19,<=1.3",  # avoid incompatibiltiy with torch+sympy with mpmath 1.4
     "scikit-learn",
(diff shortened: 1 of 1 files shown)


Balandat (commented):
Hmm, does the github CI actually test if the jaxtyping annotations work? IIUC this is for static typing and it doesn't seem from the workflow definitions that this is tested. 

cc @gpleiss, @jacobrgardner

saitcakmak:
@Balandat looks like the linear operator CI is supposed to be run with type checking (I didn't check the details of it) and passes (except for the docs, which are failing elsewhere as well). I don't know if there is anything else that specifically uses jaxtyping within GPyTorch (other than linear operators).

Balandat:
cool lezdoit

Balandat (approved):

saitcakmak:
Figured out how to run jaxtyping. `pytest --jaxtyping-packages=typeguard.typechecked` only brings up one unrelated test failure.

Issue #2534, opened 2024-06-17
[Feature Request] Gaussian Process Ordinal Regression (GPOR)


j-adamczyk:
# 🚀 Feature Request

Support for ordinal regression (ordinal classification).

## Motivation

Ordinal regression, or classification with ordered classes, naturally arises for a lot of problems where the target label are discrete preferences. This can be starts (X out of 5), ratings (X of out 10), age (ordered integer) etc. It requires a support for ordinal likelihood model to work.

## Pitch

I came upon this problem when trying to reimplement the ordinal regression from GPflow (which fully supports it) to GPyTorch: https://gpflow.github.io/GPflow/develop/notebooks/advanced/ordinal_regression.html. The basic tutorial code is:

GPflow code is very simple:
```
# construct ordinal likelihood - bin_edges is the same as unique(Y) but centered
bin_edges = np.array(np.arange(np.unique(Y).size + 1), dtype=float)
bin_edges = bin_edges - bin_edges.mean()
likelihood = gpflow.likelihoods.Ordinal(bin_edges)

# build a model with this likelihood
m = gpflow.models.VGP(
    data=(X, Y), kernel=gpflow.kernels.Matern32(), likelihood=likelihood
)

# fit the model
opt = gpflow.optimizers.Scipy()
opt.minimize(m.training_loss, m.trainable_variables, options=dict(maxiter=100))
```

As far as I can tell, this is currently impossible to reimplement in GPyTorch, as it has no `Ordinal` likelihood analogue. But this should be really the only addition needed to make this work.

bendavidsteel:
Added a pull request for this here: #2639

PR #2639, opened 2025-03-06
Adding ordinal likelihood

bendavidsteel:
I needed an ordinal likelihood for some of my own work, and saw Issue #2534, so I thought I'd make this contribution!
Tested it a bit, seems to work pretty well. It's pretty much just using the same idea as in GPflow so nothing new.

Tests and docs provided.

### DIFF
--- a/gpytorch/likelihoods/ordinal_likelihood.py
+++ b/gpytorch/likelihoods/ordinal_likelihood.py
@@ -1,68 +1,112 @@
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import torch
 from torch import Tensor
+from torch.distributions import Categorical
 
-from ..constraints import Positive
+from ..constraints import Interval, Positive
 from ..distributions import MultivariateNormal
-from .likelihood import Likelihood
+from ..priors import Prior
+from .likelihood import _OneDimensionalLikelihood
+
 
 def inv_probit(x, jitter=1e-3):
     """
     Inverse probit function (standard normal CDF) with jitter for numerical stability.
-    
+
     Args:
         x: Input tensor
         jitter: Small constant to ensure outputs are strictly between 0 and 1
-        
+
     Returns:
         Probabilities between jitter and 1-jitter
     """
     return 0.5 * (1.0 + torch.erf(x / torch.sqrt(torch.tensor(2.0)))) * (1 - 2 * jitter) + jitter
 
-class OrdinalLikelihood(Likelihood):
-    def __init__(self, bin_edges):
-        """
-        An ordinal likelihood for regressing over ordinal data.
 
-        The data are integer values from 0 to k, and the user must specify (k-1)
-        'bin edges' which define the points at which the labels switch. Let the bin
-        edges be [a₀, a₁, ... aₖ₋₁], then the likelihood is
+class OrdinalLikelihood(_OneDimensionalLikelihood):
+    r"""
+    An ordinal likelihood for regressing over ordinal data.
 
-        p(Y=0|F) = ɸ((a₀ - F) / σ)
-        p(Y=1|F) = ɸ((a₁ - F) / σ) - ɸ((a₀ - F) / σ)
-        p(Y=2|F) = ɸ((a₂ - F) / σ) - ɸ((a₁ - F) / σ)
+    The data are integer values from :math:`0` to :math:`k`, and the user must specify :math:`(k-1)`
+    'bin edges' which define the points at which the labels switch. Let the bin
+    edges be :math:`[a_0, a_1, ... a_{k-1}]`, then the likelihood is
+
+    .. math::
+        p(Y=0|F) &= \Phi((a_0 - F) / \sigma)
+
+        p(Y=1|F) &= \Phi((a_1 - F) / \sigma) - \Phi((a_0 - F) / \sigma)
+
+        p(Y=2|F) &= \Phi((a_2 - F) / \sigma) - \Phi((a_1 - F) / \sigma)
+
         ...
-        p(Y=K|F) = 1 - ɸ((aₖ₋₁ - F) / σ)
 
-        where ɸ is the cumulative density function of a Gaussian (the inverse probit
-        function) and σ is a parameter to be learned.
+        p(Y=K|F) &= 1 - \Phi((a_{k-1} - F) / \sigma)
 
-        A reference is :cite:t:`chu2005gaussian`. 
+    where :math:`\Phi` is the cumulative density function of a Gaussian (the inverse probit
+    function) and :math:`\sigma` is a parameter to be learned.
 
-        :param bin_edges: A tensor of shape (k-1) containing the bin edges.
-        """
+    From Chu et Ghahramani, Journal of Machine Learning Research, 2005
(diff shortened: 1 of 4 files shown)


gpytorch/likelihoods/ordinal_likelihood.py
```diff
@@ -0,0 +1,112 @@
...
+    def sigma(self, value: Tensor) -> None:
+        self._set_sigma(value)
+
+    def _set_sigma(self, value: Tensor) -> None:
+        if not torch.is_tensor(value):
+            value = torch.as_tensor(value).to(self.raw_sigma)
+        self.initialize(raw_sigma=self.raw_sigma_constraint.inverse_transform(value))
+
+    def forward(self, function_samples: Tensor, *args: Any, data: Dict[str, Tensor] = {}, **kwargs: Any) -> Categorical:
+        if isinstance(function_samples, MultivariateNormal):
```
gpleiss, line 96: I don't think we should ever run this line of code? Forward should only receive function samples.

gpleiss (commented):
Looks mostly good to me; see one comment.

gpytorch/likelihoods/ordinal_likelihood.py
bendavidsteel: Ah can't remember why I added that, removed.

bendavidsteel:
Comment addressed

rick-osmo:
I could use this; @gpleiss @bendavidsteel Do you need any help getting this merged?

bendavidsteel:
I believe we're just waiting on approval from @gpleiss

j-adamczyk:
@gpleiss @bendavidsteel how about merging this? GPOR would be enormously useful for our research

gpytorch/likelihoods/ordinal_likelihood.py
```diff
@@ -0,0 +1,108 @@
...
+    Inverse probit function (standard normal CDF) with jitter for numerical stability.
+
+    Args:
+        x: Input tensor
+        jitter: Small constant to ensure outputs are strictly between 0 and 1
+
+    Returns:
+        Probabilities between jitter and 1-jitter
+    """
+    return 0.5 * (1.0 + torch.erf(x / torch.sqrt(torch.tensor(2.0)))) * (1 - 2 * jitter) + jitter
```
kayween: If `x` is a GPU tensor, then this line will trigger a device error as `torch.tensor(2.0)` is always a CPU tensor.
```suggestion
    return 0.5 * (1.0 + torch.erf(x / math.sqrt(2.0))) * (1 - 2 * jitter) + jitter
```

gpytorch/likelihoods/ordinal_likelihood.py
```diff
@@ -0,0 +1,108 @@
...
+    def forward(self, function_samples: Tensor, *args: Any, data: dict[str, Tensor] = {}, **kwargs: Any) -> Categorical:
+        # Compute scaled bin edges
+        scaled_edges = self.bin_edges / self.sigma
+        scaled_edges_left = torch.cat([scaled_edges, torch.tensor([torch.inf], device=scaled_edges.device)], dim=-1)
+        scaled_edges_right = torch.cat([torch.tensor([-torch.inf], device=scaled_edges.device), scaled_edges])
+
+        # Calculate cumulative probabilities using standard normal CDF (probit function)
+        function_samples = function_samples.unsqueeze(-1)
+        scaled_edges_left = scaled_edges_left.reshape(1, -1)
+        scaled_edges_right = scaled_edges_right.reshape(1, -1)
```
kayween: Will these two lines work in batch settings where the batch shape is non-empty?

gpytorch/likelihoods/ordinal_likelihood.py
```diff
@@ -0,0 +1,108 @@
...
+        self,
+        bin_edges: Tensor,
+        batch_shape: torch.Size = torch.Size([]),
+        sigma_prior: Prior | None = None,
+        sigma_constraint: Interval | None = None,
+    ) -> None:
+        super().__init__()
+
+        self.num_bins = len(bin_edges) + 1
+        self.register_parameter("bin_edges", torch.nn.Parameter(bin_edges, requires_grad=False))
```
kayween:
```suggestion
        self.register_buffer("bin_edges", bin_edges)
```
nit: I think it makes more sense to register this as a buffer instead since we won't update the bin edges?

On the flip side, does it make sense to set `requires_grad=True` so that we learn the bin edges during model fitting? (Some packages choose to do so; see [here](https://github.com/facebookresearch/aepsych/blob/main/aepsych/likelihoods/ordinal.py).) IIUC, we only learn `sigma` here but the bin edges are fixed. I am wondering if this could limit the expressiveness of the likelihood.

gpytorch/likelihoods/ordinal_likelihood.py
```diff
@@ -0,0 +1,108 @@
...
+import torch
+from torch import Tensor
+from torch.distributions import Categorical
+
+from ..constraints import Interval, Positive
+from ..priors import Prior
+from .likelihood import _OneDimensionalLikelihood
+
+
+def inv_probit(x, jitter=1e-3):
```
kayween:
```suggestion
def inv_probit(x: Tensor, jitter: float = 1e-3):
```
Let's annotate these variables.

gpytorch/likelihoods/ordinal_likelihood.py
```diff
@@ -0,0 +1,108 @@
...
+        return self.raw_sigma_constraint.transform(self.raw_sigma)
+
+    @sigma.setter
+    def sigma(self, value: Tensor) -> None:
+        self._set_sigma(value)
+
+    def _set_sigma(self, value: Tensor) -> None:
+        if not torch.is_tensor(value):
+            value = torch.as_tensor(value).to(self.raw_sigma)
+        self.initialize(raw_sigma=self.raw_sigma_constraint.inverse_transform(value))
```
kayween: nit: We've already annotated `value` as tensor. So we could drop the if-statement here? Also, maybe we could merge this method with the sigma setter method above?

kayween:
I've merged the latest main into this PR. Multiple users have expressed interests in the ordinal likelihood implementation. So it would be great to get this merged.

It's been a while since this PR opened. @bendavidsteel apologies for the delay on our end! I left some additional comments. I am wondering if you still have the capacity to work on this PR.

gpytorch/likelihoods/ordinal_likelihood.py
bendavidsteel: I changed it to `torch.tensor(2.0, device=x.device)' to keep torch speed + fix device error

gpytorch/likelihoods/ordinal_likelihood.py
bendavidsteel: Fixed + added test to confirm

gpytorch/likelihoods/ordinal_likelihood.py
bendavidsteel: I changed the code to allow for learnable edges but default to fixed

bendavidsteel:
Addressed comments @kayween

PR #2664, opened 2025-08-30
Supporting passing full train inputs, rather than only supporting task indices to in `HadamardGaussianLikelihood`

sdaulton:
This also updates the pre-commit config, which seemed to be broken on my conda setup

### DIFF
--- a/gpytorch/likelihoods/hadamard_gaussian_likelihood.py
+++ b/gpytorch/likelihoods/hadamard_gaussian_likelihood.py
@@ -103,7 +103,7 @@ class HadamardGaussianLikelihood(_GaussianLikelihoodBase):
             noise_base_covar_matrix = noise_base_covar_matrix.squeeze(
                 -4
             )  # (*batch_shape, num_tasks, num_data, num_data)
-        all_tasks = torch.arange(self.num_tasks).unsqueeze(-1)  # (num_tasks, 1)
+        all_tasks = torch.arange(self.num_tasks, device=task_idcs.device).unsqueeze(-1)  # (num_tasks, 1)
         diag = torch.eq(all_tasks, task_idcs.mT)  # (num_tasks, num_data)
         mask = DiagLinearOperator(diag)  # (num_tasks, num_data, num_data)
         return (noise_base_covar_matrix @ mask).sum(dim=-3)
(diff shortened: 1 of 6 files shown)


gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -53,6 +53,7 @@ def __init__(
         noise_prior: Optional[Prior] = None,
         noise_constraint: Optional[Interval] = None,
         batch_shape: torch.Size = torch.Size(),
+        task_feature_index: Optional[int] = None,
```
Balandat, line 57: Let's add this to the docstring

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
```diff
@@ -81,17 +83,27 @@ def raw_noise(self, value: torch.Tensor) -> None:
...
             raise ValueError("Task indices must be provided.")
-        task_idcs = params[0][-1]
+        task_idcs = params[0] if torch.is_tensor(params[0]) else params[0][0]
+        if task_idcs.shape[-1] > 1:
+            if self.task_feature_index is None:
+                raise ValueError("Task indices must be a single dimension if task_feature_index is not provided.")
+            # handle case where full inputs are passed in,
+            # rather than just the task indices
+            task_idcs = task_idcs[..., self.task_feature_index].unsqueeze(-1)
+        task_idcs = task_idcs.long()
```
Balandat, line 97: curious why this is needed (before there was no casting to `long`)

Balandat (commented):
Overall this seems reasonable to me. 

For the pre-commit config update, I assume this changes the formatting due to the flake8 and ufmt update? If that's the case we should probably pull this out into a separate PR, and apply the proper formatting across the code base so that people don't continually run into errors with their pre-commit setup going forward.

gpytorch/likelihoods/hadamard_gaussian_likelihood.py
sdaulton: Previously this assumed that task_idcs (a long-value tensor) was passed in as params[0], but this is adding support for extracting task_idcs from the full train inputs, so we need to cast it to `long`.

sdaulton:
I added a separate PR for the precommit change https://github.com/cornellius-gp/gpytorch/pull/2665

Balandat (approved):

PR #2665, opened 2025-09-02
Update precommit

sdaulton:
see title. This runs flake8 and fixes the errors


Balandat:
Hmm surprising these are the only changes coming from a big version bump like this one.

Also what's going on with the test failures here?

sdaulton:
using `isinstance` changed the logical behavior (since it is `True` for subclasses), so I update the PR to use `is` instead`

Balandat (approved):
Thanks!

PR #2671, opened 2025-10-09
RFF/linear prediction strategy fix

gpleiss:
A small bug in the prediction strategy meant that we were sometimes computing the Cholesky factorizations of large matrices, rather than relying on the woodbury formula for RFFKernel matrices. This PR fixes that issue, and adds a unit test to prevent future regressions

It also uses this same strategy for LinearKernels, which can similarly use woodbury.


gpytorch/models/exact_prediction_strategies.py
```diff
@@ -732,19 +732,42 @@ def exact_predictive_covar(self, test_test_covar, test_train_covar):
...
+        posterior_predictive_covar = RootLinearOperator(features @ covar_cache)
+        return (posterior_predictive_mean, posterior_predictive_covar)
 
     @property
-    @cached(name="covar_cache")
-    def covar_cache(self):
+    @cached(name="mean_cache")
+    def mean_cache(self) -> tuple[Tensor, Tensor]:
+        # HACK: this method computes both the mean and the covar cache to avoid
+        # redundant computations.
```
Balandat: This is not exactly an intuitive name for what it does. Is this being called anywhere else / does this require this name or can we change this to sth like `mean_covar_cache`?

gpleiss: Good call; it's now named `mean_covar_cache`. And now the tests are all passing.

PR #2675, opened 2025-11-02
Remove unused eps parameter from relevant kernels

### LINKED ISSUE #2672: [Bug] `eps` parameter ineffective in `MaternKernel`
# 🐛 Bug

The `MaternKernel` takes as an argument the parameter `eps`, which, as described [here](https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/kernels/matern_kernel.py#L59) is "the minimum value that the lengthscale can take (prevents divide by zero errors)." However, as far as I can tell this parameter isn't actually used by the kernel (it gets set in the superclass [here](https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/kernels/kernel.py#L182), but isn't used by the `MaternKernel` to impose a lower limit on the lengthscale). Empirically, I have witnessed the lengthscale fall far below `1e-6` during optimisation as a result of this.

Perhaps I'm missing something, or perhaps this parameter is a relic which is indeed no longer used (in which case I'd request that it's removed as it's misleading). At first glance, I suppose one way of handling this would be to alter the default `lengthscale_constraint` in the base `Kernel` class to be `GreaterThan` with a sensible `lower_bound` (for instance, `1e-6`), instead of `Positive`. But if this causes issues for other kernels, then I suppose it would make sense to at least think about changing to an approach such as this in the `MaternKernel` class, and removing the ineffective `eps` parameter.

Thomas-Christie:
Fixes #2672 by removing `eps` parameter which is no longer used. At some point in the past, this was used to impose a lower bound on `lengthscale`, but such a lower bound is now enforced via `lengthscale_constraint` or similar parameters.

All unit tests still pass when running `python -m unittest`.

### DIFF
--- a/gpytorch/kernels/cosine_kernel.py
+++ b/gpytorch/kernels/cosine_kernel.py
@@ -35,9 +35,6 @@ class CosineKernel(Kernel):
             Set this if you want to apply a prior to the period length parameter.  Default: `None`
         period_length_constraint (Constraint, optional):
             Set this if you want to apply a constraint to the period length parameter. Default: `Positive`.
-        eps (float):
-            The minimum value that the lengthscale/period length can take
-            (prevents divide by zero errors). Default: `1e-6`.
 
     Attributes:
         period_length (Tensor):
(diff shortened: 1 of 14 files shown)


gpleiss:
Thanks @Thomas-Christie !

Issue #2674, opened 2025-11-02
[Bug] Rational Quadratic Kernel with Deep GPs - crashes due to shape mismatch


mrlj-hash:
# 🐛 Bug

When using the RQKernel with variational Deep GPs, a crash occurs because of a shape mismatch. 

## To reproduce

** Code snippet to reproduce **
```
###################### IMPORTS ##################################
from gpytorch.means import ConstantMean
from tqdm import tqdm
import torch
from torch.optim import Adam
import gpytorch as gpy
from torch.utils.data import TensorDataset, DataLoader
from gpytorch.models.deep_gps import DeepGP, DeepGPLayer
from gpytorch.distributions import MultivariateNormal
from gpytorch.likelihoods import GaussianLikelihood
# from gpytorch.likelihoods import MultitaskGaussianLikelihood (for comparison)
from gpytorch.optim import NGD
from gpytorch.variational import TrilNaturalVariationalDistribution
from gpytorch.variational import VariationalStrategy
from gpytorch.mlls import VariationalELBO as VELBO
from gpytorch.mlls import DeepApproximateMLL as DLL
from gpytorch.kernels import ScaleKernel, RQKernel
from gpytorch.kernels import RBFKernel, MaternKernel #(for comparison)
from scipy.io import loadmat
from math import floor

########################### SETUP ########################################
seed = 1
train_prop = 0.8
learning_rate = 0.1 # for NGD
batch_size = 64
hp_learn_rate = 0.01 # for Adam
num_ips = 20 # total number of inducing points (each GP in the network gets an equal number. If not integer, rounding)
num_tasks = 1 # number of tasks - assuming that multioutput deep GPs with 1 task is comparable to a single output DGP
num_samples = 50 # amount of sampling for deep GPs
num_hidden = 3 # number of GPs in the hidden layer
num_epochs = 20 # number of training epochs

torch.manual_seed(seed)

################ DATA SETUP ##############################################
data = torch.Tensor(loadmat('Data/elevators.mat')['data']) # the elevator dataset as used in the DGP notebook
X = data[:, :-1]
X = X - X.min(0)[0]
X = 2 * (X / X.max(0)[0]) - 1
y = data[:, -1]

train_n = int(floor(train_prop * len(X)))
train_x = X[:train_n, :].contiguous()
train_y = y[:train_n].contiguous()

test_x = X[train_n:, :].contiguous()
test_y = y[train_n:].contiguous()

[...]

PR #2677, opened 2025-11-09
Fix for #2674 - Corrected sizes for alpha in RQKernel when using Deep GPs

mrlj-hash:
When using the RQKernel for Deep GPs, a shape mismatch as described in #2674 occurs. This was identified to be a consequence of unsqueezing the alpha parameter one extra time in a for loop. The fix, ensuring consistent behaviour in deep GPs and other model families, is to unsqueeze alpha manually and conditionally based on flags diag and last_dim_is_batch.

### DIFF
--- a/test/kernels/test_rq_kernel.py
+++ b/test/kernels/test_rq_kernel.py
@@ -223,44 +223,35 @@ class TestRQKernel(unittest.TestCase, BaseKernelTestCase):
         self.assertLess(torch.norm(kernel.alpha - actual_value), 1e-5)
 
     def test_extra_batch_dim(self):
-        num_input_dims = 3
-        num_samples = 6
-        num_output_dims = 2
-        batch_shape = torch.Size([num_output_dims])
-        batch_size = 5
-        num_elements = num_samples * num_output_dims * batch_size * num_input_dims
-        # first test a and b as the same (diag distance matrix)
-        a = torch.arange(num_elements).view(num_samples, num_output_dims, batch_size, num_input_dims)
-        b = a.clone()
-        a = a.float()
-        b = b.float()
+        a = torch.linspace(0.0, 1.0, 120).view(5, 4, 3, 2)
+        b = torch.linspace(0.0, 1.0, 80).view(5, 4, 2, 2)
 
-        lengthscale = 1.0
-        alpha = 3.0
-        kernel = RQKernel(batch_shape=batch_shape, ard_num_dims=num_input_dims)
-        kernel.initialize(alpha=alpha)
-        kernel.initialize(lengthscale=lengthscale)
+        lengthscale = 0.1
+        alpha = torch.linspace(0.5, 1.0, 4)  # different `alpha` for each batch
+
+        # NOTE: Why do we pass down `batch_shape = (4,)` as opposed to `batch_shape = (5, 4)`?
+        # When training deep GPs with variational inference, there is an extra batch dimension added to the data, which
+        # corresponds to the likelihood samples. We are testing if `alpha` broadcasts properly in these cases.
+        kernel = RQKernel(batch_shape=torch.Size([4]))
+        kernel.initialize(lengthscale=lengthscale, alpha=alpha)
         kernel.eval()
 
-        res = kernel(a, b, diag=True).to_dense()
-        diff = kernel.covar_dist(a, b, square_dist=True, diag=True)
-        actual = diff.div_(2 * alpha).add_(1.0).pow(-alpha)
-        self.assertLess(torch.norm(res - actual), 1e-5)
+        # First check the diagonal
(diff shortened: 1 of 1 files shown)


test/kernels/test_rq_kernel.py
```diff
@@ -222,6 +222,18 @@ def test_initialize_alpha(self):
         actual_value = torch.tensor(3.0).view_as(kernel.alpha)
         self.assertLess(torch.norm(kernel.alpha - actual_value), 1e-5)
 
+    def test_last_layer_alpha(self):
```
kayween: This test only verifies the kernel matrix can be computed, but it does not verify if the computation is correct. Can we improve the test case by checking the kernel matrix against the ground truth (like other test cases in this file)?

gpytorch/kernels/rq_kernel.py
```diff
@@ -65,6 +65,11 @@ def postprocess_rq(dist_mat):
             alpha = self.alpha
             for _ in range(1, len(dist_mat.shape) - len(self.batch_shape)):
                 alpha = alpha.unsqueeze(-1)
+
+            # for loop above overruns by 1 in deep GPs due to additional sampling dimension
+            if len(alpha) > 1 and alpha.shape[0] != dist_mat.shape[0]:
```
kayween: My main concern with this if-statement is that it seems a bit ad hoc. This line checks if the first dimension is broadcastable, and does an unsqueeze if not.

If `alpha.shape[0]` is equal to `dist_mat.shape[0]` by accident, we might accidentally broadcast `alpha` and `dist_mat` in a wrong way and potentially compute the kernel matrix incorrectly, even in situations when an unsqueeze is indeed required.

test/kernels/test_rq_kernel.py
kayween: Can we do deterministic tests like other test cases in this file (i.e., no `torch.randn`)? Note that determinism is not guaranteed on different devices even with fixed random seeds by `torch.manual_seed`.

You could do something like
```
x1 = torch.arange(24).view(2, 2, 3, 2)
x2 = torch.arange(16).view(2, 2, 2, 2)
```

kayween (changes requested):
Hi @mrlj-hash, thank you for the PR! I left some comments. Can you also update the PR description to include a brief summary (e.g., what was the bug and how it's fixed).

gpytorch/kernels/rq_kernel.py
```diff
@@ -65,6 +65,11 @@ def postprocess_rq(dist_mat):
             alpha = self.alpha
             for _ in range(1, len(dist_mat.shape) - len(self.batch_shape)):
                 alpha = alpha.unsqueeze(-1)
```
kayween: @jacobrgardner @gpleiss 

The root cause for the shape mismatch is this line. Due to the likelihood samples in VI, `dist_mat` has an extra dimension that is not included in `batch_shape`. As a result, `alpha` was not unsqueezed into the correct shape.

In most cases (`diag=False` and `last_dim_is_batch=False`), the shape of `dist_mat` is 
```
dist_mat.shape = (num_likelihood_samples, *batch_shape, n, m)
```
Meanwhile, the shape of `alpha` is
```
alpha.shape = (*batch_shape, 1)
```
The above for-loop would do unsqueeze twice, yielding a shape `(*batch_shape, 1, 1, 1)` that is unbroadcastable with `dist_mat`.

The main challenge here is that there might be extra batch dimensions before AND after `batch_shape`. E.g., the likelihood samples would add a batch dimension before `batch_shape`, and `last_dim_is_batch=True` would add a batch dimension after `batch_shape`. So I think it makes more sense to unsqueeze `alpha` manually as opposed to the for-loop:
```suggestion
            if not diag:
                alpha = alpha.unsqueeze(-1)
            if last_dim_is_batch:
                alpha = alpha.unsqueeze(-1)
```
An implementation based on the above code snippet passes all test cases locally. Does it look sensible to you guys?

gpleiss: That seems reasonable to me!

kayween:
@mrlj-hash It's recommended to install a linter locally (e.g., flake8) or a pre-commit hook aid your development, see [CONTRIBUTING.md](https://github.com/cornellius-gp/gpytorch/blob/main/CONTRIBUTING.md). (That way those minor  lint errors in the github workflow can be fixed more easily.)

test/kernels/test_rq_kernel.py
mrlj-hash: Hi @kayween, I have modified the unit test to be deterministic following your suggestion.

kayween (approved):

kayween:
@gpleiss This PR is in good shape now. Do you want to give it a stamp before I merge it? We'll probably need to do a squash merge as there are many small commits in the PR.

PR #2681, opened 2025-11-17
RFF/linear prediction strategy fix (take 2)

gpleiss:
Attempt 2 at #2671 , but this time adding a naive fantasy strategy for BoTorch models.

The fantasy strategy is `O(d^2 (n+m))`, so it is no more efficient than re-conditioning the model. Without writing custom cuda kernels, it is unlikely to be more efficient.

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -763,13 +763,15 @@ class LinearPredictionStrategy(DefaultPredictionStrategy):
 
         # Get cached computations of the expensive components of the posterior
         mean_cache, covar_cache = self.mean_covar_cache
-        # mean_cache: ... x num_features x 1
-        # covar_cache: ... x num_features x num_features
+        # mean_cache: X^T (X X^T + D)^{-1} y, ... x num_features x 1
+        # covar_cache: chol( (I - X^T (X X^T + D)^{-1} X)^{-1} ), ... x num_features x num_features
 
         # Compute the linear model posterior
         features = test_test_covar.root.to_dense() * constant.sqrt()
         posterior_predictive_mean = (features @ mean_cache).squeeze(-1) + test_mean
-        posterior_predictive_covar = RootLinearOperator(features @ covar_cache)
+        posterior_predictive_covar = RootLinearOperator(
+            torch.linalg.solve_triangular(covar_cache, features.mT, upper=False).mT
+        )
         return (posterior_predictive_mean, posterior_predictive_covar)
 
     @property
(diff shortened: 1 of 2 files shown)


gpytorch/models/exact_prediction_strategies.py
```diff
@@ -753,39 +774,45 @@ def covar_cache(self):
...
-            self.exact_predictive_covar(test_test_covar, test_train_covar),
-        )
-
-    def exact_predictive_covar(self, test_test_covar, test_train_covar):
-        if settings.skip_posterior_variances.on():
-            return ZeroLinearOperator(*test_test_covar.size())
+    @property
+    @cached(name="mean_cache")
+    def mean_cache(self):
+        raise RuntimeError("This method should not be called!")
```
Balandat: maybe add why and that it should be `mean_covar_cache` instead?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -753,39 +774,45 @@ def covar_cache(self):
...
+    @property
+    @cached(name="covar_cache")
+    def covar_cache(self):
+        raise RuntimeError("This method should not be called!")
 
-        covar_cache = self.covar_cache
-        factor = test_test_covar.root.to_dense() * constant.sqrt()
-        res = RootLinearOperator(factor @ covar_cache)
+    def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_output, **kwargs):
+        # It's easier to just recompute the strategy from scratch
```
Balandat: Add a comment here that this is an inefficient way of doing this?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -732,19 +732,40 @@ def exact_predictive_covar(self, test_test_covar, test_train_covar):
...
+class LinearPredictionStrategy(DefaultPredictionStrategy):
     def __init__(self, train_inputs, train_prior_dist, train_labels, likelihood):
         super().__init__(train_inputs, train_prior_dist, train_labels, likelihood)
         self.train_prior_dist = self.train_prior_dist.__class__(
             self.train_prior_dist.mean, self.train_prior_dist.lazy_covariance_matrix.evaluate_kernel()
         )
 
-    def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_output, **kwargs):
-        raise NotImplementedError("Fantasy observation updates not yet supported for models using RFFs")
+    def exact_prediction(self, joint_mean, joint_covar):
```
Balandat: Add a docstring and type here?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -753,39 +774,45 @@ def covar_cache(self):
...
+        res = self.__class__(
+            train_inputs=full_inputs,
+            train_prior_dist=self.train_prior_dist.__class__(full_output.mean, full_output.lazy_covariance_matrix),
+            train_labels=full_targets,
+            likelihood=self.likelihood.get_fantasy_likelihood(**kwargs),
+        )
+        _, _ = res.mean_covar_cache  # populate the cache
         return res
 
+    def exact_predictive_mean(self, *args, **kwargs):
```
Balandat, line 866: same here, add why and what should be called instead

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -753,39 +774,45 @@ def covar_cache(self):
             constant = torch.tensor(1.0, dtype=lt.dtype, device=lt.device)
 
         train_factor = lt.root.to_dense()
-        train_train_covar = self.lik_train_train_covar
-        inner_term = (
+        train_train_covar = self.lik_train_train_covar.evaluate_kernel()  # HOTFIX here: need to resolve lazy eval!
```
Balandat: @gpleiss what's the deal with this? Why exactly is this necessary? This seems to be causing some other test failure here when running on a GPU, presumably due to changed numerical behavior: https://github.com/meta-pytorch/botorch/blob/main/test/models/test_fully_bayesian.py#L834

gpleiss: We need to resolve the LazyEvaluatedKernelTensor, because otherwise the $\boldsymbol K + \sigma^2 I$ matrix will be an `AddedDiagLinearOperator(LazyEvaluatedKernelTensor, DiagLinearOperator)` rather than a `LowRankRootAddedDiagLinearOperator(LowRankRootLinearOperator, DiagLinearOperator)`. The latter is necessary to get the structure we want.

I can remove this `evaluate_kernel` if I add the following line before calling the super constructor:

```python
train_prior_dist = train_prior_dist.__class__(
    train_prior_dist.mean, train_prior_dist.lazy_covariance_matrix.evaluate_kernel()
)
```

(which I originally added after calling the super constructor; moving it before is probably a better idea). I'm not sure if that will fix the numerical issues on the GPU test though?

gpytorch/models/exact_prediction_strategies.py
gpleiss: Done

PR #2688, opened 2025-12-18
A variational strategy optimized for large batch stochastic training

kayween:
This PR is a variational strategy implementation that is optimized for large batch stochastic training on data center GPUs (in particular A100). This yields about 3.81x speed up for large batch training on A100 compared to the current variational strategy implementation.

# What's Changed?
Let `m` be the number of inducing points and `n` be the number of test data points (i.e., the batch size). The complexity of the forward pass is O(m^2 n) assuming `m << n`.

The running time bottleneck is computing the predictive variance, which requires computing
```
k_XZ K_ZZ^{-1/2} (S - I) K_ZZ^{-1/2} k_ZX
```

**GPyTorch.** The current GPyTorch implementation computes it in this way:
```
interp_term = K_ZZ^{-1/2} k_ZX  # O(m^2 n) triangular solve
lazy_tensor = MatmulLinearOperator(
    interp_term.transpose(-2, -1),
    (S - I) @ interp_term,  # TWO O(m^2 n) matrix multiplications!
)
```
In total, there are three O(m^2 n) operations. Note that there are two matrix multiplications in the last line. This is because `S - I` is stored as a `SumLinearOperator` and `S` is a `CholLinearOperator`. The Cholesky factor of `S` invokes two matrix multiplications.

**This PR.** We could save two O(m^2 n) matrix multiplications by grouping the operations in a better way: 
```
middle_term = K_ZZ^{-1/2} @ (S - I) @ K_ZZ^{-1/2}  # O(m^3), store middle_term as a tensor
lazy_tensor = MatmulLinearOperator(k_XZ,  middle_term @ k_ZX)  # a single O(m^2 n) matmul
```
In total, there is only a single O(m^2 n) matrix operations.

The caveat is that the regrouping is not numerically friendly, and so we have to do the computation in double precision. This is not an issue because data center GPUs nowadays have very good FP64 tensor cores. Empirically, we observed that FP64 matmul is just as fast as FP32 matmul on A100.

# Benchmark
We run SVGP training on four large-scale UCI datasets. The results available in https://api.wandb.ai/links/kayween/lw2terwd

The test MAE and test NLL are virtually identical to the current GPyTorch implementation. This PR speeds up SVGP training by 3.81x on houseelectric.

### DIFF
--- a/gpytorch/variational/__init__.py
+++ b/gpytorch/variational/__init__.py
@@ -12,6 +12,7 @@ from .independent_multitask_variational_strategy import (
     IndependentMultitaskVariationalStrategy,
     MultitaskVariationalStrategy,
 )
+from .large_batch_variational_strategy import LargeBatchVariationalStrategy
 from .lmc_variational_strategy import LMCVariationalStrategy
 from .mean_field_variational_distribution import MeanFieldVariationalDistribution
 from .natural_variational_distribution import _NaturalVariationalDistribution, NaturalVariationalDistribution
@@ -29,6 +30,7 @@ __all__ = [
     "GridInterpolationVariationalStrategy",
     "IndependentMultitaskVariationalStrategy",
     "LMCVariationalStrategy",
+    "LargeBatchVariationalStrategy",
     "MultitaskVariationalStrategy",
     "OrthogonallyDecoupledVariationalStrategy",
     "VariationalStrategy",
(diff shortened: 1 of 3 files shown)


kayween:
I'll push a different branch for this implementation --- closing this PR for now.

PR #2690, opened 2025-12-19
Poisson and negative binomial 

StefanoDamato:
I provide and implementations of the PoissonLikelihood and NegativeBinomialLikelihood classes. In the latter, the GP can either parametrize the mean of the distribution or the total count parameter. The classes are built on top of Torch distributions; documentation and tests are included.

### DIFF
--- a/gpytorch/likelihoods/negative_binomial_likelihood.py
+++ b/gpytorch/likelihoods/negative_binomial_likelihood.py
@@ -14,7 +14,7 @@ class NegativeBinomialLikelihood(_OneDimensionalLikelihood):
     r"""
     A Negative Binomial likelihood for regressing over count data.
 
-    This likelihood is parameterized by :math:`k > 0`, the total number of failures (also denoted as total count
+    This likelihood is parameterized by :math:`k > 0`, the total number of failures (also named total count
     in `torch.distributions`), and :math:`p \in (0, 1)`, the probability of success.
 
     Under this parameterization, the random variable represents the number of successful independent trials,
@@ -31,12 +31,12 @@ class NegativeBinomialLikelihood(_OneDimensionalLikelihood):
 
     where :math:`f` is the GP function sample. With this choice,
     the GP function parametrizes the mean of the negative binomial distribution.
-    When :attr:`total_count_param` is True, the GP directly parametrizes :math:`k = \text{softplus}(f)`.
+    When :attr:`num_failures_param` is True, the GP directly parametrizes :math:`k = \text{softplus}(f)`.
 
     :param batch_shape: The batch shape of the learned probabilities parameter (default: []).
     :param probs_prior: Prior for probabilities parameter :math:`p`.
     :param probs_constraint: Constraint for probabilities parameter :math:`p`.
-    :param total_count_param: Whether the GP parametrizes the total count parameter :math:`k` (default: False).
+    :param num_failures_param: Whether the GP parametrizes the number of failures parameter :math:`k` (default: False).
 
     :ivar torch.Tensor probs: :math:`p` parameter (probability of success)
 
@@ -49,7 +49,7 @@ class NegativeBinomialLikelihood(_OneDimensionalLikelihood):
         batch_shape: torch.Size = torch.Size([]),
         probs_prior: Optional[Prior] = None,
         probs_constraint: Optional[Interval] = None,
-        total_count_param: bool = False,
+        num_failures_param: bool = False,
     ) -> None:
         super().__init__()
 
@@ -62,7 +62,7 @@ class NegativeBinomialLikelihood(_OneDimensionalLikelihood):
 
         self.register_constraint("raw_probs", probs_constraint)
 
-        self.total_count_param = total_count_param
+        self.num_failures_param = num_failures_param
 
     @property
     def probs(self) -> Tensor:
@@ -79,8 +79,8 @@ class NegativeBinomialLikelihood(_OneDimensionalLikelihood):
 
     def forward(self, function_samples: Tensor, *args: Any, **kwargs: Any) -> NegativeBinomial:
         probs = torch.clamp(self.probs, 1e-06, 1 - 1e-06)
-        if self.total_count_param:
-            total_count = torch.nn.functional.softplus(function_samples)
+        if self.num_failures_param:
+            num_failures = torch.nn.functional.softplus(function_samples)
         else:
-            total_count = torch.nn.functional.softplus(function_samples) * (1 - probs) / probs
-        return base_distributions.NegativeBinomial(total_count=total_count, probs=probs)
+            num_failures = torch.nn.functional.softplus(function_samples) * (1 - probs) / probs
+        return base_distributions.NegativeBinomial(total_count=num_failures, probs=probs)


gpytorch/likelihoods/poisson_likelihood.py
```diff
@@ -0,0 +1,45 @@
...
+    .. math::
+        p(y \mid f) = \text{Poisson}(\lambda)
+
+    This likelihood does not have learnable parameters and
+    enforces nonnegativity through the softplus activation function.
+
+    :ivar torch.Tensor rate: :math:`\lambda` parameter (rate)
+    """
+
+    has_analytic_marginal: bool = True
```
kayween: Same thing here. Need to override `marginal` and `log_marginal` if these two quantities can be computed in closed-forms.

gpytorch/likelihoods/poisson_likelihood.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def __init__(self):
+        super().__init__()
```
kayween, line 39: This constructor could be removed because it merely calls its parent's constructor?

gpytorch/likelihoods/negative_binomial_likelihood.py
```diff
@@ -0,0 +1,89 @@
...
+    :param probs_constraint: Constraint for probabilities parameter :math:`p`.
+    :param total_count_param: Whether the GP parametrizes the total count parameter :math:`r` (default: False).
+
+    :ivar torch.Tensor probs: :math:`p` parameter (probability of success)
+
+    **Reference:**
+        Damato et al. (2025), Forecasting intermittent time series with Gaussian Processes and Tweedie likelihood.
+    """
+
+    has_analytic_marginal: bool = True
```
kayween: If this flag is true, then we need to override `log_marginal` and `marginal`. Otherwise, these two methods would invoke numerical integration.

test/likelihoods/test_negative_binomial_likelihood.py
```diff
@@ -0,0 +1,46 @@
...
+from gpytorch.likelihoods import _OneDimensionalLikelihood, NegativeBinomialLikelihood
+from gpytorch.test.base_likelihood_test_case import BaseLikelihoodTestCase
+
+
+class TestNegativeBinomialLikelihood(BaseLikelihoodTestCase, unittest.TestCase):
+    seed = 1
+
+    def _create_targets(self, batch_shape=torch.Size([])):
+        rate = torch.sigmoid(torch.randn(*batch_shape, 5)) * 3.0
+        res = torch.poisson(rate).float()
```
kayween, line 16: Is there any particular reason why we are generating data using `torch.poisson` when testing negative binomial?

gpytorch/likelihoods/negative_binomial_likelihood.py
```diff
@@ -0,0 +1,89 @@
...
+
+    The Negative Binomial distribution is parameterized by :math:`r > 0` (total count)
+    and :math:`p \in (0, 1)` (probability of success).
+    The total count parameter is derived from the GP function samples.
+
+    We parameterize the likelihood as follows:
+
+    .. math::
+        \begin{equation*}
+            r = \text{softplus}(f) \cdot \frac{1-p}{p}, \quad p = \sigma(\text{raw\_probs})
```
kayween: nit: `\text{raw\_probs}` is rendered as raw\\_probs by sphinx.

Also, I think `r` typically denotes the number of success of negative Binomial distributions (as opposed to the total count)?

gpytorch/likelihoods/negative_binomial_likelihood.py
```diff
@@ -0,0 +1,89 @@
...
+        \end{equation*}
+
+    where :math:`f` is the GP function sample and :math:`\sigma(\cdot)` is the sigmoid function.
+    Therefore, the GP function parametrizes the mean of the negative binomial distribution.
+    When :math:`\text{total\_count\_param}` is True, GP directly parametrizes :math:`r = \text{softplus}(f)`.
```
kayween: same rendering issue here

kayween (changes requested):

gpytorch/likelihoods/poisson_likelihood.py
kayween: hmm, it seems that removing it would change how sphinx generate docs for the arguments. I'm fine with keeping it.

kayween:
@StefanoDamato Thanks for the contribution! The PR mostly looks good overall. I left some comments inline.

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: Thank you, I have fixed the error. 

It is true that `r` denote the number of successes, but I chose to keep the total count designation to be consistent with the parameters names used in torch.dsitributions.NegativeBinomial.

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: Fixed, thanks.

test/likelihoods/test_negative_binomial_likelihood.py
StefanoDamato: No, I just used this because it was simpler. 

Similarly, the GPyTorch tests for the Beta and Laplace likelihoods sample the targets using `res = torch.sigmoid(torch.randn(*batch_shape, 5)).float()`, which is not the distribution to be tested.

However,  `torch.poisson` can be replaced by something like this to sample from a negative binomial: 
```
mean = torch.sigmoid(torch.randn(*batch_shape, 5)) * 3.0
p = torch.full_like(mean, 0.5)
r = mean * p / (1 - p)  
res = NegativeBinomial(total_count=r, probs=p).sample().float()
```

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: Thank you for pointing this out. The flag has been added by mistake. I removed it.

gpytorch/likelihoods/poisson_likelihood.py
StefanoDamato: Thanks. I did not change it.

gpytorch/likelihoods/poisson_likelihood.py
StefanoDamato: Than you. I removed this line too.

gpytorch/likelihoods/negative_binomial_likelihood.py
kayween: My main concern here is that the docstring might cause some confusions for the users (including me 😅). Typically, `r` is reserved for the number of successes (e.g., the wikpedia on this topic). However, the variable `r` here in the docstring is intended to denote the total number failures.

(Correct me if I'm wrong above. PyTorch's negative binomial distribution is the distribution of successful trials **given a fixed number of failures**. So we are using `r` as the total number of failures.)

To resolve the confusion/ambiguity, my preference are as follows:
1. Explicitly mention that the negative binomial distribution this likelihood returns is parameterized by the number of failures. The docstring states that `r` is the total count parameter, but it never clarifies what is being counted. I acknowledge that we are consistent with PyTorch here. But I think it's better to make this explicit to avoid potential ambiguity. 
1. Use a different symbol other than `r` to denote the total number of failures, e.g., `k` or `num_failures`.

test/likelihoods/test_negative_binomial_likelihood.py
kayween: I guess `res` is good as long as it's a tensor of integers. I don't feel strong about this.

gpytorch/likelihoods/negative_binomial_likelihood.py
```diff
@@ -0,0 +1,87 @@
...
+        if not torch.is_tensor(value):
+            value = torch.as_tensor(value).to(self.raw_probs)
+        self.initialize(raw_probs=self.raw_probs_constraint.inverse_transform(value))
+
+    def forward(self, function_samples: Tensor, *args: Any, **kwargs: Any) -> NegativeBinomial:
+        probs = torch.clamp(self.probs, 1e-06, 1 - 1e-06)
+        if self.total_count_param:
+            total_count = torch.nn.functional.softplus(function_samples)
+        else:
+            total_count = torch.nn.functional.softplus(function_samples) * (1 - probs) / probs
```
kayween: Also, I am curious which parameterization is better empirically? I assume it depends on the data?

gpytorch/likelihoods/negative_binomial_likelihood.py
```diff
@@ -0,0 +1,87 @@
...
+        \end{equation*}
+
+    where :math:`f` is the GP function sample and :math:`\sigma(\cdot)` is the sigmoid function.
+    Therefore, the GP function parametrizes the mean of the negative binomial distribution.
+    When :math:`\text{total_count_param}` is True, the GP directly parametrizes :math:`r = \text{softplus}(f)`.
+
+    The likelihood is then:
+
+    .. math::
+        p(y \mid f) = \text{NegativeBinomial} \left( r, p \right).
```
kayween: Related to the previous discussion on the total count parameter `r`. I think it would much better to spell out the meaning of this distribution explicitly, rather than simply writing an expression `NB(r, p)`. There are many different formulations for the negative binomial distribution, and this could potentially lead to ambiguity.

gpytorch/likelihoods/negative_binomial_likelihood.py
```diff
@@ -0,0 +1,87 @@
...
+    The total count parameter is derived from the GP function samples.
+
+    We parameterize the likelihood as follows:
+
+    .. math::
+        \begin{equation*}
+            r = \text{softplus}(f) \cdot \frac{1-p}{p}
+        \end{equation*}
+
+    where :math:`f` is the GP function sample and :math:`\sigma(\cdot)` is the sigmoid function.
```
kayween: nit: we can drop the explanation on `\sigma` now that it does not appear in the equation

StefanoDamato: Done, thanks.

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: I tried to expand on this. Alternatively, we may directly write the p.m.f. of the negative binomial distribution in the docstrings.

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: You are correct: it depends. While the difference between the two parameterisations is often minimal, I observed that using the GP to directly model the number of failures can be slighly more effective on data with large amounts of zeros and large positive values, but is also more prone to numerical failures. However I choose to set the mean parameterisation as the default one because the performance is the same in most cases.

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: Thank you. I tried to be more explicit in the description of the parameterisation we used, mentioning that the parameter is known as the number of failures and denoting it with `k`. I also mention that it called total count in `torch.distributions`.
In case, we can also rename the parameter `total_count_param` in the code as `num_failures_param` (or similar).

gpytorch/likelihoods/negative_binomial_likelihood.py
kayween: That makes sense!

gpytorch/likelihoods/negative_binomial_likelihood.py
kayween: The docstring looks much clearer now!

I think it makes sense to rename the argument `total_count_param` as you suggested.

gpytorch/likelihoods/negative_binomial_likelihood.py
kayween: I am fine either way---whether we add the p.m.f or not. The new docstring hopefully provides enough context so that there's no ambiguity. But feel free to add the p.m.f formula if you feel strong about it.

This PR looks great now. We should be able to merge it once we change the argument name `total_count_param` as discussed above.

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: Done, thanks.

gpytorch/likelihoods/negative_binomial_likelihood.py
StefanoDamato: I decided not to write the pmf, since now the parameterisation is not ambiguous; moreover I saw that this is typically not done in GPyTorch. I also changed the name `total_count_param` into `num_failures_param`.

Thank you for guiding me through this process!

PR #2691, opened 2025-12-19
Adding support for load_state_dict with assign=True for priors of Transformed distributions

hvarfner:
Required changes in GPytorch to unblock https://github.com/meta-pytorch/botorch/pull/3080. When `load_state_dict` is called with `assign=True`, `setattr` is called on `_transformed` attributes of the prior at the pytorch level. 

This was not the intended use of the `_transformed` attribute, but it seems like we have to enable its modification directly.

### DIFF
--- a/gpytorch/utils/memoize.py
+++ b/gpytorch/utils/memoize.py
@@ -55,7 +55,7 @@ def _cached(method=None, name=None):
 
     @functools.wraps(method)
     def g(self, *args, **kwargs):
-        cache_name = name if name is not None else method
+        cache_name = name if name is not None else method.__qualname__
         kwargs_pkl = pickle.dumps(kwargs)
         if not _is_in_cache(self, cache_name, *args, kwargs_pkl=kwargs_pkl):
             return _add_to_cache(self, cache_name, method(self, *args, **kwargs), *args, kwargs_pkl=kwargs_pkl)
@@ -73,7 +73,7 @@ def _cached_ignore_args(method=None, name=None):
 
     @functools.wraps(method)
     def g(self, *args, **kwargs):
-        cache_name = name if name is not None else method
+        cache_name = name if name is not None else method.__qualname__
         if not _is_in_cache_ignore_args(self, cache_name):
             return _add_to_cache_ignore_args(self, cache_name, method(self, *args, **kwargs))
         return _get_from_cache_ignore_args(self, cache_name)


hvarfner:
@Balandat

hvarfner:
@SebastianAment

gpytorch/priors/prior.py
```diff
@@ -40,9 +42,17 @@ def load_state_dict(self, state_dict: Mapping[str, Any], *args, **kwargs):
...
 
     def __setattr__(self, name: str, value: Any) -> None:
+        # If setting a _transformed_ attribute, update the base attribute instead
         if hasattr(self, name) and "_transformed_" in name:
             base_attr_name = name.replace("_transformed_", "")
-            raise AttributeError(TRANSFORMED_ERROR_MSG.format(base_attr_name))
+            # Convert to Tensor if needed
+            tensor_value = (
+                value if isinstance(value, torch.Tensor) else torch.as_tensor(value)
+            )
```
Balandat: should be a noop if already a tensor...

```suggestion
            tensor_value = torch.as_tensor(value)
```

gpytorch/priors/prior.py
```diff
@@ -40,9 +42,17 @@ def load_state_dict(self, state_dict: Mapping[str, Any], *args, **kwargs):
...
+            # Update the base attribute in the base distribution
+            self.base_dist.__setattr__(base_attr_name, tensor_value)
+            # Update the transformed attribute as well
+            super().__setattr__(name, tensor_value)
```
Balandat, line 52: Wouldn't we have to save the untransformed value for the attribute of the base distribution here? Seems odd that we assign the same value to both base dist and dist...

gpytorch/priors/prior.py
hvarfner: Fixed

gpytorch/priors/prior.py
hvarfner: @Balandat  `_transformed_` is just a buffer copy of the base attribute, indicating that it comes from a `torch.distributions.TransformedDistribution`, not a transformed version. I could change the name to something like `_buffered_` instead.

hvarfner: Recall that the issue is that the base attributes (`loc`, `scale` etc.) are a `@property` of the base distribution, e.g. `torch.distributions.Normal` so we can't bufferize these on the `LogNormalPrior`. Thus, we need to bufferize an attribute containing the same info, that we can then use to set the `loc` and `scale` on the base distribution.

hvarfner: @Balandat Added a note on this in a new version:

```
# Note: "_transformed_" is just an indicator that this attribute belongs to a
        # TransformedDistribution, the value itself is not transformed.
```

gpytorch/priors/utils.py
```diff
@@ -7,30 +7,30 @@ def _bufferize_attributes(module, attributes):
...
-    transf_str = "_transformed_"
-    transformed_attrs = [attr for attr in dir(module) if transf_str in attr]
-    for transf_attr in transformed_attrs:
-        base_attr_name = transf_attr.replace(transf_str, "")
-        setattr(module.base_dist, base_attr_name, getattr(module, transf_attr))
+    buffered_str = "_buffered_"
+    buffered_attrs = [attr for attr in dir(module) if buffered_str in attr]
+    for buffered_attr in buffered_attrs:
+        base_attr_name = buffered_attr.replace(buffered_str, "")
+        setattr(module.base_dist, base_attr_name, getattr(module, buffered_attr))
```
Balandat: no need to traverse twice here
```suggestion
    for attr in dir(module):
        if buffered_str in attr:
            base_attr_name = attr.replace(buffered_str, "")
            setattr(module.base_dist, base_attr_name, getattr(module, attr))
```

gpytorch/priors/prior.py
```diff
@@ -37,16 +34,25 @@ def log_prob(self, x):
...
-            base_attr_name = name.replace("_transformed_", "")
-            raise AttributeError(TRANSFORMED_ERROR_MSG.format(base_attr_name))
-
-        elif hasattr(self, f"_transformed_{name}"):
+        # If setting a _buffered_ attribute, update the base attribute instead.
+        # Note: The _buffered_ prefix indicates this is a buffer belonging to a
+        # TransformedDistribution, NOT that the value itself is transformed.
+        # The _buffered_ buffer is simply a copy of the base_dist attribute,
+        # so we assign the same value to both.
+        if hasattr(self, name) and "_buffered_" in name:
```
Balandat: let's make `"_buffered_"` a constant so that there is a single source of truth

gpytorch/priors/prior.py
```diff
@@ -37,16 +34,23 @@ def log_prob(self, x):
...
-        elif hasattr(self, f"_transformed_{name}"):
+        # If setting a BUFFERED_PREFIX attribute, update the base attribute instead.
+        # Note: BUFFERED_PREFIX is just an indicator that this attribute belongs to a
+        # TransformedDistribution, the value itself is not transformed.
+        if hasattr(self, name) and BUFFERED_PREFIX in name:
+            base_attr_name = name.replace(BUFFERED_PREFIX, "")
+            tensor_value = torch.as_tensor(value)
+            # Update the base attribute in the base distribution
+            self.base_dist.__setattr__(base_attr_name, tensor_value)
+            # Update the transformed attribute as well
```
Balandat, line 51:
```suggestion
            # Update the buffered attribute as well
```

gpytorch/priors/utils.py
```diff
@@ -3,34 +3,39 @@
 from torch.distributions import TransformedDistribution
 
 
+# Prefix for buffered attributes in TransformedDistributions.
+# These are copies of the base distribution attributes, enabling state_dict
+# save/load since the original attributes are properties and cannot be bufferized.
+BUFFERED_PREFIX = "_buffered_"
```
Balandat, line 11: would changing this from `_transformed_` to `__buffered__` cause issues with backward compatibility of loading a state dict?

Balandat: Not sure if this would be a big deal, it's probably not super widely used so I think we could merge this even if this breaks BC.

hvarfner: Yes, it would unfortunately.

Issue #2695, opened 2026-01-07
[Bug] Incompatibility issues with numpy 2.4.0


jduerholt:
# 🐛 Bug

## To reproduce

I was setting up my software stack based on a fresh python 3.13 install using latest numpy (2.4.0). Running the tests against this setup produces the following errors:

```
FAILED test/examples/test_spectral_mixture_gp_regression.py::TestSpectralMixtureGPRegression::test_spectral_mixture_gp_mean_abs_error - AttributeError: module 'numpy' has no attribute 'trapz'. Did you mean: 'trace'?
FAILED test/kernels/test_spectral_mixture_kernel.py::TestSpectralMixtureKernel::test_active_dims_list - AttributeError: module 'numpy' has no attribute 'trapz'. Did you mean: 'trace'?
FAILED test/kernels/test_spectral_mixture_kernel.py::TestSpectralMixtureKernel::test_active_dims_range - AttributeError: module 'numpy' has no attribute 'trapz'. Did you mean: 'trace'?
FAILED test/kernels/test_spectral_mixture_kernel.py::TestSpectralMixtureKernel::test_no_batch_kernel_single_batch_x - AttributeError: module 'numpy' has no attribute 'trapz'. Did you mean: 'trace'?
FAILED test/kernels/test_spectral_mixture_kernel.py::TestSpectralMixtureKernel::test_single_batch_kernel_single_batch_x - AttributeError: module 'numpy' has no attribute 'trapz'. Did you mean: 'trace'?
FAILED test/kernels/test_spectral_mixture_kernel.py::TestSpectralMixtureKernel::test_smoke_double_batch_kernel_double_batch_x - AttributeError: module 'numpy' has no attribute 'trapz'. Did you mean: 'trace'?
```

From my point of view this is an issue with the latest numpy version, as also discussed here: https://github.com/ROCm/TheRock/issues/2674 . It looks like that `trapz` was removed as deprecated from numpy 2.4.0. Switching back to numpy 2.3.5 solves it. 

## System information

**Please complete the following information:**
-  gpytorch: current main
- . torch: 2.9.1
-  os: mac

## Additional context

Currently the test pipelines are running only on py 3.10, which is I think incompatible with latest numpy. Maybe they should be updated to include also higher versions of python.

PR #2696, opened 2026-01-07
Replace np.trapz with np.trapezoid for area calculation

jacobrgardner:
For #2695

### DIFF
--- a/gpytorch/kernels/spectral_mixture_kernel.py
+++ b/gpytorch/kernels/spectral_mixture_kernel.py
@@ -189,7 +189,7 @@ class SpectralMixtureKernel(Kernel):
             freq = freq[: M + 1]
             emp_spect = emp_spect[: M + 1]
 
-            total_area = np.trapz(emp_spect, freq)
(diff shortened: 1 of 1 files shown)


jacobrgardner:
Unit tests need an environment update too I guess: https://numpy.org/doc/2.4/reference/generated/numpy.trapezoid.html

jacobrgardner:
Hmm okay, we'd also need to bump the pytorch min version to 2.2 per this comment: https://github.com/cornellius-gp/gpytorch/blob/c019c6d94a3b55410b5a76b5ec1a598d503a7179/.github/workflows/run_test_suite.yml#L53

@Balandat botorch's pyproject currently only requires 2.0.1, so y'all are kind of in this blast radius of this one line change potentially. Thoughts about this?

Balandat:
@jacobrgardner torch 2.2 is two years old by now, I think it's fine to bump the requirement up to 2.2 we can do the same thing on the botorch side. That will help simplify some of our CI logic so that's a side benefit. cc @saitcakmak

Balandat:
Superseded by #2721

PR #2698, opened 2026-01-15
LargeBatchVariationalStrategy caches triangular solves for test time

jacobrgardner:
The new LargeBatchVariationalStrategy computes triangular solves against the inducing_values and to compute the middle_term `K_zz^{-1/2}(S - I)K_{zz}^{-1/2}`. Since neither of these depend on data, we can cache them in `eval` mode, circumventing the need for linear solves entirely at test time for both the mean and the variance.

This also simplifies onnx exporting relative to the original variational strategy (since onnx does not directly support `triangular_solve`), so I added a unit test to cover easy onnx exporting of LargeBatchVariationalStrategy GP models.

### DIFF
--- a/test/variational/test_large_batch_variational_strategy.py
+++ b/test/variational/test_large_batch_variational_strategy.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 
 import torch
 
+from gpytorch.mlls import VariationalELBO
 from gpytorch.test.base_test_case import BaseTestCase
 from gpytorch.variational.large_batch_variational_strategy import LargeBatchVariationalStrategy, QuadFormDiagonal
 from gpytorch.variational.variational_strategy import VariationalStrategy
@@ -193,7 +194,6 @@ class TestLargeBatchVariationalGP(TestVariationalGP):
         model.train()
         likelihood.train()
         optimizer = torch.optim.Adam(model.parameters(), lr=0.1)
-        from gpytorch.mlls import VariationalELBO
 
         mll = VariationalELBO(likelihood, model, num_data=train_x.size(0))
 


test/variational/test_large_batch_variational_strategy.py
```diff
@@ -111,3 +111,207 @@ def test_against_variational_strategy(self, train: bool = True):
...
+
+        # Register custom ONNX symbolics for ops not in default registry
+        def mT_symbolic(g, self):
+            # mT swaps the last two dimensions
+            tensor_type = self.type()
+            if tensor_type is not None and tensor_type.dim() is not None:
+                rank = tensor_type.dim()
+                perm = list(range(rank - 2)) + [rank - 1, rank - 2]
+                return g.op("Transpose", self, perm_i=perm)
+            return g.op("Transpose", self, perm_i=[1, 0])
```
kayween, line 249: nit: probably better `perm_i = [-1, -2]` so that it works in batch settings?

test/variational/test_large_batch_variational_strategy.py
```diff
@@ -111,3 +111,207 @@ def test_against_variational_strategy(self, train: bool = True):
...
+            one_plus_exp_neg = g.op("Add", one, exp_neg)
+            log_pos = g.op("Log", one_plus_exp_neg)
+            result_pos = g.op("Add", scaled, log_pos)
+
+            exp_scaled = g.op("Exp", scaled)
+            one_plus_exp = g.op("Add", one, exp_scaled)
+            result_neg = g.op("Log", one_plus_exp)
+
+            stable_result = g.op("Where", condition, result_pos, result_neg)
+            return g.op("Div", stable_result, beta)
```
kayween, line 271: I am not super familiar with ONNX. Does ONNX not have `softplus` so that we need to implement a custom one here? Similarly, is the transpose function necessary?

test/variational/test_large_batch_variational_strategy.py
```diff
@@ -111,3 +111,207 @@ def test_against_variational_strategy(self, train: bool = True):
...
+                return g.op("Transpose", self, perm_i=perm)
+            return g.op("Transpose", self, perm_i=[1, 0])
+
+        def softplus_symbolic(g, self, beta, threshold):
+            # Numerically stable Softplus using Where:
+            # softplus(x) = x + log(1 + exp(-x)) for x > 0
+            #             = log(1 + exp(x)) for x <= 0
+            scaled = g.op("Mul", self, beta)
+            zero = g.op("Constant", value_t=torch.tensor([0.0], dtype=torch.float64))
+            one = g.op("Constant", value_t=torch.tensor([1.0], dtype=torch.float64))
```
kayween, line 257: nit: maybe use `torch.tensor(0.0, dtype=torch.float64)` so that it does not change the shape if `x` is a scalar.

gpytorch/variational/large_batch_variational_strategy.py
```diff
@@ -75,19 +83,32 @@ def _compute_predictive_updates(
...
+        # Cache inv_chol_t_inducing_values and middle_term during inference (not training)
+        # since they don't depend on the data (through induc_data_covar)
+        if not self.training and hasattr(self, "_cached_inv_chol_t_inducing_values"):
+            inv_chol_t_inducing_values = self._cached_inv_chol_t_inducing_values
+        else:
+            inv_chol_t_inducing_values = torch.linalg.solve_triangular(
+                chol.mT, inducing_values.unsqueeze(-1), upper=True, left=True
+            )
+            if not self.training:
+                self._cached_inv_chol_t_inducing_values = inv_chol_t_inducing_values
```
kayween, line 95: This would still trigger a triangular solve once at test-time. I am not sure if this is a concern?

If we use GPyTorch's `cached` decorator like [this](https://github.com/cornellius-gp/gpytorch/blob/ff324365cdb730d41e17f4524db49bc12dc1e13e/gpytorch/variational/variational_strategy.py#L187-L190), we could avoid test-time triangular solves completely. It's probably a nit because we can do a forward pass before ONNX exporting.

test/variational/test_large_batch_variational_strategy.py
```diff
@@ -111,3 +111,207 @@ def test_against_variational_strategy(self, train: bool = True):
...
+        likelihood = likelihood.double()
+
+        # Quick training on toy data (fp64)
+        train_x = torch.rand(32, 2, dtype=torch.float64)
+        train_y = torch.sin(train_x[:, 0]) + 0.1 * torch.randn(32, dtype=torch.float64)
+
+        model.train()
+        likelihood.train()
+        optimizer = torch.optim.Adam(model.parameters(), lr=0.1)
+        from gpytorch.mlls import VariationalELBO
```
kayween: nit: move this import to the top of of the file / at the beginning of the method.

kayween:
Looks good to me. I left some inline comments. I am not super familiar with ONNX, but it looks sensible since and the tests has passed.

@jacobrgardner Can you add a short description in the PR summary?

test/variational/test_large_batch_variational_strategy.py
jacobrgardner: Onnx doesn't support negative indices, sadly.

test/variational/test_large_batch_variational_strategy.py
jacobrgardner: It has softplus, but only an fp32 version (for whatever reason).

It supports transpose, but not `.mT` specifically which is used all over the place in linear_operator and would be a much bigger fix to replace everywhere there than here.

test/variational/test_large_batch_variational_strategy.py
jacobrgardner: Yeah, this is just more onnx idiosyncrasies unfortunately.

gpytorch/variational/large_batch_variational_strategy.py
jacobrgardner: Yeah we could make it decorated and then just add a grad hook. However, the current behavior is generally in line with how other caching in GPyTorch works (e.g. exact GPs cache the first forward through in eval mode).

PR #2701, opened 2026-01-22
Modify confidence_region to use confidence parameter

vvb610:
Updated the confidence_region method to accept a confidence parameter and calculate the confidence region based on the specified confidence level.


kayween:
Hi @vvb610! Thanks for the contribution.

While supporting a more general confidence region is an interesting addition, I am not quite sure how many users will benefit from this since the two standard deviation is probably the most common use case. To keep the code lean, I am going to pass on merging this for now.

vvb610:
@kayween Sounds good! I needed something different and noted it wasn't supported so thought I'd suggest the change (with the default argument for backwards compatibility). Please feel free to close the PR or mark it a draft..., as you wish.

PR #2704, opened 2026-01-25
Fix potential bug with settings.debug.on()

benediktjohannes:
I'm not quite sure about this one and I'd like to get a review please, but I think that debug is a class and can't be "debug()"


benediktjohannes:
I'm not quite sure, but I thought of this because all the other times it's debug.on() as well without the "()", could you please review this @jacobrgardner thanks! 👍

PR #2709, opened 2026-01-25
Fix the same typo in other class as well

benediktjohannes:


gpytorch/variational/variational_strategy.py
```diff
@@ -211,8 +211,7 @@ def pseudo_points(self) -> Tuple[Tensor, Tensor]:
         if not isinstance(self._variational_distribution, CholeskyVariationalDistribution):
             raise NotImplementedError(
                 "Only CholeskyVariationalDistribution has pseudo-point support currently, ",
-                "but your _variational_distribution is a ",
-                self._variational_distribution.__name__,
+                "but your _variational_distribution is a {type(self._variational_distribution).__name__}"
```
Balandat:
```suggestion
                f"but your _variational_distribution is a {type(self._variational_distribution).__name__}"
```

benediktjohannes: Thanks

benediktjohannes:
Isn't the "," too much? This is a tuple, doesn't this have to be a string without a comma? I missed that as well, what do you think @Balandat ? And thanks for the review!

PR #2715, opened 2026-02-13
Separate covariance computations for customizable posterior inference

SebastianAment:
This commit separates out the computation of the test-set mean and covariances 
into `_get_test_prior_mean_and_covariances` and rewrites `exact_prediction` to 
accept `test_mean`, `test_test_covar`, `test_train_covar` already separated.

This enables:
1. Latent Kronecker Gaussian Processes with partial observations 
2. Posterior inference of individual components of additive models ([BoTorch PR](https://github.com/meta-pytorch/botorch/pull/3187))
3. Any application where predicted quantities differ from observed quantities, e.g. component-wise prediction of partial derivatives of kernel from observation of the function values.

### DIFF
--- a/gpytorch/models/exact_gp.py
+++ b/gpytorch/models/exact_gp.py
@@ -10,6 +10,8 @@ from copy import deepcopy
 import torch
 from torch import Tensor
 
+from gpytorch.distributions import Distribution
+
 from .. import settings
 from ..distributions import MultitaskMultivariateNormal, MultivariateNormal
 from ..likelihoods import _GaussianLikelihoodBase
@@ -300,7 +302,7 @@ class ExactGP(GP):
 
             # Get the terms that only depend on training data
             if self.prediction_strategy is None:
-                train_output = super().__call__(*train_inputs, **kwargs)
+                train_output = self._get_train_prior_distribution(train_inputs, **kwargs)
 
                 # Create the prediction strategy for
                 self.prediction_strategy = prediction_strategy(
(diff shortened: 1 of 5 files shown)


gpytorch/models/exact_gp.py
```diff
@@ -309,41 +312,110 @@ def __call__(self, *args, **kwargs):
...
+                test_shape,
+                posterior_class,
+            ) = self._get_test_prior_mean_and_covariances(train_inputs=train_inputs, test_inputs=inputs, **kwargs)
             # Make the prediction
             with settings.cg_tolerance(settings.eval_cg_tolerance.value()):
-                (
-                    predictive_mean,
-                    predictive_covar,
-                ) = self.prediction_strategy.exact_prediction(full_mean, full_covar)
+                (predictive_mean, predictive_covar,) = self.prediction_strategy.exact_prediction(
```
Balandat: code style
```suggestion
                predictive_mean, predictive_covar = self.prediction_strategy.exact_prediction(
```

gpytorch/models/exact_gp.py
```diff
@@ -309,41 +312,110 @@ def __call__(self, *args, **kwargs):
...
+            kwargs: Additional keyword arguments passed to the model's forward.
+
+        Returns:
+            A tuple of (test_mean, test_test_covar, test_train_covar, batch_shape,
+            test_shape, posterior_class).
+        """
+        # Concatenate the input to the training input
+        full_inputs = []
+        batch_shape = train_inputs[0].shape[:-2]
+        for train_input, input in length_safe_zip(train_inputs, test_inputs):
```
Balandat, line 385: Seems like we can replace `length_safe_zip` with `zip(*, strict=True)`?

gpytorch/models/exact_gp.py
```diff
@@ -309,41 +312,110 @@ def __call__(self, *args, **kwargs):
...
+        full_output = super().__call__(*full_inputs, **kwargs)
+        if settings.debug().on():
+            if not isinstance(full_output, MultivariateNormal):
+                raise RuntimeError("ExactGP.forward must return a MultivariateNormal")
+        joint_mean, joint_covar = full_output.loc, full_output.lazy_covariance_matrix
+
+        # Determine the shape of the joint distribution
+        batch_shape = full_output.batch_shape
+        joint_shape = full_output.event_shape
+        tasks_shape = joint_shape[1:]  # For multitask learning
```
Balandat: "For multitask learning" could be expanded on, not entirely clear what's happening here

gpytorch/models/exact_gp.py
```diff
@@ -309,41 +312,110 @@ def __call__(self, *args, **kwargs):
...
+
+        test_shape = torch.Size([joint_shape[0] - self.prediction_strategy.train_shape[0], *tasks_shape])
```
Balandat: Some explicit shape annotations in the comments would be helpful here

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -310,17 +308,45 @@ def num_train(self):
...
+        :param test_train_covar: Covariance matrix between test and train inputs
+            (shape: extra_batch x ... x num_test x num_train)
+        :return: A tuple (posterior_predictive_mean, posterior_predictive_covar), with shapes
+            (extra_batch x ... x num_test), and (extra_batch x ... x num_test x num_test)
+        """
+        # sum(test_train_covar.shape[-2:]) = joint_covar.size(-1)
+        if sum(test_train_covar.shape[-2:]) <= settings.max_eager_kernel_size.value():
+            # If we are calling it here, it requires two calls to "to_dense", but if we
+            # called it in the test covariance getter, then it would break the other
+            # prediction strategies (actually it already, does because we are not
```
Balandat:
```suggestion
            # prediction strategies (actually it already does because we are not
```

Balandat: should we be overwriting them?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -345,26 +375,11 @@ def exact_predictive_mean(self, test_mean: Tensor, test_train_covar: LinearOpera
...
-            # Set the columns corresponding to missing observations to 0 to ignore them during matmul.
-            mask = (~torch.isnan(mean_cache)).to(torch.float)[..., None, :]
-            test_train_covar = test_train_covar * mask
-            mean = settings.observation_nan_policy._fill_tensor(mean_cache)
-            res = (test_train_covar @ mean.unsqueeze(-1)).squeeze(-1)
-        res = res + test_mean
-
+        # Replace NaN values with 0 for the matrix multiplication.
+        # NaN values in mean_cache correspond to masked/missing observations that
+        # should not contribute to the predictive mean.
```
Balandat: Should we be handling `observation_nan_policy` somehow? Seems like this is now effectively being hard-coded to masking with zero?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -388,13 +403,13 @@ def exact_predictive_covar(
...
+            train_train_covar = self.likelihood(dist, self.train_inputs).lazy_covariance_matrix
             if settings.detach_test_caches.on():
-                train_train_covar = self.likelihood(dist, self.train_inputs).lazy_covariance_matrix.detach()
-            else:
-                train_train_covar = self.likelihood(dist, self.train_inputs).lazy_covariance_matrix
+                train_train_covar = train_train_covar.detach()
 
             test_train_covar = to_dense(test_train_covar)
             train_test_covar = test_train_covar.transpose(-1, -2)
+            # This could be using just a single cholesky solve, for dense inference (currently two)
```
Balandat: Let's implement it or add a TODO?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -681,12 +690,14 @@ def covar_cache(self):
...
-        test_mean = joint_mean[..., self.num_train :]
-        test_test_covar = joint_covar[..., self.num_train :, self.num_train :].evaluate_kernel()
-        test_train_covar = joint_covar[..., self.num_train :, : self.num_train].evaluate_kernel()
-
+    def exact_prediction(
+        self,
+        test_mean: Tensor | LinearOperator,
+        test_test_covar: Tensor | LinearOperator,
+        test_train_covar: Tensor | LinearOperator,
+    ):
```
Balandat: type this and add a docstring documenting the shape handling?>

gpytorch/models/exact_gp.py
SebastianAment: `zip(..., strict=True)` was added in Python 3.10 (PEP 618), which seems to be GPyTorch's Python version floor, so we can make this change. Since `length_safe_zip` is used throughout the code base, I'll put up a separate commit for this.

gpytorch/models/exact_prediction_strategies.py
SebastianAment: I think a better design for the prediction strategies is based on multiple-dispatch on the types of the covariance matrices. That way there wouldn't need to be a fall back if the types don't match, it would just call the correct implementation to begin with. With this in mind, I'd prefer to defer further changes to a future commit.

gpytorch/models/exact_prediction_strategies.py
SebastianAment: Good catch!

gpytorch/models/exact_gp.py
SebastianAment: @Balandat Added this in https://github.com/cornellius-gp/gpytorch/pull/2718.

SebastianAment:
@Balandat Thanks for the review, incorporated your suggestions.

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -310,17 +328,40 @@ def num_train(self):
...
-        if joint_covar.size(-1) <= settings.max_eager_kernel_size.value():
-            test_covar = joint_covar[..., self.num_train :, :].to_dense()
-            test_test_covar = test_covar[..., self.num_train :]
-            test_train_covar = test_covar[..., : self.num_train]
-        else:
-            test_test_covar = joint_covar[..., self.num_train :, self.num_train :]
-            test_train_covar = joint_covar[..., self.num_train :, : self.num_train]
+    def exact_prediction(
+        self,
+        test_mean: Tensor | LinearOperator,
```
kayween: nit: This type annotation could be stricter here since `test_mean` has to be a tensor? AFAIK, only covariance matrices in GPyTorch are allowed to be linear operators.
```suggestion
        test_mean: Tensor,
```

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -386,22 +430,29 @@ def exact_predictive_covar(
...
             if settings.detach_test_caches.on():
-                train_train_covar = self.likelihood(dist, self.train_inputs).lazy_covariance_matrix.detach()
-            else:
-                train_train_covar = self.likelihood(dist, self.train_inputs).lazy_covariance_matrix
+                train_train_covar = train_train_covar.detach()
 
             test_train_covar = to_dense(test_train_covar)
             train_test_covar = test_train_covar.transpose(-1, -2)
+            # TODO: use just a single cholesky solve, for dense inference (currently two)
             covar_correction_rhs = train_train_covar.solve(train_test_covar)
```
kayween: Here, `train_train_covar.solve` only invokes a single Cholesky solve. Does the second Cholesky solve refer to the solve against the training labels in `_mean_cache`? IIUC, it's pretty hard to batch those two solves without a major refactor right?

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -681,12 +741,33 @@ def covar_cache(self):
...
 
-    def exact_prediction(self, joint_mean, joint_covar):
-        # Find the components of the distribution that contain test data
-        test_mean = joint_mean[..., self.num_train :]
-        test_test_covar = joint_covar[..., self.num_train :, self.num_train :].evaluate_kernel()
-        test_train_covar = joint_covar[..., self.num_train :, : self.num_train].evaluate_kernel()
-
+    def exact_prediction(
+        self,
+        test_mean: Tensor | LinearOperator,
```
kayween:
```suggestion
        test_mean: Tensor,
```
Similarly to the above, `test_mean` can only be a tensor.

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -737,24 +826,43 @@ def exact_predictive_covar(self, test_test_covar, test_train_covar):
...
-            train_prior_dist.mean, train_prior_dist.lazy_covariance_matrix.evaluate_kernel()
+            train_prior_dist.mean,
+            train_prior_dist.lazy_covariance_matrix.evaluate_kernel(),
         )
         super().__init__(train_inputs, train_prior_dist, train_labels, likelihood)
 
-    def exact_prediction(self, joint_mean: Tensor, joint_covar: LinearOperator) -> tuple[Tensor, LinearOperator]:
+    def exact_prediction(
+        self,
+        test_mean: Tensor | LinearOperator,
```
kayween:
```suggestion
        test_mean: Tensor,
```

gpytorch/models/exact_prediction_strategies.py
```diff
@@ -920,14 +1051,15 @@ def get_fantasy_strategy(self, inputs, targets, full_inputs, full_targets, full_
             "Fantasy observation updates not yet supported for models using SGPRPredictionStrategy"
         )
 
-    def exact_prediction(self, joint_mean, joint_covar):
+    def exact_prediction(
+        self,
+        test_mean: Tensor | LinearOperator,
```
kayween:
```suggestion
        test_mean: Tensor,
```

gpytorch/models/exact_gp.py
```diff
@@ -309,41 +312,117 @@ def __call__(self, *args, **kwargs):
...
+
+        Returns:
+            The prior distribution evaluated on the training set.
+        """
+        return super().__call__(*train_inputs, **kwargs)
+
+    def _get_test_prior_mean_and_covariances(
+        self,
+        train_inputs: Iterable[Tensor | LinearOperator],
+        test_inputs: Iterable[Tensor | LinearOperator],
```
kayween:
```suggestion
        train_inputs: Iterable[Tensor],
        test_inputs: Iterable[Tensor],
```
The train/test inputs have to be tensors? This also maintains consistency with `_get_train_prior_distribution`

kayween:
This refactor looks good to me. Besides unblocking what's in the PR description, this PR also does quite a bit of cleanup, which is great. cc @jacobrgardner @gpleiss in case they want to weigh in here.

gpytorch/models/exact_prediction_strategies.py
SebastianAment: I meant to say triangular solve (one Cholesky solve consists of two triangular solves). We can use similar logic to the `fast_pred_var.on()` branch, even if we don't pre-compute the inverse roots. 

Also seems like we could use `self.lik_train_train_covar` here instead of evaluating `train_train_covar` from scratch. This is orthogonal to the purpose of the current commit however, so I don't plan on making these changes here.

gpytorch/models/exact_gp.py
SebastianAment: @kayween Good catch! Incorporated your suggestions.

gpytorch/models/exact_gp.py
```diff
@@ -309,41 +311,117 @@ def __call__(self, *args, **kwargs):
...
+        Override this method to customize train-train covariance computation.
+
+        Args:
+            train_inputs: The inputs in the training set.
+            kwargs: Additional keyword arguments passed to the model's forward method.
+
+        Returns:
+            The prior distribution evaluated on the training set.
+        """
+        return super().__call__(*train_inputs, **kwargs)
```
gpleiss, line 352: Should this be inside a `gpytorch.settings.prior_mode` context manager?

gpytorch/models/exact_prediction_strategies.py
kayween: Makes sense. Agreed that these changes are out of scope of this PR.

gpytorch/models/exact_gp.py
SebastianAment: The `super.__call__` is bypassing the `prior_mode` check in `ExactGP.__call__`, and calling the model's `forward`, which computes the prior, so it doesn't appear necessary here. Added an inline comment here about this.

SebastianAment:
Thanks for the review @Balandat, @kayween, @gpleiss! This is ready to be merged.

PR #2716, opened 2026-02-13
Buildthedocs fail

hvarfner:
Pin setuptools<78 in RTD config to fix docs build. m2r2 imports pkg_resources, which was removed in setuptools 78+. RTD auto-upgrades to the latest setuptools, breaking the build.


hvarfner:
@Balandat @SebastianAment

PR #2718, opened 2026-02-17
Deprecating `length_safe_zip` in favor of `zip(..., strict=True)`

SebastianAment:
`zip` supports `strict` since Python 3.10, which appears to be GPyTorch's version floor, so we can start relying on the core implementation, simplifying the implementation.


SebastianAment:
@Balandat @kayween This should be ready now as well.

kayween (approved):

PR #2734, opened 2026-03-19
Fix memory leak in DefaultPredictionStrategy cache hooks (#2631)

### LINKED ISSUE #2631: [Bug] Memory leak in fantasy models when the mean cache is not detached
# 🐛 Bug

This is a follow up to the memory leak reported in https://github.com/pytorch/botorch/issues/2728. Repeated operations on fantasized models within the `detach_test_caches(False)` context leads to indefinite increase in memory usage until the program crashes. Digging into it, I've isolated the leak to a single `detach()` call within the `DefaultPredictionStrategy._mean_cache`.

Things I've tried:
- Replacing `detach_test_caches(False)` with `detach_test_caches(True)` makes the leak go away.
- I haven't been able to reproduce with non-fantasy models. It could be that I haven't used large enough tensors to differentiate a leak from regular memory usage.
- Within `DefaultPredictionStrategy`, I've tried manually detaching each location `detach_test_caches` affects. Only the [`mean_cache`](https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/models/exact_prediction_strategies.py#L294) seems to affect the results. The other two locations didn't seem to make a meaningful difference in memory usage.
- I've tried not caching `_mean_cache` by removing the `@cached` decorator. The leak goes away if we remove caching. I then tried manually caching it as a simple attribute and that also produced the leak.

## To reproduce

```python
import torch
from gpytorch import settings as gpt_settings
from gpytorch.distributions import MultivariateNormal
from gpytorch.means import ConstantMean
from gpytorch.models import ExactGP

d = 10
mc_points = torch.rand(32, d, dtype=torch.double)


class SimpleGP(ExactGP):
    def __init__(self, train_inputs, train_targets):
        super().__init__(train_inputs, train_targets, GaussianLikelihood())
        self.mean_module = ConstantMean()
        self.covar_module = RBFKernel()

    def forward(self, x):
        mean_x = self.mean_module(x)
        return MultivariateNormal(mean_x, covar_x)


gp = SimpleGP(
    train_inputs=torch.rand(256, d, dtype=torch.double),
    train_targets=torch.rand(256, dtype=torch.double),
).eval()
gp(torch.rand(5, d, dtype=torch.double))  # set the caches before fantasize.

X = torch.rand(128, 5, d, dtype=torch.double, requires_grad=False)
Y = torch.rand(128, 5, dtype=torch.double, requires_grad=False)

for i in range(1000):
    fantasy_model = gp.get_fantasy_model(inputs=X, targets=Y).eval()
    with gpt_settings.detach_test_caches(False):
        fantasy_model(mc_points)
    print(f"{i=}")
```

Runs for a while until the process is killed.
```
i=0
i=1
i=2
i=3
i=4
...
i=294
i=295
i=296
i=297
i=298
zsh: killed     python frepro.py
```
Just before crashing, it was reporting ~100GB memory usage.

## Expected Behavior

Memory usage should not increase beyond a simple first step. The program should be able to run indefinitely.

## System information

- GPyTorch 1.14
- Linear Operator 0.6
- Reproduces on both MacOS & CentOS

saitcakmak:
Fixes #2631 and https://github.com/meta-pytorch/botorch/issues/2728

Root cause
----------
When `detach_test_caches` is False (as used by BoTorch via `propagate_grads(True)`), the `_mean_cache` and
`_exact_predictive_covar_inv_quad_form_cache` properties register a `clear_cache_hook` on the cached tensor's `grad_fn`:

    wrapper = functools.partial(clear_cache_hook, self)
    mean_cache.grad_fn.register_hook(wrapper)

This creates a reference cycle:

    prediction_strategy -> _memoize_cache -> cached tensor
        -> grad_fn (C++ object) -> hook closure -> prediction_strategy

Because PyTorch's `grad_fn` is a C++ object, Python's cycle garbage collector cannot traverse the C++/Python boundary to detect the cycle. The entire chain becomes uncollectable. For fantasy models, the cached tensor's computation graph holds references back to the parent model's caches, so each iteration adds another uncollectable chain, causing memory to grow indefinitely until OOM.

Fix
---
Added `register_cache_clear_hook(tsr, module)` to `gpytorch/utils/memoize.py`. This helper uses `weakref.ref(module)` when registering the backward hook on `tsr.grad_fn`, breaking the reference cycle. When no external strong references to the prediction strategy remain, it and its caches are garbage-collected normally, and the hook becomes a no-op.

Updated all three call sites to use the new helper:
- `DefaultPredictionStrategy._exact_predictive_covar_inv_quad_form_cache`
- `DefaultPredictionStrategy._mean_cache`
- `_variational_strategy._add_cache_hook`

Validation (ran manually)
----------
Repro script (200 iterations of fantasy model creation + evaluation with `detach_test_caches(False)`):

```python
import tracemalloc

import torch
from gpytorch import settings as gpt_settings
from gpytorch.distributions import MultivariateNormal
from gpytorch.means import ConstantMean
from gpytorch.models import ExactGP

d = 10
mc_points = torch.rand(32, d, dtype=torch.double)


class SimpleGP(ExactGP):
    def __init__(self, train_inputs, train_targets):
        super().__init__(train_inputs, train_targets, GaussianLikelihood())
        self.mean_module = ConstantMean()
        self.covar_module = RBFKernel()

    def forward(self, x):
        mean_x = self.mean_module(x)
        return MultivariateNormal(mean_x, covar_x)


gp = SimpleGP(
    train_inputs=torch.rand(256, d, dtype=torch.double),
    train_targets=torch.rand(256, dtype=torch.double),
).eval()
gp(torch.rand(5, d, dtype=torch.double))  # set the caches before fantasize.

X = torch.rand(128, 5, d, dtype=torch.double, requires_grad=False)
Y = torch.rand(128, 5, dtype=torch.double, requires_grad=False)

tracemalloc.start()

for i in range(1000):
    fantasy_model = gp.get_fantasy_model(inputs=X, targets=Y).eval()
    with gpt_settings.detach_test_caches(False):
        fantasy_model(mc_points)
    if (i + 1) % 100 == 0:
        current, peak = tracemalloc.get_traced_memory()
        print(f"Iter {i + 1}: current={current / 1024 / 1024:.1f}MB, peak={peak / 1024 / 1024:.1f}MB")
```

Before: OOM with 48GB RAM.
```bash
(botorch) saitcakmak@saitcakmak-mac gpytorch % git branch
  fix-memory-leak-mean-cache
* main
(botorch) saitcakmak@saitcakmak-mac gpytorch % python test.py
Iter 100: current=3.6MB, peak=3.6MB
Iter 200: current=6.7MB, peak=6.8MB
zsh: killed     python test.py
```

After:
```bash
(botorch) saitcakmak@saitcakmak-mac gpytorch % git branch
* fix-memory-leak-mean-cache
  main
(botorch) saitcakmak@saitcakmak-mac gpytorch % python test.py
Iter 100: current=1.1MB, peak=1.8MB
Iter 200: current=1.4MB, peak=2.0MB
Iter 300: current=1.6MB, peak=2.3MB
Iter 400: current=0.4MB, peak=2.6MB
Iter 500: current=0.9MB, peak=2.6MB
Iter 600: current=1.3MB, peak=2.6MB
Iter 700: current=1.5MB, peak=2.6MB
Iter 800: current=2.6MB, peak=2.6MB
Iter 900: current=2.8MB, peak=2.8MB
Iter 1000: current=3.0MB, peak=3.1MB
```

### DIFF
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import functools
 import string
 import warnings
 
--- a/gpytorch/models/exact_prediction_strategies.py
+++ b/gpytorch/models/exact_prediction_strategies.py
@@ -25,7 +25,7 @@ from torch import Tensor
 from .. import settings
 from ..distributions import MultitaskMultivariateNormal
 from ..lazy import LazyEvaluatedKernelTensor
-from ..utils.memoize import add_to_cache, cached, clear_cache_hook, pop_from_cache
+from ..utils.memoize import add_to_cache, cached, pop_from_cache, register_cache_clear_hook
 
 
 def prediction_strategy(train_inputs, train_prior_dist, train_labels, likelihood):
@@ -108,10 +108,7 @@ class DefaultPredictionStrategy:
         if settings.detach_test_caches.on():
             res = res.detach()
 
-        if res.grad_fn is not None:
-            wrapper = functools.partial(clear_cache_hook, self)
-            functools.update_wrapper(wrapper, clear_cache_hook)
-            res.grad_fn.register_hook(wrapper)
+        register_cache_clear_hook(res, self)
 
         return res
 
@@ -313,10 +310,7 @@ class DefaultPredictionStrategy:
         if settings.detach_test_caches.on():
             mean_cache = mean_cache.detach()
 
-        if mean_cache.grad_fn is not None:
-            wrapper = functools.partial(clear_cache_hook, self)
-            functools.update_wrapper(wrapper, clear_cache_hook)
-            mean_cache.grad_fn.register_hook(wrapper)
+        register_cache_clear_hook(mean_cache, self)
 
         return mean_cache
 
--- a/gpytorch/utils/memoize.py
+++ b/gpytorch/utils/memoize.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import functools
 import pickle
+import weakref
 
 from .errors import CachingError
 
@@ -44,6 +45,24 @@ def pop_from_cache_ignore_args(obj, name):
 
 def clear_cache_hook(module, *args, **kwargs):
     module._memoize_cache = {}
+
+
+def register_cache_clear_hook(tsr, module):
+    """Register a backward hook on tsr's grad_fn that clears module's cache.
+
+    Uses a weak reference to module to avoid creating an uncollectable
+    reference cycle through the C++ grad_fn object (which Python's cycle
+    GC cannot see through).
+    """
+    if tsr.grad_fn is not None:
+        weak_module = weakref.ref(module)
+
+        def hook(*args, **kwargs):
+            obj = weak_module()
+            if obj is not None:
+                obj._memoize_cache = {}
+
+        tsr.grad_fn.register_hook(hook)
 
 
 def _cached(method=None, name=None):
--- a/gpytorch/variational/_variational_strategy.py
+++ b/gpytorch/variational/_variational_strategy.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import functools
 from abc import ABC, abstractproperty
 from copy import deepcopy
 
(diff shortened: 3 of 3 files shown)


Balandat (approved):
thanks

PR #2742, opened 2026-03-30
Add SphericalLinear kernel

colmont:


gpytorch/kernels/spherical_linear_kernel.py
```diff
@@ -0,0 +1,147 @@
...
+
+    .. math::
+        k(\mathbf{x_1}, \mathbf{x_2}) = b_0 + b_1
+        P(z(\mathbf{x_1}))^\top P(z(\mathbf{x_2}))
+
+    where :math:`z(\mathbf x)` applies lengthscale scaling, :math:`P` is the inverse
+    stereographic projection onto a unit sphere, and :math:`(b_0, b_1)` are learned
+    mixture weights (via softmax, so :math:`b_0 + b_1 = 1`).
+
+    This kernel was proposed in `We Still Don't Understand High-Dimensional Bayesian Optimization`.
```
gpleiss, line 41: Nit: make this into a link, rather than having the arxiv link separate.

gpytorch/kernels/spherical_linear_kernel.py
```diff
@@ -0,0 +1,147 @@
...
+            ard_num_dims=ard_num_dims,
+            lengthscale_prior=lengthscale_prior,
+            lengthscale_constraint=lengthscale_constraint,
+            **kwargs,
+        )
+        self.normalize_lengthscale = normalize_lengthscale
+        self.bounds = bounds
+
+        # Learned mixture coefficients: softmax([raw_coeffs]) -> [constant, linear]
+        self.raw_coeffs = nn.Parameter(torch.zeros(*self.batch_shape, 2))
```
gpleiss, line 93: We should have getters and setters for the non-raw values, so that they can be initialized appropriately (e.g. `self.coeffs = <blah>`)

gpytorch/kernels/spherical_linear_kernel.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # Global lengthscale: sigmoid(raw_glob_ls) * max_sq_norm
+        self.raw_glob_ls = nn.Parameter(torch.zeros(*self.batch_shape, 1))
```
gpleiss, line 96: Same thing here.

gpytorch/kernels/spherical_linear_kernel.py
```diff
@@ -0,0 +1,147 @@
...
+
+    has_lengthscale = True
+
+    def __init__(
+        self,
+        bounds: Tensor,
+        ard_num_dims: int | None = None,
+        lengthscale_prior: Prior | None = None,
+        lengthscale_constraint: Interval | None = None,
+        normalize_lengthscale: bool = False,
```
gpleiss, line 70: Maybe we should set it to true? And we can add a comment that it was set to False in the original paper.

gpytorch/kernels/spherical_linear_kernel.py
```diff
@@ -0,0 +1,147 @@
...
+        # Global lengthscale: sigmoid(raw_glob_ls) * max_sq_norm
+        self.raw_glob_ls = nn.Parameter(torch.zeros(*self.batch_shape, 1))
+
+    def forward(self, x1: Tensor, x2: Tensor, diag: bool = False, **params) -> Tensor | LinearOperator:
+        if diag:
+            # The diagonal is always 1
+            return torch.ones(x1.shape[:-1], dtype=x1.dtype, device=x1.device)
+
+        if self.normalize_lengthscale:  # Enforce L2 norm = 1
+            lengthscale = torch.softmax(self.lengthscale, dim=-1).sqrt()
```
gpleiss, line 104: Perhaps it would be better if we just divided the lengthscale by its norm, so as not to distort the geometry of the lengthscale.

gpytorch/kernels/spherical_linear_kernel.py
```diff
@@ -0,0 +1,147 @@
...
+        centers = (mins + maxs) / 2.0
+        max_sq_norm = ((maxs - mins) / (2 * lengthscale)).square().sum(dim=-1, keepdim=True)
+        glob_ls = torch.sqrt(torch.sigmoid(self.raw_glob_ls[..., None]) * max_sq_norm)
+
+        # Mixture coefficients via softmax
+        coeffs = torch.softmax(self.raw_coeffs, dim=-1)
+        sqrt_const = torch.sqrt(coeffs[..., 0])
+        sqrt_linear = torch.sqrt(coeffs[..., 1])
+
+        def _featurize(x: torch.Tensor) -> torch.Tensor:
```
gpleiss, line 121: Do we need to have the nested function?

test/kernels/test_spherical_linear_kernel.py
```diff
@@ -0,0 +1,194 @@
...
+    def test_diag(self):
+        """Diagonal should always be 1 (softmax coeffs sum to 1, projection is unit norm)."""
+        kernel = SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, ard_num_dims=3)
+        kernel.eval()
+        x = torch.rand(10, 3)
+        diag = kernel(x, diag=True)
+        self.assertEqual(diag.shape, torch.Size([10]))
+        self.assertAllClose(diag, torch.ones(10), rtol=1e-5, atol=1e-5)
+
+    def test_diag_with_learned_params(self):
```
gpleiss, line 87: I don't know if I understand the purpose of this test

test/kernels/test_spherical_linear_kernel.py
```diff
@@ -0,0 +1,194 @@
...
+        effective_ls = torch.softmax(kernel.lengthscale, dim=-1).sqrt()
+        self.assertAllClose(effective_ls.norm(), torch.tensor(1.0), rtol=1e-5, atol=1e-5)
+
+    def test_project_onto_unit_sphere(self):
+        """Projected points should lie on the unit sphere."""
+        x = torch.randn(10, 5)
+        projected = project_onto_unit_sphere(x)
+        self.assertEqual(projected.shape, torch.Size([10, 6]))
+        norms = projected.norm(dim=-1)
+        self.assertAllClose(norms, torch.ones(10), rtol=1e-5, atol=1e-5)
```
gpleiss, line 127: Can you add a test that inputs that are on the unit sphere get an effectively identity mapping, to ensure we did the inverse stereographic projection correctly?

test/kernels/test_spherical_linear_kernel.py
```diff
@@ -0,0 +1,194 @@
...
+        SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=None)
+        SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=NormalPrior(0, 1))
+        self.assertRaises(TypeError, SphericalLinearKernel, UNIT_BOUNDS_3D, lengthscale_prior=1)
+
+    def test_pickle_with_prior(self):
+        """Kernel with prior should survive pickle round-trip."""
+        kernel = SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=NormalPrior(0, 1))
+        pickle.loads(pickle.dumps(kernel))
+
+    def test_consistency_square_vs_rectangular(self):
```
gpleiss, line 157: Do we need this test?

test/kernels/test_spherical_linear_kernel.py
```diff
@@ -0,0 +1,194 @@
...
+    def test_prior(self):
+        """Should accept valid priors and reject invalid ones."""
+        SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=None)
+        SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=NormalPrior(0, 1))
+        self.assertRaises(TypeError, SphericalLinearKernel, UNIT_BOUNDS_3D, lengthscale_prior=1)
+
+    def test_pickle_with_prior(self):
+        """Kernel with prior should survive pickle round-trip."""
+        kernel = SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=NormalPrior(0, 1))
+        pickle.loads(pickle.dumps(kernel))
```
gpleiss, line 155: Do we need this test?

test/kernels/test_spherical_linear_kernel.py
```diff
@@ -0,0 +1,194 @@
...
+        loss.backward()
+        for name, param in kernel.named_parameters():
+            self.assertIsNotNone(param.grad, f"No gradient for {name}")
+
+    def test_pickle_with_bounds(self):
+        """Kernel with bounds should survive pickle round-trip."""
+        bounds = torch.tensor([[0.0, 0.0], [1.0, 1.0]])
+        kernel = SphericalLinearKernel(bounds=bounds)
+        loaded = pickle.loads(pickle.dumps(kernel))
+        self.assertAllClose(loaded.bounds, bounds)
```
gpleiss, line 144: Do we need this test?

test/kernels/test_spherical_linear_kernel.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def test_prior(self):
+        """Should accept valid priors and reject invalid ones."""
+        SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=None)
+        SphericalLinearKernel(bounds=UNIT_BOUNDS_3D, lengthscale_prior=NormalPrior(0, 1))
+        self.assertRaises(TypeError, SphericalLinearKernel, UNIT_BOUNDS_3D, lengthscale_prior=1)
```
gpleiss, line 150: Do we need this test?

gpleiss (changes requested):
Generally good; a few changes

gpytorch/kernels/spherical_linear_kernel.py
colmont: Done

gpytorch/kernels/spherical_linear_kernel.py
colmont: Done

gpytorch/kernels/spherical_linear_kernel.py
colmont: Done

gpytorch/kernels/spherical_linear_kernel.py
colmont: Done

gpytorch/kernels/spherical_linear_kernel.py
colmont: I tried this on a small example (SVM benchmark in high-dim BO): dividing the lengthscale by its norm is about 10x slower than using the softmax, and performance is not necessarily better. This was just a small local test with a few iterations (20-30); I have not yet double-checked this on the cluster for more problems/seeds/iterations. What do you think, worth investigating more or do you already have a strong preference on which reparameterization to go for?

gpytorch/kernels/spherical_linear_kernel.py
colmont: I got rid of it

test/kernels/test_spherical_linear_kernel.py
colmont: I got rid of it

test/kernels/test_spherical_linear_kernel.py
colmont: Done

test/kernels/test_spherical_linear_kernel.py
colmont: I got rid of it

test/kernels/test_spherical_linear_kernel.py
colmont: I got rid of it

test/kernels/test_spherical_linear_kernel.py
colmont: I got rid of it

test/kernels/test_spherical_linear_kernel.py
colmont: I got rid of it

colmont:
I tried to address all of your comments as best as possible, and left a longer reply for the reparameterization question. When playing around with this model for the `mujoco-humanoid` problem in high-dim BO, my jobs instantly OOMed... even before acquiring the first point. I believe I found a bug in GPyTorch? I think when `d >= n`, the linear kernels should be routed to `DefaultPredictionStrategy` instead of `LinearPredictionStrategy`. I tried to fix this as best as I could in a second commit, I hope it's kinda clear!

gpytorch/kernels/spherical_linear_kernel.py
colmont: I've done some more rigorous benchmarking (the 6 HDBO problems from the paper for 10 seeds each). Softmax constraint is definitely faster than dividing by the norm, and doesn't lead to worse performance (maybe even slightly better?). I would propose keeping softmax constraint and having it on by default.

gpleiss (approved):

gpleiss:
Thanks @colmont !

PR #2782, opened 2026-09-11
Add Matern52KernelGrad sign regression vs RBFKernelGrad

### LINKED ISSUE #2563: [Bug] Sign error in Matern52KernelGrad

castor639:
## Summary

The function-to-derivative sign fix for `Matern52KernelGrad` landed in #2598 (fixes the bug in #2563). This PR adds a regression that uses the original issue inputs and checks that those covariance signs match `RBFKernelGrad`.

## Test plan

- [ ] `python -m pytest test/kernels/test_matern52_kernel_grad.py -q`

Fixes #2563.


castor639:
Closing this for now. Happy to reopen later if useful.

